PR #3, opened 2023-01-20
86 Pre-commit hooks and README

jamesobrooks:
Adds configuration of [pre-commit](https://pre-commit.com) with a baseline for [detect-secrets](https://github.com/Yelp/detect-secrets).

Adds a basic README.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,8 @@ repos:
       rev: ''  # Use the sha / tag you want to point at
       hooks:
         - id: eslint
+          files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
+          types: [ file ]
           additional_dependencies:
           -   eslint@8.32.0
           -   eslint-config-prettier@8.6.0


README.md
```diff
@@ -0,0 +1,47 @@
...
+npm start
+```
+
+This will serve the client on port 3000.
+
+# Contributing
+
+## Style and Secrets Enforcement
+
+[pre-commit](https://pre-commit.com) allows us to make use of hooks that enforce style and prevent secrets from being committed to the repo. As such, it is expected that all code pushed to the repository have had these pre-commit hooks executed on all files created or updated. The configuration can be seen in `.pre-commit-config.yaml`.
```
governmentSponsored, line 41: Took a bit, but I managed to get `pre-commit` working locally. Should have instructions on getting this installed and running. For me, it took a few steps:
1. Open admin powershell prompt, and do `pip install pre-commit` (may even want to ensure folks have python & pip installed first in readme)
2. Then I had to run both `pre-commit autoupdate` and then `pre-commit install`
3. Updated the `.pre-commit-config.yaml` to include typescript files, which are not by default.
4. Added some pws and the commit failed locally (as expected). Can also verify with a `pre-commit run --all-files`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+        -   id: check-added-large-files
+  -   repo: https://github.com/pre-commit/mirrors-eslint
+      rev: ''  # Use the sha / tag you want to point at
+      hooks:
+        - id: eslint
+          additional_dependencies:
+          -   eslint@8.32.0
+          -   eslint-config-prettier@8.6.0
+          -   eslint-plugin-prettier@4.2.1
+          -   eslint-plugin-react@7.32.1
```
governmentSponsored, line 17: Added this to include javascript and typescript files (per [this documentation](https://github.com/pre-commit/mirrors-eslint)). May want to expand as project grows.

```yaml
files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
types: [file]
```

governmentSponsored (changes requested):
see other comments for requested changes

README.md
jamesobrooks: @governmentSponsored I'd rather not duplicate install instructions that exist on the linked pre-commit site, especially considering `pip` is only one way that it can be installed. But I do think it makes sense to explicitly state that pre-commit must be installed.

governmentSponsored: @jamesobrooks that's fair, linking to install instructions and indicating it needs to be installed works for me

governmentSponsored (approved):

PR #4, opened 2023-01-25
86 react build action

fmaddenflx:
New React build action created.
Test updated to support Hello World code.


.github/workflows/react-build.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches:
+      - main
+
+jobs:
+  build:
+    defaults:
+      run:
+        working-directory: gui
```
jamesobrooks, line 19: @fmaddenflx, did this end up being necessary with the inclusion of the directory in the path to the lock file?

governmentSponsored:
@fmaddenflx please make sure to resolve those [npm install](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/4008024650/jobs/6881657493#step:4:16) step issues

.github/workflows/react-build.yml
fmaddenflx: Not totally sure I understand your question.  Line 19 was necessary in order for the action to find the lock file.

fmaddenflx:
I am working to resolve the npm install step issues now.

governmentSponsored (approved):
these changes look good, like that there are no vulnerabilities now!

.github/workflows/react-build.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+        cache-dependency-path: gui/package-lock.json
+    - run: npm install
```
jamesobrooks, line 36: Should we be running `npm ci` rather than `npm install` to leverage the package-lock.json? I think that will result in a more consistent build process.

fmaddenflx: Done.

PR #9, opened 2023-03-08
111: Add pre-commit-java hooks

jamesobrooks:
# Java Formatting

Adds the use of the Google Java Formatter.

## pre-commit

For local development, commits will trigger the formatter if the developer has `pre-commit` installed—which they should.

### Auto-fix

Files will automatically be fixed by the hook, but will need to be manually committed.

## GitHub

Because `pre-commit.ci` fails to run this check successfully, the `pretty-format-java` hook is skipped for CI in favor of running the [axel-op/googlejavaformat-action@v3 GitHub Action](https://github.com/axel-op/googlejavaformat-action). 

> Note:
> We should not be relying on either pre-commit.ci or Github Actions to run our hooks, rather they are performed on GitHub as a failsafe.

### Auto-fix

The Action should automatically fix files and push a commit.

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -3,6 +3,14 @@ name: Continuous Deployment
 on: [push]
 
 jobs:
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3 # v2 minimum required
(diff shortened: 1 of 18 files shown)


jamesobrooks:
pre-commit.ci run

jamesobrooks:
pre-commit.ci run

jamesobrooks:
pre-commit.ci run

PR #20, opened 2023-04-24
111 merged

governmentSponsored:


api/node/.eslintrc.json
```diff
@@ -0,0 +1,18 @@
+{
+    "env": {
+        "browser": true,
```
jamesobrooks, line 3: Is this needed for the backend?

jamesobrooks (approved):
As previously discussed, this review is not intended to convey complete acceptance of the code. When we remove unused versions of the backend, we'll do a more in-depth review. Generally speaking, I have no major issues with what we've got here.

PR #25, opened 2023-04-27
111: Add instructions for validating pre-commit git hook

jamesobrooks:
Adds clarification for ensuring that git hook is set up properly for pre-commit.


fmaddenflx (commented):
Looks good.

PR #32, opened 2023-05-08
111 node function

fmaddenflx:
- Added Changes to update the title for the Hearing Date and Time column to remove ambiguity.
- Added the Hearing Disposition to Description mapper and related types and updated the query results for cases.
- Added Changes to update the backend
    1) Perf Fix : Updated the SQL Query to limit the number of records to 20.
    2) Added a Correlated SubQuery to fetch only the relevant record with the latest REQ_SEQ_NBR from the Hearing History table.
    
- Added tests to bring coverage up to 90-100% for every unit found in coverage report.  Remaining tests include a few mssql database tests.  About 33% of unit lines remaining.  Almost all, if not all, code has been tested manually successfully.


magile32 (approved):
Verified changes with manual end to end testing. All unit test passing

PR #34, opened 2023-05-09
111: GUI - updated title and logo image

fmaddenflx:
GUI - updated title and logo image

Note:  The image being used is a PNG image.  We should get an SVG if there is one available.

### DIFF
--- a/gui/package.json
+++ b/gui/package.json
@@ -63,6 +63,7 @@
     "build": "react-scripts build",
     "test": "react-scripts test --watchAll=false --roots tests/ --testMatch '**/*.test.{js,jsx,ts,tsx}'",
     "test:watch": "react-scripts test --roots tests/ --testMatch '**/*.test.{js,jsx,ts,tsx}' --coverage",
+    "coverage": "npm run test -- --coverage",
     "eject": "react-scripts eject",
     "lint": "eslint .",
     "lint:fix": "eslint --fix",


jayaneti1 (approved):

gui/src/components/HeaderNavBar.tsx
```diff
@@ -8,13 +8,13 @@ export const HeaderNavBar = () => {
...
           <img
-            src="doj-logo.svg"
+            src="doj-logo.png"
             alt="U.S. Trustee Program banner"
             className="doj-logo usa-banner__header"
           ></img>
           <div className="site-title">
             U.S. Trustee Program
-            <span className="sub-title">Bankruptcy Oversight Support System</span>
+            <span className="sub-title">Case Management System (CAMS)</span>
```
jamesobrooks, line 17: @jayaneti1, Stan had stated the title is `CAse Management System`, complete with the capital A to build out the acronym. Should that be reflected here?

jayaneti1: Yes Please

jamesobrooks:
```suggestion
            <span className="sub-title">CAse Management System (CAMS)</span>
```
To keep the title casing correct, even though it's not _title cased_.

PR #39, opened 2023-05-16
CAMS-20: Add SAST to pipeline

governmentSponsored:
# Purpose

To support the security posture of our application source code and deployed application, we want a tool that will scan the source of our application code.

# Pipeline Implementation

We add the `security-sast-scan` job to the build process of GitHub Actions. It does the following:
1. zips up the contents of the repo
1. Downloads the most recent baseline file from Azure 
1. Downloads and runs the veracode `pipeline-scan-LATEST` jar file against the repo
1. Uploads the baseline file 

# Local Implementation
Locally, developers can download that same jar file for use and execute with similar after zipping up the repository.

```sh
java -jar pipeline-scan.jar \
-vid YOUR_VERACODE_API_ID \
-vkey YOUR_VERACODE_API_KEY \
-bf results-latest.json \ 
-jf results.json \
-fjf filtered_results.json \
--file cams.zip
```

# Future Considerations
We could use our own customized version of their container with the baseline file built-in or downloaded from Azure.


amorrow-flexion (approved):

.github/workflows/continuous-deployment.yml
```diff
@@ -48,6 +48,75 @@ jobs:
...
           args: "--skip-sorting-imports --replace"
 
+  accessibility-test:
+    defaults:
+      run:
+        working-directory: user-interface
+
+    runs-on: ubuntu-latest
+
+    environment: ${{ needs.Build-Info.outputs.environment }}
```
jamesobrooks, line 58: We don't have to do it here, but I've been thinking we probably can remove this line. I like the move up too, I had realized the accessibility-test job was in a weird place after it got merged.

jamesobrooks (approved):

PR #42, opened 2023-05-19
OeSA screaming architecture

jamesobrooks:
# Screaming Architecture Exercise

The exercise:
> For your system (or the one provided), review the names of your directories/packages, and modules. Make improvements to be able to understand basic architectural importance of those things just from their names.

## Backend

Rename `functions/node` -> `api/function`.

## Frontend

Rename `gui` -> `web`.


amorrow-flexion (approved):

fmaddenflx:
I don't think "web" is a good name for the frontend.  The term "web" in my mind simply refers to the internet in general and to me, the backend api is just as much a part of the "web" and the frontend is.  I believe "frontend" would be a better title, or maybe something like ... "browser" ???

fmaddenflx:
Something like "user-interface" for the gui and "application-interface" for the api is more screaming and makes better sense to me.

governmentSponsored:
@jamesobrooks @magile32 I'd like to make sure we didn't break the build/deploy into Flexion Azure. Not sure if we want to check USTP side as well before merging.

jamesobrooks:
> Something like "user-interface" for the gui and "application-interface" for the api is more screaming and makes better sense to me.

@fmaddenflx I tend to agree that the term `web` is probably not the best. Maybe we can get opinions from Jon and Phoenix?

jamesobrooks:
> @jamesobrooks @magile32 I'd like to make sure we didn't break the build/deploy into Flexion Azure. Not sure if we want to check USTP side as well before merging.

It appears that the file permissions of the deploy scripts had changed in the move. Here's a deployment from a branch I created off of this one (so I don't have to include transitory commits).

https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/5059711032/jobs/9081818780

fmaddenflx:
> Something like "user-interface" for the gui and "application-interface" for the api is more screaming and makes better sense to me.

After discussing with Phoenix and Jon, we came to the agreement that "user-interface" was good for gui, and "backend" for the api.

PR #45, opened 2023-05-23
CAMS-24: Updated front-end tests to 94% coverage

fmaddenflx:
Coverage is almost 95%.

I removed unused CRUD functionality for now.  No need to write tests for RUD if we aren't doing any yet.


user-interface/package.json
```diff
@@ -58,6 +58,13 @@
     "nth-check": "2.0.1",
     "webpack": "^5.78.0"
   },
+  "jest": {
+    "coveragePathIgnorePatterns": [
+      "/index.tsx",
+      "/reportWebVitals.ts",
+      "/configuration/apiConfiguration.ts"
+    ]
+  },
```
fmaddenflx, line 67: I added index.tsx to the ignored patterns also, since it isn't doing anything logical, but rather just attaching the app to the root html element and rendering the base react App component.

It's also running reportWebVitals(), which sets up tracking for google analytics etc.

user-interface/src/App.tsx
```diff
@@ -14,8 +14,8 @@ function App() {
         <HeaderNavBar />
         <div className="body">
           <Routes>
-            <Route path="/cases" element={<CaseList />}></Route>
             <Route path="/" element={<Home />}></Route>
+            <Route path="/cases" element={<CaseList />}></Route>
             <Route path="*" element={<NotFound />}></Route>
```
fmaddenflx, line 19: Default screen should be "/".  I don't know if defining "/cases" route first would affect the routing but I believe it handles the path expressions in a top->down manor.

governmentSponsored (approved):
test coverage looks good now, and the changes you made sense to me (as far as I understand things).

jamesobrooks (approved):

PR #61, opened 2023-06-14
Cams 49 frontend tests

fmaddenflx:
Front-end tests to support Chapter 15 and bring coverage up above 90%


user-interface/src/components/utils/http.humble.ts
jamesobrooks, line 1: I think this is more of an adapter than a humble object because we are changing how we call it. Note the following from the OeSA slides on the invasive-species rule:
> Changing semantics is the job of an adapter.

For example we are adding—by default—at least one header and the cache setting.

So I would suggest renaming this to remove `humble` from the name.

jamesobrooks: Also note that since fetch is part of the language, I don't know that we should have a humble object for it, but it's certainly reasonable to have an adapter.

jamesobrooks (changes requested):
Looks good but I think we should rename `http.humble.ts`.

governmentSponsored:
@fmaddenflx you may need to pull in main for veracode disabling in case it fails the build on your next commit

user-interface/src/models/api.test.ts
```diff
@@ -0,0 +1,79 @@
+import Api, { ObjectKeyVal } from './api';
+import * as httpHumble from '../components/utils/http.adapter';
```
jamesobrooks, line 2: Looks like we need to rename `httpHumble` to `httpAdapter` in this test file.

jamesobrooks (approved):
After changing the name we import the http adapter as in a test, this can be merged.

PR #73, opened 2023-06-29
CAMS-79: Added 'Case Assignment' link and changed 'Cases' to 'Case Li…

fmaddenflx:
Added 'Case Assignment' link and changed 'Cases' to 'Case List', both in upper right of screen.  User now has a way to click and access the Case Assignment screen without having to type in the exact URL.

### DIFF
--- a/user-interface/src/components/HeaderNavBar.tsx
+++ b/user-interface/src/components/HeaderNavBar.tsx
@@ -18,7 +18,7 @@ export const HeaderNavBar = () => {
           </div>
           <nav className="nav-bar" role="navigation" aria-label="main-navigation">
             <NavLink to="/cases" data-testid="main-nav-cases-link">
-              List Cases
+              Cases
             </NavLink>
             <NavLink to="/case-assignment" data-testid="main-nav-case-assignment-link">
               Case Assignment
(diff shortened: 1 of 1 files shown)


user-interface/src/components/HeaderNavBar.tsx
```diff
@@ -18,7 +18,10 @@ export const HeaderNavBar = () => {
           </div>
           <nav className="nav-bar" role="navigation" aria-label="main-navigation">
             <NavLink to="/cases" data-testid="main-nav-cases-link">
-              Cases
+              List Cases
```
jamesobrooks, line 21:
```suggestion
              Cases
```

jamesobrooks (approved):

user-interface/src/components/HeaderNavBar.tsx
```diff
@@ -20,6 +20,9 @@ export const HeaderNavBar = () => {
             <NavLink to="/cases" data-testid="main-nav-cases-link">
               Cases
             </NavLink>
+            <NavLink to="/case-assignment" data-testid="main-nav-case-assignment-link">
+              Case Assignment
```
jamesobrooks, line 24:
```suggestion
              Assignment
```

amorrow-flexion (approved):

PR #89, opened 2023-07-17
Cams 85 fetch attorneys list fix linting

jamesobrooks:
# Problem

During CAMS-85 and CAMS-97 work we discovered that our linting wasn't really catching everything we would have expected.

# Solution

Updating `backend/functions/.eslintrc.json` to ensure that we use the same rulesets as we are using in the frontend caused the linter to catch what it was missing.

# Other Changes

We ran the linter against all files in the backend project and corrected the issues.


fmaddenflx (approved):
Looks good to me James.  I'd love to hear any issues or suggestions you may have with my changes.

PR #92, opened 2023-07-18
CAMS-97: Repurpose PACER gateway interface

jamesobrooks:
# Purpose

Renames the PacerGatewayInterface to CasesInterface. Implements a local gateway to handle `getChapter15Cases`.

# Major Changes

- In addition to the name change, the signature for `getChapter15Cases` is modified to take an `options` object to allow for multiple optional arguments. This made testing significantly more meaningful.
- Added a `date-helper.ts` file which provides two helper functions for dealing with dates
  - `getDate` is a wrapper for `new Date()`. It takes a real month number rather than a zero-based month number.
  - `calculateDifferenceInMonths` gets the number of _completed_ months between two dates.
    - This method has several tests covering cases including spanning years (wrote the tests first actually using TDD)


fmaddenflx (approved):
Looks good.

PR #94, opened 2023-07-19
CAMS-102 Enable access to CAMS

magile32:


governmentSponsored (approved):

jamesobrooks (commented):
Please see the questions posed on the proposed code change.

.github/workflows/continuous-deployment.yml
```diff
@@ -485,3 +485,7 @@ jobs:
         run: |
           name="gha-agent-${{ env.stackName }}"
           ops/helper-scripts/dev-rm-allowed-ip.sh ${{ steps.rgApp.outputs.out }} ${{ env.stackName }} ${name:0:32}
+      - name: Enable access
+        run: |
+          az webapp config access-restriction add --resource-group ${{ steps.rgApp.outputs.out }} --name ${{ env.webappName }} --rule-name AllowAll --action Allow --ip-address 0.0.0.0/0 --priority 100 1>/dev/null || true
+          az functionapp config access-restriction add --resource-group ${{ steps.rgApp.outputs.out }} --name ${{ env.apiName }} --rule-name AllowAll --action Allow --ip-address 0.0.0.0/0 --priority 100 1>/dev/null || true
```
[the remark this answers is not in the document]
magile32: - I agree, we can improve on this by implementing this step as a separate job. It would require ensuring the new job has all the values it needs to be passed in as inputs. As we may add more jobs, we may want to consider decomposing continous-deployment into several workflows later on.
- Since we are just adding a rule to allow all ips it should not impact a running integration test if the implementation is left as is.
- I was trying to keep existing operations in tack at the moment to keep things close to how it can be executed in the USTP environment (if post deployment tests will be executed)

.github/workflows/continuous-deployment.yml
```diff
@@ -275,7 +275,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build-info, build-frontend, build-service, accessibility-test]
     environment: ${{ needs.build-info.outputs.environment }}
-    if: ((github.ref == 'refs/heads/main') || inputs.deployBranch)
+    if: ((github.ref == 'refs/heads/main') || inputs.deployBranch || true) # TODO remove boolean 'true' after validation
```
jamesobrooks, line 278: This change needs to be undone.

jamesobrooks (changes requested):

.github/workflows/continuous-deployment.yml
magile32: Updated

jamesobrooks (approved):

PR #100, opened 2023-07-25
CAMS-82 get region and office

Swathi-Darbha:
# Purpose

To fetch the region and office information for the current AUST who logs in for Chapter 15 Case Assignment.

# Major Changes

Updated the Case Assignment UI to load the region and office information from the state object.

# Testing/Validation

Verification : Login to the CAMS Website and navigate to the /case-assignment page. The page displays the region and office of the AUST who is currently logged in.

# Notes

The current story is about getting the region and office info only for the Region 2 , Manhattan office.
Future stories will add in features to pull the region and office info of the AUST, for any office in all other regions, from a backend data source.


governmentSponsored (commented):
@Swathi-Darbha Changes look good to me, however your PR is still in draft mode. Recommend moving it out. Also, you should indicate that this PR also includes adding access to the Azure WebApp since there are updates from you and Mike Ly in the `continuous-deployment.yml` file.

governmentSponsored (approved):

PR #101, opened 2023-07-27
Cams 89 logging implementation

amorrow-flexion:
# Purpose

The purpose of this PR is to implement Application Insights into each of the Azure functions within the repository and add necessary npm packages for App Insights

# Major Changes

- Implementation of app insights in each of the Azure functions. 
- addition of applicationinsights npm packages
- Additions of bicep files for app insights and log analytics workspace
- Modification of pipeline to accommodate for the new configuration settings

# Testing/Validation

Tested Validity of app deployment on a development branch and environment

# Notes

More work needs to be done with the bicep modules. This will be continued into the work on the frontend. Once merged this will add the appinsights configuration settings for the app insights resources to the main resources.


.github/workflows/continuous-deployment.yml
```diff
@@ -259,7 +258,8 @@ jobs:
           npm run build
 
       - name: Execute Tests
-        run: npm run test
+        run: |
+          export APPINSIGHTS_CONNECTION_STRING="TESTSTRING" && npm run test
```
amorrow-flexion, line 262: This change my puzzle some people. For running the tests on the service we do not need a valid connection string in the environment variables for application insights, but it does need a non null value in order for them to pass. In order to prevent any collection from the App Insights resources, we pass in an invalid string for the test components to run when the functions are started.

.github/workflows/continuous-deployment.yml
```diff
@@ -390,6 +390,8 @@ jobs:
           MSSQL_ENCRYPT=${{ secrets.AZ_MSSQL_ENCRYPT }} \
           MSSQL_TRUST_UNSIGNED_CERT=${{ secrets.AZ_MSSQL_TRUST_UNSIGNED_CERT }} \
           PACER_TOKEN_URL=${{ vars.PACER_TOKEN_URL }} \
+          APPINSIGHTS_CONNECTION_STRING=${{ secrets.API_APPINSIGHTS_CONNECTION_STRING }}
+          APPINSIGHTS_INSTRUMENTATIONKEY=${{ secrets.API_APPINSIGHTS_INSTRUMENTATION_KEY }}
```
amorrow-flexion, line 394: the secrets for main-gov are already setup

backend/functions/host.json
```diff
@@ -4,8 +4,15 @@
     "applicationInsights": {
       "samplingSettings": {
         "isEnabled": true,
-        "excludedTypes": "Request"
+        "excludedTypes": ""
       }
+    },
+    "fileLoggingMode": "always",
+    "logLevel": {
+      "default": "Trace",
```
jamesobrooks, line 12: Should the default level be `Trace`? Unless I'm misunderstanding, this is quite verbose, and could be outputting sensitive information.

https://learn.microsoft.com/en-us/azure/azure-functions/configure-monitoring?tabs=v2#configure-log-levels

.github/workflows/continuous-deployment.yml
amorrow-flexion: @jamesobrooks  Yes it would. That is a good pickup.

backend/functions/host.json
```diff
@@ -4,8 +4,15 @@
     "applicationInsights": {
       "samplingSettings": {
         "isEnabled": true,
-        "excludedTypes": "Request"
+        "excludedTypes": ""
```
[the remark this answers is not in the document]
amorrow-flexion: We don't necessarily need this at the moment. I left it in the case we would for the future when we more deeply configure logging.

backend/functions/host.json
amorrow-flexion: I set it to trace for initial ingestion, but Debug or Information levels would likely be a better option for us on our side.

jamesobrooks (approved):

PR #103, opened 2023-07-31
Cams 87 switch to vite

fmaddenflx:
# Purpose

Purpose:  To eject create-react-app (CRA) from our project due to it's being deprecated and unsupported, and setup USWDS directly, without using a 3rd party React USWDS library.  Vite looked to be an easier, more successful transition than Webpack.  Using Vite enabled us to switch away from create-react-app without the need to incorporate many of the sub-packages that came from create-react-app with all of their own issues.  CRA proved to cause a messy conversion to Webpack, and looked like it may require us to complete rewrite our test suite.

Vite on the other hand, provides a very similar api, allowing us to switch with minimal changes to our code, including our tests.

In addition to all of that, Vite builds and runs code 4-10 times faster than CRA or Webpack and produces smaller builds.

# Major Changes

Changes were minor.  

The main change to the core code was to move the index.html file out of src and into the parent folder, and then to remove any reference to build variables such as %PUBLIC_URL% or %PUBLIC_PATH%.

The other major change was in how environment variables are referenced.  With Vite we are able to ditch the use of REACT_APP_ prefix on environment variables and use whatever prefix we like.  I went with CAMS_ as an environment variable prefix.

There are some very minor changes to tests such as, rather than using jest.fn() or jest.mock(), we are using vi.fn() and vi.mock() which comes from the vitest package.

There are some minor changes to scripts in the package.json file, but not totally in the way we use them.  There is only a minor difference there.

Finally we are now using USWDS directly, rather than trusworks-uswds.

# Testing/Validation

Tests are all passing.  To run the tests, type `npm run test` in the cli.  You can also run `npm run coverage` to see the current test coverage for the app.


user-interface/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ESNext",
```
fmaddenflx, line 3: Check into this to make sure this doesn't become a problem if es is upgraded beyond what vite can handle.

governmentSponsored (approved):

jayaneti1 (approved):

PR #114, opened 2023-08-16
Cams 109 add architecture diagram

jamesobrooks:
# Purpose

Adds diagrams as code to support threat modeling.

# Major Changes

We define the architecture using the [Structurizr domain-specific-language](https://github.com/structurizr/dsl/tree/master/docs/cookbook/introduction) and then use an [accompanying CLI](https://github.com/structurizr/cli) to output [Mermaid](https://mermaid.js.org/). Then we copy the Mermaid code into markdown files for inclusion in our Docsify pages.

# Testing/Validation

GitHub Pages is currently set to deploy from this branch and behavior can be validated [here](https://us-trustee-program.github.io/Bankruptcy-Oversight-Support-Systems/#/).

# Notes

This approach allows us to generate several levels of diagram that might be used for different purposes. This is called the C4 Model and you can learn more [here](https://c4model.com/).

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -7,6 +7,7 @@ on:
     paths-ignore:
       - '*.md'
       - '**/*.md'
+      - '**/*.dsl'
   workflow_dispatch:
     inputs:
       deployBranch:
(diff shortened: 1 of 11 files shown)


jamesobrooks:
We might want to put the script in the `ops/helper-scripts` folder. We also don't have this automated yet. We could do that in this PR or a following one. There is certainly room for adding more detail to the architecture and adding other diagrams.

governmentSponsored (approved):

PR #117, opened 2023-08-21
CAMS-121 Update frontend host instance

magile32:
# Purpose

Switch from using a Windows host to Linux host for Application Service.

# Major Changes

- Update Azure App Service plan to use Linux base host
- Add flag `--os-type` for az deployment command
- Replace web.config with nginx.conf
- Update base nginx.conf to fix page routes and remove references to php

# Testing/Validation

- Provision and deploy branch changes successfully. Manually verified ui loads into web browser and navigation to existing views.
- Resolve issue found for URI routing. To test, the following steps can be taken:
  - Navigate to CAMS home page
  - Navigate to Cases or Assignment view and let it load
  - Reload page or navigating directly to the uri of the new view should not return a 404

# Notes

- For deploying to existing deployment, some manual clean up of Azure resource will be needed. The Azure app service and app service plan will need to be deleted before deploying.
- Web server on the host instance has changed from IIS to Nginx. Added nginx.conf to user-interface/public folder to include the new server configs.
- Note that php runtime was used based on Azure App Service automated selection of a container to host the web application project. A container with NodeJs runtime could be configured but that would require additional setup to include a web server since this container will not include nginx.


governmentSponsored (approved):

amorrow-flexion (approved):

user-interface/public/nginx.conf
```diff
@@ -0,0 +1,31 @@
+server {
+    #proxy_cache cache;
+    #proxy_cache_valid 200 1s;
+    listen 8080;
+    listen [::]:8080;
+
+    root /home/site/wwwroot;
+
+    index index.html index.htm;
+    server_name  example.com www.example.com;
```
fmaddenflx, line 10: This probably doesn't matter but curious to know why we left the server name "example".

magile32: At the moment, I did not what to put there.

magile32: Because there is only one server block define, the value for `server_name` should not matter.

jayaneti1 (approved):

PR #119, opened 2023-08-23
CAMS-93 Import dxtr data

magile32:
# Purpose

Update sql db import process to include support for new dxtr database and table.

# Major Changes

- Add loader class for new table
- Refactor existing logic to work with new database

# Testing/Validation

WIP

# Notes

### DIFF
--- a/ops/dataloader/src/main/java/gov/doj/Driver.java
+++ b/ops/dataloader/src/main/java/gov/doj/Driver.java
@@ -8,7 +8,11 @@ import java.util.Properties;
 
 public class Driver {
 
-  private static LoaderMap loaderMap = new LoaderMap();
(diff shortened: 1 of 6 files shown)


ops/dataloader/src/main/java/gov/doj/Driver.java
```diff
@@ -8,7 +8,11 @@
 
 public class Driver {
```
magile32, line 9: At the moment, this class still needs refactoring to include executing loading logic for AODATEX database.

magile32:
Related changes for CAMS-93 https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/pull/120

PR #127, opened 2023-09-07
CAMS-98 Get cases from DXTR

jamesobrooks:
# Purpose

Replace use of PACER with query to DXTR database.

# Major Changes

Removes all code related to the use of PACER.
Adds a DXTR gateway for performing queries.
Refactoring some code that was named for PACER but was actually not PACER-specific.
Updated documentation to reflect newer environment variables required to run locally.

# Testing/Validation

Wrote automated tests. Manually validated that when not mocking the database we get cases from DXTR for the `Assignments` page.


.github/workflows/continuous-deployment.yml
```diff
@@ -16,16 +16,16 @@ on:
         default: 'false'
         type: choice
         options:
-        - 'false'
-        - 'true'
+          - 'false'
```
magile32, line 19: Notice mostly formatting updates to this file. Is there a linter or formatter that the team could use?

backend/functions/lib/adapters/gateways/cases.azure.sql.gateway.ts
```diff
@@ -64,15 +69,20 @@ class Chapter11ApiGateway implements Chapter11GatewayInterface {
       });
     }
 
-    const queryResult: QueryResults = await executeQuery(context, query, input);
+    const queryResult: QueryResults = await executeQuery(
+      context,
+      context.config.acmsDbConfig,
```
magile32, line 74: I think we either just past the context or just what is needed from context here.

A third option here could be just to pass in config here. If the reason for passing in context is for logging, we could let the error message propagate and log it in this class.

Another option which would probably take the most work though is to create a repository object to query anything related to cases. Also refactor log object as an object instead of a static methods. We should talk about this one if interested in this approach.

backend/functions/lib/adapters/gateways/cases.dxtr.gateway.ts
```diff
@@ -0,0 +1,52 @@
...
+      input,
+    );
+
+    if (queryResult.success) {
+      log.debug(context, MODULENAME, `Results received from DXTR ${queryResult}`);
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return (queryResult.results as mssql.IResult<any>).recordset;
+    } else {
+      throw Error(queryResult.message);
```
magile32, line 49: I prefer to log error message as soon as possible. Since log and context is available here I think we should log the error.

backend/functions/lib/adapters/gateways/cases.local.inmemory.gateway.test.ts
magile32, line 1: Notice two classes named ./backend/functions/lib/adapters/gateways/cases.local.gateway.ts and ./backend/functions/lib/adapters/gateways/cases.local.inmemory.gateway.ts

Naming is a bit confusing. What does 'local' mean? We used local elsewhere and my interpretation is that it is a class to help with testing and not the actual gateway.

docs/running.md
```diff
@@ -28,18 +31,21 @@ This will serve the client on port 3000.
...
 
 ```
-REACT_APP_BASE_PATH={the base path of the backend, if any}
-REACT_APP_SERVER_HOSTNAME={the TLD of the backend}
-REACT_APP_SERVER_PORT={the port the backend is served on}
-REACT_APP_SERVER_PROTOCOL=http[s]
+CAMS_BASE_PATH={the base path of the backend, if any}
+CAMS_SERVER_HOSTNAME={the fully qualified domain name (FQDN) of the backend}
+CAMS_SERVER_PORT={the port the backend is served on}
+CAMS_SERVER_PROTOCOL=http[s]
```
magile32, line 41:
```suggestion
CAMS_SERVER_PROTOCOL=http[s]
CAMS_APPLICATIONINSIGHTS_CONNECTION_STRING={optional instrumentation key for extended logging features}
```

magile32 (commented):
Pointed out some area of confusion and suggestions.

backend/functions/lib/adapters/gateways/cases.azure.sql.gateway.ts
jamesobrooks: The reason we pass `context.config.acmsDbConfig` is because that is how we communicate to the `executeQuery` function which database to run the query against. In other places we pass `context.config.dxtrDbConfig`. I would prefer to remove chapter 11 stuff entirely and then we don't need to keep track of two different sql configs.

backend/functions/lib/adapters/gateways/cases.local.inmemory.gateway.test.ts
jamesobrooks: This is a good point. Fritz and I were talking about that as a place where some refactoring is warranted, but I forgot about it. I believe the difference is that one is used exclusively for testing and the other is used for local development (and possibly some tests as well).

jamesobrooks: Correction to my last. `cases.local.gateway.ts` is what used to be the local PACER gateway and is what we use for chapter 15. `cases.local.inmemory.gateway.ts` is chapter-11-specific.

magile32 (approved):
Comments will be address in DevEx/OpEx tasks.

backend/functions/lib/adapters/gateways/cases.local.inmemory.gateway.test.ts
fmaddenflx: File name changed to chapter11.local.gateway.ts

Affected files included the data-access.proxy.test.ts.  This was testing the cases local inmemory gateway using the data-access proxy.  However, chapter 11 actually doesn't use that proxy any longer.  The only endpoint still using it is the users endpoint.  So I updated that test to test against users instead of chapter 11 cases.

PR #131, opened 2023-09-08
CAMS-130 Updates for USTP deployment

magile32:
# Purpose

During deployment of backend code, seeing intermittent failure cause by a container restart during deployment. The container restart occurs because of management operations to add/update app settings to the Azure functionapp.

# Major Changes

- Update deployment script to run management operation after code deployment. 

# Testing/Validation

- Tested via deployment in USTP environment
- Test that intermittent issues with access issues for integration and smoke tests are no longer an issue
- Tested that intermittent failure with SCM container restart are no longer an issue

# Notes


jamesobrooks:
> Tested via deployment in USTP environment

Could you provide some more detail here? Was this deployment run successfully enough times to be confident that the change will reduce the occurrence of the failure? Are both the change in order and the longer sleep time required?

ops/helper-scripts/az-func-deploy.sh
```diff
@@ -116,3 +105,14 @@ if [[ -n ${identities} ]]; then
...
+# configure Application Settings
+if [[ -n ${app_settings} ]]; then
+    echo "Set Application Settings for ${app_name}"
+    for item in ${app_settings}; do
+        az functionapp config appsettings set -g $app_rg -n $app_name --settings "${item}" --query "[?name=='${item%=*}'].name | [0]" --output tsv
+    done
+fi
+
+# Gives some time for prior management operation to complete before going forward
+sleep 45s
```
amorrow-flexion, line 118: I don't think this will function quite as expected since each operation updating the function app config will cause a restart and this is outside of the for loop.

magile32: I'm hoping that container will restart once for these set of changes. 

Looking through documentation and a search, I couldn't find any definitive answer on how Azure will behave. I think the best way to test this is to run this thru the pipeline.

magile32: Add jira task CAMS-144 to address the looping. There should be a way to pass in a list of settings and have it execute once. We just need to adjust the output and make sure no secrets is logged

governmentSponsored (approved):

PR #132, opened 2023-09-08
Cams 98 format date

jamesobrooks:
# Purpose

The date format coming from PACER was `yyyy-MM-dd`, but from DXTR is an ISO date string. Our frontend code was splitting the `date filed` on the `-` and as a result was displaying a messed up ISO date instead of the `MM-dd-yyyy` format we were looking for.

# Major Changes

We format the date filed field in our query from DXTR. For the local gateway which was originally based on PACER data, we leave the source data the same, but add a helper function which converts the date. 

# Testing/Validation

Added a test for the new helper function. Validated the date format running locally with both the DXTR and local data sources.

# Notes

If we find the new helper function useful for more use cases, some validation would be needed. For now, it is only being used on hard coded data, so that would add unnecessary complexity.


fmaddenflx (approved):
Looks good.

governmentSponsored (approved):

PR #138, opened 2023-09-14
CAMS-148: Get count of cases assigned to attorneys

jamesobrooks:
# Purpose

We had generated random numbers for case count on the assignment modal and needed to get real numbers.

# Major Changes

When we retrieve the list of attorneys from the backend we now also get a count of how many chapter 15 case assignments they have.

# Testing/Validation

Validated locally by hitting the databases deployed in Azure.

Unit testing included.

# Further Work

We are just returning an empty array from `case.assignment.local.repository.ts` (for local development) when we get the assignments for an attorney. We need to decide on an approach for this to let us develop locally without hitting deployed databases.

# Notes

Renamed `attorneyName` to `name` in the `CaseAttorneyAssignment` class—and consequently the Cosmos DB data for the `assignments` container.

### DIFF
--- a/backend/functions/lib/adapters/types/attorney.class.ts
+++ b/backend/functions/lib/adapters/types/attorney.class.ts
@@ -1,56 +1,52 @@
+import { ObjectKeyVal } from './basic';
+
 export class Attorney {
-  private _firstName: string;
-  private _middleName: string;
-  private _lastName: string;
-  private _generation: string;
-  private _courtId: string;
-  private _office: string;
+  firstName: string;
+  middleName: string;
+  lastName: string;
+  generation: string;
+  courtId: string;
+  office: string;
+  caseLoad: number;
 
-  get firstName(): string {
-    return this._firstName;
+  constructor(attorney?: ObjectKeyVal) {
+    if (attorney) {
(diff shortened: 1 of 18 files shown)


backend/functions/lib/adapters/types/attorney.class.ts
```diff
@@ -1,56 +1,52 @@
...
+    return full;
+  }
+
+  getAsObjectKeyVal(): ObjectKeyVal {
+    return {
+      firstName: this.firstName,
+      middleName: this.middleName,
+      lastName: this.lastName,
+      generation: this.generation,
+      courtId: this.courtId,
```
jharcey-flexion, line 47: Why is the courtId here? The trial attorneys are not tied to a court, they are tied to an USTP office.

jharcey-flexion: I thought we had already removed that.

jamesobrooks: Good catch. It is now removed.

fmaddenflx (approved):
All looks good to me.

magile32 (approved):

PR #139, opened 2023-09-15
CAMS-143 Fix post deployment test failures

magile32:
# Problem

Pipeline run experiencing intermittent smoke test and integration failures. Determine that failure is most likely cause by non-unique names created for allow rules for making network request to webapp and api. The other cause is that integration test executes before the webapp is online.

# Solution

- Modify approach to add rule to allow target GHA runner access to execute test request to webapp and api. Rule name created should be unique
- Move integration test after smoke test. 

# Testing/Validation

- Manual execution of pipeline with all tests passing

# Other Changes

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -453,7 +453,7 @@ jobs:
         working-directory: backend/functions
 
     runs-on: ubuntu-latest
-    needs: [build-info, deploy-webapp, deploy-service]
+    needs: [build-info, smoke-test-application]
     environment: ${{ needs.build-info.outputs.environment }}
     env:
       stackName: ${{ needs.build-info.outputs.stackName }}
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-deployment.yml
```diff
@@ -453,11 +453,12 @@ jobs:
         working-directory: backend/functions
 
     runs-on: ubuntu-latest
-    needs: [build-info, deploy-webapp, deploy-service]
+    needs: [build-info, smoke-test-application]
     environment: ${{ needs.build-info.outputs.environment }}
     env:
       stackName: ${{ needs.build-info.outputs.stackName }}
       apiName: ${{ needs.build-info.outputs.apiName }}
+      ruleNumber: 100
```
jamesobrooks, line 461: Why do we call this `ruleNumber` here and `priority` in the other file?

fmaddenflx (approved):
Looks good.

amorrow-flexion (approved):

PR #148, opened 2023-09-26
CAMS-47: update pa11y tests

jamesobrooks:
# Purpose

Ensure the case assignment screen and modal are tested for accessibility.

# Major Changes

Adds a test of the case assignment screen with an action to open the modal so that Pa11y validates both the main screen and the modal.

# Testing/Validation

The change is itself a test which passes locally and in the [GitHub Actions workflow](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/6319376601/job/17160183722).


user-interface/src/components/CaseAssignment.tsx
```diff
@@ -72,12 +72,10 @@ export const CaseAssignment = () => {
...
-          if (caseNode.assignments) {
-            if (caseNode.assignments.length > 0) {
-              assignmentList.push(caseNode);
-            } else {
-              nonAssignmentList.push(caseNode);
-            }
+          if (caseNode.assignments && caseNode.assignments.length > 0) {
+            assignmentList.push(caseNode);
+          } else {
+            nonAssignmentList.push(caseNode);
```
jamesobrooks, line 78: This change allows us to cover the condition where `caseNode.assignments` is falsy, where before we were not covering that condition. Semantically speaking this condition should not happen because the backend sends an empty array, but that isn't really a business requirement, just something we do.

amorrow-flexion (approved):

PR #152, opened 2023-09-27
CAMS-47: Set the case assignment screen as the home screen

jamesobrooks:
# Purpose

Set the home view to be the case assignment screen.

# Major Changes

Add the CaseAssignment component to the Home component.

# Testing/Validation

Ran frontend locally, validated that the Case Assignment screen is displayed as the home page.


magile32 (approved):
Passing test in build now

PR #154, opened 2023-09-27
CAMS-144 improve func app deploy

magile32:
# Purpose

There was a concern with the CAMS functionapp deployment script, `az-func-deploy.sh`, where multiple calls to update app settings may cause the scm container restart because of management operations. Reducing the number of request via the az cli is also desired.

# Major Changes

- Update script to make a single call to config app settings
- Update query to display name of app settings

# Testing/Validation

- Local testing to confirm scripts works as expected to deploy backend code.
- Verified no secrets outputted to standard out
- Verified all the above again in branch deployment

# Notes

### DIFF
--- a/ops/helper-scripts/az-func-deploy.sh
+++ b/ops/helper-scripts/az-func-deploy.sh
@@ -109,18 +109,5 @@ fi
 # configure Application Settings
 if [[ -n ${app_settings} ]]; then
     echo "Set Application Settings for ${app_name}"
-    # for item in ${app_settings}; do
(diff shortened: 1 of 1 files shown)


governmentSponsored (approved):

governmentSponsored:
@magile32 did the deployment confirm that the deployment was faster or the app restarted fewer times?

magile32:
> @magile32 did the deployment confirm that the deployment was faster or the app restarted fewer times?

I didn't have a way to confirm the prior implementation caused multiple restarts but this change should guarantee that it needs to only restart the container once. I did remove the `sleep` operation in the script which is no longer needed so it would speed the job execution time.

fmaddenflx (approved):
Looks good.

PR #158, opened 2023-10-03
CAMS-188: Pin node version to 18.17 across backend and user-interface

btposey:
# Purpose

Support consistent use of Node.js versions across environments using Node Version Manager (NVM).

# Major Changes

No new or updated feature.

Added NVM resource files and updated README.md files. Pinned the Node.js version to 18.17 independently in the backend and user-interface directories. 

# Testing/Validation

No feature to test.


user-interface/README.md
```diff
@@ -2,6 +2,61 @@
 
 This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
jamesobrooks, line 3: Not necessary for you to fix here, but since we got rid of create-react-app, we should probably remove this.

btposey: That was going to be my question to the team.

backend/functions/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+See: https://github.com/nvm-sh/nvm#install--update-script
+
+### `.nvmrc`
+
+The Node.js version used by the project is specified in the `.nvmrc` resource file.
+
+### Installing a Node.js version
+
+Use `nvm` to install a Node.js version.
```
governmentSponsored, line 17: @jamesobrooks @btposey @magile32 @amorrow-flexion think we should update the build on USTP ADO as well to match this?

governmentSponsored:
@btposey @jamesobrooks Which jira issue is this tied to? I don't see that reflected in commits or branch name.

governmentSponsored (changes requested):
Need to see which issue(s) this is tied to. Also, not sure I understand why we need this change now?

backend/functions/README.md
btposey: This is an opportunistic improvement without any urgent need to be merged. No JIRA ticket, so one can be added to track this. Commit messages can be modified to reference the ticket if and when one is created.

The node version of the Azure function runtime in use should drive the version in the "backend" directory. The opposite may be true with the React app. The Github runner can/should be able to use NVM to use the specified node version to build the UI bundle.

governmentSponsored (approved):

jamesobrooks (approved):

PR #166, opened 2023-10-09
CAMS-189: Added Back button and Icon uswds component

fmaddenflx:
# Purpose

Add Back Button to return to case assignment screen from case detail.

# Major Changes

Link added and USWDS Icon component created.

# Testing/Validation

Testing coverage did not change.


jamesobrooks (approved):
I think maybe we should consider a way to make the Icon component more clear what is available. As is, it seems a bit like an invasive species rule smell.

PR #181, opened 2023-10-18
CAMS-197: Fixed font loading issues, optimized image load

fmaddenflx:
Jira ticket: CAMS-197

# Purpose

Fonts were not being referenced correctly in the css.

# Major Changes

I resolved this by having the build copy the USWDS fonts over to the expected directory.

In the process I noticed that the images could be loaded the same way, and by making the switch, the image use for the sprite can be simplified.  Therefore I also updated the way images are copied on build.  The result is that all images from uswds are now included, as well as all fonts.  This may be seen as a bad thing if we do not need to utilize all of those images.  This can be refined later if need be.

# Testing/Validation

We may not need any testing for this.

### DIFF
--- a/common/src/feature-flags.d.ts
+++ b/common/src/feature-flags.d.ts
@@ -1,0 +1,4 @@
+export interface FeatureFlagSet {
+  [key: string]: boolean | string | number;
+}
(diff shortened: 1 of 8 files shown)


amorrow-flexion:
image

PR #185, opened 2023-10-20
CAMS-193 dod accessibility

fmaddenflx:
# Purpose

Fix Tab stop and scrolling accessibility issues in the modal.

# Major Changes

Added focus handling to the AssignAttorneyModal to handle scrolling updates.
Added modal fucus on initial load of modal.  This was handled in the generic modal.

# Testing/Validation

Our coverage has fallen below 90%.  We need to tackle testing for the modal components next week. (anticipating difficult testing)

### DIFF
--- a/user-interface/src/components/uswds/modal/Modal.tsx
+++ b/user-interface/src/components/uswds/modal/Modal.tsx
@@ -99,9 +99,7 @@ function ModalComponent(props: ModalProps, ref: React.Ref<ModalRefType>) {
   }));
 
   useEffect(() => {
-    console.log('Outside If');
     if (isVisible && modalShellRef.current) {
-      console.log('Inside If for UseEffect');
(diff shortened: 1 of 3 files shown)


user-interface/src/components/uswds/modal/Modal.tsx
```diff
@@ -39,6 +41,9 @@ function ModalComponent(props: ModalProps, ref: React.Ref<ModalRefType>) {
     if (!props.forceAction) {
       if (e.key === 'Escape') {
         close(e);
+      } else if (e.key === 'Tab') {
+        //tabbed to button to accept or close modal when tabbed all the way through modal
+        return;
```
fmaddenflx, line 46: This can probably be removed, since we are now handling the tabs in the assignAttorneyModal rather than here.

jamesobrooks (approved):

PR #189, opened 2023-10-24
Cams 193 contain tab within modal

fmaddenflx:
# Purpose

Fix accessibility issue with tabbing out of the modal going back to the document.  

# Major Changes

We added event handlers to the modal close button and the modal body container that cause the tab stops to loop back to the first or last element in the modal, while the modal is open.

# Testing/Validation

Test coverage is already above 90%, barely. 

Testing has proven difficult for this... yet again.  It seems the testing libraries don't handle tabbing and focus events properly or at all in their jestdom.  I began a test, but it needs work to get it to pass.  I'm not sure we'll be able to successfully test these changes in this branch.

### DIFF
--- a/user-interface/src/components/uswds/modal/Modal.test.tsx
+++ b/user-interface/src/components/uswds/modal/Modal.test.tsx
@@ -43,7 +43,7 @@ describe('Test Modal component', () => {
               modalId={modalId}
               ref={modalRef}
               heading={'Test Heading'}
-              content={<button data-testid="test-content-button">Test Content</button>}
+              content={'Test Content'}
               actionButtonGroup={actionButtonGroup}
               onClose={closeModal}
               onOpen={onOpenModal}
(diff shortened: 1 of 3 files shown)


amorrow-flexion (approved):
Verified this functionality correctly tabbing through and back through with shift tab

user-interface/src/components/uswds/modal/Modal.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('Test Modal component', () => {
               modalId={modalId}
               ref={modalRef}
               heading={'Test Heading'}
-              content={'Test Content'}
+              content={<button data-testid="test-content-button">Test Content</button>}
```
jamesobrooks, line 46: Might want to revert this back to its original.

user-interface/src/components/uswds/modal/Modal.tsx
```diff
@@ -37,24 +37,53 @@ function ModalComponent(props: ModalProps, ref: React.Ref<ModalRefType>) {
...
+  function closeModal() {
     hide();
     if (props.onClose) {
       props.onClose();
     }
-    e.preventDefault();
+    setKeyboardAccessible(100000);
+  }
+
+  const closeButtonKeyDown = (ev: React.KeyboardEvent<HTMLButtonElement>) => {
```
amorrow-flexion, line 63: rename these methods to something like handleShiftTab**

PR #190, opened 2023-10-25
CAMS-192 Include retry logic in curl commands invoked in deployment script.

magile32:
Jira ticket: CAMS-192

# Purpose

Noticed some failed deployment due to curl command that did not make successful request the first try.

# Major Changes

- Update deployment scripts that used curl command to include retry options

# Testing/Validation

- Verified snippet of command locally with options set. Spoof url to simulate unreachable service.

# Notes


fmaddenflx (approved):
Looks good

PR #192, opened 2023-10-27
CAMS-139 add actions linter

jamesobrooks:
# Purpose

Add GitHub Action Workflow definition validation.

# Major Changes

Add the use of a `pre-commit` hook to act as a linter for our GitHub Actions Workflows.

# Testing/Validation

Tested several possible incorrect workflow definitions and validated that all GitHub Actions syntax violations were caught. A possible exception is that when the `|` is omitted from a multi-line `run` command, the linter does not identify this, but it will not work. I'm not sure if this is a GitHub Actions syntax violation per se, or if maybe it's more of a runner/shell issue.

The `pre-commit.ci` tool also runs this check, and [the successful check can be seen here](https://results.pre-commit.ci/run/github/590534939/1698373868.TVu5htrURAq1DXulqc2wvA).

> Check the commit messages for information about what passes the yaml check but fails this check.

# Notes

The linter does not necessarily catch everything, so care should always be taken to validate the correct behavior of the workflow.


.github/workflows/test-action-linter.yml
jamesobrooks, line 1: This is obviously not intended to be merged. You can see the history for a glimpse into what the linter can/cannot do, but mostly the git commit history contains error messages received with various bad syntax.

.github/workflows/continuous-deployment.yml
```diff
@@ -10,6 +10,7 @@ on:
       - "**/*.md"
     branches-ignore:
       - "mob/**"
+      - "CAMS-139-add-actions-linter"
```
jamesobrooks, line 13: This will be removed prior to merge.

jayaneti1: Good

jayaneti1 (commented):
Are you deleting the whole branch or only test-action-linter.yml?

jamesobrooks:
> Are you deleting the whole branch or only test-action-linter.yml?

Only the one file.

governmentSponsored (approved):

PR #202, opened 2023-11-01
CAMS-181 view debtor name

jamesobrooks:
# Purpose

We want to display the debtor name on the case detail screen.

# Major Changes

- Adds retrieval and display of the debtor's name on the case detail screen. Joint debtors are not included in this change.
- Extracts the getFullName function out to the common project so that we do not have identical code in the frontend and backend projects.
- Updates the test-data dev tool to include debtors in the creation of a case and a parties fixture to allow the addition of parties to existing cases.

# Testing/Validation

- Updated and added automated tests.
- Validated that names are displayed correctly whether the debtor is a company or an individual (with or without middle name).


mrinsin (approved):

dev-tools/test-data/tables/AO_PY.ts
```diff
@@ -0,0 +1,188 @@
...
+	[PY_FAX_PHONE] [varchar](20) NULL,
+	[PY_E_MAIL] [varchar](60) NULL,
+	[PY_PROSE] [char](1) NULL,
+	[PY_END_DATE] [date] NULL,
+	[SSN_EVENT] [char](1) NULL,
+	[NAME_EVENT] [char](1) NULL,
+	[ADDRESS_EVENT] [char](1) NULL
+) ON [PRIMARY]
+
+*/
```
fmaddenflx, line 33: We probably don't need this create statement in here.  It's commented out.

fmaddenflx (approved):
Approved.  There is one commented code block that can be removed.

PR #203, opened 2023-11-01
VER-11 Add content security policy

magile32:
# Problem

See issue VER-11 for details

# Solution

Define initial content-security-policy

# Testing/Validation

- Verify in branch deployment with included header updates in nginx.conf
- Verify finding resolved with Veracode DAST scan

# Other Changes


jamesobrooks:
@magile32, it might be a good idea to deploy this and ensure that the Jira issue collector still works when accessing from USTP laptops. From reading the [CWE definition](https://cwe.mitre.org/data/definitions/829.html), it seems that the goal is to prevent running of scripts from untrusted sources. I wonder if Jira's script `issueCollectorBootstrap.js` is considered trusted and will function properly.

magile32:
> @magile32, it might be a good idea to deploy this and ensure that the Jira issue collector still works when accessing from USTP laptops. From reading the [CWE definition](https://cwe.mitre.org/data/definitions/829.html), it seems that the goal is to prevent running of scripts from untrusted sources. I wonder if Jira's script `issueCollectorBootstrap.js` is considered trusted and will function properly.

Yes,  good point. We need to verify that the CSP doesn't block that.

magile32:
> > @magile32, it might be a good idea to deploy this and ensure that the Jira issue collector still works when accessing from USTP laptops. From reading the [CWE definition](https://cwe.mitre.org/data/definitions/829.html), it seems that the goal is to prevent running of scripts from untrusted sources. I wonder if Jira's script `issueCollectorBootstrap.js` is considered trusted and will function properly.
> 
> Yes, good point. We need to verify that the CSP doesn't block that.

Update. Issue collector script is failing the CSP with the following message:
`Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'"`

Will need to research on the solution for this one. We most likely don't want to use `unsafe-inline`. The other options are to use a hash or nonce to enable inline execution.

governmentSponsored (approved):

jamesobrooks (approved):

PR #243, opened 2023-11-08
CAMS-215 Corrected multiple assignment bug

amorrow-flexion:
Jira Ticket: CAMS-215

# Problem

Assign Attorney button on case would not pop modal on a second assignment

# Solution

Corrected the modal state to allow the modal after first assignment

# Testing/Validation

- ran application locally to validate changes
- ran tests to verify they still pass

### DIFF
--- a/user-interface/src/components/AssignAttorneyModal.tsx
+++ b/user-interface/src/components/AssignAttorneyModal.tsx
@@ -16,7 +16,6 @@ export interface AssignAttorneyModalProps {
   attorneyList: Attorney[];
   bCase: Chapter15Type | undefined;
   modalId: string;
-  openerId: string;
   callBack: (props: CallBackProps) => void;
 }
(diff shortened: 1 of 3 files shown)


user-interface/src/components/AssignAttorneyModal.tsx
```diff
@@ -16,7 +16,6 @@ export interface AssignAttorneyModalProps {
   attorneyList: Attorney[];
   bCase: Chapter15Type | undefined;
   modalId: string;
-  openerId: string;
```
fmaddenflx, line 19: I'd love to hear the rationale of removing openerId, and why that was causing the issue.  Have we thought through all of the possible effects of removing this or perhaps why it was put in in the first place?

fmaddenflx (approved):
Looks good.

PR #267, opened 2023-11-09
CAMS-208 judicial district

fmaddenflx:
# Purpose

Add Court Name (Judicial District) to Case Details screen

# Major Changes

Judicial District (Court Name) pulled in with a JOIN from AO_COURT.
Displayed on case details screen.  Screen slightly reformatted to fit Figma design.

# Testing/Validation

Tests updated and passing > 90%


btposey (approved):
I fixed the Pa11y test data so the accessibility test would pass.

PR #277, opened 2023-11-10
CAMS-213 type of debtor

fmaddenflx:
# Purpose

Add Debtor type label to Case Details screen under category of debtor.

# Major Changes

Backend API includes a hard coded Label look up map of codes to labels.
Dxtr Helper now includes a parseDebtorTypeLabel to grab the code out of the Transaction Record from AO_TX table and return a label based on the Label look up map.

# Testing/Validation

Testing for backend and front end added and coverage is above 90%


backend/functions/lib/adapters/gateways/debtor-type-gateway.ts
jamesobrooks, line 1: I've realized that we no longer talk about gateways in the OeSA workshop, but I don't see this as a gateway. In the _Clean Architecture_ book, gateways are described as `polymorphic interfaces that contain methods for every create, read, update, or delete operation that can be performed by the application on the database`. The way I've landed on describing them is that they are the interface between database semantics and application semantics. While what this is doing is certainly in the realm of my definition, it doesn't meet the CA definition. I would include this in either the cases gateway or maybe pull out a transactions gateway where it reads the code from the transaction and then just returns the label.

backend/functions/lib/adapters/gateways/dxtr/dxtr.gateway.helper.ts
```diff
@@ -30,3 +31,12 @@ export function parseTransactionDate(record: DxtrTransactionRecord): Date {
     parseInt(transactionDateDay),
   );
 }
+
+// Examples:
+// 1081201013220-10132            15JC
+// 1081231056523-10565            15IB00-0000000
+export function parseDebtorType(record: DxtrTransactionRecord): string {
+  const debtorTypeMatch = record.txRecord.match(/\d{13}-\d{5}\s+\d{2}\w{2}/);
```
jamesobrooks, line 39: Why do we not just take the two characters from their place in the fixed width string as we did with the transaction dates previously?

fmaddenflx: I just felt like this was a better way to grab the ID.  We can discuss.  Maybe it's because I didn't want to keep track of character columns.  I'm being lazy perhaps?

fmaddenflx: Now that I've given it more thought, I believe my original use of regex was because, different types of transactions seem to have very different record formats.  If all we do is grab characters at a certain location, we are likely to be grabbing the wrong information.

jamesobrooks: My main concern is with the relative strictness of the pattern. 13 digits, a dash, 5 digits, some amount of whitespace*, 2 digits, 2 word characters. With the exception of the whitespace, this is perhaps always correct for valid data. But is some of that whitespace placeholder? So is there any valid record in the real database that has some data in place of some of that whitespace?

jamesobrooks (approved):

PR #317, opened 2023-11-22
CAMS-222 Sort docket entries

btposey:
# Purpose

Sort docket entries by sequence number as a stand in for date.

# Major Changes

Added sort toggle button to the user case docket screen.
Sorted docket by the toggled sort direction.

# Testing/Validation

Added unit tests to confirm the sorting worked as expected.


governmentSponsored (approved):
code looks good from what I can tell. merged to stg, looks good there too.

PR #320, opened 2023-11-22
CAMS-250 null summary bug

btposey:
# Problem

The case docket screen does not render when a summary text is null in the docket

# Solution

Added null checks in the SQL to ensure text is always returned from the API.

# Testing/Validation

Added the data integrity issue to a case in the Flexion DXTR database to reproduce the error then tested to ensure the page renders as expected.


governmentSponsored (approved):
Tested this branch out in stg and it worked as expected.

PR #323, opened 2023-11-27
CAMS-294: Added Scroll to top button to every screen

fmaddenflx:
Jira ticket: CAMS-294

# Purpose

Add scroll-to-top button to case docket list

# Major Changes

Creation of ScrollToTopButton component and added functionality to App.tsx to display button when scrollTop of App is more than 100px and hide button when scrollTop is less than 100px.

Scroll button animates when it first appears.

# Testing/Validation

Tests were added to App.test.tsx to detect that the button is displaying correctly and that screen is scrolled to the top when button is clicked.

# Notes

Since I had to add events to the entire screen to detect when the whole screen was scrolled, I had to add this component to the App.tsx. Therefore, it's available on every screen.

### DIFF
--- a/user-interface/src/App.scss
+++ b/user-interface/src/App.scss
@@ -6,8 +6,20 @@
 @use 'styles/abstracts/colors';
 @use 'styles/abstracts/tables';
 
+body {
+  overflow-y: hidden;
+}
 .App {
+  position: absolute;
   padding: 0;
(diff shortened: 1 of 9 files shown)


btposey (approved):
The bounce on the return to top button is a fun touch.

amorrow-flexion (approved):

PR #324, opened 2023-11-28
CAMS-238 Added Scroll to top button to every screen

fmaddenflx:
# Purpose

Add scroll-to-top button to case docket list

# Major Changes

Creation of ScrollToTopButton component and added functionality to App.tsx to display button when scrollTop of App is more than 100px and hide button when scrollTop is less than 100px.

Scroll button animates when it first appears.

# Testing/Validation

Tests were added to App.test.tsx to detect that the button is displaying correctly and that screen is scrolled to the top when button is clicked.

# Notes

Since I had to add events to the entire screen to detect when the whole screen was scrolled, I had to add this component to the App.tsx. Therefore, it's available on every screen.


user-interface/src/case-detail/panels/CaseDetailCourtDocket.scss
```diff
@@ -1,25 +1,33 @@
 @import '../../styles/abstracts/colors';
 
 #case-detail-court-docket-panel {
-  .docket-entry {
-    padding-bottom: 2rem;
+  #searchable-docket {
```
fmaddenflx, line 4: Check.  I believe this can be reset since the button is no longer located on this screen and this section no longer scrolls.

user-interface/src/case-detail/panels/CaseDetailCourtDocket.tsx
```diff
@@ -54,90 +54,92 @@ export default function CaseDetailCourtDocket(props: CaseDetailCourtDocketProps)
...
-                    Sort ({sortDirection})
-                    <Icon
-                      className="sort-button-icon"
-                      name={sortDirection === 'Newest' ? 'arrow_upward' : 'arrow_downward'}
-                    />
-                  </div>
-                </button>
-              </div>
-            </section>
+    <>
```
fmaddenflx, line 57: This container is no longer needed due to the button being in App.tsx instead of this file.  You can probably reset this file to the last commit.

user-interface/src/lib/hooks/UseWindowSize.ts
```diff
@@ -0,0 +1,14 @@
...
+  useLayoutEffect(() => {
+    function updateWindowSize() {
+      setWindowSize([window.innerWidth, window.innerHeight]);
+    }
+    window.addEventListener('resize', updateWindowSize);
+    updateWindowSize();
+    return () => window.removeEventListener('resize', updateWindowSize);
+  }, []);
+  return windowSize;
+}
```
fmaddenflx, line 14: Stash somewhere, if it's needed at some point.  It doesn't need to be committed to the repo.

amorrow-flexion (approved):

PR #326, opened 2023-11-30
CAMS-246 view case header while scrolling

fmaddenflx:
# Purpose

Case Details Screen header, which contains the Case Title, number, location, and chapter number and type, should become fixed to the top of the screen whenever the screen is scrolled to the point where the header touches the top of the browser window.

Likewise the Case Detail Navigation menu at the left should become fixed.

The header and navigation should then go back to normal once the screen is scrolled down enough for the BLUE app header peaks out under the Case Details Screen Header.

# Major Changes

A lot... ?  Handling for screen manipulation via useEffect and watching scroll events to either render a new component to replace one on screen, or add css classes to components.  The animation and style changes are handled in SCSS files.

# Testing/Validation

Testing was added to ensure that components were properly replaced and/or css classes were properly added and removed after scrolling to various points on the screen.

# Notes

In the tests, we had to override getBoundingClientRect() for various elements, due to jsdom not updating the information properly.

### DIFF
--- a/user-interface/src/App.tsx
+++ b/user-interface/src/App.tsx
@@ -18,6 +18,7 @@ function App() {
   const bodyElement = document.querySelector('.App');
 
   function documentScroll(ev: React.UIEvent<HTMLElement>) {
+    console.log((ev.currentTarget as Element).scrollTop);
     if ((ev.currentTarget as Element).scrollTop > 100) {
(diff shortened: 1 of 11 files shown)


btposey (approved):
Reservations about the use of .d.ts files, but not going to hold up the PR.

amorrow-flexion (approved):

PR #329, opened 2023-12-01
CAMS-14 sql with managed identity

magile32:
# Purpose

Switch from using SQL auth to User Managed Identity to grant functionapp access to sql server database
# Major Changes

Describe what has changed.

- Update and refactoring existing database client and config logic to support connecting with managed identity. Also, improved on error logging.
- Update GHA deployment and biciep to provision managed identity for read access and assign to functionapp instance as a user identity
- Check for sql database access via healthcheck endpoint
- Additional unit tests added
- Documentation updated

# Testing/Validation

- Confirm database access via user managed identity from functionapp instance
- Verify automation of managed identity creation and assignment to functionapp
- Validate configuration in Flexion environment only

# Notes

- Manual step needed to grant access to managed identity. Because this is SQL server will need administrator with privileges to execute queries. Steps documented.


governmentSponsored (commented):
@magile32 am I correct in assuming that these would be steps to test on USTP side?
1. create the managed identity for sql
2. manually or programmatically add that value to kv
3. run that SQL script to add permissions to the DB

backend/functions/lib/configs/application-configuration.ts
```diff
@@ -30,40 +27,48 @@ export class ApplicationConfiguration implements AppConfig {
...
-        trustServerCertificate: Boolean(process.env.MSSQL_TRUST_UNSIGNED_CERT),
-      },
+    const server = process.env.MSSQL_HOST;
+    const port: number = Number(process.env.MSSQL_PORT) || 1433;
+    const encrypt: boolean = Boolean(process.env.MSSQL_ENCRYPT);
+    const trustServerCertificate: boolean = Boolean(process.env.MSSQL_TRUST_UNSIGNED_CERT);
+    const type = process.env.MSSQL_AUTH_TYPE || 'azure-active-directory-default';
+    const user = process.env.MSSQL_USER;
+    const password = process.env.MSSQL_PASS;
+    const clientId = process.env.MSSQL_CLIENT_ID; // User Identity client
```
jamesobrooks, line 37: I would prefer to use a more meaningful variable name than have to include the comment describing what it is.

docs/running.md
```diff
@@ -116,6 +118,24 @@ To interact with the Cosmos database from your local machine you will need to se
...
+
+Existing Bicep deployment can automate the creation of the user managed identity and assign the identity to the functionapp instance. The below are some manual steps to handle.
+
+Grant access to a managed identity with the following sql query
+```sql
+CREATE USER [userAssignedIdentityName] FROM EXTERNAL PROVIDER;
+ALTER ROLE db_datareader ADD MEMBER [userAssignedIdentityName];
+// Remember to leverage principle of least privilege
+// ALTER ROLE db_datawriter ADD MEMBER [userAssignedIdentityName];
+// ALTER ROLE db_ddladmin ADD MEMBER [userAssignedIdentityName];
```
jamesobrooks, line 133: Are these intended to be comments? If so, `//` should be replaced with `--`.

backend/functions/lib/configs/application-configuration.ts
jamesobrooks: As Fritz noted, `type` is probably also not a good choice for variable name because that is a reserved word in TypeScript. So maybe something like `authType` and `userIdentityClientId`.

jamesobrooks (approved):

PR #343, opened 2023-12-05
CAMS-223 filter docket

magile32:
# Purpose

Add mechanism for filtering list of dockets by summary text.

# Major Changes

- Added MultiSelect component used to filter docket list
- Include new react-select dependency
- Override default react-select component style
- Moved sort, search and filter ui components to the left sidebar
- Updated pa11y test and include automated unit tests

# Testing/Validation

- Maintain unit test coverage above 90%
- pa11y test pass
- Manual inspection of web application

# Notes

### DIFF
--- a/user-interface/src/case-detail/CaseDetailScreen.test.tsx
+++ b/user-interface/src/case-detail/CaseDetailScreen.test.tsx
@@ -4,6 +4,7 @@ import { CaseDetail } from './CaseDetailScreen';
 import { getCaseNumber } from '@/lib/utils/formatCaseNumber';
 import { CaseDetailType, Debtor, DebtorAttorney } from '@/lib/type-declarations/chapter-15';
 import { BrowserRouter, MemoryRouter } from 'react-router-dom';
+import React from 'react';
 
 const caseId = '101-23-12345';
 const brianWilsonName = 'Brian Wilson';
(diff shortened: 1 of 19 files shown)


user-interface/src/case-detail/CaseDetailScreen.test.tsx
```diff
@@ -666,4 +680,364 @@ describe('Case Detail screen tests', () => {
...
+      expect(fn(left, right)).toEqual(expectedValue);
+    });
+    test('should return the expected sort direction for Oldest sort', () => {
+      const fn = docketSorterClosure('Oldest');
+      const expectedValue = -1;
+      expect(fn(left, right)).toEqual(expectedValue);
+    });
+  });
+
+  describe('old tests we may have to rewrite', () => {
```
jamesobrooks, line 780: Now that we're done with these tests, we should rename the describe. 
```suggestion
  describe('sort, search, and filter tests', () => {
```

user-interface/src/case-detail/panels/CaseDetailNavigation.test.tsx
```diff
@@ -44,29 +43,4 @@ describe('Navigation tests', () => {
 
     expect(result).toEqual(NavState.COURT_DOCKET);
   });
-
-  test('should render with css class name of "case-details-navigation fixed" when ref.fix() is called and "case-details-navigation" when ref.loosen() is called', async () => {
```
jamesobrooks, line 48: I'm checking now to make sure we put a similar test in the CaseDetailScreen tests.

jamesobrooks: The test `should fix when scrolled at top of viewport` on line 708 of CaseDetailScreen.test.tsx may be sufficient, but it doesn't have the same approach. @fmaddenflx, do you think we should have a test like this that calls the `fix()` and `loosen()` functions?

user-interface/src/lib/components/MultiSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+export interface MultiSelectProps {
+  onChange?: (newValue: MultiSelectOptionList<Record<string, string>>) => void;
+  className?: string;
+  closeMenuOnSelect?: boolean;
+  options?: Record<string, string>[];
+  label: string;
+}
+
+export default function MultiSelect(props: MultiSelectProps) {
+  //
```
jamesobrooks, line 17:
```suggestion
 ```

jamesobrooks (approved):

PR #400, opened 2023-12-13
CAMS-278 clear all filters

fmaddenflx:
# Purpose

Add a Clear-All button to allow the user to clear all filters.  Filter fields should visually clear and results should be displayed unfilterd.

# Major Changes

Added button and on-click to clear all fields.
Stateful values in caseDetailsScreen.tsx are cleared.
Each form component needed a clearValue() method to clear it's own state.
clearValue() methods were created for each element and added to a ref which was then exposed to the parent component.
After clearing caseDetailsScreen.tsx component state for each form element, ref.current?.clearValue() is called for each form component.

# Testing/Validation

A test was written for the case details screen component to click the button and clear all fields.

test coverage is between 93 and 94 % depending on the coverage type.


user-interface/src/case-detail/CaseDetailScreen.tsx
```diff
@@ -179,18 +181,18 @@ export const CaseDetail = (props: CaseDetailProps) => {
   const [documentNumber, setDocumentNumber] = useState<number | null>(null);
   const [sortDirection, setSortDirection] = useState<SortDirection>('Newest');
   const leftNavContainerRef = useRef<CaseDetailScrollPanelRef>(null);
-
+  // const [clearAllFilters, setClearAllFilters] = useState<boolean>(false);
```
jamesobrooks, line 184: Can we remove this?

user-interface/src/case-detail/CaseDetailScreen.tsx
```diff
@@ -240,6 +242,17 @@ export const CaseDetail = (props: CaseDetailProps) => {
     }
     setDocumentNumber(newDocumentNumber);
   }
+  function clearFilters() {
```
jamesobrooks, line 245: Nit-pick, but could we add a newline above?
```suggestion

  function clearFilters() {
```

user-interface/src/lib/components/MultiSelect.tsx
```diff
@@ -11,9 +14,13 @@ export interface MultiSelectProps {
   closeMenuOnSelect?: boolean;
   options?: Record<string, string>[];
   label: string;
+  id: string;
 }
 
-export default function MultiSelect(props: MultiSelectProps) {
+function MultiSelectComponent(props: MultiSelectProps, ref: React.Ref<MultiSelectRef>) {
+  //const multiSelectRef =
+  //  React.useRef<SelectInstance<MultiSelectProps, true, GroupBase<MultiSelectProps>>>();
```
jamesobrooks, line 22: Can we remove these?
```suggestion
```

user-interface/src/lib/components/uswds/date-range-picker.d.ts
jamesobrooks, line 1: From an OeSA standpoint—specifically dependency and invasive species rules—I'm thinking that interfaces like these probably should go somewhere outside the `uswds` directory. The USWDS implementation is the _less important_ code compared to the component using it. I'd be happy to discuss.

user-interface/src/lib/components/uswds/icon-input.d.ts
jamesobrooks, line 1: See the comment on `date-range-picker.d.ts`.

user-interface/src/case-detail/CaseDetailScreen.tsx
jamesobrooks:
```suggestion
```

user-interface/src/lib/components/uswds/date-range-picker.d.ts
```diff
@@ -0,0 +1,3 @@
+export interface DateRangePickerRef {
```
amorrow-flexion, line 1: Since we are doing the same thing on each input field, would it be better to move all these refs into a single file? like input-fields.d.ts. 

Unless we think we will be expanding on these individually I think we should combine them

fmaddenflx: More than that, we could actually use the same interface for each of the refs, if they are all the same.  Then later if we ever need something more specific for one of these components, then we can create one specific for *that* component.

user-interface/src/case-detail/CaseDetailScreen.tsx
fmaddenflx: Yes I'll do that now.

user-interface/src/lib/components/MultiSelect.tsx
fmaddenflx: Yes I'll do that now.

user-interface/src/lib/components/uswds/date-range-picker.d.ts
fmaddenflx: Let's discuss

user-interface/src/lib/components/uswds/date-range-picker.d.ts
fmaddenflx: Let's discuss

user-interface/src/lib/components/MultiSelect.tsx
```diff
@@ -68,6 +75,21 @@ export default function MultiSelect(props: MultiSelectProps) {
     }),
   };
 
+  function clearValue() {
+    if (
+      multiSelectRef.current &&
+      Object.prototype.hasOwnProperty.call(multiSelectRef.current, 'clearValue')
```
btposey, line 81: Since `hasOwnProperty` is on the object prototype we should be able to shorten this to `multiSelectRef.current.hasOwnProperty('clearValue')`. I believe we can also test if the clearValue function exists with a truthy expression like `&& multiSelectRef.current.clearValue`.

btposey (changes requested):

btposey (approved):

PR #403, opened 2023-12-14
CAMS-90 Edit case assignment

btposey:
# Purpose

Assign and un-assign attorneys from cases.

# Major Changes

* Updated assignment document model to include timestamps.
* Added update assignment capability to assignment repo.
*  Added logic to add and remove assignments to cases.

# Testing/Validation

Increased lacking test coverage for backend tests related case assignment.


backend/functions/lib/adapters/gateways/case.assignment.cosmosdb.repository.ts
```diff
@@ -51,22 +52,54 @@ export class CaseAssignmentCosmosDbRepository implements CaseAssignmentRepositor
...
+  }
+
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   getAssignment(assignmentId: string): Promise<CaseAttorneyAssignment> {
     throw new Error('Method not implemented.');
   }
 
   async findAssignmentsByCaseId(caseId: string): Promise<CaseAttorneyAssignment[]> {
+    let query = '';
+    query = 'SELECT * FROM c WHERE c.caseId = @caseId AND NOT IS_DEFINED(c.unassignedOn)';
```
magile32, line 91:
```suggestion
    const query = 'SELECT * FROM c WHERE c.caseId = @caseId AND NOT IS_DEFINED(c.unassignedOn)';
```
`let` was used before because there was some conditional statements to update the query. We could refactor this way since conditional been remoted.

backend/functions/lib/adapters/gateways/case.assignment.local.repository.test.ts
```diff
@@ -1,23 +1,133 @@
...
+    const caseId = '123';
+    const currentDate = new Date().toISOString();
+    const assignment = {
+      caseId,
+      name: 'Susan Arbeit',
+      role: CaseAssignmentRole['TrialAttorney'],
+      assignedOn: currentDate,
+    };
+
+    // First create a record to update.
```
jamesobrooks, line 101: Looks like we copied this comment from above but it doesn't mean the same thing here.

backend/functions/lib/adapters/gateways/case.assignment.local.repository.test.ts
```diff
@@ -1,23 +1,133 @@
...
```
jamesobrooks, line 125: Looks like we copied this comment from above but it doesn't mean the same thing here.

backend/functions/lib/adapters/gateways/case.assignment.local.repository.test.ts
```diff
@@ -1,23 +1,133 @@
...
+      new UnknownError('LOCAL-ASSIGNMENT-REPOSITORY', {
+        message:
+          'Unable to update assignment. Please try again later. If the problem persists, please contact USTP support.',
+        originalError: new Error('Can not find record'),
+        status: 500,
+      }),
+    );
+  });
+
+  test('Should return assignment by id', async () => {
```
jamesobrooks, line 86:
```suggestion
  test('Should return assignment by case id', async () => {
```
I'm just thinking we can avoid future confusion about which id it should be found by.

backend/functions/lib/adapters/gateways/case.assignment.local.repository.ts
```diff
@@ -32,7 +51,11 @@ export class CaseAssignmentLocalRepository implements CaseAssignmentRepositoryIn
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
jamesobrooks, line 53:
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

backend/functions/lib/adapters/types/case.assignment.role.ts
```diff
@@ -1,3 +1,3 @@
-export enum CaseAssignmentRole {
-  TrialAttorney = 1,
-}
+export const CaseAssignmentRole = {
+  TrialAttorney: 'TrialAttorney',
+} as const;
```
jamesobrooks, line 3: I like it!

backend/functions/lib/adapters/types/case.attorney.assignment.ts
```diff
@@ -1,14 +1,8 @@
...
-  role: CaseAssignmentRole;
-
-  constructor(caseId: string, name: string, role: string) {
-    this.caseId = caseId;
-    this.name = name;
-    this.role = CaseAssignmentRole[role];
-  }
+  role: string;
+  assignedOn: string;
+  unassignedOn?: string;
```
jamesobrooks, line 7: I like this too!

backend/functions/lib/cosmos-humble-objects/fake.cosmos-client-humble.ts
```diff
@@ -28,6 +29,24 @@ export default class FakeCosmosClientHumble {
...
+              }
+              assignment.id = `assignment-id-${Math.round(Math.random() * 1000)}`;
+              this.caseAssignments.push(assignment);
+              return {
+                item: {
+                  ...assignment,
+                },
+              };
+            },
+            replace: (assignment: CaseAttorneyAssignment) => {
```
jamesobrooks, line 43: Is there a replace on `items`? I don't see one in the library.

backend/functions/lib/use-cases/case.assignment.ts
```diff
@@ -4,72 +4,81 @@ import { CaseAttorneyAssignment } from '../adapters/types/case.attorney.assignme
...
-      return existingAssignments.length === 0;
-    } catch (e) {
-      throw new AssignmentError(MODULE_NAME, {
-        message:
-          'Unable to determine whether assignments already exist. Please try again later. If the problem persists, please contact USTP support.',
-        originalError: e,
-        status: 500,
+    log.info(applicationContext, MODULE_NAME, 'New assignments:', newAssignments);
+
+    // Add any attorney from the new assignment list to the case that is not already assigned.
```
jamesobrooks, line 56: Do we feel we need this comment? If so, maybe we should just extract the relevant code into a named function so that we don't need a comment.

user-interface/src/case-assignment/AssignAttorneyModal.tsx
```diff
@@ -71,25 +89,65 @@ function AssignAttorneyModalComponent(
...
-    let localCheckListValues = checkListValues;
+    let localCheckListValues = [...checkListValues];
     if (ev.target.checked && !checkListValues.includes(name)) {
       localCheckListValues.push(name);
     } else if (!ev.target.checked && checkListValues.includes(name)) {
       localCheckListValues = checkListValues.filter((theName) => theName !== name);
     }
-    modalRef.current?.buttons?.current?.disableSubmitButton(localCheckListValues.length === 0);
+    // TODO: This should look for a difference between the current selection list
+    // compare bCase.assignments to localCheckListValues
```
jamesobrooks, line 143: I believe the todo is todone. Also, do we need the comment?

user-interface/src/case-assignment/AssignAttorneyModal.tsx
```diff
@@ -129,7 +187,7 @@ function AssignAttorneyModalComponent(
         // that is system wide and always displays in the same place.  Probably an alert triggered by
```
jamesobrooks, line 187: I know we didn't add it here, but is this necessary?

mrinsin (approved):
Looks good to me! My only nit would be I see comments here and there, not sure if we they're something we want keep around or clean up

user-interface/src/lib/components/uswds/DateRangePicker.tsx
```diff
@@ -27,6 +27,7 @@ function DateRangePickerComponent(props: DateRangePickerProps, ref: React.Ref<In
   const [startDateValue, setStartDateValue] = useState<string | undefined>(undefined);
   const [endDateValue, setEndDateValue] = useState<string | undefined>(undefined);
 
+  // set timeout added to solve some funky weirdness with the date type input handling keyboard events after a reset.
```
jamesobrooks, line 30: Is this like a debounce issue? If so, maybe we could define a debounce function so it's clear why we're doing the setTimeout.

jamesobrooks (approved):

PR #425, opened 2023-12-29
CAMS-262 fix tests

jamesobrooks:
# Purpose

Tests were hanging in the CI/CD pipeline.

# Major Changes

We remove `userEvent` and `fireEvent` actions from `act()` and `waitFor()`. `act()` is unnecessary because `testing-library` already wraps `render` and these events in `act()`. We suspect the actions being triggered from within `waitFor()` was causing a memory leak.

We also found after this change that the ScrollToTopButton test failed because the way the button was scrolling to the top was not triggering the `onScroll` function during tests. We now trigger an event on the app rather than directly editing the scroll position.

# Testing/Validation

We ran the tests locally several ways and the failure—which had become constant—went away in the CI/CD pipeline.

### DIFF
--- a/user-interface/src/App.test.tsx
+++ b/user-interface/src/App.test.tsx
@@ -11,13 +11,19 @@ describe('App', () => {
     );
 
     const app = document.querySelector('.App');
-    expect(app).not.toHaveClass('header-scrolled-out');
+    await waitFor(() => {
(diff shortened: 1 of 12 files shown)


fmaddenflx (commented):
Looks good.

PR #431, opened 2024-01-04
CAMS-160 zero downtime deployment

amorrow-flexion:
# Purpose

Implement deployment slots within Azure to allow for seemingly zero downtime deployments

# Major Changes

- implemented slot creation and deployment functionality within pipeline
- added 'staging' slot testing functionality prior to swap
- added 'Production' slot testing functionality post swap
- modified healthcheck output service ibfo
- updated documentation
- added/modified deployment scripts to allow slot deployments
- updated location of scripts to follow screaming architecture
- added deployment parameter for workflow_dispatch in GHA
- Updated backend api healthcheck to return git sha to determine version of code deployed
- Include meta tag to include git sha to determine version of code deployed

# Testing/Validation

- deployed slots through pipeline 
- allowed testing to complete on the slots and verified slot endpoint was hit prior to switching
- deployed without slot to verify existing deployment works
- post switch testing completed on 'Production' slot
- Validated full deployment and slot deployment in Flexion environment
- Validated slot deployment workflow within the USTP environment

# Notes

- We will need to make heavy updates to the USTP environment to allow slots to work prior to merging this
- we will need to update the security sast scan baseline file
- The baseline file for the sast scan will need updated, it is just a line change for one of the existing errors CWE-201 cases.integration.tests.tsx line 13 moved to line 16
- Need to verify that updates to main branch will always trigger a slot deployment. To be done after merging PR to main.


.github/workflows/continuous-deployment.yml
```diff
@@ -345,6 +369,7 @@ jobs:
     outputs:
       cosmosDbClientId: ${{ steps.azure-db-deploy.outputs.cosmosDbClientId }}
       cosmosDbPrincipalId: ${{ steps.azure-db-deploy.outputs.cosmosDbPrincipalId }}
+      vnetName: ${{ steps.azure-db-deploy.outputs.cosmosDbPrincipalId }}
```
magile32, line 372: This doesn't seem right. Do we need this line?

backend/functions/cases/cases.integration.test.ts
```diff
@@ -4,13 +4,16 @@ dotenv.config();
...
   beforeAll(() => {
     functionUrl = process.env.CASES_FUNCTION_URL;
+    slotName = process.env.SLOT_NAME ?? 'self';
   });
 
   test('cases azure function should return success when retrieving cases', async () => {
     let casesResponse;
-    await fetch(`${functionUrl}/cases`, {
+    const targetURL = `${functionUrl}/cases?x-ms-routing-name=${slotName}`;
+    await fetch(targetURL, {
```
magile32, line 16: This gets flagged by Veracode, a possible refactor is to default to self and check a boolean to determine if the slot deployment environment is not production. If not explicitly pass in the slotName as staging.

.github/workflows/continuous-deployment.yml
amorrow-flexion: This can be removed. Fortunately, we use the ref for vnetName correctly within this job when creating slot resources for the webapp. On line 455

docs/operations/slots.md
```diff
@@ -0,0 +1,40 @@
...
+- Add a header to all requests during validation. An Azure rep indicated that this would be an option. This would
+involve setting up something very similar to a proxy to capture requests and add a header, or changing code to add
+the header, but we would have to have some way to not do so from the users' machines.
+- Use a `x-ms-routing-name` cookie with the value set to the name of the staged slot (i.e. `staging`). This seems
+to be a viable option. A request to `https://api-hostname.azurewebsites.us/?x-ms-routing-name=staging` will trigger
+the creation of a cookie so subsequent requests will be routed to the staged API. This does require that at least
+temporarily we set up the routing to distribute some traffic to the staged slot. This can be done by running:
+```shell
+az webapp traffic-routing set --distribution staging=1 --name <func app name> --resource-group <resource group name>
+## This same command with staging=0 can be run after and the effects should persist across deployments according to Azure rep
```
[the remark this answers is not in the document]
amorrow-flexion: Okay, so we can change the value to 0, then once the swap is completed we clear traffic-routing rules. Is that correct?

backend/functions/cases/cases.integration.test.ts
magile32: Tried several refactoring but the problem seems to be related to the functionUrl that needs to be passed in. @amorrow-flexion mention that this was an existing known Veracode finding that was ignored. Baseline needs to be updated.

docs/operations/slots.md
```diff
@@ -0,0 +1,40 @@
...
+### Options
+
+There are many options for how to ensure the validation of the staged client is interacting with the staged API.
+They generally involve tradeoffs so determining which method to use is not simple.
+
+- Get the hostname from configuration at runtime. This involves providing some way for the client to get a
+configuration value, and providing a different value only for use in validation. The client runs in browsers
+on users' machines, so the configuration value would need to come from a call to some API and only have value
+during deployment.
+- Proxy requests in our validation runner (GitHub Actions or Azure DevOps) to hit the staged API instead of the
```
amorrow-flexion, line 25: We may want to move this into an ADR as considerations vs implementation.

magile32 (approved):

jamesobrooks (approved):

PR #442, opened 2024-01-11
CAMS-263

jamesobrooks:
# Purpose

The order sync function runs in USTP but does not seem to be retrieving any order from DXTR.

# Major Changes

- Adds logging so that we can see more about what is going on when we query DXTR for orders and documents
- Gets all documents for the orders we are looking at

# Testing/Validation

Automated testing is already in place.

# Other Changes

Adds brackets and dbo to table names in the queries for consistency.


backend/functions/lib/adapters/gateways/dxtr/orders.dxtr.gateway.ts
```diff
@@ -58,42 +92,21 @@ export class DxtrOrdersGateway implements OrdersGateway {
           return rawOrder satisfies Order;
         })
         .sort(dxtrOrdersSorter);
+      context.logger.info(
```
magile32, line 95: This is getting flag by Veracode. Using the log service should resolve this
`import log from '../lib/adapters/services/logger.service';`

amorrow-flexion (approved):

PR #455, opened 2024-01-19
Cams 300 reject order

fmaddenflx:
# Purpose

User should be able to reject a transfer order by clicking on red Reject button.  Confirmation modal should display with an optional "Reason for rejection" field, which they can submit.

Orders screen should update indicating that the order was rejected and form should be unavailable on the order in the accordion.

# Major Changes

Added user interface elements, button, modal, etc.
Also added a confirmation modal for Approval process.

# Testing/Validation

Tests have been updated to ensure that coverage stays above 90%.

### DIFF
--- a/user-interface/src/review-orders/TransferOrderAccordion.tsx
+++ b/user-interface/src/review-orders/TransferOrderAccordion.tsx
@@ -44,11 +44,10 @@ export function isValidOrderTransfer(transfer: OrderTransfer) {
 }
 
 function safeToInt(s: string) {
-  try {
-    return parseInt(s).toString();
-  } catch {
-    return s;
-  }
+  const intVal = parseInt(s);
+  if (isNaN(intVal)) return s;
+
+  return intVal.toString();
 }
 
 export function validateNewCaseIdInput(ev: React.ChangeEvent<HTMLInputElement>) {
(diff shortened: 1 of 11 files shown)


user-interface/src/review-orders/TransferOrderAccordion.tsx
```diff
@@ -158,186 +184,342 @@ export function TransferOrderAccordion(props: TransferOrderAccordionProps) {
...
+            timeOut: 8,
+          },
+          {
+            ...order,
+            ...rejection,
+          },
+        );
+      })
+      .catch((reason) => {
+        // TODO: make the error message more meaningful
```
amorrow-flexion, line 220: What do we need to do to resolve this TODO block?

user-interface/src/review-orders/TransferOrderAccordion.tsx
```diff
@@ -158,186 +184,342 @@ export function TransferOrderAccordion(props: TransferOrderAccordionProps) {
...
+                fromCaseId={order.caseId}
+                toCaseId={orderTransfer.newCaseId}
+                fromDivisionName={order.courtDivisionName}
+                toDivisionName={orderTransfer.newCourtDivisionName}
+                fromCourtName={order.courtName}
+                toCourtName={orderTransfer.newCourtName}
+                onCancel={cancelUpdate}
+                onConfirm={confirmAction}
+              ></ConfirmationModal>
+              {/* <ConfirmationModal
```
amorrow-flexion, line 490: Still need to get rid of this commented out modal

user-interface/src/review-orders/TransferOrderAccordion.tsx
```diff
@@ -367,3 +549,157 @@ function CaseSelection(props: CaseSelectionProps) {
     </>
   );
 }
+
+interface ConfirmationModalProps {
```
amorrow-flexion, line 553: We should probably extract the Confirmation Modal into its own file

jamesobrooks: I'm okay with it being here for now, but yeah, we'll almost certainly end up with this being extracted at some point in the future.

jamesobrooks (approved):

PR #486, opened 2024-02-02
CAMS-324: Fix same court exclusion

jamesobrooks:
# Purpose

We are incorrectly excluding cases transferred to another division within the same court.

# Major Changes

Modify the exclusion condition to look for both court and division to be the same.

# Testing/Validation

Ran query in USTP environment against the case where we saw the issue.


fmaddenflx (approved):
looks good.

btposey (approved):

PR #502, opened 2024-02-14
CAMS-332 Fix RangeError on docket search

magile32:
# Purpose

Application crashed when entering in value for docket search in case details view for cases with large number of docket entries. Investigation shows approximately 9k+ docket entries for this particular case and webapp throws a **RangeError** when invoking `handleHighlight`

# Major Changes

- Catch error to prevent application to break
- Set a min limit before invoking handleHighlight
- Resolve warning related to IconInput

# Testing/Validation

- Seed database with large number of dockets for case id 96-45961 in test database and confirm that problem can be recreated
- Manually verify fix in deployed webapp 

# Notes

- Noticed degraded performance with entering text into the search field for cases with large number of dockets. Min search text size is set to 3. This should alleviate the webapp from crashing and a way to improve on this with this fix is to increase the min text size.

### DIFF
--- a/user-interface/src/case-detail/panels/CaseDetailCourtDocket.tsx
+++ b/user-interface/src/case-detail/panels/CaseDetailCourtDocket.tsx
@@ -31,7 +31,7 @@ export default function CaseDetailCourtDocket(props: CaseDetailCourtDocketProps)
 
   const [alertOptions, setAlertOptions] = useState<AlertOptions | null>(null);
 
-  const minSearchTextSize = 3;
+  const MINIMUM_SEARCH_CHARACTERS = 3;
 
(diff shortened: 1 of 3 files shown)


magile32:
The **security-sast-scan** started failing as of recently. Note this change is not the root cause of the finding. Looks like Veracode policy might have been updated for scans going forward.

btposey (changes requested):
I think we can do a better job by hardening the handleHightlight function against failure. I think that would also include moving the minimum search string length into that function so the logic is reusable elsewhere we may use the highlight API in the future. Reach out to me and we can collaborate on this.

jayaneti1 (approved):

PR #533, opened 2024-03-18
CAMS-269 Fix grid layout for transfer accordian buttons

magile32:
# Purpose

Noticed issue with how the `Cancel` and `Approve` buttons get overlay during implementation of test using Cypress. The right gutter column overlay the Approve button and prevented it being clicked by the test framework.

# Major Changes

- Modified the grid layout

# Testing/Validation

- Manual inspection
- Confirm passing Cypress test after change

# Notes

### DIFF
--- a/user-interface/src/data-verification/TransferOrderAccordion.scss
+++ b/user-interface/src/data-verification/TransferOrderAccordion.scss
@@ -1,4 +1,8 @@
 .accordion-content {
+  div.align-right {
+    text-align: right; // aligns content in div to the right
+  }
+
   .transfer-text {
     padding: 1rem 0;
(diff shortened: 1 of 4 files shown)


amorrow-flexion (approved):

user-interface/src/data-verification/TransferOrderAccordion.scss
```diff
@@ -1,4 +1,8 @@
 .accordion-content {
+  div.align-right {
+    text-align: right; // aligns content in div to the right
```
jamesobrooks, line 3: Should we be using `text-align` for the alignment of React components?

fmaddenflx (approved):

PR #542, opened 2024-03-20
CAMS-272 consolidation type

fmaddenflx:
# Purpose

Added Consolidation Type to Consolidation Modal, API, and Cosmos

# Testing/Validation

Over 90% test coverage

### DIFF
--- a/user-interface/src/data-verification/ConsolidationOrderAccordion.tsx
+++ b/user-interface/src/data-verification/ConsolidationOrderAccordion.tsx
@@ -9,7 +9,6 @@ import {
   ConsolidationOrder,
   ConsolidationOrderActionApproval,
   ConsolidationOrderCase,
-  OrderStatus,
 } from '@common/cams/orders';
 import Button, { ButtonRef, UswdsButtonStyle } from '@/lib/components/uswds/Button';
 import SearchableSelect, { SearchableSelectOption } from '@/lib/components/SearchableSelect';
@@ -21,6 +20,7 @@ import { AttorneyInfo } from '@/lib/type-declarations/attorneys';
 import {
   ConsolidationOrderModal,
   ConfirmationModalImperative,
+  ConfirmActionResults,
 } from '@/data-verification/ConsolidationOrderModal';
 import useFeatureFlags, { CONSOLIDATIONS_ADD_CASE_ENABLED } from '@/lib/hooks/UseFeatureFlags';
 import Alert, { UswdsAlertStyle } from '@/lib/components/uswds/Alert';
@@ -93,14 +93,15 @@ export function ConsolidationOrderAccordion(props: ConsolidationOrderAccordionPr
     approveButtonRef.current?.disableButton(true);
   }
 
-  function confirmAction(status: OrderStatus, _reason: string = '', leadCaseNumber?: string): void {
+  function confirmAction({ status, leadCaseId, consolidationType }: ConfirmActionResults): void {
     if (status === 'approved') {
       const data: ConsolidationOrderActionApproval = {
         ...order,
+        consolidationType,
         approvedCases: selectedCases
           .map((bCase) => bCase.caseId)
-          .filter((caseId) => caseId !== leadCaseNumber),
-        leadCase: order.childCases.find((bCase) => bCase.caseId === leadCaseNumber)!,
+          .filter((caseId) => caseId !== leadCaseId),
+        leadCase: order.childCases.find((bCase) => bCase.caseId === leadCaseId)!,
       };
 
       api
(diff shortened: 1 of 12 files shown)


user-interface/src/data-verification/ConsolidationOrderModal.test.tsx
```diff
@@ -101,17 +101,29 @@ describe('ConsolidationOrderModalComponent', () => {
...
+
+    fireEvent.click(radioSubstantiveClickTarget!);
 
-    fireEvent.click(radioSubstantive!);
-    expect(radioSubstantive).toBeChecked();
-    expect(radioAdministrative).not.toBeChecked();
+    // await waitFor(() => {
+    //   expect(radioSubstantive).toBeChecked();
+    //   expect(radioAdministrative).not.toBeChecked();
+    // });
```
fmaddenflx, line 126: Probably remove these lines.

user-interface/src/data-verification/ConsolidationOrderModal.tsx
```diff
@@ -66,11 +80,14 @@ function ConsolidationOrderModalComponent(
...
-          reasonRef.current?.value,
-          `${leadCaseDivisionCode}-${leadCaseNumber}`,
-        );
+        onConfirm({
+          status: options.status,
+          rejectionReason: reasonRef.current?.value,
+          leadCaseId: `${leadCaseDivisionCode}-${leadCaseNumber}`,
+          // onConfirm should never be called unless the button is enabled.
+          // The button should never be enabled unless a consolidationType is selected
+          consolidationType: consolidationType as ConsolidationType,
```
btposey, line 89: Maybe use a ! operator rather than using the 'as' keyword.

user-interface/src/data-verification/ConsolidationOrderAccordion.tsx
```diff
@@ -86,21 +81,23 @@ export function ConsolidationOrderAccordion(props: ConsolidationOrderAccordionPr
     throw new Error('Function not implemented.');
   }
 
-  function cancelUpdate(): void {
-    courtSelectionRef.current?.clearValue();
-    caseIdRef.current?.resetValue();
-    clearIncludedCases();
+  function clearInputs(): void {
+    //courtSelectionRef.current?.clearValue();
```
amorrow-flexion, line 85: remove these?

user-interface/src/data-verification/ConsolidationOrderModal.test.tsx
```diff
@@ -101,17 +101,29 @@ describe('ConsolidationOrderModalComponent', () => {
...
-    expect(radioAdministrative).toBeChecked();
-    expect(radioSubstantive).not.toBeChecked();
+    fireEvent.click(radioAdministrativeClickTarget!);
+
+    await waitFor(() => {
+      expect(radioAdministrative).toBeChecked();
+      expect(radioSubstantive).not.toBeChecked();
+    });
+
+    fireEvent.click(radioSubstantiveClickTarget!);
```
jamesobrooks, line 121: I would suggest picking between 114 and 121 and removing the rest of these lines. Note that the expect is currently written in favor of line 121.

user-interface/src/data-verification/ConsolidationOrderModal.test.tsx
```diff
@@ -140,7 +152,13 @@ describe('ConsolidationOrderModalComponent', () => {
     await waitFor(() => {
       expect(modal).toHaveClass('is-hidden');
     });
-    expect(onConfirmSpy).toHaveBeenCalled();
+
+    expect(onConfirmSpy).toHaveBeenCalledWith({
+      status: 'approved',
+      courtDivision: undefined,
+      leadCaseId: `${courts[0].courtDivision}-11-11111`,
```
jamesobrooks, line 159: I think we should use some better constants in this test. We've defined `11-11111` in an array and then again in a const, but then use it again here literally.

user-interface/src/data-verification/ConsolidationOrderModal.tsx
```diff
@@ -104,20 +121,36 @@ function ConsolidationOrderModalComponent(
...
-      modalRef.current?.buttons?.current?.disableSubmitButton(false);
     } else {
-      modalRef.current?.buttons?.current?.disableSubmitButton(true);
+      setLeadCaseNumber('');
     }
   }
 
+  useEffect(() => {
+    modalRef.current?.buttons?.current?.disableSubmitButton(
+      consolidationType === null || leadCaseNumber.length !== 8,
```
jamesobrooks, line 150: We don't need to necessarily, but if we're doing something like this length check in other places we might define a well-named constant to indicate why we're checking for a length of 8.

amorrow-flexion (approved):

PR #605, opened 2024-03-26
Clean up .mob file a bit

jamesobrooks:
# Purpose

We want to have an easier time managing ensemble processes. Really just looking for thoughts.

# Major Changes

- Removes MOB_TIMER_USER to be set on each dev's machine at `~/.mob`
- Adds a default MOB_TIMER_ROOM
  - We could alternatively delete this line and set it locally at `~/.mob` for whatever work we're doing to better support more than one timer

## Commands
- `mob [start | timer] 15` will push the user's name to the timer and be visible at `https://timer.mob.sh/{MOB_TIMER_ROOM}`
- `mob break 15` will push a break to the timer
  - at the end of the break, mob.sh plays a `mob start` audio notification
- `mob timer open` opens the timer in a browser


governmentSponsored:
do we have a way to track goals with this tool?

PR #611, opened 2024-03-28
CAMS-345 Fix for missing branch hash id for main CICD deployment

magile32:
# Purpose

Fix deployment failure related to empty string passed to database deployment script. This only occurs on main branch.

# Major Changes

- Pass expected non-empty string to script
- Update smoke test

# Testing/Validation

- Successful Deploy database step in branch deployment 

# Notes

Need to verify in main branch CICD workflow

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -688,7 +688,7 @@ jobs:
 
           if [[ $webStatusCode = "200" && $apiStatusCode = "200" ]]; then
             echo "Print api healthcheck response"
-            curl https://${{ inputs.apiName }}.azurewebsites${{ inputs.azureHostnameSuffix }}/api/healthcheck?x-ms-routing-name=${{ inputs.slotName }}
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-deployment.yml
```diff
@@ -491,7 +491,7 @@ jobs:
             --analyticsWorkspaceId ${{ secrets.AZ_ANALYTICS_WORKSPACE_ID }} \
             --actionGroupResourceGroup ${{ secrets.AZ_ANALYTICS_RG }} \
             --actionGroupName ${{ secrets.AZ_ACTION_GROUP_NAME }} \
-            --branchHashId ${{ needs.build-info.outputs.environmentHash }}
+            --branchHashId ${{ needs.build-info.outputs.environmentHash || 'DOES_NOT_EXIST' }}
```
amorrow-flexion, line 494: Didn't this work with an empty branch has previously?

magile32: This needs to be tested on main branch.

magile32: The original error is because the script expects branchHashId to be passed in with an additional argument to the script. This just can't be an empty string. For the main deployment, 'DOES_NOT_EXIST' value will be ignored.

amorrow-flexion (approved):

PR #628, opened 2024-04-05
CAMS-237 Refactoring Github Action workflow

magile32:
# Purpose

The main GHA workflow file (continuous-deployment.yml) has grown in size and complexity. To improve readability and maintainability job steps will be group into reusable workflows.

# Major Changes

- Implement reusable workflow for security related jobs
- Implement reusable workflow for automated unit testing
- Implement reusable workflow for build and upload jobs
- Reusable workflow for build-info and update existing references to build-info outputs
- Group cloud deployment jobs 
- Group production code deployment jobs
- Group slot deployment jobs

# Testing/Validation

- Run full and slot deployment workflows

# Notes

- Some script changes may impact USTP. Will need to verify that these changes do not impact ADO environment


ops/scripts/pipeline/endpoint-test.sh
```diff
@@ -2,7 +2,7 @@
 
 # Title:        az-app-deploy.sh
```
magile32, line 3: Let's update the title and description

ops/scripts/pipeline/endpoint-test.sh
```diff
@@ -44,44 +40,39 @@ while [[ $# -gt 0 ]]; do
...
     esac
 done
 
+webStatusCode=""
+apiStatusCode=""
+targetApiURL="https://${api_name}.azurewebsites${host_suffix}/api/healthcheck"
+targetWebAppURL="https://${webapp_name}.azurewebsites${host_suffix}"
 # shellcheck disable=SC1083 # REASON: Wants to quote http_code
-webCmd="curl -q -o -I -L -s -w "%{http_code}" --retry 5 --retry-delay 60 --retry-connrefused -f https://${webapp_name}.azurewebsites${host_suffix}"
+webCmd="curl -q -o /dev/null -I -L -s -w "%{http_code}" --retry 5 --retry-delay 60 --retry-connrefused -f ${targetWebAppURL}"
```
magile32, line 54: We'll need to confirm if this works on the USTP side. We were notified that USTP had updated their build agent so this will most likely work.

jamesobrooks: @magile32, okay. As it was, the `-I` flag was being ignored because the `-o` flag requires a `<file>` argument to follow it. So we were seeing odd behavior running locally and were exiting early without actually validating the healthcheck response. Similarly with the `apiCmd`, but with the additional oddity that the `-I` flag was causing a problem there.

magile32: Original code leverage --retry-all-errors
`curl -q -o -I -L -s -w "%{http_code}" --retry 5 --retry-delay 60 --retry-all-errors -f https://${{ env.webappName }}.azurewebsites${{ vars.AZ_HOSTNAME_SUFFIX }}`

Let's modify the curl command to leverage `--retry 5 --retry-delay 60 --retry-all-errors ` instead of --retry-connfefused

test/e2e/playwright/fixture/urlQueryString.ts
```diff
@@ -9,7 +9,7 @@ export const test = base.extend({
       url += 'x-ms-routing-name=staging';
       return goto(url, options);
     }
-
+    //TODO: When doing an initial deployment we want to be able to test the Production slot
```
magile32, line 12: Do we want to take care of this in CAMS-237?

ops/scripts/pipeline/endpoint-test.sh
```diff
@@ -13,75 +13,92 @@
...
+  webStatusCode=$($webCmd)
+  apiStatusCode=$($apiCmd)
+  targetApiURL+="?x-ms-routing-name=${slot_name}"
+
+  if [[ -n "${expected_git_sha}" ]]; then
+    echo "Expect sha ${expected_git_sha}"
+    retry=0
+    currentGitSha=""
+    while [ "${expected_git_sha}" != "${currentGitSha}" ] && [ ${retry} -le 2 ]; do
+      echo "Debug 01"
```
magile32, line 79: TODO: Need to add a sleep here

test/e2e/playwright/fixture/urlQueryString.ts
magile32: Will address at a later time

ops/scripts/pipeline/check-deploy-bicep.sh
```diff
@@ -19,9 +19,9 @@ branch=$(git branch --show-current)
 lastMergeCommitSha=$(git log ${branch} --first-parent --pretty=format:"%H" --merges -n 1)
```
magile32, line 19: Not sure if this works as expected.

ops/scripts/pipeline/check-deploy-bicep.sh
```diff
@@ -3,7 +3,7 @@
 # Title:        check-deploy-bicep.sh
 # Description:  Helper script to check if bicep changes have been made and if bicep should be deployed
 
-set -euo pipefail # ensure job step fails in CI pipeline when error occurs
+set -xeuo pipefail # ensure job step fails in CI pipeline when error occurs
 
 app_rg=$1
 deploy_flag=$2 #workflow dispatch parameter enableBicepDeployment from GHA
```
magile32, line 9: I don't think need to pass this into the script. OR we could short circuit the logic if true is passed in. But I really don't think we need to pass this value. A false doesn't do anything and if its true we don't really need to run the script.

governmentSponsored:
@magile32 @amorrow-flexion I think this type of change warrants some updates to the ADR.

amorrow-flexion (approved):

ops/scripts/pipeline/endpoint-test.sh
magile32: Confirmed

PR #629, opened 2024-04-05
CAMS-301 Reject consolidation

btposey:
# Purpose

Reject a consolidation order.

# Major Changes

* Enable the rejection button in the consolidation order accordion
* Render the rejection modal
* Call the rejection endpoint from the UI
* Update the order buffer in the UI
* Update status on the consolidation order in Cosmos

# Testing/Validation

* Confirm documents are correctly written to Cosmos
* Test coverage above threshold

### DIFF
--- a/dev-tools/cosmos-records/delete-orders.ts
+++ b/dev-tools/cosmos-records/delete-orders.ts
@@ -13,6 +13,4 @@ export default function deleteOrders() {
   deleteDocuments('consolidations', 'consolidationId', 'SELECT * FROM c');
 }
 
-if (require.main === module) {
-  deleteOrders();
-}
+deleteOrders();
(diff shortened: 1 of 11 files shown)


dev-tools/cosmos-records/delete-orders.ts
```diff
@@ -13,6 +13,4 @@ export default function deleteOrders() {
   deleteDocuments('consolidations', 'consolidationId', 'SELECT * FROM c');
 }
 
-if (require.main === module) {
-  deleteOrders();
-}
+deleteOrders();
```
fmaddenflx, line 16: What was this about?  Why were we doing this check to begin with?

btposey: I added that guard a couple tickets ago. I broke `npm run reload-orders` with it. So this reverts it.

fmaddenflx (approved):

PR #632, opened 2024-04-08
CAMS-366 Limit court divisions

btposey:
# Purpose

Limit the lead case court selection for case consolidation modal

# Major Changes

Add a filter for the court list to find offices that are peer with the division noted on the consolidation case.

# Testing/Validation

Test coverage above threshold.


fmaddenflx (approved):
Looks good

PR #637, opened 2024-04-09
CAMS-366 - Amend default court selection logic

btposey:
# Purpose

Amend how we select the default court for consolidations

# Major Changes

* Fixed mapping issue between sync and presentation. Changed mapping of divisionCode => courtDivisionCode.
* Added function to find the unique division code from a set of cases.

# Testing/Validation

Added unit test for function that finds the default court.


backend/functions/lib/adapters/gateways/dxtr/orders.dxtr.gateway.ts
```diff
@@ -330,7 +330,7 @@ export class DxtrOrdersGateway implements OrdersGateway {
         CS.CS_DIV + '-' + CS.CASE_ID AS caseId,
         CS.CS_SHORT_TITLE AS caseTitle,
         FORMAT(CS.CS_DATE_FILED, 'yyyy-MM-dd') AS dateFiled,
-        CS.CS_DIV as divisionCode,
```
jamesobrooks, line 333: I think because we have `courtDivisionName`, maybe this should be `courtDivisionCode` so that it is very clear that there is a difference.

btposey: I agree. This is a bit more of a refactor, but I'm cool with that.

jamesobrooks (approved):

amorrow-flexion (approved):

PR #643, opened 2024-04-10
CAMS-299 - Show consolidation audit history

btposey:
# Purpose

Show consolidation audit history.

# Major Changes

Update case detail audit screen to show consolidation audit history.

# Testing/Validation

Test coverage above threshold.


btposey:
E2E and pa11y tests pass.

user-interface/src/case-detail/panels/CaseDetailAuditHistory.tsx
```diff
@@ -12,6 +17,82 @@ export interface CaseDetailAuditHistoryProps {
...
+    return (
+      <tr key={idx}>
+        <td>Staff</td>
+        <td data-testid={`previous-assignment-${idx}`}>
+          {history.before.length === 0 && <>(none)</>}
+          {history.before
+            .map((assignment) => {
+              return assignment.name;
+            })
+            .join(', ')}
```
jamesobrooks, line 30: Should we use that function we created recently that prepares assignee names for display?

jamesobrooks: If so, the after column should get the same treatment.

user-interface/src/case-detail/panels/CaseDetailAuditHistory.tsx
```diff
@@ -12,6 +17,82 @@ export interface CaseDetailAuditHistoryProps {
...
+      switch (history.documentType) {
+        case 'AUDIT_ASSIGNMENT':
+          return showCaseAssignmentHistory(history, idx);
+        case 'AUDIT_TRANSFER':
+          return showCaseTransferHistory(history, idx);
+        case 'AUDIT_CONSOLIDATION':
+          return showCaseConsolidationHistory(history, idx);
+      }
+    });
+  }
```
jamesobrooks, line 94: I like this improvement. It breaks up the html into more consumable chunks.

user-interface/src/lib/components/CaseNumber.tsx
```diff
@@ -18,7 +18,7 @@ export function CaseNumber(props: CaseNumberProps) {
     </span>
   );
   if (renderAs === 'link') {
-    const target = openLinkIn === 'new-window' ? '_blank' : '_self';
+    const target = openLinkIn === 'new-window' ? `CAMS-case-detail-${caseId}` : '_self';
```
jamesobrooks, line 21: Why do we use a named target here?

jamesobrooks (approved):
One suggestion for improvement but otherwise looks good.

user-interface/src/case-detail/panels/CaseDetailAuditHistory.tsx
btposey: Probably not here because we are not rendering a sentence.

user-interface/src/lib/components/CaseNumber.tsx
btposey: It opens a given case in the same window/tab each time. I think I'm going to revert it though because it felt awkward.

PR #646, opened 2024-04-11
CAMS-348 required fields

fmaddenflx:
# Purpose

Add required * to all required input fields

# Major Changes

All input fields have been rebuilt to accommodate and several input fields have been consolidated, due to the similarity between them.

# Testing/Validation

All tests are passing and new tests were added.


common/src/date-helper.ts
```diff
@@ -13,3 +13,7 @@ export function sortDatesReverse(dateA: Date | string, dateB: Date | string): nu
   //Sort ASC
   return sortDates(dateA, dateB) * -1;
 }
+
+export function checkDateStringFormat(value: string | undefined) {
```
btposey, line 17: Test

user-interface/src/lib/components/uswds/DateRangePicker.test.tsx
```diff
@@ -93,4 +93,50 @@ describe('Test DateRangePicker component', async () => {
     expect(startDateText).toHaveAttribute('max', '01/01/2035');
     expect(endDateText).toHaveAttribute('min', '01/01/2020');
   });
+
+  /*
```
btposey, line 97: Should we just .skip this?

user-interface/src/lib/components/uswds/todo.txt
```diff
@@ -0,0 +1,9 @@
+Missing USWDS Form Components:
+  1. Combo box
+  3. File input
```
btposey, line 3: Do we need this file??

btposey (changes requested):

user-interface/src/lib/components/uswds/Checkbox.tsx
```diff
@@ -60,16 +58,18 @@ const CheckboxComponent = (props: CheckboxProps, ref: React.Ref<CheckboxRef>) =>
   const checkboxTestId = `checkbox-${props.id}`;
   const labelTestId = `checkbox-label-${props.id}`;
   return (
-    <div className={classes}>
+    <div className={`usa-form-group uswds-form usa-checkbox ${props.className ?? ''}`}>
```
fmaddenflx, line 61: This is funny.  So in other places we moved the class stuff out from inline and set up the class name outside of the jsx.  Here we did the opposite.  As long as we can test it I guess it doesn't matter.

Also, I don't think we need "uswds-form" anywhere any longer.

user-interface/src/lib/components/uswds/DatePicker.tsx
fmaddenflx, line 1: We need to give people the option to pass in a class name in a prop.

user-interface/src/lib/components/uswds/DateRangePicker.test.tsx
fmaddenflx: I really do feel like we should complete this test.  It's bugging me.  I know it should be simple, and I'm not sure why it's not.

If we punt on it, then yes we should skip it and leave a comment as to why.

user-interface/src/lib/components/uswds/todo.txt
fmaddenflx: I suppose not.  If not, then I'll save the file somewhere.  I think it's good to have a record of what isn't covered.  There was a group that was forming at Flexion to create a standard react and vue library for USWDS.  I imagine we could contribute our uswds components to it.

user-interface/src/lib/components/uswds/DatePicker.tsx
fmaddenflx: Will wait for axis of change.

btposey (approved):

PR #652, opened 2024-04-16
CAMS-237 e2e update

amorrow-flexion:
Jira ticket: CAMS-237

# Purpose

Only deploy e2e database if e2e tests will be run on slots

# Major Changes

Logic to only deploy e2e databases if slots are being deployed

# Testing/Validation

Deployed to environment without slot, deployed to environment with slots

# Notes

This is something I noticed during pipeline runs and felt it was not expected functionality.


magile32 (approved):
LGTM

PR #659, opened 2024-04-17
CAMS-370: Add EA team and general code ownership

jamesobrooks:
# Purpose

We want the EA team to be required approvers for files affecting deployment in USTP environments.

# Major Changes

- Add EA as code owners for all `.bicep` files, and the directories where we store scripts and other files relating to deployment that affect USTP environments.


CODEOWNERS
jamesobrooks, line 1: @governmentSponsored should we the ustp-team as owner of this file?

governmentSponsored: oh of the CODEOWNERS file? Yeah I like that idea.

PR #660, opened 2024-04-17
CAMS-368 Prevent multiple consolidation leads

btposey:
# Purpose

Prevent consolidation child cases from being assigned as lead cases on a different consolidation.

# Major Changes

* Added new local repo implementations to aid in testing use cases with complex preconditions.
* Added guard stop consolidation approval logic in the API to stop any consolidation with an existing child case.
* Added logic to the UI to use the associated cases endpoint to determine if cases are child cases.
* Added warning messages to consolidation order cases that are child cases of another consolidation
* Added an error message to the consolidation order modal if a lead case is a child case of another consolidation.
* Added a guard to disable the approval button if any selected case is a child case of another consolidation.

# Testing/Validation

* Manual testing
* Unit testing above coverage threshold.


backend/functions/lib/adapters/gateways/cosmos/cosmos.repository.test.ts
```diff
@@ -4,9 +4,9 @@ import { ApplicationContext } from '../../types/basic';
 import { getCosmosDbCrudRepository } from '../../../factory';
 import { AggregateAuthenticationError } from '@azure/identity';
 import { ServerConfigError } from '../../../common-errors/server-config-error';
-import { CosmosDbRepository } from './cosmos.repository';
 import { ErrorResponse } from '@azure/cosmos';
 import { ID_ALREADY_EXISTS } from './cosmos.helper';
+import { DocumentRepository } from '../../../use-cases/gateways.types';
```
fmaddenflx, line 9: What was the reason for switching to DocumentRepository instead of CosmosDbRepository?

user-interface/src/data-verification/ConsolidationOrderAccordion.test.tsx
```diff
@@ -353,11 +363,18 @@ describe('ConsolidationOrderAccordion tests', () => {
...
+      .mockResolvedValue({
+        message: '',
+        count: 1,
+        body: [MockData.getAttorneyAssignment()],
+      })
+      .mockResolvedValue({
+        message: '',
+        count: 1,
+        body: [],
+      });
```
fmaddenflx, line 377: This test seems to be failing because the attorneys never seem to get loaded.  The approval button will not become enabled until after the attorneys finish loading because it's loading the associated cases.  But since the loading never resolves, the button is never enabled.

btposey: Yeah, we added a new call to the associated cases endpoint and we have to get the sequence right in the mocks.

btposey: James replaced this with a more elegant solution.

backend/functions/lib/use-cases/orders/DELETE_ME.md
```diff
@@ -0,0 +1,12 @@
+1. We approve a consolidation of the following cases:
```
btposey, line 1: Delete

test/e2e/playwright/consolidation-orders.spec.ts
```diff
@@ -29,9 +29,12 @@ test.describe('Consolidation Orders', () => {
     page,
   }) => {
     // get pending consolidation order id
-    const pendingConsolidationOrder: Order = orderResponseBody.find(
-      (o) => o.orderType === 'consolidation' && o.status === 'pending',
-    );
+    const pendingConsolidationOrder = orderResponseBody
```
btposey, line 32: I think this should be reverted. The e2e test fails locally if we have consolidated cases in the first accordion. Workaround is we need to reload orders before running the e2e test.

user-interface/src/data-verification/ConsolidationCasesTable.test.tsx
```diff
@@ -105,4 +106,45 @@ describe('test ConsolidationCasesTable component', () => {
...
+
+  test('Should display "Already part of consolidation" alert if a case in a consolidation order is a part of another consolidation', () => {
+    const props = {
+      cases: [
+        MockData.getConsolidatedOrderCase({
+          override: {
+            associations: [
+              MockData.getConsolidationReference({
+                override: {
+                  caseId: '11-1111',
```
btposey, line 118: This needs to have a `documentType: 'CONSOLIDATION_FROM'` set in the override. Probably remove the `as ConsolidationFrom`.

fmaddenflx (approved):
Looks good to me.

amorrow-flexion (approved):

PR #665, opened 2024-04-22
Dependency updates auto

jamesobrooks:
# Problem

Node packages need to be kept up to date.

# Major Changes

- Add a workflow to schedule running dependency updates from GitHub Actions
- Update the script to handle running in CI/CD context

# Testing/Validation

Automated tests.

# Notes

The workflow is not tested yet as it will not execute unless it is in main. We will check it out tomorrow if it has merged, or we can change the workflow to run on push/pr.


jamesobrooks:
The pipeline-scan if failing and we are creating an issue with VeraCode to figure out why. We get a 500 response when trying to upload the zip for scanning, even locally. Ignoring this error for now.

governmentSponsored (approved):

PR #666, opened 2024-04-23
CAMS-367 Edit case assignment

btposey:
# Purpose

Add trial attorney assignments to the case detail screen.

# Major Changes

* Add assignment modal from case assignment screen to case detail screen.
* Add assignment business rules to the API when handling consolidation cases.

# Testing/Validation

* Unit tests above threshold.


common/src/cams/events.test.ts
```diff
@@ -0,0 +1,54 @@
...
+describe('Event domain helper functions', () => {
+  describe('isJointAdministrationLeadCase', () => {
+    test('should return true', () => {
+      expect(isJointAdministrationLeadCase(jointAdministrationLeadCase)).toBeTruthy();
+    });
+    test('should return false for all other cases', () => {
+      expect(isJointAdministrationLeadCase([])).toBeFalsy();
+      expect(isJointAdministrationLeadCase(undefined)).toBeFalsy();
+      expect(isJointAdministrationLeadCase(jointAdministrationChildCase)).toBeFalsy();
+      expect(isJointAdministrationLeadCase(substantiveConsolidationLeadCase)).toBeFalsy();
```
jamesobrooks, line 38: Should we also be testing `substantiveConsolidationChildCase` and `nonConsolidatedCase`?

backend/functions/lib/use-cases/case.assignment.ts
```diff
@@ -21,12 +21,41 @@ export class CaseAssignmentUseCase {
...
+
+    return {
+      success: true,
+      message: 'Trial attorney assignments created.',
+      count: assignmentIds.length,
+      body: assignmentIds,
+    };
+  }
+
+  private async assignTrialAttorneys(
```
fmaddenflx, line 52: On line 77 - 101 (not a change in this PR), it looks like we are performing the exact same loop 2 times, but just doing something different based on the same filter process.  I'd like to understand if there is reasoning behind that and I'm wondering if there is a more efficient way to handle this.

common/src/cams/events.test.ts
btposey: Looks good per conversation.

user-interface/src/case-assignment/AssignAttorneyModal.test.tsx
```diff
@@ -1,61 +1,66 @@
...
 
+const susan = new Attorney('Susan', 'Arbeit', 'Manhattan');
+const mark = new Attorney('Mark', 'Bruh', 'Manhattan');
+const shara = new Attorney('Shara', 'Cornell', 'Manhattan');
+const brian = new Attorney('Brian', 'Masumoto', 'Manhattan', { middleName: 'S' });
+const joe = new Attorney('Joe', 'Cornell', 'Manhattan');
+const bob = new Attorney('Bob', 'Cornell', 'Manhattan');
+const frank = new Attorney('Frank', 'Cornell', 'Manhattan');
+const sally = new Attorney('Sally', 'Cornell', 'Manhattan');
+const may = new Attorney('May', 'Cornell', 'Manhattan');
```
jamesobrooks, line 21: These are real attorneys. Unless a test is getting attorneys from `attorneys.mock.ts`, which is only in the backend.

jamesobrooks (approved):

PR #698, opened 2024-05-06
CAMS-387: setup clearly visibile environment when not in production

amorrow-flexion:
Jira ticket: CAMS-387

# Purpose

Clearly state environment when the user is not within the production application

# Major Changes

- Change color of banner dependent on environment
- Display environment centered within banner for development and staging
- Edited workflow to include CAMS_LAINCH_DARKLY_ENV as an env var

# Testing/Validation

- ran build locally with all 3 LaunchDarkly environments as env var CAMS_LAUNCH_DARKLY_ENV= (development, staging, production) and verified the app functioned as expected

# Notes

We can modify this however we want. these were just my initial thoughts

We will also need to add the var for CAMS_LAUNCH_DARKLY_ENV to the USTP STG and PRD AzDO variable libraries
- 'staging' within CAMS_STG

### DIFF
--- a/user-interface/src/lib/components/Header.tsx
+++ b/user-interface/src/lib/components/Header.tsx
@@ -41,9 +41,8 @@ export const Header = () => {
   const flags = useFeatureFlags();
   const transferOrdersFlag = flags[TRANSFER_ORDERS_ENABLED];
   const caseSearchFlag = flags[CASE_SEARCH_ENABLED];
-
+  const launchDarklyEnvironment = import.meta.env['CAMS_LAUNCH_DARKLY_ENV'];
   const [activeNav, setActiveNav] = useState<NavState>(mapNavState(location.pathname));
-
   useEffect(() => {
     setActiveNav(mapNavState(location.pathname));
   }, [location]);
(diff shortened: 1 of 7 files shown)


user-interface/src/lib/components/uswds/Banner.scss
```diff
@@ -2,4 +2,22 @@
   .usa-banner__inner {
     max-width: none;
   }
+  .development {
+    background-color: #419ce1;
+    color: white;
+  }
+
+  .staging {
```
jamesobrooks, line 10: The environment name in LaunchDarkly is `test`.
```suggestion
  .test {
```

user-interface/src/lib/components/Header.tsx
```diff
@@ -41,9 +41,7 @@ export const Header = () => {
   const flags = useFeatureFlags();
   const transferOrdersFlag = flags[TRANSFER_ORDERS_ENABLED];
   const caseSearchFlag = flags[CASE_SEARCH_ENABLED];
-
   const [activeNav, setActiveNav] = useState<NavState>(mapNavState(location.pathname));
-
```
jamesobrooks, line 46: I think we would generally have a blank line between the variables and the functions. I'd also be okay with the one above added back in to separate the React state stuff from the regular variables.

user-interface/src/lib/components/uswds/Banner.scss
```diff
@@ -2,4 +2,22 @@
   .usa-banner__inner {
     max-width: none;
   }
+  .development {
```
jamesobrooks, line 5: I would prefer consistent use of blank lines (whether there or not).

jamesobrooks (changes requested):
Let's get the name for the staging environment corrected.

@jamesobrooks We can't update the name of the LaunchDarkly environment in place, only the display name. It wouldn't be much effort to create a new environment and get that setup with the proper name 'Staging. Is that something we would want to do?

user-interface/src/lib/components/Header.tsx
amorrow-flexion: Added a break above and below. I am going to see if I can enforce line space rules locally so this is consistent

jamesobrooks (approved):

PR #705, opened 2024-05-08
CAMS-382: initial push to simplify and segregate bicep templates

amorrow-flexion:
Jira ticket: CAMS-382

# Purpose

Simplify and clearly separate the components of the application infrastructure into more straightforward bicep templates. This inintial push is focused primarily on the networking configuration

# Major Changes

- updates to main.bicep adding the ustp-cams-network.bicep
- updates to backend-api-deploy.bicep removing networking other than private endpoint creation
- updates to frontend-webapp-deploy.bicep removing networking other than private endpoint creation
- creation of ustp-cams-network.bicep to deploy networking components prior to other operations
- updates to many other templates reflecting the single use principle

# Testing/Validation

- intent to deploy on a development environment and test that the application is function as it should be

# Notes

There is still more to be done


jamesobrooks (commented):
It seems like mostly extracting a bunch of stuff into `ops/cloud-deployment/ustp-cams-network.bicep`, but it's hard to tell for sure. It might be best to walk through this if we can get someone's time from EA.

PR #706, opened 2024-05-10
Cams 382: bicep improvements - network

amorrow-flexion:
Jira ticket: CAMS-382

# Purpose

Simplify and clearly separate the components of the application infrastructure into more straightforward bicep templates. This inintial push is focused primarily on the networking configuration

# Major Changes

- updates to main.bicep adding the ustp-cams-network.bicep
- updates to backend-api-deploy.bicep removing networking other than private endpoint creation
- updates to frontend-webapp-deploy.bicep removing networking other than private endpoint creation
- creation of ustp-cams-network.bicep to deploy networking components prior to other operations
- updates to many other templates reflecting the single use principle
- updates to azure-deploy.sh to allow for shared params to be clearly passed in

# Testing/Validation

- intent to deploy on a development environment and test that the application is function as it should be

# Notes

- Will need updates on USTP side before merging.
- This will reconfigure the Flexion and USTP Staging network and is expected to cause downtime in our environments


.github/workflows/continuous-deployment.yml
```diff
@@ -134,15 +134,14 @@ jobs:
...
-      [
+    needs: [
         setup,
         build,
-        accessibility-test,
-        security-scan,
-        unit-test-frontend,
-        unit-test-backend,
-        unit-test-common,
+        # accessibility-test,
```
governmentSponsored, line 140: can we delete these if they are commented out or are they still needed?

ops/cloud-deployment/main.bicep
```diff
@@ -1,241 +1,214 @@
...
+@description(' Vnet Name. USTP Required')
 param virtualNetworkName string = 'vnet-${appName}'
+@description('Array of Vnets to link to DNS Zone. USTP Required')
 param linkVnetIds array = []
 
-@description('Set to true to deploy web module resources. This should be set to false for Azure slot deployments.')
+@description('Private Endpoint Subnet Name. USTP Required')
+param privateEndpointSubnetName string = 'snet-${appName}-private-endpoints'
+@description('Private Endpoint Subnet address prefix. USTP Required')
+param privateEndpointSubnetAddressPrefix string = '10.10.12.0/28'
```
jamesobrooks, line 22: I don't really see value in descriptions like this. If we were to put this CIDR in a variable or move it to the GitHub Workflow, we wouldn't need the `USTP Required` part because it would always be required, and the rest of the description is redundant.

ops/cloud-deployment/main.bicep
```diff
@@ -1,241 +1,214 @@
...
 param privateDnsZoneSubscriptionId string = subscription().subscriptionId
 
 @description('Name of deployment slot for frontend and backend')
 param slotName string = 'staging'
-param azHostSuffix string = '.net'
+param azHostSuffix string = '.us'
 
-@secure()
-param databaseConnectionString string = ''
+@description('SQL Server Name. Flexion and USTP Required')
```
jamesobrooks, line 64: If this is required, why does it have the default value? Could we not just remove the default and not need the description at all?

ops/cloud-deployment/main.bicep
amorrow-flexion: That is a fair statement. I had added these descriptions prior to reevaluating the azure-deploy.sh script, which also detracts from the value of these.

ops/scripts/pipeline/azure-deploy.sh
```diff
@@ -125,12 +159,97 @@ while [[ $# -gt 0 ]]; do
...
+    exit 10
+fi
+if [[ -z "${deploy_webapp}" ]]; then
+    echo "Error: Missing deployWebapp"
+    exit 10
+fi
+if [[ -z "${action_group_name}" && "${create_alerts}" == true ]]; then
+    echo "Create Alerts: ${create_alerts} but no actionGroupName supplied"
+    exit 10
+fi
```
jamesobrooks, line 235: This seems extremely verbose. Could we implement something else to check for required variables?

amorrow-flexion: Intent to handle this better in the second iteration of bicep improvements

governmentSponsored (approved):

jamesobrooks (approved):

brandonjohnsen (approved):

PR #730, opened 2024-05-29
Cams 391 filter usability

amorrow-flexion:
# Purpose

Make Filter buttons within the data verification screen more usable and understandable

# Major Changes

- Container added for the Event Status and Event Type
- Added styles based on the enabled/disabled properties of filters

# Testing/Validation

Ran the application locally and validated the filter were functioning properly


btposey (changes requested):
The positions of Event Type and Event Status need to be swapped to match the prototype.

btposey (approved):

PR #740, opened 2024-06-04
CAMS-347 Single Sign-On - MOCK ONLY

btposey:
# Purpose

Add single sign on capability to the web app.

# Major Changes

* Added screens supporting the SSO workflow
* Added DOJ authorized use only screens per regulatory requirements
* Added good fences for SSO integrations per OEA
* Added Azure AD OIDC provider
* Added mock provider

# Testing/Validation

Unit test coverage above threshold.


user-interface/src/login/providers/azure/authConfig.ts
```diff
@@ -0,0 +1,65 @@
...
+      loggerCallback: (level: LogLevel, message: string, containsPii: boolean) => {
+        if (containsPii) {
+          return;
+        }
+        switch (level) {
+          case LogLevel.Error:
+            console.error(message);
+            return;
+          case LogLevel.Info:
+            // console.info(message);
```
amorrow-flexion, line 29: drop these logs

PR #776, opened 2024-06-27
CAMS-401 Mock Oauth2

btposey:
# Purpose

To provide a cohesive mock login provider solution for the front and back ends.

# Major Changes

* Added Oauth2 endpoint to API
* Integrated MockLogin component to use Oauth2 endpoint for JWT tokens

# Testing/Validation

Just mock code.


common/src/cams/mock-auth.ts
```diff
@@ -0,0 +1,13 @@
+import { CamsUser } from './session';
+
```
btposey, line 2: Rename file mock role.

amorrow-flexion (approved):

PR #781, opened 2024-07-01
CAMS-401 remove USTP urls from mock issuer

amorrow-flexion:
Jira ticket: CAMS-401

# Purpose

We have Okta for USTP now, so we do not need mockIssuer URLS for those 2 environments any more

# Major Changes

removal of USTP Urls

# Testing/Validation

N/A

### DIFF
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -178,6 +178,7 @@ jobs:
       environmentHash: ${{ needs.setup.outputs.environmentHash }}
       cosmosDbClientId: ${{ needs.deploy.outputs.cosmosDbClientId }}
       cosmosDbPrincipalId: ${{ needs.deploy.outputs.cosmosDbPrincipalId }}
+      mockAuth: ${{ vars.MOCK_AUTH }}
     secrets: inherit # pragma: allowlist secret
 
   initial-slot-resource-deployment: ##setup slot resources on initial deployment
--- a/.github/workflows/continuous-deployment.yml
+++ b/.github/workflows/continuous-deployment.yml
@@ -134,7 +134,6 @@ jobs:
       webappName: ${{ needs.setup.outputs.webappName }}
       environment: ${{ needs.setup.outputs.ghaEnvironment }}
       launchDarklyEnvironment: ${{ vars.CAMS_LAUNCH_DARKLY_ENV }}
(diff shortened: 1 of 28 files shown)


.github/workflows/continuous-deployment.yml
```diff
@@ -178,6 +178,7 @@ jobs:
       environmentHash: ${{ needs.setup.outputs.environmentHash }}
       cosmosDbClientId: ${{ needs.deploy.outputs.cosmosDbClientId }}
       cosmosDbPrincipalId: ${{ needs.deploy.outputs.cosmosDbPrincipalId }}
+      mockAuth: ${{ vars.MOCK_AUTH }}
```
amorrow-flexion, line 181: I don't see a GHA var setup for this. What is the intent of it?

backend/functions/lib/configs/authorization-configuration.ts
```diff
@@ -3,13 +3,16 @@ import { AuthorizationConfig } from '../adapters/types/authorization';
 
 dotenv.config();
 
+const doMockAuth = process.env.MOCK_AUTH === 'true';
```
amorrow-flexion, line 6: Why bother setting up another env when we can leverage provider and do

`const doMockAuth = process.env.CAMS_LOGIN_PROVIDER === 'mock';`

.github/workflows/sub-deploy-code-slot.yml
```diff
@@ -112,7 +115,7 @@ jobs:
           -n ${{ inputs.apiName }} \
           --src ./${{ inputs.apiName }}.zip \
           --slotName ${{ inputs.slotName }} \
-          --settings "INFO_SHA=${{ github.sha }}"
+          --settings "INFO_SHA=${{ github.sha }} MOCK_AUTH=${{ inputs.mockAuth }}"
```
amorrow-flexion, line 118: I'm not sure we need this here. We read from React env on the frontend, I don't think we would ever access this setting

backend/functions/lib/configs/authorization-configuration.ts
btposey: We could have done it that way too. There was code already in place to look for the MOCK_AUTH env var so we built off that.

btposey (approved):

btposey (approved):

PR #788, opened 2024-07-03
CAMS-401 Added key value utilities to common

btposey:
# Purpose

Plan B to replace JSON environment variables

# Major Changes

Added a library to turn a pipe-delimited string if key-value pairs into a Record or Map.

# Testing/Validation

100% unit test coverage

### DIFF
--- a/common/src/cams/utilities.test.ts
+++ b/common/src/cams/utilities.test.ts
@@ -28,6 +28,13 @@ describe('Key Value utils', () => {
       expect(actual).toEqual(expected);
     });
 
+    test('should omit a pair without a key', () => {
+      const encoded = '=value';
+      const expected = {};
+      const actual = keyValuesToRecord(encoded);
+      expect(actual).toEqual(expected);
+    });
+
     test('should omit empty pairs', () => {
       const encoded = '||ONE=1|||||';
       const expected = {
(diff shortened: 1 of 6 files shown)


btposey:
This is probably tech debt already without any usage.

common/src/cams/utilities.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      const actual = keyValuesToMap(encoded);
+      expect(actual).toEqual(expected);
+    });
+
+    test('should omit a pair without a equal sign delimiter', () => {
+      const encoded = 'foo|ONE=1|bad';
+      const expected = new Map<string, string>([['ONE', '1']]);
+      const actual = keyValuesToMap(encoded);
+      expect(actual).toEqual(expected);
+    });
```
jamesobrooks, line 74: Should it omit the pair, or should it create an object like this:
```
{
  foo: undefined,
  ONE: '1',
  bad: undefined,
}
```

common/src/cams/utilities.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    test('should trim keys and values', () => {
+      const encoded = ' ONE = 1 | TWO      = 2         ';
+      const expected = new Map<string, string>([
+        ['ONE', '1'],
+        ['TWO', '2'],
+      ]);
+      const actual = keyValuesToMap(encoded);
+      expect(actual).toEqual(expected);
+    });
+  });
```
jamesobrooks, line 92: I would like to add a test that covers something where the value includes an equal sign, maybe something like a connection string:
```
connectionString=Server=tcp:app-name.database.usgovcloudapi.net,1433;Initial Catalog=DATABASE_NAME;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;Authentication="Active Directory Default";|foo=true
```

PR #812, opened 2024-07-25
Update diagrams

jamesobrooks:
# Purpose

We want our diagrams to be up to date.

# Major Changes

Updated architecture diagrams.


amorrow-flexion:
Should we choose to auto-update diagrams, we can use https://github.com/marketplace/actions/structurizr-cli

amorrow-flexion (approved):

PR #855, opened 2024-08-23
CAMS-356 brian experiment

btposey:
# Purpose

Eliminate boiler plate code in Azure function scripts.

# Major Changes

Added builder utility function which outputs a Azure function handler containing all boilerplate code


backend/functions/lib/controllers/controller-types.ts
```diff
@@ -1,3 +1,4 @@
+export interface CamsController {}
```
btposey, line 1: Don't need this change.

PR #870, opened 2024-09-04
CAMS-418 - Return a list of offices for a group designator

btposey:
# Purpose

Return a list of offices per a given group designator

# Major Changes

Changed the function that used to return a single office to return a list of offices.

# Testing/Validation

Testing passes


fmaddenflx (approved):
Looks good.

PR #901, opened 2024-09-16
Cams 282 auditable

jamesobrooks:
# Purpose

We want to document the reason for the creation and proper use of `auditable.ts`.

# Major Changes

Add an ADR.

### DIFF
--- a/backend/functions/lib/use-cases/case-assignment.ts
+++ b/backend/functions/lib/use-cases/case-assignment.ts
@@ -79,9 +79,9 @@ export class CaseAssignmentUseCase {
           role: CamsRole[role],
           assignedOn: currentDate,
         },
-        context.session,
-        { updatedOn: currentDate },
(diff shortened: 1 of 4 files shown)


btposey (approved):

backend/functions/lib/use-cases/case-assignment.ts
```diff
@@ -79,9 +79,9 @@ export class CaseAssignmentUseCase {
           role: CamsRole[role],
           assignedOn: currentDate,
         },
-        context.session,
-        { updatedOn: currentDate },
+        context.session?.user,
       );
+      assignment.updatedOn = currentDate;
```
fmaddenflx, line 84: Curious why we decided to update "updatedOn" after creating the audit record.  I see we did this also on line 127.  I must have missed the reasoning behind that.

fmaddenflx (approved):

PR #970, opened 2024-10-10
CAMS-420 case search critical accessibility 

amorrow-flexion:
CAMS-420 case search critical accessibility

#  Purpose

Update Case Search screen in accordance with accessibility findings

# Major Changes

- added functionality that properly renders and reads alerts in case search in screen reader
- added search button instead of automatically applying searches as it was impacting accessibility
- updated labels and roles of search screen in accordance with accessibility
- for specifics, see google doc for this ticket

# Testing/Validation

- ran application locally with screen reader to verify accessibility updates
- ran pa11y tests locally and in the pipeline to verify accessibility updates

# Notes

- decided on submit button instead of automatically applying searches

# Definition of Done:

- [x] Acceptance criteria met - The system meets all acceptance criteria
- [x] Feature toggles created - Features that are deployed but not released have toggles
- [ ] Design QA passed - All visual UI looks and works the way it’s supposed to
- [ ] Usability validated - Team ensures the working code is easy to use
- [x] Accessibility tests passed - System passes automated and manual accessibility tests and supports applicable devices
- [x] Dependency Rule followed - More important code doesn’t directly depend on less important code
- [ ] Source code merged - Code has been merged into the main branch
- [x] 90% unit test coverage achieved - Automated test coverage tools indicate coverage of >= 90%
- [x] Code reviewed - Code is reviewed by at least two other team members before being merged
- [x] Code quality checks passed - Code passes all automated quality checks
- [x] Security scans passed - Code passes expected scans for vulnerabilities and compliance issues
- [x] Threat model updated - The threat model incorporates new threats and remediations
- [x] Build process updated - Automated build process include code supporting the user story
- [x] Load/performance tests passed - Performance tests include the functionality of the user story
- [x] Documents are updated - All required documentation is up to date and version controlled
- [ ] Code is deployed - Code is deployed to the highest production like environment

### DIFF
--- a/user-interface/src/lib/components/CaseNumberInput.tsx
+++ b/user-interface/src/lib/components/CaseNumberInput.tsx
@@ -100,19 +100,17 @@ function CaseNumberInputComponent(props: CaseNumberInputProps, ref: React.Ref<In
   }, [isDisabled]);
 
   return (
-    <>
-      <Input
-        {...otherProps}
-        ref={forwardedRef}
-        onChange={handleOnChange}
-        onKeyDown={handleEnter}
-        onFocus={handleOnFocus}
-        includeClearButton={true}
-        ariaDescription="For example, 12-34567"
-        placeholder="__-_____"
-        aria-placeholder=""
-      ></Input>
-    </>
+    <Input
+      {...otherProps}
+      ref={forwardedRef}
+      onChange={handleOnChange}
+      onKeyDown={handleEnter}
+      onFocus={handleOnFocus}
+      includeClearButton={true}
+      ariaDescription="For example, 12-34567"
+      placeholder="__-_____"
+      aria-placeholder=""
+    ></Input>
   );
 }
 
--- a/user-interface/src/lib/components/CaseNumberInput.tsx
+++ b/user-interface/src/lib/components/CaseNumberInput.tsx
@@ -110,6 +110,7 @@ function CaseNumberInputComponent(props: CaseNumberInputProps, ref: React.Ref<In
         includeClearButton={true}
         ariaDescription="For example, 12-34567"
         placeholder="__-_____"
+        aria-placeholder=""
       ></Input>
     </>
   );
--- a/user-interface/src/lib/components/CaseNumberInput.tsx
+++ b/user-interface/src/lib/components/CaseNumberInput.tsx
@@ -72,13 +72,17 @@ function CaseNumberInputComponent(props: CaseNumberInputProps, ref: React.Ref<In
   useImperativeHandle(ref, () => ({ clearValue, resetValue, setValue, getValue, disable }));
 
   return (
-    <Input
-      {...otherProps}
-      ref={forwardedRef}
-      onChange={handleOnChange}
-      onKeyDown={handleEnter}
-      includeClearButton={true}
-    ></Input>
+    <>
+      <Input
+        {...otherProps}
+        ref={forwardedRef}
+        onChange={handleOnChange}
+        onKeyDown={handleEnter}
+        includeClearButton={true}
+        ariaDescription="For example, 12-34567"
+        placeholder="__-_____"
+      ></Input>
+    </>
   );
 }
 
--- a/user-interface/src/lib/components/CaseNumberInput.tsx
+++ b/user-interface/src/lib/components/CaseNumberInput.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useImperativeHandle, useRef } from 'react';
+import React, { forwardRef, useEffect, useImperativeHandle, useRef, useState } from 'react';
 import Input, { InputProps } from './uswds/Input';
 import { InputRef } from '../type-declarations/input-fields';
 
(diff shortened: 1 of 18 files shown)


user-interface/src/staff-assignment/screen/StaffAssignmentScreen.test.tsx
```diff
@@ -31,16 +37,11 @@ describe('StaffAssignmentScreen', () => {
     };
 
     const SearchResults = vi
-      .spyOn(searchResultsModule, 'SearchResults')
```
amorrow-flexion, line 34: when looking at the coverage, this was not needed as part of the coverage for StaffAssignmentScreen.

user-interface/src/lib/components/Pill.tsx
```diff
@@ -39,7 +39,7 @@ export function Pill(props: PillProps) {
       onClick={() => props.onClick(props.value)}
       onKeyDown={(ev) => handleKeyDown(ev)}
       tabIndex={0}
-      aria-label={`${props.ariaLabelPrefix} - ${props.label} is currently selected.`}
+      aria-label={`${props.ariaLabelPrefix} - ${props.label} is currently selected. Press Enter key to unselect.`}
```
jamesobrooks, line 42: Why do we need this additional verbiage? Space bar also works, so it seems like default behavior for a button.

user-interface/src/search/SearchScreen.tsx
```diff
@@ -167,9 +199,12 @@ export default function SearchScreen() {
                   className="new-court__select"
                   label="District (Division)"
                   ariaLabelPrefix="District (Division)"
+                  ariaDescription="Select multiple Districts (Divisions)."
```
jamesobrooks, line 202: Are we sure on this wording? "Select multiple" could be perceived as requiring more than one selection. I would think we would want to indicate that zero, one, or more selections are valid. Or maybe just say something like "multi-select [field]"?

user-interface/src/lib/components/Pill.tsx
amorrow-flexion: It is default interaction behavior, but the button removing itself isn't completely self explanatory. maybe something like `${props.ariaLabelPrefix} - ${props.label} filter. Select to remove` ?

user-interface/src/search/SearchScreen.tsx
jamesobrooks: @shalliedragon

user-interface/src/lib/components/Pill.tsx
jamesobrooks: @shalliedragon

ShallieDragon: I'm okay with having that. Though I'd say "deselect" instead of "unselect." Unselect isn't a word.

jamesobrooks: @ShallieDragon, sorry, I just want to clarify. You're okay with having what, the current text or Arthur's suggestion. Note that both the enter and space bar keys perform the same action.

ShallieDragon: I think it'd be good to have a programmatic indicator that says something like "click to deselect it". I think saying "click" specifically might be okay, because I think screen reader users know to interpret "click" as "press the interact keyboard button".

user-interface/src/lib/components/CaseNumberInput.tsx
```diff
@@ -69,16 +85,33 @@ function CaseNumberInputComponent(props: CaseNumberInputProps, ref: React.Ref<In
...
+
   return (
-    <Input
-      {...otherProps}
-      ref={forwardedRef}
-      onChange={handleOnChange}
-      onKeyDown={handleEnter}
-      includeClearButton={true}
-    ></Input>
+    <>
```
fmaddenflx, line 103: What's the purpose of putting this input inside a fragment?

user-interface/src/search-results/SearchResults.tsx
```diff
@@ -40,11 +40,20 @@ export type SearchResultsProps = {
...
+  const {
+    id,
+    searchPredicate: searchPredicateProp,
+    onStartSearching,
+    onEndSearching,
+    noResultsMessage: noResultsMessageProp,
+    header: Header,
+    row: Row,
+    ...otherProps
+  } = props;
```
fmaddenflx, line 52: I would just like to say about this general practice -- This may may provide for less typing when referring to properties throughout our components, but it also muddies the waters in regards to name spacing.  It becomes unclear as to the difference between variables defined as native to the component and those fed in through properties.  I would like to perhaps keep this practice to a minimum and only for variables that really need it.

I suppose on the flip side of that, there is the need to pull certain properties out that you need passed down to subcomponents and sometimes those are native DOM attributes that we don't necessarily know about... so in that case it's good to de-structure props because you can exclude certain things from "otherProps" that you don't want added to subcomponents.

user-interface/src/search/SearchScreen.tsx
```diff
@@ -13,25 +13,33 @@ import { courtSorter } from '@/data-verification/DataVerificationScreen';
...
 import ComboBox, { ComboOption } from '@/lib/components/combobox/ComboBox';
-import { isValidSearchPredicate, SearchResults } from '@/search-results/SearchResults';
+import SearchResults, { isValidSearchPredicate } from '@/search-results/SearchResults';
 import { SearchResultsHeader } from './SearchResultsHeader';
 import { SearchResultsRow } from './SearchResultsRow';
 import { useGlobalAlert } from '@/lib/hooks/UseGlobalAlert';
 import DocumentTitle from '@/lib/components/cams/DocumentTitle/DocumentTitle';
 import { MainContent } from '@/lib/components/cams/MainContent/MainContent';
+import Button, { ButtonRef } from '@/lib/components/uswds/Button';
+import { UswdsButtonStyle } from '../lib/components/uswds/Button';
```
fmaddenflx, line 23: Looks like these are all coming from the same file.  We should add UswdsButtonStyle to the line above it and also try to use the '@' whenever we are going up to a parent folder or beyond.

fmaddenflx (commented):
Overall looks good.

user-interface/src/lib/components/CaseNumberInput.tsx
amorrow-flexion: I am not sure, it works fine without the fragment. I see your name on the commit from last week. We can just remove it

fmaddenflx (approved):

PR #1029, opened 2024-11-12
CAMS-473 - Identify legacy and invalid divisions and filter where necessary

btposey:
Jira ticket: CAMS-473

# Purpose

Identify legacy and invalid divisions and filter where necessary

Remove legacy Islip division from transfer court list.

# Major Changes

* Added lists of invalid and legacy division codes.
* Decorated the division model with tags.
* Added INVALID and LEGACY tags to divisions in the lists.

# Testing/Validation

Manual testing of the API


backend/functions/lib/adapters/gateways/dxtr/offices.dxtr.gateway.ts
```diff
@@ -14,6 +14,11 @@ import { buildOfficeCode, getOfficeName } from '../../../use-cases/offices/offic
 
 const MODULE_NAME = 'OFFICES-GATEWAY';
 
+// Remove invalid divisions at the gateway rather than forcing the
+// more important use case code to include logic to remove them.
```
jamesobrooks, line 18: For legacy codes, I would say this is more of a business logic issue than a persistence issue, but we can leave it for an axis of change in the future to decide whether/how to treat them differently than the invalid ones.

backend/functions/lib/adapters/gateways/storage/local-storage-gateway.ts
```diff
@@ -55,10 +61,33 @@ function getRoleMapping(): Map<string, CamsRole> {
...
+  });
+}
+
+function getUstpDivisionMeta(): Map<string, UstpDivisionMeta> {
+  if (!metaMapping) {
+    metaMapping = new Map<string, UstpDivisionMeta>();
+    addUstpDivisionMetaToMap(metaMapping, { isLegacy: true }, LEGACY_DIVISION_CODES);
+  }
+  return metaMapping;
+}
```
jamesobrooks, line 84: I am not seeing what this code is doing for us. It seems that we're never adding any codes to `LEGACY_DIVISION_CODES` so line 69 has nothing to iterate over.

backend/functions/lib/use-cases/offices/offices.ts
```diff
@@ -17,8 +17,23 @@ const MODULE_NAME = 'OFFICES_USE_CASE';
...
+    const offices = await officesGateway.getOffices(context);
+
+    const storageGateway = getStorageGateway(context);
+    const metas = storageGateway.getUstpDivisionMeta();
+
+    offices.forEach((ustpOffice) => {
+      ustpOffice.groups.forEach((group) => {
+        group.divisions.forEach((division) => {
+          if (metas.has(division.divisionCode)) {
+            division.isLegacy = metas.get(division.divisionCode).isLegacy;
```
jamesobrooks, line 30: With the current definition of UstpDivisionMeta, this assignment seems to violate the type:
```
export type UstpDivisionMeta = {
  isLegacy?: true;
};
```
So if `isLegacy` exists, it should be `true`, but we're potentially setting it to `false`.

backend/functions/lib/controllers/orders/orders.controller.test.ts
```diff
@@ -21,6 +21,7 @@ import { mockCamsHttpRequest } from '../../testing/mock-data/cams-http-request-h
 import { ResponseBody } from '../../../../../common/src/api/response';
 import { NotFoundError } from '../../common-errors/not-found-error';
 import { BadRequestError } from '../../common-errors/bad-request';
+import * as crypto from 'crypto';
```
jamesobrooks, line 24: Why is this added?

btposey: This missing import was causing the test to fail for me locally.

backend/functions/lib/use-cases/offices/offices.ts
btposey: I intend for it to be absent if not applicable.

backend/functions/lib/adapters/gateways/storage/local-storage-gateway.ts
btposey: I did have "070" in here, but Jon confirmed that the division had no legacy cases so I moved it to the "invalid" list in the gateway. I left the legacy logic in place because I'm pretty sure we will be adding legacy division codes in the near term.

backend/functions/lib/adapters/gateways/dxtr/offices.dxtr.gateway.ts
```diff
@@ -14,6 +14,11 @@ import { buildOfficeCode, getOfficeName } from '../../../use-cases/offices/offic
...
+const INVALID_DIVISION_CODES = ['070', '990', '991', '992', '993', '994', '995', '996', '999'];
```
fmaddenflx, line 19: Curious to know where we got this list.

btposey: From our analysis of divisions exported from DXTR. Some are obvious. Then 070 is for Islip. We have more to come in the future as the analysis is completed.

PR #1057, opened 2024-12-09
CAMS-461 brian

btposey:
Exploring multiple paths


amorrow-flexion (commented):
I was following a very similar approach to you Brian, as a high level note though, should we think about renaming the `backend/functions` directory to `backend/function-apps`?

amorrow-flexion (approved):

jamesobrooks (approved):

fmaddenflx:
Looks great.

PR #1090, opened 2024-12-27
CAMS-393 Replace eslint & prettier with Biome

amorrow-flexion:
# Purpose

The intent is to replace eslint & prettier as it is becoming a pain to update with their breaking changes

# Major Changes

- removed eslint, eslint plugins, and prettier
- installed biome at the top level
- added biome.json in each sub directory
- changed precommit to use biome-lint:write for now
- updated package.json scripts accordingly

# Testing/Validation

- ran all scripts locally with package.json file. 
- ran the precommit locally

# Notes

- We will still need to do some playing around and tweaking of Biome but this seems to fit our needs well
- have not used format yet as we will want to visit that
- there are some exclusions within code that we may want to visit, namely in the frontend
- [Biome](https://biomejs.dev/guides/getting-started/)


package.json
```diff
@@ -14,5 +15,8 @@
     "build:frontend": "pushd user-interface && npm run build && popd",
     "build:common": "pushd common && npm run build && popd",
     "clean-and-build": "npm run clean:all && npm run ci:all && npm run build"
+  },
+  "devDependencies": {
+    "@biomejs/biome": "1.9.4"
```
jamesobrooks, line 20: Biome 2.0 is [supposed to support plugins](https://github.com/biomejs/biome/discussions/1649). That said, I'm a bit concerned by the fact that as of yet, the only way to get new rules [is for the maintainers to implement the rules in Rust](https://github.com/biomejs/biome/discussions/1649#discussioncomment-11476815). To be clear, we don't often add rules. I've been wanting to add a rule (from a plugin or custom) to enforce newlines after functions/methods, but have been waiting until we handle the outdated package. We recently added `a11y` linting. So while we don't often add new rules, it does happen.

user-interface/src/login/broadcast-logout.test.ts
```diff
@@ -10,7 +10,7 @@ describe('Broadcast Logout', () => {
     const postMessageSpy = vi
       .spyOn(BroadcastChannelHumble.prototype, 'postMessage')
       .mockReturnValue();
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+    // biome-ignore lint: do not use Function as a type
```
jamesobrooks, line 13: This is the other thing that I don't really like, and actually fairly strongly. The fact that the ignore cannot target specific rules means that we could end up violating rules we actually care about because we are trying to intentionally ignore a different rule in just this one case.

jamesobrooks: https://biomejs.dev/linter/#ignore-code

amorrow-flexion:
Cancelling this, we want to revisit at biome 2.0

amorrow-flexion:
Closing this. We decided against but we want to revisit at a later date when Biome 2.0 is released to see what is available

PR #1091, opened 2024-12-28
CAMS-393: Upgrade eslint with modular configs

jamesobrooks:
# Purpose

We were behind by a major version and 17 minor versions since then.

# Major Changes

- Upgrade all projects to use v9.17.0 of eslint
- Ensure that `pre-commit` is using the latest eslint as well
- Use one top-level config file that applies modular configs as sources

# Testing/Validation

Spot checked by modifying files to ensure that both eslint and prettier rule violations are caught when running both from npm scripts and from `pre-commit`.

# Notes

The recommended eslint rules include rules for frontend frameworks, but it may still make sense to use the `eslint-plugin-react` rules. There are also some other related plugins we may want to consider such as `eslint-plugin-react-hooks` and `@vitest/eslint-plugin`. 

Since we have to have a root-level config, it is possible that the ruleset for any given file would be different when the sub-project config is used compared to the root-level one. I have an idea for a test script that we should probably add. If the output of `eslint --print-config foo.ts` is deterministic, we could print the config to a file from both levels and do a diff. If the exit code of the diff is 0, then the rulesets are the same. If for some reason the output is non-deterministic (i.e. the same rules exist but are printed in random order) then we can still accomplish this but would need to use something like `jq` and/or a JavaScript script to compare them as json objects.

# Definition of Done:

- [ ] Acceptance criteria met - The system meets all acceptance criteria
- [ ] Feature toggles created - Features that are deployed but not released have toggles
- [ ] Design QA passed - All visual UI looks and works the way it’s supposed to
- [ ] Usability validated - Team ensures the working code is easy to use
- [ ] Accessibility tests passed - System passes automated and manual accessibility tests and supports applicable devices
- [ ] Dependency Rule followed - More important code doesn’t directly depend on less important code
- [ ] Source code merged - Code has been merged into the main branch
- [ ] 90% unit test coverage achieved - Automated test coverage tools indicate coverage of >= 90%
- [ ] Code reviewed - Code is reviewed by at least two other team members before being merged
- [ ] Code quality checks passed - Code passes all automated quality checks
- [ ] Security scans passed - Code passes expected scans for vulnerabilities and compliance issues
- [ ] Threat model updated - The threat model incorporates new threats and remediations
- [ ] Build process updated - Automated build process include code supporting the user story
- [ ] Load/performance tests passed - Performance tests include the functionality of the user story
- [ ] Documents are updated - All required documentation is up to date and version controlled
- [ ] Code is deployed - Code is deployed to the highest production like environment


common/jest-eslint.config.mjs
```diff
@@ -0,0 +1,32 @@
...
+  },
+  {
+    rules: {
+      ...jest.configs['flat/recommended']['rules'],
+    },
+  },
+  {
+    settings: {
+      jest: {
+        version: '29.7.0',
```
jamesobrooks, line 26: I tried to determine this version dynamically, but it doesn't work in the `pre-commit.ci` run. There's probably some automation we can do with the dependency updates script to notice a change and either directly update here or at least leave a PR comment letting us know that it needs to be updated.

eslint.config.mjs
```diff
@@ -0,0 +1,43 @@
...
+}));
+const jestConfig = jestEslintConfig.map((configObject) => ({
+  files: [
+    'backend/**/*.test.ts',
+    'common/**/*.test.ts',
+    'dev-tools/**/*.test.ts',
+    'test/e2e/**/*.test.ts',
+  ],
+  ...configObject,
+}));
```
jamesobrooks, line 26: Treating these—source and tests—differently may not be necessary. I was having a problem that turned out to be related to eslint not being able to determine what version of Jest we are using. As it is, it's possible that if one of the sub-project config files changes, the root would produce a different set of rules for files in that sub-project than the lower level would. I do have an idea for how to test for that though, which honestly we should probably add at some point regardless.

If we don't care to separate source and test rulesets, we would need to modify this and the sub-project configs.

common/ts-eslint.config.mjs
jamesobrooks, line 1: This needs to move up one directory.

jamesobrooks: Nevermind.

amorrow-flexion (approved):

amorrow-flexion:
We still want to revisit Biome, we should probably mention this somewhere

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -10,12 +10,11 @@ import { UnknownError } from '../../../common-errors/unknown-error';
...
-    { chapter: '11', querySubString: null, inputVariable: '11' },
-    { chapter: '12', querySubString: null, inputVariable: '12' },
-    { chapter: '13', querySubString: null, inputVariable: '13' },
-    { chapter: '15', querySubString: null, inputVariable: '15' },
-    { chapter: '7', querySubString: "IN ('7A', '7N')", inputVariable: null },
+    { chapter: '9', inputVariable: '09' },
+    { chapter: '11', inputVariable: '11' },
+    { chapter: '12', inputVariable: '12' },
+    { chapter: '13', inputVariable: '13' },
+    { chapter: '15', inputVariable: '15' },
```
fmaddenflx, line 17: Interesting.  Why was chapter 7 removed from this list?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -40,21 +39,43 @@ describe('ACMS gateway tests', () => {
...
+
+    const context = await createMockApplicationContext();
+    const gateway = new AcmsGatewayImpl(context);
+    await gateway.getLeadCaseIds(context, predicate);
+
+    expect(spy).toHaveBeenCalledWith(
+      context,
+      expect.stringContaining("IN ('7A', '7N')"),
+      expect.any(Array),
+    );
```
fmaddenflx, line 78: This probably answers my question above ;-)

fmaddenflx (approved):

common/eslint-jest.config.mjs
```diff
@@ -0,0 +1,38 @@
+import eslintTsConfig from './eslint-ts.config.mjs';
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+
+const tsEslint = require('typescript-eslint');
+const jest = require('eslint-plugin-jest');
+
```
fmaddenflx, line 7: Interesting.  What was the reason for this?

btposey (approved):

common/eslint-js.config.mjs
```diff
@@ -0,0 +1,16 @@
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+
+const tsEslint = require('typescript-eslint');
+const eslint = require('@eslint/js');
+const eslintPluginPrettierRecommended = require('eslint-plugin-prettier/recommended');
```
fmaddenflx, line 6: Same thing here.  createRequire() interesting.  A backward compatibility shim?

common/src/cams/actions.test.ts
```diff
@@ -5,14 +5,17 @@ describe('Actions', () => {
     const resource1 = {
       _actions: [Actions.ManageAssignments],
     };
+    /* eslint-disable-next-line jest/no-standalone-expect */
     expect(Actions.contains(resource1, Actions.ManageAssignments)).toBeTruthy();
```
fmaddenflx, line 9: Interesting.  What's wrong with a stand-alone expect?

common/eslint-jest.config.mjs
btposey: The dependencies are packaged as commonjs, not ES modules.

common/eslint-js.config.mjs
btposey: The dependencies are packaged as commonjs, not ES modules.

common/src/cams/courts.test.ts
```diff
@@ -36,6 +36,7 @@ describe('common court library tests', () => {
     expect(newOfficeList).toEqual(expect.arrayContaining([...expectedOffices]));
   });
 
+  /* eslint-disable-next-line jest/no-identical-title */
   test('should filter court offices list by court division', async () => {
```
fmaddenflx, line 40: We shouldn't have an identical title in here.  Why do we?

common/src/cams/actions.test.ts
btposey: Bad form I guess. I would like to wrap the stand alone expects in a test block.

common/src/name-helper.test.ts
```diff
@@ -46,6 +46,7 @@ describe('Name helper tests', () => {
     ).toEqual('First Middle Last Jr.');
   });
 
+  /* eslint-disable-next-line jest/no-identical-title */
```
fmaddenflx, line 49: We do have an identical title.  Why?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Indeed it does.

fmaddenflx (approved):

PR #1103, opened 2025-01-03
Dependency update fix

amorrow-flexion:
# Purpose

properly trigger deployment on npm updates auto branch

# Major Changes

- addition o workflow-dispatch step

# Testing/Validation

- ran dependency updates auto and verified continuous-deployment.yml was triggered with correct inputs

# Notes

Note: GITHUB_TOKEN auth does not trigger workflow builds in any aspects to prevent infinite looping of workflows. So we have to explcitly call

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -50,4 +50,5 @@ jobs:
         uses: benc-uk/workflow-dispatch@v1
         with:
             workflow: continuous-deployment.yml
-            ref: 'refs/head/dependency-updates-auto'
+            repo: US-Trustee-Program/Bankruptcy-Oversight-Support-Systems
+            ref: ${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/update-dependencies.yml
```diff
@@ -45,3 +45,11 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: ./ops/scripts/utility/audit.sh -c
+
+      - name: Trigger Continuous Deployment
+        uses: benc-uk/workflow-dispatch@v1
```
jamesobrooks, line 50: Can we not simply call `continuous-deployment.yml` as a reusable workflow?

amorrow-flexion: You know what, you're right. Do we want that tacked onto this workflow though? If you have no problem with it, I'll give that a shot

amorrow-flexion: The only caveat there is the parameters for `enableBidepDeployment` and `deployBranch` as below:
`
on: 
   workflow_dispatch:
       enableBicepDeployment: ""
       deployBranch: ""
   workflow_call:
       enableBicepDeployment: ""
       deployBranch: ""
`

jamesobrooks (approved):

PR #1107, opened 2025-01-03
CAMS-491 - Ambiguous transfer indication

fmaddenflx:
# Purpose

Display information on Case Details screen about Ambiguous Transfers

# Major Changes

* Modified Cases Dxtr gateway to return transfer date.
* Mofified Case Detail screen to display ambiguous transfer information

# Testing/Validation

Added tests to backend and frontend

### DIFF
--- a/backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.ts
+++ b/backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.ts
@@ -21,6 +21,7 @@ const MODULE_NAME = 'CASES-DXTR-GATEWAY';
 const closedByCourtTxCode = 'CBC';
 const dismissedByCourtTxCode = 'CDC';
 const reopenedDateTxCode = 'OCO';
+const orderToTransferCode = 'CTO';
 
 export function getCaseIdParts(caseId: string) {
   const parts = caseId.split('-');
@@ -35,11 +36,12 @@ export default class CasesDxtrGateway implements CasesInterface {
       ...caseSummary,
     };
 
-    const transactionDates = await this.queryTransactions(
+    const transactionDates = await this.queryTransactionDates(
       applicationContext,
       bCase.dxtrId,
       bCase.courtId,
     );
+
     if (transactionDates.closedDates.length > 0) {
       bCase.closedDate = getMonthDayYearStringFromDate(transactionDates.closedDates[0]);
     }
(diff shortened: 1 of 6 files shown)


user-interface/src/case-detail/panels/CaseDetailOverview.tsx
```diff
@@ -43,6 +43,13 @@ export default function CaseDetailOverview(props: CaseDetailOverviewProps) {
     openModalButtonRef.current?.focus();
   }
 
+  const isAmbiguousTransferIn =
+    !!caseDetail.petitionCode && ['TI', 'TV'].includes(caseDetail.petitionCode);
```
amorrow-flexion, line 47: should we export `transferCodes` from petition-gateway.ts and use them here?

btposey (approved):

user-interface/src/case-detail/panels/CaseDetailOverview.test.tsx
```diff
@@ -281,6 +281,34 @@ describe('Case detail basic information panel', () => {
...
+    test('should display information about case being transfered in when there is no verified transfer', async () => {
+      const transferredCase = {
+        ...BASE_TEST_CASE_DETAIL,
+        petitionCode: 'TI',
+      };
+      renderWithProps({ caseDetail: transferredCase });
+
+      expect(screen.queryByTestId('verified-transfer-header')).not.toBeInTheDocument();
+      const ambiguousTransferText = screen.queryByTestId('ambiguous-transfer-text');
+      expect(ambiguousTransferText?.textContent).toEqual(
```
amorrow-flexion, line 307: we can probably just use `expect(ambiguousTransferText?).toHaveTextContent()` instead here

btposey: Better. Agreed.

user-interface/src/case-detail/panels/CaseDetailOverview.tsx
btposey: Yeah. If we have string constants defined we should use them.

backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.ts
```diff
@@ -21,6 +21,7 @@ const MODULE_NAME = 'CASES-DXTR-GATEWAY';
 const closedByCourtTxCode = 'CBC';
 const dismissedByCourtTxCode = 'CDC';
 const reopenedDateTxCode = 'OCO';
```
amorrow-flexion, line 23: I feel like we should put these somewhere and reference them as exports

btposey: Good idea. But punting today.

user-interface/src/case-detail/panels/CaseDetailOverview.tsx
btposey: But.... Gateway specific and in backend only. Not in FE or common. Punting for now.

amorrow-flexion (approved):

PR #1111, opened 2025-01-06
Add testing library linting

jamesobrooks:
# Purpose

We want to use the `testing-library` package as intended most of the time.

# Major Changes

- Add the eslint plugin provided by the package authors.
  - Configure the default rules leaving some turned off for further consideration before enabling.
  - Correct new linting errors.

# Testing/Validation

Automated tests still pass.

# Future Work

We could still add the [vitest plugin](https://www.npmjs.com/package/@vitest/eslint-plugin).

### DIFF
--- a/user-interface/eslint-ui-test.config.mjs
+++ b/user-interface/eslint-ui-test.config.mjs
@@ -1,0 +1,31 @@
+import eslintUiConfig from './eslint-ui.config.mjs';
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+
+const tsEslint = require('typescript-eslint');
+const testingLibrary = require('eslint-plugin-testing-library');
+
+/**
+ * eslintUiTestConfig
+ *
+ * This ConfigArray is intended to be the eslint configuration for TypeScript
+ * family test files (e.g. `.test.ts`, `.test.tsx`).
+ */
+const eslintUiTestConfig = tsEslint.config(
+  eslintUiConfig,
+  {
+    plugins: testingLibrary.configs['flat/react']['plugins'],
+  },
+  {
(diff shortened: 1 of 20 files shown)


user-interface/eslint-ui-test.config.mjs
```diff
@@ -0,0 +1,44 @@
...
+    'testing-library/no-unnecessary-act': 'error',
+    'testing-library/no-wait-for-multiple-assertions': 'warn', // default: error
+    'testing-library/no-wait-for-side-effects': 'warn', // default: error
+    'testing-library/no-wait-for-snapshot': 'error',
+    'testing-library/prefer-find-by': 'error',
+    'testing-library/prefer-presence-queries': 'warn', // default: error
+    'testing-library/prefer-query-by-disappearance': 'error',
+    'testing-library/prefer-screen-queries': 'warn', // default: error
+    'testing-library/render-result-naming-convention': 'error',
+  },
```
jamesobrooks, line 41: Using the provided flat config causes a lot of errors I'm not sure we care about. For example, `no-node-access` produces a lot of errors for DOM traversal which I'm not sure we want to follow their rule. There are others that may be good to move toward, but would cause major refactoring of some of our tests such as `no-wait-for-multiple-assertions`. I've marked all rules that deviate from the default with a comment.

jamesobrooks: If we decide we want to not turn on a rule we could leave it here and set to `off`, or we could remove it completely here.

amorrow-flexion:
We will want to continue this work by using the suggestions the linter found for the frontend
specifically 
- avoiding multiple assertions inside of a waitFor
- fireEvent inside of waitFor blocks

amorrow-flexion (approved):

PR #1113, opened 2025-01-06
CAMS-492: Only get chapter 15 and enabled chapters

jamesobrooks:
# Problem

The staff assignment screen should only look for cases for enabled chapters which means 15 by default and 11 and/or 12 if the feature flags are turned on.

# Solution

We modify the search predicate to only include chapter 15 by default and other chapters if enabled by feature flag.

# Testing/Validation

Automated tests modified to test this.


jamesobrooks:
Note that I was getting some errors when running the user-interface tests. We may have some improperly handled async functions.

amorrow-flexion (approved):

PR #1118, opened 2025-01-07
CAMS-497 - Retry JWT verification when ECONNRESET encountered

btposey:
Jira ticket: CAMS-497

# Purpose

Retry JWT verification when a ECONNRESET error is encountered.

# Major Changes

Added retry logic around JWT verification code


backend/lib/adapters/gateways/okta/okta-gateway.ts
```diff
@@ -22,6 +22,40 @@ type OktaUserInfo = {
...
+): Promise<object> {
+  try {
+    return await verifyAccessToken(issuer, token, audience);
+  } catch (originalError) {
+    if (retry && isJwtParseError(originalError) && originalError.innerError.code === 'ECONNRESET') {
+      return await verifyAccessToken(issuer, token, audience);
+    }
+    throw originalError;
+  }
+}
```
jamesobrooks, line 57: @btposey, I wonder if we should try a few times with an exponential backoff. Since we're not seeing this behavior in other environments, I'm not sure I'm confident with a single retry.

jamesobrooks (approved):

PR #1139, opened 2025-01-10
CAMS-502: Properly handle modified result for upserts

jamesobrooks:
# Problem

Upserts that modify documents are incorrectly identified as having failed.

# Solution

Properly identify modified count as an indicator of success for upserts. Extend the ttl back out past 24 hours to ensure that documents exist at least until the next sync runs.

# Testing/Validation

Updated all related automated tests. Also exercised the endpoint to ensure that upserts and replacements work.

### DIFF
--- a/backend/lib/factory.ts
+++ b/backend/lib/factory.ts
@@ -56,6 +56,7 @@ let orderSyncStateRepo: RuntimeStateRepository<OrderSyncState>;
 let officeStaffSyncStateRepo: RuntimeStateRepository<OfficeStaffSyncState>;
 
 let mockOrdersRepository: MockMongoRepository;
+let mockOfficesRepository: MockMongoRepository;
 let mockConsolidationsRepository: MockMongoRepository;
 let mockCasesRepository: MockMongoRepository;
 let mockUserSessionCacheRepository: MockMongoRepository;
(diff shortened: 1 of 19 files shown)


backend/lib/factory.ts
```diff
@@ -116,10 +117,10 @@ export const getOfficesGateway = (context: ApplicationContext): OfficesGateway =
 
 export const getOfficesRepository = (context: ApplicationContext): OfficesRepository => {
   if (context.config.get('dbMock')) {
-    if (!mockOrdersRepository) {
-      mockOrdersRepository = MockMongoRepository.getInstance(context);
+    if (!mockOfficesRepository) {
```
jamesobrooks, line 120: Note that a previous PR for CAMS-500 replaced the MockOfficesRepository with the MockMongoRepository in an attempt to let the Flexion environment use staff from the db rather than hard-coded staff. This is not currently working. If we want to revert to the previous use of the hard-coded staff, we can for now.

btposey (approved):

amorrow-flexion (approved):

backend/lib/factory.ts
jamesobrooks: Reverting to prior behavior for now.

PR #1145, opened 2025-01-14
CAMS-500 - Provide admin to augment user access

btposey:
# Purpose

To provide a way to augment privileged users with additional permissions and to modify attorneys assigned to offices.

# Major Changes

* Added support to augment users
* Added support to add augmented users as office attorneys.

# Testing/Validation

* Unit test coverage


backend/lib/adapters/gateways/storage/local-storage-gateway.ts
```diff
@@ -1,19 +1,21 @@
...
+import { NotFoundError } from '../../../common-errors/not-found-error';
+
+const MODULE_NAME = 'LOCAL_STORAGE_GATEWAY';
 
 let roleMapping: Map<string, CamsRole>;
 export const ROLE_MAPPING_PATH = '/rolemapping.csv';
 const ROLE_MAPPING =
   'ad_group_name,idp_group_name,cams_role\n' +
+  'USTP_CAMS_Super_User,USTP CAMS Super User,SuperUser\n' +
+  'USTP_CAMS_Augmentable_User,USTP CAMS Augmentable User,AugmentableUser\n' +
```
governmentSponsored, line 13: I think this should be `USTP_CAMS_Privileged_Identity_Management,USTP CAMS Privileged Identity Management...`

common/src/cams/users.ts
```diff
@@ -25,6 +25,14 @@ export type CamsUserGroup = {
   users?: CamsUser[];
 };
 
+export type PrivilegedIdentityUser = CamsUserReference & {
+  documentType: 'PRIVILEGED_IDENTITY_USER';
+  claims: {
+    groups: string[];
+  };
+  expires?: string;
```
btposey, line 33: Add logic to make use of the `expires` property?

backend/lib/use-cases/user-session/user-session.ts
```diff
@@ -55,6 +62,28 @@ export class UserSessionUseCase {
...
+          // Silently log the failure so the user can continue without augmentation.
+          context.logger.error(
+            MODULE_NAME,
+            `Failed to augment user ${user.name} (${user.id}).`,
+            error.message,
+          );
+        }
+      }
+
+      // TODO: Maybe delete this 'restrict-case-assignment' feature flag.
```
btposey, line 86: Remove this feature flag?

backend/lib/use-cases/user-session/user-session.test.ts
```diff
@@ -139,6 +259,7 @@ describe('user-session.gateway test', () => {
     await expect(gateway.lookup(context, jwtString, provider)).rejects.toThrow(ServerConfigError);
   });
 
+  // TODO: This should be removed once the feature flag logic is removed.
```
btposey, line 262: ...and remove this along with the feature flag if the feature flag is removed.

backend/lib/adapters/gateways/storage/local-storage-gateway.ts
```diff
@@ -1,19 +1,23 @@
...
 export const ROLE_MAPPING_PATH = '/rolemapping.csv';
 const ROLE_MAPPING =
   'ad_group_name,idp_group_name,cams_role\n' +
+  'USTP_CAMS_Super_User,USTP CAMS Super User,SuperUser\n' +
+  'USTP_CAMS_PRIVILEGED_IDENTITY_USER,USTP CAMS Privileged Identity User,PrivilegedIdentityUser\n' +
   'USTP_CAMS_Case_assignment_Manager,USTP CAMS Case Assignment Manager,CaseAssignmentManager\n' +
   'USTP_CAMS_Trial_Attorney,USTP CAMS Trial Attorney,TrialAttorney\n' +
   'USTP_CAMS_Data_Verifier,USTP CAMS Data Verifier,DataVerifier\n';
 
+// TODO: We should delete this. How are we handling AD group from Seattle mapped to SE and AK groups??
```
btposey, line 18: Need to do some research here. See what we've changed. Fix any breakage for the Seattle office edge case.

backend/lib/controllers/admin/staff-admin.controller.ts
```diff
@@ -22,10 +21,8 @@ export class AdminController implements CamsController {
     }
 
     try {
-      if (procedure === deleteMigrations && context.request.method === 'DELETE') {
-        await useCase.deleteMigrations(context);
-        return { statusCode: 204 };
-      } else if (procedure === createStaff && context.request.method === 'POST') {
+      // TODO: we should probably now treat this more RESTfully and not use procedure
```
btposey, line 24: Make this RESTful.

backend/lib/use-cases/admin/admin.ts
```diff
@@ -69,4 +80,108 @@ export class AdminUseCase {
       });
     }
   }
+
+  // TODO: Expose the PrivilegedIdentityUser functions through the web API to drive drop down in the UI.
```
btposey, line 84: Still need the admin controller and function for privileged identity user functions.

backend/lib/adapters/gateways/dxtr/offices.dxtr.gateway.ts
```diff
@@ -40,6 +40,7 @@ type DxtrFlatOfficeDetails = {
   staff?: CamsUserReference[];
 };
 
+// TODO: Maybe we need to add configuration options here for the Seatle => SE+AK mapping edge case.
```
btposey, line 43: See comments about the Seattle edge case.

btposey (commented):
We need a JIRA ticket to refactor gross mock singletons logic in the `factory.ts` file.

backend/lib/adapters/gateways/storage/local-storage-gateway.ts
btposey: Done.

btposey (commented):
Need to integrate feature flag into API for the privilege logic.

common/src/cams/users.ts
jamesobrooks: Done.

backend/lib/use-cases/user-session/user-session.ts
jamesobrooks: Removed.

backend/lib/use-cases/user-session/user-session.test.ts
jamesobrooks: Removed.

backend/lib/use-cases/admin/admin.ts
jamesobrooks: Done.

backend/lib/controllers/admin/privileged-identity-admin.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+    context: ApplicationContext,
+  ): Promise<CamsHttpResponseInit<object | undefined>> {
+    try {
+      if (!context.featureFlags['privileged-identity-management']) {
+        throw new ForbiddenError(MODULE_NAME, { message: NOT_ENABLED });
+      }
+
+      if (!context.session.user.roles.includes(CamsRole.SuperUser)) {
+        throw new ForbiddenError(MODULE_NAME);
+      }
```
jamesobrooks, line 26: Note this from the RFC:
> If authentication credentials were provided in the request, the server considers them insufficient to grant access. The client SHOULD NOT automatically repeat the request with the same credentials.
I've applied that to the lack of SuperUser. Unauthorized also seems to be reasonable, but really because of the indication that the client should not retry with the same credentials, Forbidden seems to be indicated. This also matches my experience with most API's, with 401 being applied when there is some issue with the user's credentials generally, but 403 when the credentials are good but do not provide access to the specific resource.

But also notice this:
> However, a request might be forbidden for reasons unrelated to the credentials.
I've applied this to the feature flag being turned off. Unauthorized doesn't feel right for a feature that is turned off. Since the user should not be able to make such a request without the flag turned on, Forbidden seems to fit because it means the user has circumvented our client code to force the request.

[HTTP RFC](https://www.rfc-editor.org/rfc/rfc9110#name-403-forbidden)

backend/lib/controllers/admin/staff-admin.controller.ts
btposey: We are deleting the staff admin controller and function.

jamesobrooks:
> We need a JIRA ticket to refactor gross mock singletons logic in the `factory.ts` file.

I disagree with putting it off to another ticket. We should be eliminating debt. I think we should handle this in another PR before we move on to something else.

backend/lib/testing/mock-gateways/mock-user-session-use-case.ts
```diff
@@ -2,18 +2,17 @@ import * as jwt from 'jsonwebtoken';
...
 const cache = new Map<string, CamsSession>();
 
 export class MockUserSessionUseCase {
   async lookup(
-    context: ApplicationContext,
+    _context: ApplicationContext,
     accessToken: string,
     provider: string,
   ): Promise<CamsSession> {
+    // TODO: MAYBE we overlay the PIM record using the user helper function.
```
btposey, line 15: I think we agreed to not worry about overlaying the PIM record with the "mock/local" user sessions. Testing would be done in dev with Okta enabled as the auth provider.
```suggestion
```

fmaddenflx (approved):

fmaddenflx (approved):

PR #1161, opened 2025-01-24
CAMS-302: Only use default flag values for testing

jamesobrooks:
# Purpose

We do not want to default to features being turned on simply to ensure that a11y tests can validate features that are behind flags.

# Major Changes

- Only return feature flag values as true by default if we are testing.

# Testing/Validation

- Updated automated tests.
- Validated locally in the following scenarios:
  - Valid client id for Launch Darkly.
  - Invalid client id for Launch Darkly.
  - `CAMS_PA11Y` set to true.
  - `CAMS_PA11Y` set to false and no client id for Launch Darkly.

# Notes

This will also prevent the behavior we see on refresh where a Case Assignment Manager who refreshes the screen while viewing the Staff Assignment screen will see chapter 11 cases despite that feature flag being turned off.

### DIFF
--- a/user-interface/src/lib/hooks/UseFeatureFlags.ts
+++ b/user-interface/src/lib/hooks/UseFeatureFlags.ts
@@ -1,6 +1,6 @@
 import { useFlags } from 'launchdarkly-react-client-sdk';
-import { defaultFeatureFlags, FeatureFlagSet } from '@common/feature-flags';
-import { getFeatureFlagConfiguration } from '../../configuration/featureFlagConfiguration';
+import { FeatureFlagSet, testFeatureFlags } from '@common/feature-flags';
+import { getFeatureFlagConfiguration } from '@/configuration/featureFlagConfiguration';
 
 export const CHAPTER_ELEVEN_ENABLED = 'chapter-eleven-enabled';
 export const CHAPTER_TWELVE_ENABLED = 'chapter-twelve-enabled';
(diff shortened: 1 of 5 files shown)


user-interface/src/lib/hooks/UseFeatureFlags.ts
```diff
@@ -12,10 +12,9 @@ export const PRIVILEGED_IDENTITY_MANAGEMENT = 'privileged-identity-management';
 
 export default function useFeatureFlags(): FeatureFlagSet {
   const config = getFeatureFlagConfiguration();
-  if (!config.clientId) return defaultFeatureFlags;
+  if (import.meta.env['CAMS_PA11Y'] === 'true') return testFeatureFlags;
```
jamesobrooks, line 15: We might want something like this instead:
```suggestion
  if (import.meta.env['CAMS_PA11Y'] === 'true' || import.meta.env['DEFAULT_FEATURE_FLAGS_TRUE'] === 'true') return testFeatureFlags;
```

This would allow us to test locally easier while still being able to hit a locally-running backend. However, that would potentially cause issues with the flag values for the frontend and backend not matching. I'd probably just default to hitting Launch Darkly and changing the values through the portal there, which is a better test anyway, although potentially a worse developer experience.

fmaddenflx (approved):
Looks good.  We could implement your condition suggestion perhaps.

PR #1170, opened 2025-01-30
CAMS-421 Change migration function name to dataFlow

btposey:
Purpose

Rename migration function app name references to pipelines in package.json and biceps.

# Major Changes

Simple renaming.


.github/workflows/reusable-backend-slot.yml
```diff
@@ -54,19 +54,19 @@ jobs:
...
-              --migrationFunctionName ${{ inputs.migrationFunctionName }} \
+              --pipelinesFunctionName ${{ inputs.pipelinesFunctionName }} \
               --webappName ${{ inputs.webappName }} \
               --slotName ${{ inputs.slotName }} \
               --kvIdName ${{ secrets.AZ_KV_APP_CONFIG_MANAGED_ID }} \
               --sqlIdName ${{ secrets.AZ_SQL_IDENTITY_NAME }} \
-              --apiStorageAccName ${migrationSlotStorageAccountName:0:24} \
-              --migrationStorageAccName ${apiSlotStorageAccountName:0:24} \
+              --apiStorageAccName ${pipelinesSlotStorageAccountName:0:24} \
+              --pipelinesStorageAccName ${apiSlotStorageAccountName:0:24} \
```
amorrow-flexion, line 70: I know the name pipelines fits well for our use case... but in the terms of a storage Account I could see this causing confusion, especially should we choose to move towards and hosted artifact based deployment structure rather than what we do currently which is utilization of GitHub pipeline artifacts

amorrow-flexion: I know its long but I think pipelinesFunctionStorageAccName might be the best fit if we want to continue with the naming

btposey: I corrected the swapped, mismatched argument values.

amorrow-flexion (approved):

PR #1176, opened 2025-01-31
CAMS-516 UseOutsideClick now acts on mousedown, not click

fmaddenflx:
# Purpose

Bug Fix for -- comboBox outside click action was swallowing up the click event, such that if a user has the comboxOpen, and then clicks on a button, the comboBox closes but the button click is not handled.

# Major Changes

Adjusted the UseOutsideClick hook to handle a "mousedown" event rather than a "click" event.

# Testing/Validation

Manually tested using mouse and verified that keyboard interaction still works as expected.

### DIFF
--- a/user-interface/src/lib/hooks/UseOutsideClick.ts
+++ b/user-interface/src/lib/hooks/UseOutsideClick.ts
@@ -1,5 +1,11 @@
 import React from 'react';
 
+// NOTE:  This hook acts on "mousedown" rather than "click" so that it does
+// not disturb other click events. For instance, if a comboBox dropdown
+// popup is open, and a user clicks on a button, we don't want to swallow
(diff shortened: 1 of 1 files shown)


user-interface/src/lib/hooks/UseOutsideClick.ts
fmaddenflx, line 1: Does it make sense to rename this hook, since it's no longer acting on "click" but rather "mousedown" ?

I think that perhaps we shouldn't rename it, because it's easier to remember that "hey I need to add an outside click handler to this thing" than it is to remember "hey I need to add a handler for mousedown to this thing."

jamesobrooks (approved):

jamesobrooks:
I'm not sure we need the comment, but looks good.

amorrow-flexion (approved):

PR #1177, opened 2025-01-31
Cosmos Database Deployment and bicep Updates

amorrow-flexion:
# Purpose

when working with an error we were seeing related to deployment of the users collection, I noticed there was an opportunity to streamline the deployment workflow for the cosmos e2e database and push it into one template and simplify the process. to keep us up to date I also switched all of our az/login GHA usages to v2 instead of v1

# Major Changes

- update all azure/login@v1 to v2
- minimize and simplify cosmos database deployment bicep

# Testing/Validation

- ran through pipeline and tested deployment of database

# Notes

- I left branchHaseId in the cosmos deployment script for now to avoid having to update USTP but we can remove it and pass in the already generated name to the script if we choose to.

### DIFF
--- a/ops/scripts/pipeline/slots/az-func-slot-deploy.sh
+++ b/ops/scripts/pipeline/slots/az-func-slot-deploy.sh
@@ -52,7 +52,7 @@ while [[ $# -gt 0 ]]; do
 done
 
 # set rule_name in case of exit or other scenarios
-rule_name="agent-${app_name:0:26}" # rule name has a 32 character limit
+rule_name="agent-slot-${app_name:0:26}" # rule name has a 32 character limit
 
 function on_exit() {
     # always try to remove temporary access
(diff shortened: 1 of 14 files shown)


ops/scripts/pipeline/slots/az-func-slot-deploy.sh
```diff
@@ -52,7 +52,7 @@ while [[ $# -gt 0 ]]; do
 done
 
 # set rule_name in case of exit or other scenarios
-rule_name="agent-${app_name:0:26}" # rule name has a 32 character limit
+rule_name="agent-slot-${app_name:0:21}" # rule name has a 32 character limit
```
amorrow-flexion, line 55: This was interfering on occassion by incidentally emoving a rule required by deployment of api and api slot at the same time for a new branch

fmaddenflx (approved):

PR #1186, opened 2025-02-05
CAMS-515 login 404

btposey:
# Problem

Users encounter the "404" screen after returning to the app following a session timeout.

# Solution

* A new GoHome component was added to the routing configuration.
* The GoHome component encapsulates logic for where to navigate the user when the route is ambiguous.
* The GoHome component currently sends all users to the my cases screen.
* The GoHome component is configured for the root/index and wildcard routes.

# Testing/Validation

* Manual testing
* Unit testing


btposey:
This PR also contains a fix to the Okta humble to return a CAMS Unauthorized error when a 401 response is encountered from the Okta API. We believe this fixes CAMS-513.

backend/lib/humble-objects/okta-humble.ts
```diff
@@ -98,12 +99,20 @@ export class OktaHumble {
...
-          message: 'Failed to retrieve user.',
-          module: MODULE_NAME,
-        },
-      });
+      if (
+        isOktaApiError(originalError) &&
+        originalError.status === 401 &&
+        originalError.errorCode === 'E0000011'
+      ) {
+        throw new UnauthorizedError(MODULE_NAME, { originalError });
```
btposey, line 107: This is likely the fix for CAMS-513.

btposey: This selectively returns a 401 to the client whereas the prior solution cast ANY error as a 401 to the client. So we are still leaving the client in a ambiguous state when. we return a 500 for any other error. We should specifically handle the 500 condition.

user-interface/src/login/Session.tsx
```diff
@@ -73,6 +73,8 @@ export function Session(props: SessionProps) {
     if (LOGIN_PATHS.includes(location.pathname)) navigate(LOGIN_SUCCESS_PATH);
   }, [state.isLoaded === true]);
 
+  useEffect(() => {}, [state.isError]);
```
btposey, line 76: This is an incomplete thought. Remove or work it into a more thorough solution to handle the error state.

backend/lib/humble-objects/okta-humble.ts
```diff
@@ -98,12 +99,20 @@ export class OktaHumble {
...
-      throw getCamsErrorWithStack(originalError, MODULE_NAME, {
-        camsStackInfo: {
-          message: 'Failed to retrieve user.',
-          module: MODULE_NAME,
-        },
-      });
+      if (
+        isOktaApiError(originalError) &&
+        originalError.status === 401 &&
+        originalError.errorCode === 'E0000011'
```
btposey, line 105: This may be too specific. Maybe just map all 401 status to the Unauthorized error.
```suggestion
```

user-interface/src/login/Session.tsx
fmaddenflx: Removed.  Doesn't seem necessary.

backend/lib/humble-objects/okta-humble.ts
fmaddenflx: Agreed.  Done.

backend/function-apps/azure/application-context-creator.ts
```diff
@@ -84,6 +84,12 @@ async function getApplicationContextSession(context: ApplicationContext) {
     accessToken,
     context.config.authConfig.provider,
   );
+  // TODO: I don't think we are handling errors from sessionUseCase.lookup properly.
+  // There is also no indication to the end user that anything has gone wrong, and
+  // there doesn't seem to be anyway to trigger a new login prompt in the UI. So user
+  // gets a never ending spinner and even if you go to /logout, there is still no way
+  // to get a login prompt.
+  // The error I am getting seems to be a timeout error.
```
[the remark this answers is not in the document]
fmaddenflx: Yup.

user-interface/src/login/AccessDenied.tsx
```diff
@@ -11,10 +10,10 @@ export type AccessDeniedProps = {
...
 
 export function AccessDenied(props: AccessDeniedProps) {
-  const navigate = useNavigate();
-
   function handleLoginRedirect() {
-    navigate(LOGIN_PATH);
+    const { host, protocol } = window.location;
+    const logoutUri = protocol + '//' + host + LOGIN_PATH;
+    window.location.assign(logoutUri);
   }
```
btposey, line 17: Why was the use of the useNavigate hook removed? Also rename `logoutUri` to `loginUri` if this code is retained.

user-interface/src/lib/components/GoHome.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useNavigate } from 'react-router-dom';
+
+export type GoHomeProps = {
+  path?: string;
+};
+
+/**
+ * GoHome
+ *
+ * FUTURE: This component can be crafted to redirect to legacy routes that no longer exist in the app.
```
btposey, line 11:
```suggestion
 * FUTURE: This component can be crafted to redirect to new routes from legacy routes that no longer exist in the app.
```

user-interface/src/lib/components/GoHome.tsx
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * FUTURE: This component dynamically route based on the user's role in CAMS.
```
btposey, line 13:
```suggestion
 * FUTURE: This component could dynamically route to a "home" based on the user's role in CAMS.
```

user-interface/src/login/AccessDenied.tsx
fmaddenflx: Navigate was not working, even adding an option to navigate to force it to do refresh didn't work.  window.location properly forces a reload.

Also Renamed variable to loginUri.

fmaddenflx (approved):

PR #1191, opened 2025-02-10
Cams 514 reenable veracode sca scans on all commits

governmentSponsored:
# Purpose

Reenable the veracode SCA scan steps

# Major Changes

- Veracode action version
- names of artifacts

# Testing/Validation

Ran action several times to make sure it would complete successfully

# Notes

I'm not sure if the version `artifactname` needs to be kept or if we can resume using the normal versioning convention. Will follow-up with support.

### DIFF
--- a/.github/workflows/sub-security-scan.yml
+++ b/.github/workflows/sub-security-scan.yml
@@ -7,30 +7,35 @@ jobs:
     uses: ./.github/workflows/reusable-sca-scan.yml
     with:
       path: user-interface
+      artifactname: ui
     secrets: inherit # pragma: allowlist secret
 
   sca-scan-backend-lib:
     uses: ./.github/workflows/reusable-sca-scan.yml
     with:
       path: backend
+      artifactname: backend
     secrets: inherit # pragma: allowlist secret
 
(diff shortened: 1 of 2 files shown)


amorrow-flexion (approved):

.github/workflows/sub-security-scan.yml
```diff
@@ -7,30 +7,35 @@ jobs:
     uses: ./.github/workflows/reusable-sca-scan.yml
     with:
       path: user-interface
+      artifactname: ui
```
amorrow-flexion, line 10: can we renamme this to user-interface for consistency?

.github/workflows/sub-security-scan.yml
```diff
@@ -7,30 +7,35 @@ jobs:
...
     secrets: inherit # pragma: allowlist secret
 
   sca-scan-backend-lib:
     uses: ./.github/workflows/reusable-sca-scan.yml
     with:
       path: backend
+      artifactname: backend
```
amorrow-flexion, line 17: some here, rename to backend-lib

PR #1192, opened 2025-02-10
updated e2e tests to reflect our new changes

amorrow-flexion:
Jira ticket: CAMS-515 && 513

# Problem

The changes we introduced has altered the way session routes the browser and redirects the page. This was not previously corrected in the e2e tests so they continued to fail/

# Solution

- Update routing to prevent redundant routing
- Correct previously improper routing
- Update E2E tests to properly exercise functionality

# Testing/Validation

- ran through a deployment

# Other Changes

# Note
- We are still seeing session being cleared when the user directly manipulates the url within mock mode. This is specific to the hosted application, locally it seems to be okay.

### DIFF
--- a/user-interface/src/App.tsx
+++ b/user-interface/src/App.tsx
@@ -2,16 +2,12 @@ import { Routes, Route } from 'react-router-dom';
 import { Header } from './lib/components/Header';
 import { AppInsightsErrorBoundary } from '@microsoft/applicationinsights-react-js';
 import { useAppInsights } from './lib/hooks/UseApplicationInsights';
-import { createContext, useEffect, useRef, useState } from 'react';
+import { createContext, useRef } from 'react';
 import { withLDProvider } from 'launchdarkly-react-client-sdk';
 import { getFeatureFlagConfiguration } from './configuration/featureFlagConfiguration';
 import CaseDetailScreen from './case-detail/CaseDetailScreen';
 import ScrollToTopButton from './lib/components/ScrollToTopButton';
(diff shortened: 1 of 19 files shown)


governmentSponsored (approved):

user-interface/src/App.tsx
```diff
@@ -21,49 +21,70 @@ import GlobalAlert, { GlobalAlertRef } from './lib/components/cams/GlobalAlert/G
...
   const { reactPlugin } = useAppInsights();
   const flags = useFeatureFlags();
 
   const globalAlertRef = useRef<GlobalAlertRef>(null);
+  const [transfersEnabled, setTransfersEnabled] = useState<boolean>(false);
+  const [privilegedIdentityEnabled, setPrivilegedIdentityEnabled] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsLoading(false);
+  }, [Object.keys(flags).length > 0]);
```
fmaddenflx, line 40: I wonder if there will ever be a time when we don't have any flags at all.  This may not be a good condition to set isLoading to false.  Happy to get some ideas from others.

jamesobrooks, line 40: What happens if for some reason flags length _is_ 0? This shouldn't happen, but we no longer use a default, so if for some reason we fail to retrieve flags from LaunchDarkly, the portions of CAMS not behind flags should still work.

amorrow-flexion: We should push this down into their respective components instead of handling feature flags at the higher level.

PR #1200, opened 2025-02-17
CAMS-522 fixed Regex on isValidUserInput 

amorrow-flexion:
Jira ticket: CAMS-522

# Problem

We were seeing an issue where normal words were being rejected and not saved with notes. 

# Solution

- The JAVACSRIPT regex was causing issues within the sanitization. I temporarily commented it out until we figure out what we actually want to put there.
- The rejection post calls were also doing an odd order of operations for valid user input and sanitization that was oppposite of the way we really wanted. So That was fixed as well
# Testing/Validation

- updated unit tests validating the change
- ran locally and verified

# Other Changes

to get this nailed down we just need to decide what goes into the JAVASCRIPT regex and wrap up the tests for it

### DIFF
--- a/common/src/cams/sanitization.ts
+++ b/common/src/cams/sanitization.ts
@@ -1,7 +1,9 @@
 const MONGO_CONSOLE_INJECTED_PATTERN = RegExp(
   /\b(?:db\.[a-zA-Z]+|mongo\.[a-zA-Z]+|(?:find|insert|update|delete|aggregate|create|drop|remove|replace|count|distinct|mapReduce|save)\b\s*(?:\(|{))/i,
 );
-const JAVASCRIPT_INJECTED_PATTERN = RegExp(/<script[\s\S]*?>[\s\S]*?<\/script>/i);
+const JAVASCRIPT_INJECTED_PATTERN = RegExp(
+  /(eval\s*\(.*?\)|window\.\w+|document.\.\w+|<script[\s\S]*?>[\s\S]*?<\/script>)/i,
+);
 const MONGO_QUERY_INJECTED_PATTERN = RegExp(
   /\$(eq|ne|gt|gte|lt|lte|in|nin|and|or|not|regex|exists|type|mod|text|where|all|elemMatch|size)/,
 );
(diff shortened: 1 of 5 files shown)


common/src/cams/sanitization.ts
```diff
@@ -1,9 +1,7 @@
 const MONGO_CONSOLE_INJECTED_PATTERN = RegExp(
   /\b(?:db\.[a-zA-Z]+|mongo\.[a-zA-Z]+|(?:find|insert|update|delete|aggregate|create|drop|remove|replace|count|distinct|mapReduce|save)\b\s*(?:\(|{))/i,
 );
-const JAVASCRIPT_INJECTED_PATTERN = RegExp(
```
amorrow-flexion, line 4: Potentially find a way to be more targeted with this pattern. e.g eval()

common/src/cams/sanitization.ts
```diff
@@ -2,7 +2,7 @@ const MONGO_CONSOLE_INJECTED_PATTERN = RegExp(
   /\b(?:db\.[a-zA-Z]+|mongo\.[a-zA-Z]+|(?:find|insert|update|delete|aggregate|create|drop|remove|replace|count|distinct|mapReduce|save)\b\s*(?:\(|{))/i,
 );
 const JAVASCRIPT_INJECTED_PATTERN = RegExp(
-  /\b(?:eval|Function|with|document\.|window\.|alert|prompt|confirm|fetch\s*\(|setTimeout|setInterval)|<script[\s\S]*?>[\s\S]*?<\/script>/i,
+  /\b(?:fetch\s*\(.*?\)|eval\s*\(.*?\)|window\.[a-zA-Z]+|document\.[a-zA-Z]+|(?:<script[\s\S]*?>[\s\S]*?<\/script>))/i,
```
fmaddenflx, line 5: Yes.  Exactly what I was thinking.

fmaddenflx: This was failing tests however.  I changed the regex so that tests all pass.

jamesobrooks (approved):

btposey (approved):

PR #1201, opened 2025-02-17
Add an index for our case search

jamesobrooks:
# Purpose

We want the query to be performant.

# Major Changes

Add an index for the cases collection to aid in searching.

# Testing/Validation

We will deploy to staging and validate there.

### DIFF
--- a/ops/cloud-deployment/lib/cosmos/mongo/cosmos-collections.bicep
+++ b/ops/cloud-deployment/lib/cosmos/mongo/cosmos-collections.bicep
@@ -188,3 +188,37 @@ resource usersCollection 'Microsoft.DocumentDB/databaseAccounts/mongodbDatabases
     }
   }
 }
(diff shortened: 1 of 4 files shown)


ops/cloud-deployment/lib/cosmos/mongo/cosmos-collections.bicep
```diff
@@ -188,3 +188,37 @@ resource usersCollection 'Microsoft.DocumentDB/databaseAccounts/mongodbDatabases
     }
   }
 }
+
```
amorrow-flexion, line 191: A fair warning, this change is going to cause a conflict in Azure with the ShardKey difference and force us to delete all `cases` data collections

amorrow-flexion (approved):

fmaddenflx (approved):

PR #1240, opened 2025-02-28
CAMS-421 Upgrade case sync to use last sync date

btposey:
# Purpose

Upgrade case sync to use new sync date in DXTR

# Major Changes

Replaced sync case ID query using the AO_TX with new query to update date in AO_CS.
Stored last sync date in case sync runtime state.

# Testing/Validation

* Unit tests
* Manual testing

# Definition of Done:

- [X] Code refactored for clarity - Developers can understand the work simply by reviewing the code
- [X] Dependency rule followed - More important code doesn’t directly depend on less important code
- [X] Development debt eliminated - UX and code aligns to the team’s latest understanding of the domain

## Summary by Sourcery

Updates the case synchronization process to use the last synchronization date instead of the last transaction ID. This change improves the efficiency and accuracy of case synchronization by focusing on changes since the last successful sync.

Enhancements:
- Replaces the case ID query using the AO_TX table with a new query that updates the date in the AO_CS table.
- Stores the last sync date in the case sync runtime state.
- Updates the sync cases use case to retrieve case IDs based on the last sync date.
- Removes the retrieval of max transaction ID from the CasesLocalGateway as it is no longer needed.
- Updates the CasesSyncState type to use lastSyncDate instead of txId.
- Updates the sync cases function app to pass the lastSyncDate to the getCaseIds use case and store the lastSyncDate in the runtime state.

### DIFF
--- a/backend/lib/use-cases/cases/cases.interface.d.ts
+++ b/backend/lib/use-cases/cases/cases.interface.d.ts
@@ -26,7 +26,7 @@ export interface CasesInterface {
 
   getSuggestedCases(applicationContext: ApplicationContext, caseId: string): Promise<CaseSummary[]>;
 
-  getUpdatedCaseIds(applicationContext: ApplicationContext, caseId: string): Promise<string[]>;
+  getUpdatedCaseIds(applicationContext: ApplicationContext, start: string): Promise<string[]>;
 
   findTransactionIdRangeForDate(
     context: ApplicationContext,
--- a/backend/lib/use-cases/cases/cases.interface.d.ts
+++ b/backend/lib/use-cases/cases/cases.interface.d.ts
@@ -26,10 +26,7 @@ export interface CasesInterface {
 
   getSuggestedCases(applicationContext: ApplicationContext, caseId: string): Promise<CaseSummary[]>;
 
-  getCaseIdsAndMaxTxIdToSync(
-    applicationContext: ApplicationContext,
-    lastTxId: string,
-  ): Promise<CasesSyncMeta>;
+  getUpdatedCaseIds(applicationContext: ApplicationContext, caseId: string): Promise<string[]>;
 
   findTransactionIdRangeForDate(
     context: ApplicationContext,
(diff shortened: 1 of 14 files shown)


backend/lib/use-cases/cases/cases.interface.d.ts
```diff
@@ -26,10 +26,7 @@ export interface CasesInterface {
 
   getSuggestedCases(applicationContext: ApplicationContext, caseId: string): Promise<CaseSummary[]>;
 
-  getCaseIdsAndMaxTxIdToSync(
-    applicationContext: ApplicationContext,
-    lastTxId: string,
-  ): Promise<CasesSyncMeta>;
+  getUpdatedCaseIds(applicationContext: ApplicationContext, caseId: string): Promise<string[]>;
```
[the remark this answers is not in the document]
amorrow-flexion: I dont think this is necessarily correct. We do want `start: string` to be the param

common/src/feature-flags.ts
```diff
@@ -15,4 +15,5 @@ export const testFeatureFlags: FeatureFlagSet = {
   'case-search-enabled': true,
   'case-notes-enabled': true,
   'privileged-identity-management': true,
+  'enhanced-case-sync': true,
```
jamesobrooks, line 18:
```suggestion
```

backend/lib/use-cases/dataflows/sync-cases.ts
```diff
@@ -4,34 +4,38 @@ import Factory, { getCasesGateway } from '../../factory';
...
     }
 
     const casesGateway = getCasesGateway(context);
-    const { caseIds, lastTxId } = await casesGateway.getCaseIdsAndMaxTxIdToSync(
-      context,
-      syncState.txId,
-    );
+    const start = syncState.lastSyncDate;
+    if (!start) {
+      throw new CamsError(MODULE_NAME);
```
btposey, line 30: We should add a description of the error here.
```suggestion
      throw new CamsError(MODULE_NAME, {message: 'Last sync date not available.'});
```

backend/lib/use-cases/gateways.types.ts
```diff
@@ -177,7 +177,7 @@ export type OrderSyncState = RuntimeState & {
 
 export type CasesSyncState = RuntimeState & {
   documentType: 'CASES_SYNC_STATE';
-  txId: string;
+  lastSyncDate?: string;
```
btposey, line 180: I don't think this should be optional. But then see the thrown error message above if it is always required.

PR #1254, opened 2025-03-11
CAMS-528 Case Note Archive

amorrow-flexion:
# Purpose

- Provide the user the ability to soft delete case Notes

# Major Changes

Describe what has changed.

# Testing/Validation

# Notes

# Definition of Done:

- [ ] Dependency rule followed - More important code doesn’t directly depend on less important code

## Summary by Sourcery

Adds the ability to soft delete case notes, allowing users to remove notes they've created. Implements a modal confirmation for the deletion action and updates the UI to include a remove button for authorized users.

New Features:
- Implements soft deletion of case notes, allowing users to remove notes they've created.
- Adds a confirmation modal to prevent accidental deletion of case notes.
- Introduces a remove button in the UI for authorized users to trigger the soft deletion process.

Tests:
- Adds tests for the case note removal functionality, including modal interactions and API calls.

### DIFF
--- a/common/src/cams/test-utilities/mock-data.ts
+++ b/common/src/cams/test-utilities/mock-data.ts
@@ -538,7 +538,7 @@ function getCaseNoteDeletionRequest(
 }
 
 function getCaseNoteDeletion(override: Partial<CaseNote> = {}): Partial<CaseNote> {
-  const archivedOn = getTodaysIsoDate();
+  const archivedOn = new Date().toISOString();
   return {
     id: randomId(),
     caseId: randomCaseId(),


amorrow-flexion:
> Hey @amorrow-flexion - I've reviewed your changes - here's some feedback:
> 
> **Overall Comments**:
> 
> * Consider adding a loading state or disabling the archive button while the archival request is in progress to prevent multiple clicks.
> * The DELETE request in `_Api2.patchCaseNoteArchival` should be a PATCH request.
> 
> Here's what I looked at during the review
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

We do have a loading state after the Archive request is sent. The loading state is based on the retrieval of CaseNotes afterwards. 

We also intentionally chose DELETE for this request because that is technically what we are doing but due to data retention policies, we soft delete the record

user-interface/src/lib/models/api2.ts
```diff
@@ -266,6 +266,19 @@ async function postCaseNote(note: CaseNoteInput): Promise<void> {
   }
 }
 
+async function deleteCaseNote(note: Partial<CaseNote>) {
+  await api().delete<Partial<CaseNote>>(
+    `/cases/${note.caseId}/notes/${note.id}/${note.updatedBy!.id}`,
+    /*
```
amorrow-flexion, line 272: remove

backend/lib/use-cases/gateways.types.ts
```diff
@@ -52,6 +58,7 @@ interface Reads<R> {
 }
 
 interface Updates<T, R = void> {
+  //TODO: why is this named update when we do a replaceOne?
```
amorrow-flexion, line 61: This

backend/lib/controllers/case-notes/case.notes.controller.ts
```diff
@@ -41,8 +49,18 @@
         return httpSuccess({
           statusCode: HttpStatusCodes.CREATED,
         });
+      } else if (context.request.method === 'DELETE') {
+        const id = context.request.params.noteId;
+        const caseId = context.request.params.caseId;
```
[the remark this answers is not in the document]
amorrow-flexion: I like this and we should find other usages

backend/lib/adapters/gateways/mongo/utils/mongo-adapter.ts
```diff
@@ -161,6 +162,42 @@ export class MongoCollectionAdapter<T> implements DocumentCollectionAdapter<T> {
...
+
+    try {
+      const result = await this.collectionHumble.updateOne(mongoQuery, mongoItem);
+      const unknownError = new UnknownError(this.moduleName, {
+        message: 'Failed to insert document into database.',
+      });
+      const notFoundError = new NotFoundError(this.moduleName, {
+        message: 'No matching item found.',
+      });
+      // const unknownMatchError = new NotFoundError(this.moduleName, {
```
amorrow-flexion, line 177: remove

backend/lib/adapters/gateways/mongo/case-notes.mongo.repository.test.ts
```diff
@@ -55,12 +79,22 @@ describe('case notes repo tests', () => {
...
+
     test('should handle error on getNotesByCaseId', async () => {
       const caseId = '12-12345';
       jest.spyOn(MongoCollectionAdapter.prototype, 'find').mockRejectedValue(error);
       await expect(() => repo.getNotesByCaseId(caseId)).rejects.toThrow(
         getCamsError(error, 'CASE_NOTES_MONGO_REPOSITORY', 'Unable to retrieve case note.'),
       );
     });
+
+    //test('should handle error on archiveCaseNote', {});
```
amorrow-flexion, line 98: remove

backend/lib/controllers/case-notes/case.notes.controller.ts
```diff
@@ -89,4 +105,34 @@ export class CaseNotesController implements CamsController {
...
+
+    if (!request['userId']) {
+      badParams.push('userId');
+    }
+
+    if (badParams.length > 0) {
+      const isPlural = badParams.length > 1;
+      const message = `Required ${isPlural ? 'parameters' : 'parameter'} ${badParams.join(', ')} ${isPlural ? 'are' : 'is'} absent.`;
+      messages.push(message);
+    }
```
amorrow-flexion, line 133: I think james might want a space here....

user-interface/src/case-detail/panels/CaseNotes.tsx
```diff
@@ -153,6 +205,29 @@ export default function CaseNotes(props: CaseNotesProps) {
...
+        <div className="case-note-toolbar" data-testid={`case-note-toolbar-${idx}`}>
+          {userCanArchive(note) && (
+            <OpenModalButton
+              className="remove-button"
+              id={`case-note-remove-button-${idx}`}
+              buttonIndex={`${idx}`}
+              uswdsStyle={UswdsButtonStyle.Unstyled}
+              modalId={archiveConfirmationModalId}
+              modalRef={archiveConfirmationModalRef}
+              ref={openModalButtonRefs.current[idx]}
```
fmaddenflx, line 217: I took great pains to implement this unique ref for every button but don't think I actually tested to see if it solved the problem I was observing.  The problem is as follows:

- Let's say we have 3 notes with remove buttons
- Clicking on the first remove button brings up the modal
- Canceling the modal is supposed to focus on the button that opened it.
- However, what I was seeing is that the very last button was focused.
- This seemed to be because (my hypothesis) all the buttons were sharing the same ref, so when calling ref.current.focus(), it was focusing the very last component that was assigned that ref. (or maybe it was focusing all of them in turn, which placed the final focus on the last one?)

We need to verify that this solved the issue.

user-interface/src/case-detail/panels/CaseNotes.tsx
```diff
@@ -96,24 +142,29 @@ export default function CaseNotes(props: CaseNotesProps) {
...
-          }
-          disableFormFields(false);
-        });
+      if (session?.user) {
+        const caseNoteInput: CaseNoteInput = {
+          caseId: props.caseId,
+          title: formData.title,
+          content: formData.content,
+          updatedBy: getCamsUserReference(session?.user),
+        };
```
fmaddenflx, line 151: So we discussed this but don't think we actually implemented it:

We should use local storage to populate the fields when we initially come to the screen (i.e. initially load from saved data).

However, when we actually go to send the data to the API, we should grab the data from the form input fields directly.

This would be a safety measure if say for example, the system has local cache disabled.

user-interface/src/case-detail/panels/CaseNotes.tsx
```diff
@@ -41,7 +50,16 @@ export default function CaseNotes(props: CaseNotesProps) {
   const contentInputRef = useRef<TextAreaRef>(null);
   const submitButtonRef = useRef<ButtonRef>(null);
   const clearButtonRef = useRef<ButtonRef>(null);
+  const archiveConfirmationModalRef = useRef<ModalRefType>(null);
+  const openModalButtonRefs = useRef<React.RefObject<OpenModalButtonRef>[]>([]);
+  useMemo(() => {
+    openModalButtonRefs.current =
+      caseNotes?.map((_, index) => openModalButtonRefs.current[index] ?? { current: null }) || [];
+  }, [caseNotes]);
```
fmaddenflx, line 58: So I wrote this using the guidance of ChatGPT.  I can't say I understand useMemo() at this point.  I need to research useMemo.  But essentially, this provides us the ability to create a unique ref for ever one of the openModalButton(s).  This is only necessary for the functionality that focuses on the button that opened the modal, when the modal is closed (an accessibility requirement)

user-interface/src/case-detail/panels/CaseNotes.test.tsx
```diff
@@ -249,6 +264,69 @@ describe('case note tests', () => {
...
+    session.user.id = userId;
+    session.user.name = userFullName;
+    vi.spyOn(LocalStorage, 'getSession').mockReturnValue(session);
+    vi.spyOn(Api2, 'getCaseNotes').mockResolvedValue({
+      data: caseNotes,
+    });
+    const patchSpy = vi
+      .spyOn(Api2, 'patchCaseNoteArchival')
+      .mockResolvedValueOnce()
+      .mockRejectedValueOnce(new Error());
```
fmaddenflx, line 279: Since we're no longer patching but deleting (from a REST verb perspective), we need to rename anything using the word patch.

user-interface/src/case-detail/panels/CaseNotes.test.tsx
```diff
@@ -249,6 +264,69 @@ describe('case note tests', () => {
...
+    const expectedFirstPatchArgument = {
+      id: caseNotes[0].id,
+      caseId: caseNotes[0].caseId,
+      updatedBy: expectedUser,
+    };
+    const expectedSecondPatchArgument = {
+      id: caseNotes[2].id,
+      caseId: caseNotes[2].caseId,
+      updatedBy: expectedUser,
+    };
```
fmaddenflx, line 294: expectedFirstDeleteArgument
expectedSecondDeleteArgument

or something better if you like.

user-interface/src/lib/models/api2.ts
```diff
@@ -370,6 +376,7 @@ export const _Api2 = {
   getCaseHistory,
   postCaseNote,
   getCaseNotes,
+  patchCaseNoteArchival: deleteCaseNote,
```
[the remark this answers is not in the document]
amorrow-flexion: Yep definitely need a rename

user-interface/src/case-detail/panels/CaseNotes.tsx
amorrow-flexion: Would we ever have a case where the system has local cache disabled?  However I do see your point about validation prior to send the data, because the data could also be manipulated directly in the stored object in localstorage

user-interface/src/case-detail/panels/CaseNotes.tsx
amorrow-flexion: I verified that this behavior is proper with many case notes available as the Bert user. It does return focus to the previously focused element

backend/lib/controllers/case-notes/case.notes.controller.ts
fmaddenflx: Space instead of commas?  What do you mean exactly?

jamesobrooks: An empty line. Mostly just because each of the other if blocks has a newline after it. @amorrow-flexion was indeed correct, that I would prefer to follow the same pattern or to remove the empty lines. Alternatively if there are logical groupings of the blocks, they could be grouped.

backend/lib/use-cases/gateways.types.ts
```diff
@@ -89,6 +96,7 @@ export interface CaseAssignmentRepository<T = CaseAssignment>
 
 export interface CaseNotesRepository<T = CaseNote> extends Creates<T, string> {
   getNotesByCaseId(caseId: string): Promise<CaseNote[]>;
+  archiveCaseNote(archiveNote: Partial<CaseNote>): Promise<UpdateResult>; //TODO: replace the current update function with replace? and create new update function
```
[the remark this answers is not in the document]
jamesobrooks: I could actually see having three separate ones. `Upserts`, `Replaces`, and `Patches`. I think `Patches` might be better than `Updates` because it has [semantic meaning](https://www.rfc-editor.org/rfc/rfc5789) outside of whatever a particular db technology defines `update` to mean.

backend/lib/use-cases/gateways.types.ts
jamesobrooks: Our gateways are boundaries between CAMS application semantics and the third-party semantics of the underlying tech. We can call the functions whatever makes the most sense for the way we want to refer to things and then the implementation can call whatever function from the library or make whatever http verb requests the third-party presents to perform the operation we need—or combination of calls if a single one does not suffice.

backend/lib/use-cases/gateways.types.ts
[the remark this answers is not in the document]
jamesobrooks: Note that even if we make changes to the `Updates` interface—and/or add others—it may still be appropriate to have an `archiveCaseNote` function separate from those to ensure that any other changes to notes do not archive the note and that no archiving action edits anything else about the note.

user-interface/src/case-detail/panels/CaseNotes-globalAlert.test.tsx
```diff
@@ -15,13 +17,16 @@ function renderWithProps(props?: Partial<CaseNotesProps>) {
     caseNotes: [],
     searchString: '',
     onNoteCreation: vi.fn(),
+    onNoteDelete: vi.fn(),
```
fmaddenflx, line 20: I'm wondering if on the frontend we should be using the term Remove rather than Delete since in the UI we're using the term Remove everywhere.  The only place we actually Need to use Delete is API2 when we make the REST call.

backend/lib/controllers/case-notes/case.notes.controller.ts
```diff
@@ -11,10 +11,18 @@ import { isValidUserInput } from '../../../../common/src/cams/sanitization';
...
+const VALID_ID_PATTERN = RegExp(/^[\dA-Za-z]+(-[\dA-Za-z]+)*$/);
+const INVALID_ID_MESSAGE = 'case note ID must be provided.';
 const INVALID_CASEID_MESSAGE = 'caseId must be formatted like 111-01-12345.';
 const INVALID_NOTE_MESSAGE = 'Note content contains invalid keywords.';
 const INVALID_NOTE_TITLE_MESSAGE = 'Note title contains invalid keywords.';
 
+type ArchiveRequest = {
+  id?: string;
+  caseId?: string;
+  archive?: boolean;
```
btposey, line 23: Not used

backend/lib/use-cases/case-notes/case-notes.ts
```diff
@@ -19,14 +25,26 @@ export class CaseNotesUseCase {
         id: user.id,
         name: user.name,
       },
-      updatedOn: new Date().toISOString(),
+      updatedOn: getTodaysIsoDate(),
```
btposey, line 28: This is just the date YYYY-MM-DD, not the full timestamp. We should add a common helper function like `getNowTimstamp` that would return `new Date().toISOString()`.

backend/lib/use-cases/gateways.types.ts
```diff
@@ -33,6 +33,12 @@ export type ReplaceResult = {
   upsertedCount: number;
 };
 
+export type UpdateResult = {
+  // id: string;
```
btposey, line 37: Not used? Remove.

common/src/cams/test-utilities/mock-data.ts
```diff
@@ -523,6 +524,30 @@ function getCaseNote(override: Partial<CaseNote> = {}): CaseNote {
...
+    id: crypto.randomUUID(),
+    caseId: randomCaseId(),
+    userId,
+    sessionUser: getCamsUserReference({ id: userId }),
+    ...override,
+  };
+}
+
+function getCaseNoteDeletion(override: Partial<CaseNote> = {}): Partial<CaseNote> {
+  const archivedOn = getTodaysIsoDate();
```
btposey, line 541:
```suggestion
  const archivedOn = new Date().toISOString();
```

user-interface/src/case-detail/panels/CaseNotes.tsx
fmaddenflx: So Memo, short for Memoization, is just a caching optimization here.

governmentSponsored (approved):

btposey (approved):

PR #1263, opened 2025-03-14
CAMS-537 Externalize configuration of enabled data flows

btposey:
# Purpose

Enable data flows with an env var.

# Major Changes

- Added a common scheme to export the name and setup of data flow modules
- Added helper class to register and enable data flow modules
- Refactored to simplify logger and app context creation.

# Definition of Done:

- [x] Dependency rule followed - More important code doesn’t directly depend on less important code

## Summary by Sourcery

Externalizes the configuration of enabled data flows using an environment variable. This allows enabling or disabling data flows without modifying the code. It also introduces a common scheme to export the name and setup of data flow modules, and a helper class to register and enable them.

Enhancements:
- Introduces a common scheme to export the name and setup of data flow modules.
- Adds a helper class to register and enable data flow modules.


backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,83 @@
...
+      this.set(dataflow.MODULE_NAME, dataflow.setup);
+    }
+  }
+
+  list() {
+    return [...this.keys()];
+  }
+
+  enable(...dataflowNames: string[]) {
+    const cleanNames = dataflowNames.map((name) => name.trim().toUpperCase());
```
btposey, line 29: We should add `cleanNames` to a set to make sure we do not setup a data flow more than once.

btposey, line 29: We could move `.map((name) => name.trim().toUpperCase())` to the end of the .`split(',')` function after we read the names from the env var.

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,83 @@
...
+    for (const dataflow of dataflows) {
+      this.set(dataflow.MODULE_NAME, dataflow.setup);
+    }
+  }
+
+  list() {
+    return [...this.keys()];
+  }
+
+  enable(...dataflowNames: string[]) {
```
btposey, line 28: Maybe `enable` should be renamed to `setup`.

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,83 @@
...
+console.log(MODULE_NAME, 'Registered Dataflows', registeredDataflows);
+
+// Enable the data flows specified in from the configuration env var.
+const envVar = process.env.CAMS_ENABLE_DATAFLOWS;
+const dataflowNames = envVar.split(',');
+const status = dataflows.enable(...dataflowNames);
+
+// Log the status of each registered data flow.
+status.forEach((s) => {
+  console.log(MODULE_NAME, 'Enabled', s);
```
btposey, line 68: If the `enable` function is renamed to `setup` then change what is logged.
```suggestion
  console.log(MODULE_NAME, s);
```

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,83 @@
...
+
+const MODULE_NAME = 'DATAFLOWS_SETUP';
+
+type DataflowSetup = {
+  MODULE_NAME: string;
+  setup: () => void;
+};
+
+class DataflowSetupMap extends Map<string, () => void> {
+  add(...dataflows: DataflowSetup[]) {
```
btposey, line 18: Maybe `add` could be renamed `register`.

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,86 @@
...
-setupSyncOrders();
+// Register data flows.
+dataflows.add(SyncCases, SyncOfficeStaff, SyncOrders, MigrateCases, MigrateConsolidations);
+
+// Log the list of registered data flows.
+const registeredDataflows = dataflows.list().join(', ');
+logger.info(MODULE_NAME, 'Registered Dataflows', registeredDataflows);
+
+// Enable the data flows specified in from the configuration env var.
+const envVar = process.env.CAMS_ENABLE_DATAFLOWS;
```
btposey, line 65: Use the nullish operator to make sure we have a string.
```suggestion
const envVar = process.env.CAMS_ENABLE_DATAFLOWS ?? '';
```

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,83 @@
...
+import MigrateConsolidations from './import/migrate-consolidations';
+
+const MODULE_NAME = 'DATAFLOWS_SETUP';
+
+type DataflowSetup = {
+  MODULE_NAME: string;
+  setup: () => void;
+};
+
+class DataflowSetupMap extends Map<string, () => void> {
```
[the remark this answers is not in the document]
jamesobrooks: Why do we not want it to behave like a Map? We could certainly implement a `list()` function that returns the 
```suggestion
class DataflowSetupMap extends Map<string, () => void> {
  functionMap: Map<string, () => void> = new Map();
  
  // Modify existing functions to use this.functionMap.[set, get, etc.]
  
  list() {
    return this.functionMap.list();
  }
}
```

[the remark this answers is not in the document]
btposey: We do want the `Map`, we could just make it a private variable. When using the object we only care about the `register`, `list` and `setup` functions.

backend/function-apps/dataflows/import/sync-orders.ts
```diff
@@ -4,12 +4,9 @@ import { OrdersController } from '../../../lib/controllers/orders/orders.control
 import { toAzureError } from '../../azure/functions';
 import { buildFunctionName, buildHttpTrigger } from '../dataflows-common';
 
-const MODULE_NAME = 'SYNC_ORDERS';
+const MODULE_NAME = 'SYNC-ORDERS';
```
amorrow-flexion, line 7: We have a mismatch in a lot of places where the module name either has an underscore, or a hyphen in different module, maybe we should update that with one or the other to keep it consistent. e.g. courts.function.ts has an underscore

jamesobrooks: A quick search finds 23 with `_` and 89 with `-`.

btposey: Which is the preference. I've been converting to hyphens as I encounter them.

btposey: We also have other constants that aren't module names that use underscore. I've been leaving those alone.

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -1,24 +1,88 @@
...
+const status = dataflows.setup(...names);
+
+// Log the status of each registered data flow.
+status.forEach((s) => {
+  logger.info(MODULE_NAME, s);
+});
+
+/*
+
+Sample log output on startup:
```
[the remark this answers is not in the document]
btposey: Dunno. Documentation lies, but it informs the reader that vital information is available on startup.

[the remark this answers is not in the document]
btposey: ...so go check appinsights.

backend/function-apps/dataflows/import/sync-orders.ts
btposey: `MODULE_NAME = '\w+_` returns 4 instances of `MODULE_NAME` with underscores. Fixing each of them.

.github/workflows/reusable-deploy.yml
```diff
@@ -67,6 +67,7 @@ jobs:
       - name: Deploy Azure resources
         id: azure-deploy
         if: inputs.deployBicep == 'true'
+        ## NOTE: enabledDataflows requires quotes here
```
amorrow-flexion, line 70: remove

amorrow-flexion (approved):

governmentSponsored (approved):

PR #1270, opened 2025-03-25
CAMS-527 Hot Fix

amorrow-flexion:
# Purpose

Correct Edit behavior on Case Note and update use case tests and e2e tests to properly exercise functionality

# Major Changes

- update case-notes use case to properly spread in the new note for create
- update e2e test to validate modal
- update case-notes use case tests to properly expect on call

# Testing/Validation

ran locally

# Notes

n/a

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where edits to case notes were not being correctly applied.

### DIFF
--- a/backend/lib/use-cases/case-notes/case-notes.ts
+++ b/backend/lib/use-cases/case-notes/case-notes.ts
@@ -81,8 +81,8 @@ export class CaseNotesUseCase {
     const dateOfEdit = new Date().toISOString();
     const newNote: CaseNote = {
       ...existingNote,
+      ...note,
       id: randomUUID(),
-      documentType: 'NOTE',
       previousVersionId: note.id,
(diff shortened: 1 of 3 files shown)


backend/lib/use-cases/case-notes/case-notes.ts
```diff
@@ -81,8 +81,8 @@ export class CaseNotesUseCase {
     const dateOfEdit = new Date().toISOString();
     const newNote: CaseNote = {
       ...existingNote,
+      ...note,
       id: randomUUID(),
-      documentType: 'NOTE',
```
jamesobrooks, line 85: I know I said we don't need this, but it probably should stay. If the client were to change the documentType, the new "note" would be invalid.

jamesobrooks (approved):

PR #1285, opened 2025-04-02
CAMS-544: Use a better order for courts query

jamesobrooks:
# Purpose

The order of the courts in the district (division) filter on the search screen is difficult to use.

# Major Changes

Instead of ordering by group designator and office code—which the users do not see—we now order by state, then court name, then office name display.

# Testing/Validation

Validated the query results make sense.

# Definition of Done:

- [x] Dependency rule followed - More important code doesn’t directly depend on less important code

## Summary by Sourcery

Enhancements:
- Modify the SQL query ordering to prioritize state, court name, and office name display for better readability and usability


btposey (approved):

jharcey-flexion (commented):
Is the apostrophe after OFFICE_NAME_DISPLAY breaking the order by? Running the query in MS SQL Studio, I had to remove it.

PR #1314, opened 2025-04-21
Use pre-commit to verify untested changes

jamesobrooks:
# Purpose

We want to promote the updating of tests where appropriate (discussed in a prior team retrospective where an action item was created).

# Major Changes

Checks whether source files were modified and test files were not and then prompts the committer for a response as to whether this was intended. If any tests were changed or only non-source files were changed, no challenge is made.

# Testing/Validation

Validated locally that the new hook contents produce the desired behavior.

# Notes

This does not attempt to match test changes to source changes in any way. I think that would be option-limiting, but it does mean that it's certainly possible to end up with tests that fail or are inadequate, particularly with larger commits or with changes to code that affects more than just one area of the codebase.

# Definition of Done:

- [ ] Dependency rule followed - More important code doesn’t directly depend on less important code


jamesobrooks:
> Hey @jamesobrooks - I've reviewed your changes - here's some feedback:
> 
> **Overall Comments**:
> 
> * Consider adding a configuration option to disable the hook in certain cases, like when making documentation-only changes.
> 
> Here's what I looked at during the review
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

The hook already ignores changes where only non-source files are changed and allows the committer to answer `y` to the prompt if they intended to not update tests.

governmentSponsored (approved):

PR #1449, opened 2025-08-01
CAMS-555 Replace 1-column gutters with constrained main content container

En-8:
# Purpose

To reduce duplicative management of 1-column gutters with a single container of constrained width, making layout and responsiveness somewhat simpler to manage.

# Major Changes

- Remove empty 1-column gutter divs, absorbing that width in other columns of the same row.
- Style cams-content with constrained width

# Testing/Validation

Visual inspection on the UI.

# Notes

# Definition of Done:

### DIFF
--- a/user-interface/src/App.scss
+++ b/user-interface/src/App.scss
@@ -59,8 +59,8 @@ body {
     flex: 1;
     padding: 0 0 50px 0;
     margin: 0 auto;
-    width: calc(10 / 12 * 100%); // to stay consistent with previous 1 + 10 + 1 layout
-    max-width: 64rem;
+    width: 83.3333333333333%; // 10/12 * 100 to stay consistent with previous 1 + 10 + 1 layout
+    max-width: 82rem;
 
     h1,
     h2 {
--- a/user-interface/src/App.scss
+++ b/user-interface/src/App.scss
@@ -45,21 +45,25 @@ body {
     Use this class anywhere you need screen readers to get more context, such as with aria-describedby sections.
   */
   .screen-reader-only {
-      position: absolute;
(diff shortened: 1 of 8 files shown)


user-interface/src/App.scss
```diff
@@ -45,21 +45,25 @@ body {
...
+    white-space: nowrap;
+    border: 0;
   }
 
   .cams-content {
     flex: 1;
     padding: 0 0 50px 0;
+    margin: 0 auto;
+    width: 83.3333333333333%; // 10/12 * 100 to stay consistent with previous 1 + 10 + 1 layout
+    max-width: 82rem;
```
En-8, line 63: This is the real change in this file. The rest is formatting

jamesobrooks:
My only issue with this change is that it squeezes the 12 columns into the same space as the previous 10 columns. Maybe that's fine, and maybe it's even unavoidable without having to do what we were struggling with in the first place. Probably worth a conversation though.

jamesobrooks (dismissed):
My only issue with this change is that it squeezes the 12 columns into the same space as the previous 10 columns. Maybe that's fine, and maybe it's even unavoidable without having to do what we were struggling with in the first place. Probably worth a conversation though.

En-8:
> My only issue with this change is that it squeezes the 12 columns into the same space as the previous 10 columns. Maybe that's fine, and maybe it's even unavoidable without having to do what we were struggling with in the first place. Probably worth a conversation though.

@jamesobrooks This is a totally fair point. I chatted with Katie about this to talk through trade-offs and any concerns she may have, and we landed on continuing with the approach in this PR.

PR #1475, opened 2025-08-20
CAMS-582 manage trustees spec

btposey:
# Purpose

Enable users to add and update trustee information.

# Major Changes

* Add trustee repo
* Add trustee API endpoint
* Add trustee list to UI
* Add trustee form to add/update trustee information

# Testing/Validation

* Unit tests
* E2E tests
* A11y tests

# Definition of Done:


backend/lib/controllers/trustees/trustees.controller.ts
```diff
@@ -0,0 +1,136 @@
...
+  public async handleRequest(
+    context: ApplicationContext,
+  ): Promise<CamsHttpResponseInit<Trustee | Trustee[]>> {
+    try {
+      // Check feature flag
+      if (!context.featureFlags['trustee-management']) {
+        return {
+          statusCode: 404,
+        };
+      }
```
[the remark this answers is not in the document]
btposey: This endpoint would not exist if the feature is off.

.github/workflows/sub-deploy-code-slot.yml
```diff
@@ -161,25 +161,10 @@ jobs:
...
-      branchHashId: ${{ inputs.environmentHash }}
-      e2eCosmosDbExists: ${{ inputs.e2eCosmosDbExists }}
-    secrets: inherit # pragma: allowlist secret
-
   swap-webapp-deployment-slot:
     if: ${{ inputs.initialDeployment != 'true' }}
     runs-on: ubuntu-latest
-    needs: [deploy-webapp-slot, deploy-api-slot, deploy-dataflows-slot, execute-e2e-test]
+    needs: [deploy-webapp-slot, deploy-api-slot, deploy-dataflows-slot]
     environment: ${{ inputs.ghaEnvironment }}
```
fmaddenflx, line 168: We need to put these e2e tests back in.  They were removed temporarily.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  // Helper function to convert district codes to readable names
+  function formatDistricts(districts: string[] | undefined): string {
+    if (!districts || districts.length === 0) {
+      return 'No districts assigned';
+    }
+
+    const readableDistricts = districts
+      .map((code) => courtMap.get(code) || code) // Fall back to code if not found
+      .sort();
```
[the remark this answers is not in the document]
btposey: The fallback code is basically there because `get` can return a falsy value. The code would be readily recognizable as a code rather than a named district and can report that code to fix the inconsistent data. Otherwise the end users would just see one or more "unknown district".

backend/function-apps/api/courts/courts.function.ts
```diff
@@ -19,7 +19,8 @@ export default async function handler(
     });
 
     const controller = new CourtsController();
-    context.session = await ContextCreator.getApplicationContextSession(context);
+    // Courts endpoint doesn't require authentication for dropdown data
+    // context.session = await ContextCreator.getApplicationContextSession(context);
```
fmaddenflx, line 24: Was this temporary?

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -0,0 +1,77 @@
...
+      }
+
+      // Prepare trustee for creation with audit fields
+      const userReference = getCamsUserReference(context.session.user);
+      const trusteeForCreation: TrusteeInput = {
+        ...trustee,
+        status: trustee.status || 'active',
+        districts: trustee.districts || [],
+        chapters: trustee.chapters || [],
+      };
```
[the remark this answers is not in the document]
btposey: This code improves data quality. Empty arrays of districts and chapters is semantically possible during a transition period when records are being added or modified. The record should always have an empty array by default before being persisted to the document DB.

user-interface/src/lib/hooks/UseTrusteeFormValidation.ts
```diff
@@ -0,0 +1,170 @@
...
+  TrusteeFormData,
+  ValidationError,
+  FormValidationResult,
+  TrusteeFormValidation,
+} from './UseTrusteeFormValidation.types';
+
+/**
+ * Validates individual form fields with specific business rules
+ */
+function validateField(field: string, value: string | string[] | undefined | null): string | null {
```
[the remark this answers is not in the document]
fmaddenflx: Probably something we should add.

backend/function-apps/api/courts/courts.function.ts
jamesobrooks: I don't know. I also don't know that we would want to remove the gate. Generally speaking, if a request is coming from a backend process it will have the api key which is a secret, and all other requests are treated as privileged. I think before we decide to have unprivileged endpoints accessible via any client—healthcheck is our only exception I think—we should strongly consider how we want to do this to prevent leaking sensitive information in the future or providing a foothold from a security perspective.

backend/function-apps/api/trustees/trustees.function.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  test('should handle successful trustee request', async () => {
+    const mockResponse: Partial<CamsHttpResponseInit<Trustee>> = {
+      statusCode: 200,
+      body: { data: [] as unknown as Trustee },
+    };
+    const mockAzureResponse = { status: 200 };
+
+    mockController.handleRequest.mockResolvedValue(mockResponse);
+    (toAzureSuccess as jest.Mock).mockReturnValue(mockAzureResponse);
```
jamesobrooks, line 56: This feels different than previous approaches to tests. For example, we don't typically mock `functions.ts` and its functions at all. Here we're mocking it on line 14 but then mocking a particular function in a way we don't typically mock anything. Do we like this approach?

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
jamesobrooks, line 1: I think the mocking in this test is quite a bit more complex than it has to be. I'm trying out something to simplify.

.gitignore
```diff
@@ -11,6 +11,7 @@
 *.iws
 *.ipr
 .idea/
+qodana.yaml
```
fmaddenflx, line 14: What is qodana?

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -0,0 +1,163 @@
...
+  useEffect(() => {
+    async function fetchData() {
+      try {
+        setLoading(true);
+
+        // Fetch both trustees and court information
+        const [trusteesResponse, courtsResponse] = await Promise.all([
+          api.getTrustees(),
+          api.getCourts(),
+        ]);
```
fmaddenflx, line 35: I had changed this to an arrow function.  I'm not sure why this is either changed back... or as if I hadn't changed it.  ???

backend/lib/controllers/trustees/trustees.controller.ts
```diff
@@ -0,0 +1,136 @@
...
+    const { url } = context.request;
+    const urlParts = url.split('/');
+    const trusteeId = urlParts[urlParts.length - 1];
+
+    // If the last part of the URL is 'trustees', this is a list request
+    // If it's something else, it's an individual trustee request
+    if (trusteeId === 'trustees') {
+      return await this.listTrustees(context);
+    } else {
+      return await this.getTrustee(context, trusteeId);
```
btposey, line 90: Why aren't we using the `params` dictionary on the `context.request` object? Why are we rolling our own?

user-interface/src/trustees/TrusteesList.tsx
btposey: I refactored this block of code so this is no longer relevant.

common/src/cams/parties.ts
```diff
@@ -1,29 +1,155 @@
...
+export type TrusteeInput = Party & {
+  districts?: string[];
+  chapters?: ChapterType[];
+  status?: TrusteeStatus;
 };
+
+export type Trustee = Auditable &
+  TrusteeInput & {
+    assistant?: TrusteeAssistant;
+    id?: string;
```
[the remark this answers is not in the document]
btposey: We refactored the domain model.

ops/cloud-deployment/ustp-cams-cosmos-temp.bicep
```diff
@@ -0,0 +1,131 @@
...
+param resourceGroupName string = 'bankruptcy-oversight-support-systems'
+@description('CosmosDb account name')
+param accountName string = 'cosmos-mongo-ustp-cams-dev'
+@description('CosmosDb database name')
+param databaseName string = 'cams'
+
+@description('Cosmos E2E database name')
+param e2eDatabaseName string = ''
+
+param deployE2eDatabase bool = false
```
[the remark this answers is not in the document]
btposey: Just nuke the file.

user-interface/src/trustees/TrusteeDetailScreen.tsx
```diff
@@ -0,0 +1,176 @@
...
+import useApi2 from '@/lib/hooks/UseApi2';
+import { useGlobalAlert } from '@/lib/hooks/UseGlobalAlert';
+import { Trustee } from '@common/cams/parties';
+import { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { MainContent } from '@/lib/components/cams/MainContent/MainContent';
+import DocumentTitle from '@/lib/components/cams/DocumentTitle/DocumentTitle';
+import Tag, { UswdsTagStyle } from '@/lib/components/uswds/Tag';
+
+export default function TrusteeDetailScreen() {
```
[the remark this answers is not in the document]
jamesobrooks: I'd rather wait for an axis of change to prompt that.

user-interface/src/lib/hooks/UseTrusteeFormValidation.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    case 'state':
+      return trimmedValue ? null : 'State is required';
+
+    case 'zipCode':
+      if (!trimmedValue) {
+        return 'ZIP code is required';
+      }
+      // ZIP code must be exactly 5 digits
+      if (!/^\d{5}$/.test(trimmedValue)) {
```
jamesobrooks: Shouldn't it be exactly 5 or exactly 9 digits with a dash?

PR #1764, opened 2025-12-30
CAMS-666 extended date verification

fmaddenflx:
# Purpose

Add additional date validation routines and error messages.

# Major Changes

    Simplified date validating for invalid dates.
    
    Do to the fact that the html5 native date picker already does internal
    validating, and entering an incomplete date, though the value in the
    field looks like "02/dd/yyyy" for example, the actual value of the input
    field is an empty string "".  So the only way to validate that a valid date
    was entered correctly is to check input.validity.badInput for a value of
    true.
    
    This simplifies our code, but also limits the messages that we can
    provide to the user.  Luckily, the only message we want to provide to
    the user is "Must be a valid date mm/dd/yyyy."

    Added new validation properties to Date pickers
    
    You have the ability to pass validators into Date Range Picker and Date
    Picker and it will evaluate the hard coded validators followed by the
    validators passed to the component.  This allows us to keep the date
    pickers generic enough for uswds component library but also pass in our
    own custom validators unique to our application.
    

# Testing/Validation

All unit testing updated and passing.

# Definition of Done:

## Summary by Sourcery

Add customizable date validation to date picker components and simplify invalid date handling.

New Features:
- Allow DatePicker to accept custom validators for additional domain-specific date rules.
- Allow DateRangePicker to accept separate custom validators for start and end dates.

Enhancements:
- Simplify invalid date detection by relying on native Date parsing instead of a shared helper.
- Update CaseDetailScreen to enforce a minimum allowed date on docket search range inputs via shared validators.

Tests:
- Add unit tests verifying custom validators run on blur, handle multiple validators, clear errors on valid input, and are correctly wired through DateRangePicker.

### DIFF
--- a/user-interface/src/case-detail/CaseDetailScreen.tsx
+++ b/user-interface/src/case-detail/CaseDetailScreen.tsx
@@ -6,7 +6,7 @@ import Icon from '@/lib/components/uswds/Icon';
 import Input from '@/lib/components/uswds/Input';
 import { UswdsAlertStyle } from '@/lib/components/uswds/Alert';
 import DateRangePicker from '@/lib/components/uswds/DateRangePicker';
-import { DateValidator } from '@/lib/components/uswds/DatePicker';
+import { ValidatorFunction } from 'common/src/cams/validation';
 import {
   ComboBoxRef,
   DateRange,
@@ -283,10 +283,11 @@ export default function CaseDetailScreen(props: Readonly<CaseDetailProps>) {
 
   const globalAlert = useGlobalAlert();
 
-  const minDateValidator: DateValidator = (value: string) => {
+  const minDateValidator: ValidatorFunction = (value: unknown) => {
+    if (typeof value !== 'string') return { reasons: ['Must be a string'] };
     const date = new Date(value);
     const minDate = new Date('1978-12-31');
-    return date <= minDate ? 'Must be later than 1978.' : null;
+    return date <= minDate ? { reasons: ['Must be later than 1978.'] } : { valid: true as const };
   };
 
   async function fetchCaseBasicInfo() {


user-interface/src/lib/components/uswds/DatePicker.test.tsx
```diff
@@ -792,4 +792,74 @@ describe('DatePicker edge case coverage', () => {
...
+  test('should run multiple custom validators and concatenate error messages', async () => {
+    const minDateValidator = (value: string) => {
+      const date = new Date(value);
+      const minDate = new Date('1978-01-01');
+      return date < minDate ? 'Date must be on or after 01/01/1978.' : null;
+    };
+
+    const maxDateValidator = (value: string) => {
+      const date = new Date(value);
+      const maxDate = new Date('2030-12-31');
```
fmaddenflx: Test should validate that these functions ran, but does not expect them to have been called.

user-interface/src/case-detail/CaseDetailScreen.tsx
```diff
@@ -282,6 +283,12 @@ export default function CaseDetailScreen(props: Readonly<CaseDetailProps>) {
 
   const globalAlert = useGlobalAlert();
 
+  const minDateValidator: DateValidator = (value: string) => {
+    const date = new Date(value);
+    const minDate = new Date('1978-12-31');
```
mattstankey: I thought Jon had mentioned that we have cases going back to 1978, does this need to be changed to 1977 instead?

user-interface/src/lib/components/uswds/DatePicker.tsx
```diff
@@ -9,9 +9,8 @@ import React, {
   useEffect,
   type JSX,
 } from 'react';
-import DateHelper from '@common/date-helper';
 
-const { isInvalidDate } = DateHelper;
+export type DateValidator = (value: string) => string | null;
```
btposey: This has me considering whether we should extend the validators library...

btposey (changes requested):
Let's consider whether we should extend Validators and integrate with the DateRangePicker, DatePicker props.

PR #1770, opened 2026-01-02
CAMS-666 accessibility updates

fmaddenflx:
# Purpose

Fix weird error messages and problems with needing to have both dates valid in order to search. We should only need to have 1 date, and if no dates are cleared after having a date, then it should search.

# Major Changes

-.-

# Testing/Validation

Tests all pass and coverage is goot.  Not good.  Just goot.

# Definition of Done:

## Summary by Sourcery

Relax date range validation so searches can run with a single valid date or no dates while improving error messaging and callback behavior.

Bug Fixes:
- Allow callbacks to fire when only a start date or only an end date is provided, or when both dates are cleared, instead of requiring both dates to be valid.
- Prevent invalid dates on one side of the range from blocking callbacks for valid dates on the other side.
- Ensure range validation only runs when both dates are present and individually valid, avoiding incorrect range errors.

Enhancements:
- Leverage validator-provided error messages for out-of-range dates to improve accessibility and clarity of validation feedback.
- Adjust validation tests to reflect the new single-date and clear-range behaviors and updated error messages.


fmaddenflx:
@sourcery-ai review

mattstankey (approved):

PR #2230, opened 2026-04-08
CAMS-725 Display trustee appointment info on case detail panel

cruzjone-flexion:
# Purpose

Surfaces trustee appointment information on the case detail page so users can quickly see who the assigned trustee is, when they were appointed, and how to contact them — without navigating away from the case.

# Major Changes

* Added `appointedDate` capture in the DXTR sync pipeline, reading the YYMMDD value from `AO_TX.REC` and converting it to ISO format before persisting in CosmosDB
* Added new `GET /cases/{caseId}/trustee-appointment` endpoint that returns the active `CaseAppointment` (or 404 if none exists)
* Added `useCaseAppointment` hook and `Api2.getCaseTrusteeAppointment` to fetch the active appointment from the frontend
* Added `CaseDetailTrusteePanel` with a trustee name heading, appointed date (formatted as MM/DD/YYYY), and side-by-side contact cards (public and internal)
* Reused `ContactInformationCard` and `MeetingOfCreditorsInfoCard` on the new panel for consistent layout with the trustee profile page
* Added navigation entry for the trustee panel in `CaseDetailNavigation`

# Testing/Validation

Implemented using TDD. Unit tests cover the new use case, controller, gateway changes, hook, and panel components. BDD-style integration coverage follows existing patterns for case detail panels.

# Notes

The `appointedDate` is derived from the 6-character `YYMMDD` field in `AO_TX.REC` (positions 24–29) and converted to ISO format in the gateway before being forwarded through the sync event. The frontend formats it as MM/DD/YYYY for display, consistent with date conventions used elsewhere in the app.

The trustee panel layout mirrors the trustee profile page — cards sit side by side with flexbox wrapping — so the UI is familiar to users who visit both.

# Definition of Done:

## Summary by Sourcery

Surface trustee appointment details on case detail, including appointed date and trustee contact info, backed by a new appointment endpoint and data backfill.

New Features:
- Expose appointedDate from DXTR and propagate it through trustee appointment sync into case appointments.
- Add a GET /cases/{caseId}/trustee-appointment API and controller to return the active trustee appointment for a case.
- Introduce a trustee information panel on the case detail screen that shows trustee name, appointed date, and public/internal contact and 341 meeting info.
- Add a useCaseAppointment frontend hook and Api2 client method to fetch the active trustee appointment for a case.

Enhancements:
- Gate trustee name links behind TrusteeAdmin role and optionally open them in a new tab with a launch icon.
- Make trustee contact and 341 meeting info cards reusable without edit handlers and align internal contact heading copy.
- Add runtime state and an Azure dataflow to backfill appointedDate for existing case appointment documents using DXTR data, with cursor-based pagination and retry handling.

Tests:
- Add unit and integration tests for the new case trustee appointment use case, controller, Azure function, backfill use case, and DXTR date handling.
- Add frontend tests for the trustee panel, trustee name behavior, navigation entry, and useCaseAppointment hook.


cruzjone-flexion:
@sourcery-ai review

backend/function-apps/dataflows/migrations/backfill-case-appointment-dates.ts
jamesobrooks: Did we need traces and metrics for this story?

cruzjone-flexion: Yes, the metrics will be its own PR.

backend/lib/controllers/cases/case-trustee-appointment.controller.ts
```diff
@@ -0,0 +1,26 @@
+import { ApplicationContext } from '../../adapters/types/basic';
+import { CamsHttpResponseInit, httpSuccess } from '../../adapters/utils/http-response';
+import { NotFoundError } from '../../common-errors/not-found-error';
+import { CaseTrusteeAppointmentUseCase } from '../../use-cases/cases/case-trustee-appointment.use-case';
+import { CaseAppointment } from '@common/cams/trustee-appointments';
+
+const MODULE_NAME = 'CASE-TRUSTEE-APPOINTMENT-CONTROLLER';
+
+export class CaseTrusteeAppointmentController {
+  async handleRequest(context: ApplicationContext): Promise<CamsHttpResponseInit<CaseAppointment>> {
```
jamesobrooks, line 12: This controller is missing a role-based authorization check. Other trustee controllers (e.g. `TrusteeAppointmentsController`, `TrusteeAssignmentsController`) enforce `CamsRole.TrusteeAdmin` and throw an `UnauthorizedError` when the role is absent. Without this check, any authenticated CAMS user can retrieve trustee appointment data regardless of role.

backend/lib/controllers/cases/case-trustee-appointment.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+    const caseId = context.request.params['caseId'];
+    const useCase = new CaseTrusteeAppointmentUseCase(context);
+    const appointment = await useCase.getActiveCaseAppointment(caseId);
+    if (!appointment) {
+      throw new NotFoundError(MODULE_NAME, {
+        message: `No active trustee appointment for case ${caseId}`,
```
jamesobrooks: The raw `caseId` route parameter is interpolated into the `NotFoundError` message before any format validation. Since `toAzureError` includes `error.message` in the response body, an arbitrarily crafted `caseId` value is reflected back to the caller and appears in logs. Consider validating the format (e.g. using `getCaseIdParts`) before use, and omitting `caseId` from the error message body.

backend/lib/use-cases/cases/case-trustee-appointment.use-case.ts
```diff
@@ -0,0 +1,23 @@
+import { ApplicationContext } from '../../adapters/types/basic';
+import { CaseAppointment } from '@common/cams/trustee-appointments';
+import factory from '../../factory';
+import { getCamsError } from '../../common-errors/error-utilities';
+
```
jamesobrooks, line 5: This class stores `ApplicationContext` as a constructor field, which diverges from the pattern used by every other use case in the codebase — including `BackfillCaseAppointmentDatesUseCase` and `SyncTrusteeAppointments` in this same PR — which pass `context` as a parameter on each function call. Storing request-scoped context on the instance creates a subtle hazard if the class is ever cached or reused across requests. The module-object pattern used elsewhere avoids this.

backend/lib/use-cases/dataflows/backfill-case-appointment-dates.ts
```diff
@@ -0,0 +1,204 @@
...
+        originalError,
+        MODULE_NAME,
+        'Failed to backfill appointment dates for batch.',
+      ),
+    };
+  }
+}
+
+/**
+ * Reads the current backfill state from the runtime-state collection.
```
jamesobrooks, line 218: There is a per-record `getActiveCaseAppointment` Cosmos read inside this loop (line 134), meaning up to 100 sequential reads per page. The appointments in this batch were already fetched from Cosmos by `getPageNeedingBackfill` — if the full document were returned from that query rather than the trimmed `{ _id, caseId, trusteeId }` projection, the re-fetch would be unnecessary and `updateCaseAppointment` could use the document directly.

backend/lib/use-cases/dataflows/backfill-case-appointment-dates.ts
```diff
@@ -0,0 +1,204 @@
...
+        });
+      }
+    }
+
+    return { data: results };
+  } catch (originalError) {
+    return {
+      error: getCamsError(
+        originalError,
+        MODULE_NAME,
```
jamesobrooks, line 133: These `warn`-level logs fire once per case with no `apt_date` in DXTR, which is expected for many older cases. At scale this could generate a large volume of Application Insights events and trigger any alert rules scoped to `warn` and above. Consider downgrading to `debug` and emitting a single summary `info` after the loop (e.g. `Skipped N appointments with no DXTR apt_date`).

backend/lib/use-cases/dataflows/backfill-case-appointment-dates.ts
```diff
@@ -0,0 +1,204 @@
...
+  },
+): Promise<MaybeData<CaseAppointmentDateBackfillState>> {
+  try {
+    const repo = factory.getCaseAppointmentDateBackfillStateRepo(context);
+    const now = new Date().toISOString();
+
+    let existingState: CaseAppointmentDateBackfillState | null = null;
+    try {
+      existingState = await repo.read('CASE_APPOINTMENT_DATE_BACKFILL_STATE');
+    } catch (originalError) {
```
jamesobrooks: The state document is read a second time inside `updateBackfillState` (line 172) to retrieve `startedAt` and `id`, even though the caller has already read state via `readBackfillState` earlier in the same invocation. This results in two Cosmos reads per page and introduces a window where concurrent invocations could see different state between the two reads. Passing the existing state as a parameter would eliminate the second read.

backend/function-apps/dataflows/migrations/backfill-case-appointment-dates.ts
```diff
@@ -0,0 +1,327 @@
...
+  logger.info(
+    MODULE_NAME,
+    existingState
+      ? `Resuming backfill from cursor ${lastId}. Already processed ${processedCount} appointments.`
+      : 'Starting fresh case appointment date backfill migration.',
+  );
+
+  const cursorMessage: CursorMessage = { lastId };
+  invocationContext.extraOutputs.set(PAGE, cursorMessage);
+}
```
jamesobrooks, line 115: To prevent duplicate trigger races, consider writing `status: 'IN_PROGRESS'` to the state document here before queuing the first page. A second concurrent POST to the HTTP trigger would then read `IN_PROGRESS` and know work is already underway, rather than independently queueing another page from the same cursor.

backend/function-apps/dataflows/migrations/backfill-case-appointment-dates.ts
```diff
@@ -0,0 +1,327 @@
...
+async function handleError(event: BackfillEvent, invocationContext: InvocationContext) {
+  const logger = ApplicationContextCreator.getLogger(invocationContext);
+
+  logger.error(
+    MODULE_NAME,
+    `Error encountered backfilling appointment for case ${event.caseId}: ${event.error?.message ?? 'Unknown error'}.`,
+  );
+
+  delete event.error;
+  invocationContext.extraOutputs.set(RETRY, [event]);
```
jamesobrooks, line 191: Mutating the incoming event object in place before queuing it is fragile. If the DLQ trigger were ever retried by the Azure Functions runtime, the `error` field would already be gone. Prefer spreading to a new object: `const { error: _, ...retryEvent } = event; invocationContext.extraOutputs.set(RETRY, [retryEvent]);`

backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.ts
```diff
@@ -25,6 +25,17 @@ import { Trustee } from '@common/cams/trustees';
 
 const MODULE_NAME = 'CASES-DXTR-GATEWAY';
 
+export function parseDxtrDate(yymmdd: string | undefined): string | undefined {
```
jamesobrooks, line 28: This validates length and numeric format but doesn't validate the month (positions 2-3) or day (positions 4-5) ranges. A value like `261399` would pass and produce `2026-13-99`, which becomes `Invalid Date` silently when parsed downstream. Consider adding: `const month = parseInt(s.substring(2, 4), 10); const day = parseInt(s.substring(4, 6), 10); if (month < 1 || month > 12 || day < 1 || day > 31) return undefined;`

user-interface/src/case-detail/panels/cards/CaseTrusteeCard.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Trustee } from '@common/cams/trustees';
+import { TrusteeName } from '../TrusteeName';
+import FormattedContact from '@/lib/components/cams/FormattedContact';
+
+interface CaseTrusteeCardProps {
+  trustee: Trustee;
+  trusteeId: string;
+}
+
+export default function CaseTrusteeCard({ trustee, trusteeId }: Readonly<CaseTrusteeCardProps>) {
```
jamesobrooks: This component is only used in one place (`CaseDetailTrusteePanel`). The extraction into a separate `cards/` directory adds indirection without a second caller to justify it. Consider inlining it into the panel unless reuse is anticipated.

backend/lib/use-cases/dataflows/backfill-case-appointment-dates.ts
```diff
@@ -0,0 +1,204 @@
...
+import { MaybeData } from './queue-types';
+import { CaseAppointmentDateBackfillState } from '../gateways.types';
+
+const MODULE_NAME = 'BACKFILL-CASE-APPOINTMENT-DATES-USE-CASE';
+
+export type BackfillAppointment = {
+  _id: string;
+  caseId: string;
+  trusteeId: string;
+};
```
jamesobrooks: `_id` is a MongoDB document identifier — a storage-layer artifact. Defining a type that augments `CaseAppointment` with `_id` here in the use-case layer lets MongoDB internals leak across the abstraction boundary. This may be unavoidable given the cursor-based pagination approach, but it's worth being intentional about.

diabagatekelly (approved):

diabagatekelly:
## Review Questions

Great work on this PR! The implementation is solid with excellent test coverage. I have a few clarifying questions:

### 1. Date Timezone Assumptions
Can you confirm that DXTR `aptDate` values (from `AO_TX.REC`) are truly UTC dates and not local times? 

The frontend formatting explicitly uses UTC timezone:
```typescript
const appointedDateFormatter = new Intl.DateTimeFormat('en-US', {
  year: 'numeric',
  month: '2-digit',
  day: '2-digit',
  timeZone: 'UTC',
});
```

And appends `T00:00:00Z` when parsing: `new Date(\`${isoDate}T00:00:00Z\`)`

If DXTR dates are local times, this could display incorrect dates to users.

### 2. Year 2100 Edge Case
The date parser assumes 20XX:
```typescript
const year = `20${s.substring(0, 2)}`;  // 99 → 2099, 00 → 2000
```

Is the system expected to operate past 2099? If so, should we implement a smarter year parsing strategy (e.g., pivot year logic: 50-99 → 1950-1999, 00-49 → 2000-2049)?

### 3. Date Validation
Should we validate that parsed dates are calendar-valid (e.g., reject "2026-02-31")? Currently the parser trusts DXTR data quality and doesn't check if the resulting date is real.

### 4. Internal Contact Card Heading
I noticed the internal contact card heading text changed. Was this intentional for reuse between trustee profile and case detail, or should each context have different copy?

---

Overall this is excellent work - no blocking issues, just these clarifications for documentation and potential future enhancements.

jalba-flexion:
Addressed all open review comments (excluding the CamsRole authorization check, which will be handled separately):

- **`parseDxtrDate` month/day range validation** — added `monthNum`/`dayNum` range guards so values like `261399` return `undefined` instead of `2026-13-99`. Added 4 new tests covering invalid month and day values.
- **`caseId` reflected in error message body** — validate format with `getCaseIdParts` before use; throw `BadRequestError` with a safe message if invalid. Omit raw `caseId` from the `NotFoundError` message body.
- **`ApplicationContext` as constructor field** — removed constructor storage; `context` is now passed as a parameter on each `getActiveCaseAppointment` call, matching the pattern used by every other use case in the codebase.
- **Per-record `getActiveCaseAppointment` in backfill loop** — changed `BackfillAppointment` to carry the full `CaseAppointment & { _id }` from the page query. `backfillAppointmentDates` now uses the passed document directly, eliminating up to 100 sequential Cosmos reads per page.
- **`warn` logs per case with no DXTR date** — downgraded to `debug` per-case; added a single `info` summary after the loop (e.g. "Skipped N appointments with no DXTR apt_date.").
- **Double Cosmos read in `updateBackfillState`** — added optional `existingState` param; `handleStart` and `handlePage` pass the state they already read, avoiding a redundant round-trip per page.
- **Write `IN_PROGRESS` before queuing first page** — `handleStart` now persists `IN_PROGRESS` status before enqueuing the first cursor message, preventing a concurrent HTTP trigger from independently starting a second run.
- **Mutating incoming event before retry queue** — replaced `delete event.error` with a destructured spread (`{ error: _, ...retryEvent }`) so the original event object is not modified in place.
- **`CaseTrusteeCard` only used in one place** — inlined the JSX into `CaseDetailTrusteePanel`, merged the SCSS into `CaseDetailTrusteePanel.scss`, migrated the unique tests (profile link href, public address) into `CaseDetailTrusteePanel.test.tsx`, and deleted the `cards/CaseTrusteeCard` files.

jalba-flexion:
Addressed @sourcery-ai's complexity suggestion:

- **`BackfillEvent` → `BackfillRetryMessage`** — removed `error?: Error` from the queue message type; `Error` objects don't serialize reliably over storage queues. Replaced with `lastErrorMessage?: string`. `handleError` no longer needs to strip the field before forwarding to RETRY; `handleRetry` stores a string on failure rather than reconstructing an `Error`.

- **`processBackfillPage` added to use-case** — moved state-read → page-fetch → backfill → state-update coordination into the use-case layer. `handlePage` is now a thin queue I/O wrapper (~30 lines vs ~70). Added 5 tests for `processBackfillPage` covering empty page, hasMore, last-page, state-read failure, and per-appointment write failures.

PR #2235, opened 2026-04-08
CAMS-727 - Added select option for Lead Trial Attorney

mattstankey:
# Purpose

Add support for designating a Lead Trial Attorney on a case, distinct from regular Trial Attorney assignments.

# Major Changes

  Common (common/)
                                                                                                                                                                                                       
  src/cams/roles.ts
  - Added LeadTrialAttorney to both CamsRole and AssignableRole enums                                                                                                                                  
                                                                                                                                                                                                       
  src/cams/cases.ts                                                                                                                                                                                    
  - Changed leadTrialAttorneys?: CamsUserReference[] → leadTrialAttorney?: CamsUserReference on CaseBasics                                                                                             
                                                                                                                                                                                                       
  src/cams/assignments.ts                                                                                                                                                                              
  - Widened StaffAssignmentAction.role to AssignableRoleType to allow LeadTrialAttorney                                                                                                                
                                                                                                                                                                                                       
  ---                                                                                                                                                                                                  
  Backend (backend/)                                                                                                                                                                                   
                                                                                                                                                                                                       
  lib/use-cases/case-assignment/case-assignment.ts                                                                                                                                                     
  - When assigning a LeadTrialAttorney, uses TrialAttorney as the "assignable role" for staff validation (they must be a trial attorney to be a lead)                                                  
  - Stores the assignment with role: LeadTrialAttorney                                                                                                                                                 
  - Filters existing assignment records by role to prevent cross-role unassignment                                                                                                                     
                                                                                                                                                                                                       
  lib/use-cases/cases/case-management.ts                                                                                                                                                               
  - getCaseDetail: splits assignments — TrialAttorney → caseDetails.assignments, first LeadTrialAttorney → caseDetails.leadTrialAttorney                                                               
  - searchCases: same split for case list results                                                                                                                                                      
  - getCaseAssignments: added ?? [] fallback so .filter() doesn't throw when a case has no assignments                                                                                                 
                                                                                                                                                                                                       
  ---                                                                                                                                                                                                  
  Frontend (user-interface/)                                                                                                                                                                           
                                                                                                                                                                                                       
  src/staff-assignment/modal/AssignAttorneyModal.tsx                                                                                                                                                   
  - Added leadTrialAttorney state (initially null)                                                                                                                                                     
  - Pre-populates from bCase.leadTrialAttorney when a case is loaded                                                                                                                                   
  - Auto-selects: if exactly 1 attorney is checked, auto-sets them as lead                                                                                                                             
  - Disables submit button when no lead is selected                                                                                                                                                    
                                                                                                                                                                                                       
  src/staff-assignment/modal/AssignAttorneyModalView.tsx                                                                                                                                               
  - Replaced the ComboBox with a native USWDS <select> scoped to only the checked attorneys                                                                                                            
                                                                                                                                                                                                       
  src/staff-assignment/modal/assignAttorneyModalUseCase.ts                                                                                                                                             
  - submitValues posts two API calls: one for TrialAttorney (full list) and one for LeadTrialAttorney (single)                                                                                         
  - show pre-populates checkboxes from both assignments and leadTrialAttorney, deduped                                                                                                                 
  - Both success and error callbacks include leadTrialAttorney                                                                                                                                         
                                                                                                                                                                                                       
  src/case-detail/panels/CaseDetailTrusteeAndAssignedStaff.tsx                                                                                                                                         
  - Displays lead trial attorney in its own <li> with "Lead Trial Attorney" role label                                                                                                                 
  - Deduplicates: lead attorney is excluded from the regular Trial Attorney list                                                                                                                       
  - (unassigned) placeholder only shows when both assignments and leadTrialAttorney are absent                                                                                                         
                                                                                                                                                                                                       
  src/case-detail/CaseDetailScreen.tsx                                                                                                                                                                 
  - Passes leadTrialAttorney through the assignment callback update 

# Definition of Done:

## Summary by Sourcery

Add explicit support for designating and handling a single Lead Trial Attorney per case across shared types, backend case assignment/case management flows, and UI assignment/case-detail screens.

New Features:
- Introduce LeadTrialAttorney as an assignable role alongside TrialAttorney and expose it on case basics and case detail views.
- Allow users to select a Lead Trial Attorney in the Assign Attorney modal, including auto-selection when a single attorney is checked.

Enhancements:
- Split mixed trial-attorney assignments into regular TrialAttorney and a distinct leadTrialAttorney field in case detail and search results.
- Ensure assigning a LeadTrialAttorney validates against trial-attorney eligibility, avoids removing regular TrialAttorney assignments, and scopes unassignment by role.
- Pre-populate and deduplicate assignment checklists and case-detail displays when a lead trial attorney is present, and hide the unassigned placeholder when any attorney is assigned as lead.

Tests:
- Extend backend and frontend tests to cover LeadTrialAttorney validation, assignment persistence, case-detail/search splitting, modal behavior, and deduplication logic.

### DIFF
--- a/backend/lib/use-cases/case-assignment/case-assignment.ts
+++ b/backend/lib/use-cases/case-assignment/case-assignment.ts
@@ -79,15 +79,19 @@ export class CaseAssignmentUseCase {
     const { officeCode } = divisionCodeMap.get(bCase.caseId.substring(0, 3));
 
     const officesRepo = factory.getOfficesRepository(context);
+    const assignableRole = role === CamsRole.LeadTrialAttorney ? CamsRole.TrialAttorney : role;
     const calls = [];
     const validatedAssignments: CamsUserReference[] = [];
     newAssignees.forEach((assignee) => {
       calls.push(
         officesRepo
-          .search({ officeCode, userId: assignee.id, role })
+          .search({ officeCode, userId: assignee.id, role: assignableRole })
           .then((response) => {
             return response.find((staff) => {
-              if (staff.roles.includes(role as CamsRoleType) && staff.name === assignee.name) {
+              if (
+                staff.roles.includes(assignableRole as CamsRoleType) &&
+                staff.name === assignee.name
+              ) {
                 return true;
               }
             });
@@ -143,7 +147,9 @@ export class CaseAssignmentUseCase {
     const existingAssignmentRecordsMap = await assignmentRepo.getAssignmentsForCases([
       bCase.caseId,
     ]);
-    const existingAssignmentRecords = existingAssignmentRecordsMap.get(bCase.caseId) ?? [];
+    const existingAssignmentRecords = (existingAssignmentRecordsMap.get(bCase.caseId) ?? []).filter(
+      (a) => a.role === role,
+    );
     for (const existingAssignment of existingAssignmentRecords) {
       const stillAssigned = listOfAssignments.find((newAssignment) => {
         return (
--- a/backend/lib/use-cases/case-assignment/case-assignment.ts
+++ b/backend/lib/use-cases/case-assignment/case-assignment.ts
@@ -47,6 +47,11 @@ export class CaseAssignmentUseCase {
         message: 'User does not have appropriate access to create assignments for this office.',
       });
     }
+    if (role === CamsRole.LeadTrialAttorney && newAssignees.length > 1) {
+      throw new AssignmentError(MODULE_NAME, {
+        message: 'Only one Lead Trial Attorney may be assigned to a case.',
+      });
+    }
     await this.assignTrialAttorneys(context, caseId, newAssignees, role);
 
     // Reassign all member cases if this is a joint administration lead case.
--- a/user-interface/src/staff-assignment/modal/assignAttorneyModalUseCase.ts
+++ b/user-interface/src/staff-assignment/modal/assignAttorneyModalUseCase.ts
@@ -2,7 +2,7 @@ import { AttorneyUser, CamsUserReference } from '@common/cams/users';
 import { deepEqual } from '@common/object-equality';
 import Api2 from '@/lib/models/api2';
 import { ResponseBody } from '@common/api/response';
-import { CamsRole } from '@common/cams/roles';
+import { CamsRole, AssignableRole } from '@common/cams/roles';
 import { getCamsUserReference } from '@common/cams/session';
 import {
   AssignAttorneyModalStore,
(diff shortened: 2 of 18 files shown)


jalba-flexion (approved):

user-interface/src/staff-assignment/modal/assignAttorneyModalUseCase.ts
```diff
@@ -98,10 +101,16 @@ const assignAttorneyModalUseCase = (
           attorneyList: finalAttorneyList,
           role: CamsRole.TrialAttorney,
         });
+        await Api2.postStaffAssignments({
```
diabagatekelly: If one of the API calls fails, there might be an inconsistent state. We might want to wrap both API calls in a single try/catch and if one fails, roll back the other, or something along these lines.

backend/lib/use-cases/case-assignment/case-assignment.ts
```diff
@@ -143,7 +147,9 @@ export class CaseAssignmentUseCase {
     const existingAssignmentRecordsMap = await assignmentRepo.getAssignmentsForCases([
       bCase.caseId,
     ]);
-    const existingAssignmentRecords = existingAssignmentRecordsMap.get(bCase.caseId) ?? [];
+    const existingAssignmentRecords = (existingAssignmentRecordsMap.get(bCase.caseId) ?? []).filter(
```
diabagatekelly, line 155: Just checking, since this is an array, does this mean on the DB side more than one lead attorney can be assigned?

diabagatekelly:
Some remaining questions:                                                                                                                                                                               
 - Should joint admin member cases inherit the lead from the lead case?                                                                                                                               
- Will adding leadTrialAttorney require a migration?

mattstankey:
> Some remaining questions:
> 
> * Should joint admin member cases inherit the lead from the lead case?
> * Will adding leadTrialAttorney require a migration?

Some feedback from Claude:

1.  They already do. Lines 52–63 of case-assignment.ts propagate any assignment (including LeadTrialAttorney) to all joint admin member cases automatically — the same assignTrialAttorneys call is looped over memberCaseIds via CONSOLIDATION_FROM administrative references.

2.  No migration needed. leadTrialAttorney is an optional field derived at read time from existing ASSIGNMENT records with role === 'LeadTrialAttorney'. Since no existing records have that role, existing cases will simply have undefined for the field (shown as unassigned in the UI). When a user designates a lead via the modal, the record is created.

backend/lib/use-cases/case-assignment/case-assignment.ts
mattstankey: From Claude:
```
Correct that there's no DB-level unique constraint. However, the application layer enforces single-lead semantics in two ways:                                                               
  - The UI always sends attorneyList: [leadTrialAttorney] (single item) for the LeadTrialAttorney role                                                                                                 
  - assignTrialAttorneys (line 150) filters existing records by role and marks any not in the new list as unassignedOn, effectively replacing the previous lead
```

I also just added an additional guard to throw an error if we ever come back with multiple lead trial attorneys somehow.

diabagatekelly (approved):

PR #2242, opened 2026-04-10
CAMS-721: No match trustee

diabagatekelly:
# Looking to create a bug?

Please go to the `Preview` tab and select the appropriate template. **Otherwise, please delete this section.**

* [Bug](?expand=1&template=bug.md)

# Purpose

Describe the reason why this was developed.

# Major Changes

Describe what has changed.

# Testing/Validation

# Notes

# Definition of Done:

## Summary by Sourcery

Add trustee phonetic search capabilities and manual trustee resolution for verification orders, including backfill tooling and UI support.

New Features:
- Expose a new trustee search API supporting name and phonetic matching, consumed by a UI modal for manually selecting trustees on verification orders.
- Store phonetic search tokens on trustee records and index them in Cosmos to enable phonetic and exact-name lookup.
- Enrich trustee match verification orders with resolved trustee names and court names so the UI can display meaningful labels instead of raw IDs.
- Add a dataflow function to backfill phonetic tokens for existing trustees and a script command to fully clean trustee-related data from dev Cosmos DB.

Enhancements:
- Update trustee match verification approval flows end-to-end to persist and propagate the resolved trustee name alongside the ID.
- Refine trustee verification UI to use in-place search buttons instead of navigation links and to support manual trustee search and approval flows.

Build:
- Add a Cosmos DB index on trustee phoneticTokens to support efficient phonetic searches.

Deployment:
- Register the new trustee search HTTP function and the trustee phonetic backfill dataflow in the Azure Functions app configuration.

Documentation:
- Extend trustee migration testing documentation with instructions for the new clean-all option and full-reset workflows.

Tests:
- Add comprehensive unit and integration tests covering trustee phonetic search, manual trustee resolution, court-name enrichment, the new backfill use case, and updated UI flows.

Chores:
- Extend seed-test-trustees tooling with a clean-all mode that deletes all trustee-related data and runtime state from the dev database.


backend/lib/controllers/trustee-match-verification/trustee-match-verification.controller.ts
```diff
@@ -75,7 +76,38 @@ export class TrusteeMatchVerificationController {
...
-        return { ...verification, matchCandidates: enrichedCandidates };
+        // For approved records where the resolved trustee was manually selected (not a candidate),
+        // fetch their name so the UI can display it instead of the raw trustee ID.
+        let resolvedTrusteeName = verification.resolvedTrusteeName;
+        if (
+          verification.status === 'approved' &&
+          verification.resolvedTrusteeId &&
+          !resolvedTrusteeName &&
+          !enrichedCandidates.find((c) => c.trusteeId === verification.resolvedTrusteeId)
+        ) {
```
jalba-flexion: Complicated/long conditional. Strip out to separate variables?

backend/lib/controllers/trustee-match-verification/trustee-match-verification.controller.ts
```diff
@@ -75,7 +76,38 @@ export class TrusteeMatchVerificationController {
...
+          try {
+            const resolved = await trusteesRepo.read(verification.resolvedTrusteeId);
+            resolvedTrusteeName = resolved.name;
+          } catch {
+            // name unavailable — UI falls back to trustee ID
```
jalba-flexion: Nothing is needed in this catch block?

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.ts
```diff
@@ -318,6 +354,20 @@ export class TrusteesMongoRepository extends BaseMongoRepository implements Trus
     }
   }
 
+  async setPhoneticTokens(trusteeId: string, tokens: string[]): Promise<void> {
+    try {
+      const doc = using<TrusteeDocument>();
+      const query = and(doc('documentType').equals('TRUSTEE'), doc('trusteeId').equals(trusteeId));
+      await this.getAdapter<TrusteeDocument>().updateOne(query, {
+        $set: { phoneticTokens: tokens },
+      } as unknown as Partial<TrusteeDocument>);
```
diabagatekelly: Not a huge deal, but we could create a dedicated type instead of as unknown as Partial<TrusteeDocument>).

user-interface/src/data-verification/trustee-verification/TrusteeMatchVerificationAccordion.tsx
```diff
@@ -325,16 +366,32 @@ export function TrusteeMatchVerificationAccordion(props: TrusteeMatchVerificatio
           {viewMode === 'resolved' ? (
             <p className="resolved-statement" data-testid="resolved-statement">
               Trustee{' '}
-              {order.matchCandidates.find((c) => c.trusteeId === order.resolvedTrusteeId)
-                ?.trusteeName ?? order.resolvedTrusteeId}{' '}
+              {order.resolvedTrusteeName ??
+                order.matchCandidates.find((c) => c.trusteeId === order.resolvedTrusteeId)
```
jalba-flexion: Make this a variable for readability?
matchResolvedTrusteeId?.trusteeName ?? order.resolvedTrusteeId

backend/lib/controllers/trustee-match-verification/trustee-match-verification.controller.ts
mattstankey: Reply from Claude:
`Not worth changing. The empty catch is intentional — if the trustee lookup fails, the name is just unavailable and the UI falls back to the trustee ID. The comment explains this. Sourcery also noticed it and agreed the comment is sufficient. I'd just reply to the comment explaining the intent. `

backend/lib/use-cases/trustees/trustee-search.use-case.ts
```diff
@@ -0,0 +1,103 @@
...
+      // Cap results
+      const capped = merged.slice(0, MAX_RESULTS);
+
+      // Fetch appointments and apply court filter
+      const results: TrusteeSearchResult[] = [];
+      await Promise.all(
+        capped.map(async ({ trustee, matchType }) => {
+          const appointments = await appointmentsRepo.getTrusteeAppointments(trustee.trusteeId);
+
+          if (courtId && !appointments.some((appt) => appt.courtId === courtId)) {
```
diabagatekelly, line 53: Instead of fetching all appointments, then filtering by court Id, would it make sense to get the trusteeIDs with the correct courtId first, then fetch only the appointments we care about? It might be a bit more efficient.

backend/lib/controllers/trustee-match-verification/trustee-match-verification.controller.ts
mattstankey: Done — split into two named booleans: `isApprovedWithUnresolvedName` and `notAlreadyInCandidates`.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.ts
mattstankey: Good catch — and there was actually a bug hiding here. The `$set` wrapper was double-applied: the repository passed `{ $set: { phoneticTokens } }` to the adapter, which passed it to the humble object, which wrapped it again as `{ $set: { $set: { phoneticTokens } } }`. Fixed by dropping the `$set` in the repository and passing `{ phoneticTokens: tokens }` directly as `Partial<TrusteeDocument>`, which is type-safe since `phoneticTokens` is defined on `Trustee`. Updated the test assertions to match.

user-interface/src/data-verification/trustee-verification/TrusteeMatchVerificationAccordion.tsx
mattstankey: Done — extracted to `matchedCandidateName`, so the return is a clean three-way fallback on one line.

user-interface/src/data-verification/trustee-verification/TrusteeSearchModal.tsx
```diff
@@ -0,0 +1,210 @@
...
+      courts.find((c) => c.courtDivisionCode === selectedCourtId))
+    : undefined;
+
+  async function handleFilterChange(value: string) {
+    if (!value || value.length < 2) {
+      setSearchResults([]);
+      return;
+    }
+    try {
+      const response = await Api2.searchTrustees(value, selectedCourtEntry?.courtId);
```
diabagatekelly: We might want to consider using our useDebounceCallback here so we don't hit the API on every keystroke.

backend/lib/use-cases/trustees/trustee-search.use-case.ts
mattstankey: From Claude:
```
The repository only exposes getTrusteeAppointments(trusteeId) — no court-based query. Implementing diabagatekelly's suggestion would require:                                                        
                                                                                                                                                                                                       
  1. A new repository method on TrusteeAppointmentsRepository — something like getTrusteeIdsByCourtId(courtId) that queries the appointments collection filtered by court                              
  2. New implementation in the Mongo repository (and mock)                                                                                                                                             
  3. Changed use case logic: get qualifying trustee IDs upfront, intersect with the capped trustees, then fetch full appointments only for matches                                                     
                                                                                                                                                                                                       
  Current approach: up to 25 appointment fetches, some discarded                                                                                                                                       
  Proposed approach: 1 upfront appointments query by court + ≤25 appointment fetches for matches only                                                                                                  
                                                                                                                                                                                                       
  The optimization is real but modest given the 25-result cap. It would also need test coverage for the new repository method.                                                                         
                                                                                                                                                                                                       
  My read: the suggestion is sound but it's scope — new interface method, implementation, mock, and tests. Given the 25-result cap makes the current approach acceptable, I'd lean toward replying that it's a good idea for a follow-up rather than doing it here.
```

We also don't have a way to get appointments by courtID.

jalba-flexion (approved):

mattstankey (approved):

PR #2269, opened 2026-04-20
CAMS-730 Harden sub-security-scan.yml to use least privilege

fmaddenflx:
# Purpose

Reduce the blast radius of a compromised security scan workflow by eliminating long-lived Azure credentials and replacing broad secret inheritance with explicit, minimal secret declarations.

# Major Changes

- Replaced `secrets: inherit` in `continuous-deployment.yml` with explicit secret pass-through listing only the 4 secrets the workflow actually needs
- Added `workflow_call.secrets:` block to `sub-security-scan.yml` declaring only required secrets
- Added `permissions:` blocks to both scan jobs (`contents: read`, `security-events: write` for sca-scan, `id-token: write` for OIDC)
- Migrated `azure/login` from long-lived `AZURE_CREDENTIALS` to OIDC Workload Identity Federation using scoped app registration
- Switched Azure Storage commands from account-key auth to `--auth-mode login`, eliminating the need to fetch storage keys (and the `AZ_SECURITY_SCAN_RG` secret entirely)
- Added runbook script (`ops/scripts/utility/setup-security-scan-federated-credential.sh`) for creating the Azure federated credential and Storage Blob Data Contributor role assignment scoped to the security scan storage account

# Testing/Validation

Workflow syntax validated by pre-commit hook. Azure federated credential and role assignment created via the runbook and verified in Azure portal. GitHub Actions repository variables (`AZ_SECURITY_SCAN_CLIENT_ID`, `AZ_TENANT_ID`, `AZ_SUBSCRIPTION_ID`) set for the OIDC login.

# Notes

The `AZURE_CREDENTIALS` secret declaration is retained in the workflow with `required: false` to avoid breaking callers during transition, but is no longer referenced by any step. It can be removed entirely in a follow-up once confirmed clean.

`AZ_SECURITY_SCAN_RG` is no longer passed to or used by this workflow — `--auth-mode login` resolves the storage endpoint from the account name alone without needing the resource group.

This change applies to the Flexion environment only; USTP infrastructure is unaffected.

# Definition of Done:

## Summary by Sourcery

Harden the security scan workflow by switching to OIDC-based Azure auth and least-privilege secret usage, and update documentation to reflect the workflow’s reusable status and new inputs.

Enhancements:
- Limit sub-security-scan workflow inputs to explicitly declared secrets and remove reliance on broad secrets inheritance.
- Migrate Azure login in security scan jobs from long-lived AZURE_CREDENTIALS to OIDC Workload Identity Federation using repository variables for identity.
- Use Azure CLI login-based authentication for storage operations instead of account keys, removing the need for the security scan resource group secret.
- Tighten GitHub Actions job permissions for the security scan jobs to only those required for contents, security events, and id-token.
- Add a runbook script to provision the federated credential and scoped role assignment for the security scan storage account.

Documentation:
- Update the workflow diagram documentation to classify sub-security-scan.yml as a reusable workflow and to document its new secrets inputs and relationships.

### DIFF
--- a/.github/workflows/sub-security-scan.yml
+++ b/.github/workflows/sub-security-scan.yml
@@ -3,10 +3,6 @@ name: Security
 on:
   workflow_call:
     secrets:
-      SNYK_OAUTH_CLIENT_ID:
-        required: true
-      SNYK_OAUTH_CLIENT_SECRET:
-        required: true
       AZ_SECURITY_SCAN_CLIENT_ID:
         required: true
       AZ_TENANT_ID:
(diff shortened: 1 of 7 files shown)


mattstankey (approved):

.github/workflows/sub-security-scan.yml
```diff
@@ -35,28 +48,25 @@ jobs:
       - uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
         if: always()
         with:
-          creds: ${{ secrets.AZURE_CREDENTIALS }}
+          client-id: ${{ vars.AZ_SECURITY_SCAN_CLIENT_ID }}
+          tenant-id: ${{ vars.AZ_TENANT_ID }}
+          subscription-id: ${{ vars.AZ_SUBSCRIPTION_ID }}
```
fmaddenflx: I updated the skill to place these in secrets rather than variables.  Perhaps we want to update this to do the same?

ops/scripts/utility/setup-security-scan-federated-credential.sh
```diff
@@ -0,0 +1,121 @@
...
+fi
+
+echo ""
+echo "==> Done."
+if [[ -z "$EXISTING_APP_ID" ]]; then
+  echo "    Add the following as GitHub Actions repository variables (not secrets):"
+  echo "    AZ_SECURITY_SCAN_CLIENT_ID = $APP_ID"
+  echo "    AZ_TENANT_ID               = $TENANT_ID"
+  echo "    AZ_SUBSCRIPTION_ID         = $SUBSCRIPTION_ID"
+fi
```
fmaddenflx: This should change.  We decided they should be secrets.

PR #2277, opened 2026-04-21
CAMS-596 Fix bug caused by static db connection held by abstract mssql client

btposey:
# Purpose

Fix bug caused by static db connection held by abstract mssql client

# Major Changes

* Added a map to the abstract sql client that holds connections for specific "server + port + dbname" tuples.

# Testing/Validation

* Unit tests written to confirm function of the map.

## Summary by Sourcery

Switch the abstract MSSQL client to manage separate connection pools per database configuration instead of a single static pool, and add tests to verify the new behavior.

Bug Fixes:
- Fix incorrect sharing of a single static database connection pool across clients using different MSSQL database configurations.

Enhancements:
- Introduce keyed connection pool management in the abstract MSSQL client using a map keyed by server, port, and database name.

Tests:
- Add unit tests to ensure separate pools are created for different database configs and reused for identical configs in the abstract MSSQL client.

### DIFF
--- a/backend/lib/adapters/gateways/abstract-mssql-client.test.ts
+++ b/backend/lib/adapters/gateways/abstract-mssql-client.test.ts
@@ -52,12 +52,15 @@ vi.mock('mssql', async (importOriginal) => {
   };
 });
 
+class TestDbClient extends AbstractMssqlClient {
+  constructor(_context: ApplicationContext, config: IDbConfig, childModuleName: string) {
+    super(config, childModuleName);
+  }
+}
+
 describe('Abstract MS-SQL client', () => {
   test('should get appropriate results', async () => {
-    const applicationContext = await createMockApplicationContext();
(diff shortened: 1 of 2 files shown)


btposey:
Thanks for the review @sourcery-ai.

**Suggestion 1 (pool key too narrow):** Acknowledged. In CAMS, each environment uses a single credential per database, so this is not a foreseeable problem today. We'll revisit the key structure if we encounter an axis of change that requires it (e.g., per-request credentials or multi-tenant auth).

**Suggestion 2 (unbounded pool accumulation):** Not a concern for this codebase. Databases are mocked in tests so pool state doesn't accumulate there. In production, Azure Function instances have short lifecycles and the existing `deferClose` shutdown hook closes all accumulated pools on teardown.

jamesobrooks (approved):

PR #2281, opened 2026-04-22
CAMS-596 Replace legacy SQL client with latest abstract MSSQL client

btposey:
# Purpose

Tech Debt: Replace legacy SQL client with latest abstract MSSQL client

# Major Changes

Replace legacy SQL client with latest abstract MSSQL client

# Testing/Validation

* Improved unit tests

## Summary by Sourcery

Refactor DXTR-related MSSQL access to use the shared AbstractMssqlClient and update consumers and tests accordingly.

Enhancements:
- Convert DXTR cases, orders, case docket, offices gateways, and the SQL healthcheck to extend AbstractMssqlClient instead of using the legacy database utility and factory SQL connections.
- Adjust factory wiring to construct DXTR gateways with the application context and remove the legacy SQL connection accessor.
- Tighten error handling paths in DXTR orders and case docket gateways to surface failed query scenarios consistently.
- Simplify AbstractMssqlClient.executeQuery to return recordsets directly and mark its connection pool map as readonly.

Tests:
- Update DXTR gateway tests to spy on AbstractMssqlClient.executeQuery, align with the new method signatures, and expand coverage of whitespace collapsing, additional identifier handling, and error conditions.
- Adapt healthcheck, case docket, orders, cases, and offices tests to the revised gateway construction pattern and query result shape.
- Remove obsolete tests tied to the deprecated database utility layer.


btposey:
Thanks for the review @sourcery-ai.

**Suggestion 1 (results shape inconsistency):** Confirmed and fixed. `AbstractMssqlClient.executeQuery` was returning `results: result.recordset` (the bare array) while all DXTR gateway callers expected `results` to be the full `mssql.IResult<T>` object so they could access `.recordset` on it. We've changed `executeQuery` to return `results: result` (the full IResult), bringing the contract in line with what all callers expect. Also updated ACMS and ATS gateways which had the inverse assumption (treating `results` as a bare array) and updated all affected test mocks accordingly.

**Suggestion 2 (tighter QueryResults typing):** Acknowledged as valid. Fully generic `QueryResults<T>` would require updating `handleQueryResult`, all gateway callbacks, and every call site — a broader refactor with good ROI but outside the scope of this PR. Tracked as a follow-on improvement.

btposey:
Update: The previous comment was slightly off. After deeper investigation, the actual fix was the reverse — `AbstractMssqlClient.executeQuery` was correctly returning `results: result.recordset` (the bare array), but DXTR, ACMS, and ATS gateways had all been written expecting `results` to be the full `mssql.IResult<T>` object so they could access `.recordset` on it again. We standardised by changing `executeQuery` to return `results: result` (the full `IResult`), which is consistent with what all gateway callers expected, and updated ACMS and ATS gateways that had the inverse assumption. All tests updated and passing.

Suggestion 2 (tighter typing) remains a valid follow-on — tracked for a future cleanup.

jalba-flexion (approved):

PR #2314, opened 2026-04-27
CAMS-742 Split trustee name into first, middle, and last fields

cruzjone-flexion:
# Purpose

Support structured name data for trustees so staff can view, sort,
and edit names using separate first, last, and middle name fields
rather than a single concatenated string.

# Major Changes

- Added `firstName`, `lastName`, and `middleName` to `TrusteeCore` type; `name` is now computed on write via `computeTrusteeName()`
- Updated ATS mapping to populate separate name fields from source (`FIRST_NAME`, `MIDDLE`, `LAST_NAME`)
- Trustee list displays names as "Last, First Middle" and sorts alphabetically by `lastName` by default
- Trustee detail page shows "First Middle Last" in the header with individually labeled overview fields; middle name hidden when absent
- Add/edit form replaced single name input with three separate fields (first and last required, middle optional) with per-field length validators
- Form labels prefixed with "Trustee" (e.g., "Trustee First Name") for clarity

# Testing/Validation

Implemented using TDD. Unit tests cover the new type, validators, formatter utility, ATS mapping, repository writes, and all UI components. Manual testing steps cover list display/sort, detail page with and without middle name, add/edit form validation, and pre-population on edit.

# Notes

The `name` field is retained as a denormalized field in MongoDB to support existing regex-based search queries (`findTrusteesByName`, `searchTrusteesByName`, `findTrusteeByNameAndState`). It is always kept in sync on write — users never edit it directly. Removing it is a future cleanup once search queries are migrated to query `firstName`/`lastName` directly.

Existing trustees in MongoDB without `firstName`/`lastName` will be handled via a full delete-and-reimport from ATS before this ships. The `formatTrusteeListName` utility includes a `fallbackName` parameter for safe display of any legacy records during the transition.

# Definition of Done:

## Summary by Sourcery

Introduce structured trustee name fields and update UI, validation, and persistence to use first, middle, and last names while preserving the legacy concatenated name for compatibility.

New Features:
- Support separate first, middle, and last name fields for trustees in domain types, ATS imports, and API payloads.
- Add utilities to compute full trustee names and format list display names with legacy-safe fallbacks.
- Enable client-side sorting of the trustees list by last name with an accessible, toggleable Name column header.

Enhancements:
- Refine trustee list and detail views to display names in structured formats, including middle names when present and fallbacks for legacy records.
- Adjust trustee create/edit flows to send minimal patch payloads, including explicit middleName nulling on removal, and to avoid persisting empty nested objects.
- Tighten Trustee MongoDB repository handling to recompute denormalized name and search tokens on write based on structured name parts.

Tests:
- Extend unit, integration, and accessibility tests across common utilities, ATS mapping, use cases, repositories, UI components, validation, and API wrappers to cover structured trustee names, sorting behavior, and new field constraints.


backend/lib/adapters/gateways/mongo/trustees.mongo.repository.ts
```diff
@@ -260,9 +267,11 @@ export class TrusteesMongoRepository extends BaseMongoRepository implements Trus
       const doc = using<TrusteeDocument>();
       const query = and(doc('documentType').equals('TRUSTEE'), doc('trusteeId').equals(id));
 
+      const computedName = computeTrusteeName(input.firstName, input.middleName, input.lastName);
```
mattstankey: from Claude: 

`input` is typed `Partial<TrusteeInput>`, so input.firstName is `string | undefined`. computeTrusteeName's signature takes firstName: string — calling `.trim()` on undefined is a runtime crash.

TypeScript should flag this as a type error. If it isn't caught by the compiler it will fail at runtime for any PATCH that doesn't include name fields (e.g., updating only address info).

**Suggestion**: Guard with nullish coalescing or adjust the signature: `computeTrusteeName(input.firstName ?? '', input.middleName, input.lastName ?? '')`.

user-interface/src/trustees/forms/TrusteePublicContactForm.tsx
```diff
@@ -92,7 +95,9 @@ function TrusteePublicContactForm(props: Readonly<TrusteePublicContactFormProps>
     const trimmedCompanyName = formData.companyName?.trim();
 
     return {
-      name: formData.name,
+      firstName: formData.firstName,
+      lastName: formData.lastName,
+      middleName: isCreate ? formData.middleName : (formData.middleName ?? null),
```
mattstankey: `TrusteeInput.middleName` is typed as `string | undefined`, not `string | null | undefined`. Sending null is intentional (PATCH semantics: explicitly clear vs. omit), but the type should reflect this — either as `string | null | undefined` on the PATCH input type, or through a separate TrusteePatch type.

user-interface/src/trustees/panels/TrusteeOverviewCard.scss
```diff
@@ -13,6 +13,22 @@
         margin-bottom: 0.4rem;
         line-height: 1.5;
       }
+
+      .trustee-name-fields {
```
mattstankey: Are these styles used at all?  This seems to be dead CSS.

mattstankey (dismissed):

cruzjone-flexion:
@sourcery-ai review

mattstankey (commented):
Looks good now.

common/src/cams/trustees.ts
```diff
@@ -73,14 +73,39 @@ export type ZoomInfo = {
...
 };
 
+export function computeTrusteeName(
+  firstName: string,
+  middleName: string | undefined,
+  lastName: string,
+): string {
+  const parts = [firstName.trim(), middleName?.trim(), lastName.trim()].filter(Boolean);
+  return parts.join(' ');
+}
```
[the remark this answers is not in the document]
mattstankey: We don't need these due to the Person object making only middleName optional.

diabagatekelly (approved):

diabagatekelly (approved):

mattstankey (commented):
Updates look good 👍

PR #2334, opened 2026-05-01
CAMS-737 Add trustee name search filter

jalba-flexion:
# Purpose

Allow users to quickly find a specific trustee on the Trustee Management list by searching by name, without having to scroll through the full list.

# Major Changes

* Added a "Trustee Name" plain text input as the first filter control inside the existing Filters accordion
* Name search calls the existing `GET /api/trustee-search` endpoint (debounced 300ms, 2-character minimum), reusing the same phonetic/fuzzy search used by the Trustee Search Modal
* Search results are intersected client-side with the already-loaded trustee list so district and chapter filters continue to compose correctly
* Clear button follows the same `filter-control-header` / `filter-clear-link` pattern as the existing District and Chapter filters
* Live ARIA announcement fires after first name filter interaction (mirrors chapter filter behavior)
* AppInsights telemetry tracks `Trustee Name Filter Changed` and `Trustee Name Filter Cleared` events

# Testing/Validation

Implemented using TDD. 57 tests pass across the two affected test files (8 new in `TrusteesList.test.tsx`, 5 new in `TrusteeDistrictFilter.test.tsx`). Tests cover: debounce behavior, 2-character minimum, composition with district/chapter filters, Clear button, live region announcement, and AppInsights events.

# Notes

The name filter uses server-side search (same `Api2.searchTrustees` as the Trustee Search Modal) rather than client-side substring matching, giving users phonetic/fuzzy matching for free. The returned `trusteeId`s are intersected with the already-loaded `TrusteeListItem[]` so no additional API call is needed for the list data.

The `filterTrustees` function was intentionally left unchanged — name filtering happens as a post-filter step in the `useMemo`, keeping the existing function pure and its tests unaffected.

# Definition of Done:

## Summary by Sourcery

Add a trustee name text filter to the Trustees list and integrate it with existing filtering, accessibility, and telemetry.

New Features:
- Introduce a trustee name text input filter within the Trustees list filters accordion that narrows results using the existing trustee search API and composes with district and chapter filters.

Enhancements:
- Announce filtered trustee counts via the existing ARIA live region when the name filter is used and track AppInsights events for name filter changes and clears.

Tests:
- Extend TrusteesList and TrusteeDistrictFilter tests to cover name filtering behavior, debounce and minimum length rules, clear button behavior, ARIA announcements, and AppInsights telemetry.

### DIFF
--- a/user-interface/src/trustees/TrusteesList.tsx
+++ b/user-interface/src/trustees/TrusteesList.tsx
@@ -2,6 +2,7 @@ import './TrusteesList.scss';
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { NavLink } from 'react-router-dom';
 import { TrusteeListItem } from '@common/cams/trustees';
+import useDebounce from '@/lib/hooks/UseDebounce';
 import {
   formatChapterType,
   formatAppointmentType,
@@ -63,8 +64,13 @@ export default function TrusteesList() {
   const [sortDirection, setSortDirection] = useState<'asc' | 'desc'>('asc');
   const [selectedChapters, setSelectedChapters] = useState<ComboOption[]>([]);
   const [liveAnnouncement, setLiveAnnouncement] = useState<string>('');
+  const [nameSearch, setNameSearch] = useState('');
+  const [nameSearchIds, setNameSearchIds] = useState<Set<string> | null>(null);
   const filterRef = useRef<TrusteeDistrictFilterRef>(null);
   const pageLoadStart = useRef(performance.now());
+  const isNameFilterInteracted = useRef(false);
+  const previousNameSearchRef = useRef('');
+  const debounce = useDebounce();
 
   useEffect(() => {
     const fetchTrustees = () => {
@@ -110,8 +116,30 @@ export default function TrusteesList() {
     setLiveAnnouncement('');
   };
 
+  const handleFilterName = (name: string) => {
+    isNameFilterInteracted.current = true;
+    setNameSearch(name);
+    if (name.length < 2) {
+      setNameSearchIds(null);
+      return;
+    }
+    debounce(async () => {
+      try {
+        const response = await Api2.searchTrustees(name);
+        const ids = new Set(response.data.map((r) => r.trusteeId));
+        setNameSearchIds(ids);
+      } catch {
+        setNameSearchIds(new Set());
+      }
+    }, 300);
+  };
+
   const { filteredTrustees } = useMemo(() => {
-    const filtered = filterTrustees(trustees, selectedDistricts, selectedChapters);
+    let filtered = filterTrustees(trustees, selectedDistricts, selectedChapters);
+
+    if (nameSearchIds !== null) {
+      filtered = filtered.filter((t) => nameSearchIds.has(t.trusteeId));
+    }
 
     const sorted = [...filtered].sort((a, b) => {
       const lastCmp = (a.lastName ?? '').localeCompare(b.lastName ?? '', undefined, {
@@ -135,7 +163,7 @@ export default function TrusteesList() {
     return {
       filteredTrustees: sortedWithAppointments,
     };
-  }, [trustees, selectedDistricts, selectedChapters, sortDirection]);
+  }, [trustees, selectedDistricts, selectedChapters, nameSearchIds, sortDirection]);
 
   useEffect(() => {
     if (!isDefaultApplied.current) return;
@@ -158,7 +186,7 @@ export default function TrusteesList() {
   }, [selectedDistricts, selectedChapters, trustees]);
 
   useEffect(() => {
-    if (!isChapterFilterInteracted.current) return;
+    if (!isChapterFilterInteracted.current && !isNameFilterInteracted.current) return;
     setLiveAnnouncement(`${filteredTrustees.length} Trustees`);
   }, [filteredTrustees]);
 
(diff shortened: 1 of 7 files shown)


user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -63,8 +64,16 @@ export default function TrusteesList() {
   const [sortDirection, setSortDirection] = useState<'asc' | 'desc'>('asc');
   const [selectedChapters, setSelectedChapters] = useState<ComboOption[]>([]);
   const [liveAnnouncement, setLiveAnnouncement] = useState<string>('');
+  const [nameSearch, setNameSearch] = useState('');
```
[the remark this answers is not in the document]
jalba-flexion: Thanks for the detailed suggestion. The consolidation is a reasonable long-term direction, but the scope exceeds this PR. Deferring to a follow-up refactor task.

btposey (approved):

cruzjone-flexion (approved):

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -2,6 +2,7 @@ import './TrusteesList.scss';
 import { useEffect, useMemo, useRef, useState } from 'react';
```
mattstankey, line 2: test

mattstankey (commented):
I found three bugs worth addressing before merge.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -110,8 +119,39 @@ export default function TrusteesList() {
...
+    setNameSearch(name);
+  };
+
+  useEffect(() => {
+    if (nameSearch.length < 2) {
+      nameSearchQueryLengthRef.current = 0;
+      setNameSearchIds(new Set());
+      return;
+    }
+    nameSearchCountRef.current += 1;
```
mattstankey: **Bug: `sessionSearchCount` overcounts**

This counter increments on every keystroke that reaches the effect (every change where `nameSearch.length >= 2`), but the debounce means only one API call fires per 300ms window. Typing "Smith" increments the counter 4 times but fires 1 API call, so `sessionSearchCount` in the telemetry will be inflated and misleading.

Move the increment inside the debounce callback, after the API call succeeds:

```ts
debounce(async () => {
  try {
    const response = await Api2.searchTrustees(nameSearch);
    nameSearchCountRef.current += 1;  // ← here, once per actual search
    ...
  } catch { ... }
}, 300);
```

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -110,8 +119,39 @@ export default function TrusteesList() {
...
+    const searchStart = performance.now();
+    debounce(async () => {
+      try {
+        const response = await Api2.searchTrustees(nameSearch);
+        const ids = new Set(response.data.map((r) => r.trusteeId));
+        nameSearchStartRef.current = performance.now() - searchStart;
+        setNameSearchIds(ids);
+      } catch {
+        nameSearchStartRef.current = null;
+        setNameSearchIds(new Set());
```
mattstankey, line 145: **Bug: API failure silently shows 0 results**

On a search API error, `nameSearchIds` is set to an empty `Set` while `nameSearch.length >= 2` remains true. The `useMemo` then filters out every trustee from the already-loaded list, showing users "0 Trustee(s)" with no explanation.

The safest fix is to reset `nameSearch` back to empty on error, which also clears the input and returns the full list:

```ts
} catch {
  setNameSearch('');
  setNameSearchIds(new Set());
}
```

Alternatively, surface the error via `useGlobalAlert` so the user knows the search failed.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -110,8 +119,39 @@ export default function TrusteesList() {
...
+
+  useEffect(() => {
+    if (nameSearch.length < 2) {
+      nameSearchQueryLengthRef.current = 0;
+      setNameSearchIds(new Set());
+      return;
+    }
+    nameSearchCountRef.current += 1;
+    nameSearchQueryLengthRef.current = nameSearch.length;
+    const searchStart = performance.now();
```
mattstankey: **Bug: `searchResponseMs` measurement includes the debounce delay**

`searchStart` is captured here, before the 300ms debounce fires. By the time `performance.now() - searchStart` is computed on line 140, the measurement includes the ~300ms wait. Every recorded value will be inflated by the debounce delay, making the metric unreliable as a proxy for API latency.

Move `searchStart` inside the debounce callback so it captures the moment the actual HTTP request begins:

```ts
debounce(async () => {
  const searchStart = performance.now();  // ← inside, not outside
  try {
    const response = await Api2.searchTrustees(nameSearch);
    nameSearchStartRef.current = performance.now() - searchStart;
    ...
  }
}, 300);
```

jalba-flexion:
Addressed the three bugs flagged by @mattstankey:
                                                                                                                                                 
  - sessionSearchCount overcounting — moved nameSearchCountRef.current += 1 inside the debounce callback so it increments once per completed API
  call, not per keystroke.                                                                                                                       
  - API failure showing 0 results — added setNameSearch('') in the catch block so a failed search clears the input and restores the full trustee
  list instead of silently filtering everything out.                                                                                             
  - searchResponseMs including debounce delay — moved searchStart = performance.now() inside the debounce callback so timing starts when the HTTP
   request actually begins, not 300ms earlier.                                                                                                   
                                                                       
  A regression test was added for each scenario.

cruzjone-flexion (approved):

PR #2337, opened 2026-05-01
CAMS-596 Fix zoom import deduplication and metrics accuracy

btposey:
# Purpose

Fix discrepancies with the Zoom info import report

# Major Changes

Resolves two critical bugs in the zoom import dataflow:

1. Deduplication bug: Multiple ATS TRU_IDs that map to the same CAMS trustee were incorrectly marked as ambiguous. Now tracks trustees by CAMS trusteeId to properly handle deduplicated records.

2. Metrics accuracy: In-memory counters were diverging from actual report file outcomes. Now counts outcomes directly from the written report lines, making the report the authoritative source of truth.

Impact: Enables 72 records with multiple TRU_IDs to match correctly and ensures App Insights metrics accurately reflect import results.

## Summary by Sourcery

Fix Zoom CSV import to correctly deduplicate trustees and align metrics with the generated report.

Bug Fixes:
- Prevent multiple ATS TRU_IDs mapping to the same CAMS trustee from being incorrectly treated as ambiguous by deduplicating on CAMS trusteeId.
- Ensure Zoom import metrics reflect the actual outcomes written to the report by deriving counts from the report lines instead of in-memory counters.

### DIFF
--- a/backend/lib/use-cases/dataflows/import-zoom-csv.test.ts
+++ b/backend/lib/use-cases/dataflows/import-zoom-csv.test.ts
@@ -271,72 +271,6 @@ describe('import-zoom-csv', () => {
       );
     });
 
-    test('should write unmatched report when unmatched records exist', async () => {
-      vi.mocked(mockObjectStorage.readObject).mockResolvedValue(SAMPLE_MATCHED_TSV);
-      vi.spyOn(MockMongoRepository.prototype, 'findTrusteeByLegacyTruId')
-        .mockResolvedValueOnce(MOCK_TRUSTEE)
-        .mockResolvedValueOnce(null);
-      vi.spyOn(MockMongoRepository.prototype, 'findTrusteesByName').mockResolvedValue([]);
-      vi.spyOn(MockMongoRepository.prototype, 'searchTrusteesByName').mockResolvedValue([]);
-      vi.spyOn(MockMongoRepository.prototype, 'searchTrusteesByPhoneticTokens').mockResolvedValue(
-        [],
-      );
-      vi.spyOn(MockMongoRepository.prototype, 'updateTrustee').mockResolvedValue(MOCK_TRUSTEE);
-
-      await importZoomCsv(context);
-
-      expect(mockObjectStorage.writeObject).toHaveBeenCalledWith(
-        expect.any(String),
-        'zoom-import-unmatched-report.tsv',
-        expect.stringContaining('Zoom Name\tZoom Email\tMeeting ID'),
-      );
-      expect(mockObjectStorage.writeObject).toHaveBeenCalledWith(
-        expect.any(String),
-        'zoom-import-unmatched-report.tsv',
-        expect.stringContaining('Jane Smith'),
-      );
-    });
-
-    test('should use computed outcome in unmatched report, not original TSV outcome', async () => {
-      vi.mocked(mockObjectStorage.readObject).mockResolvedValue(SAMPLE_MATCHED_TSV);
-      vi.spyOn(MockMongoRepository.prototype, 'findTrusteeByLegacyTruId').mockResolvedValue(null);
-      vi.spyOn(MockMongoRepository.prototype, 'findTrusteesByName').mockResolvedValue([]);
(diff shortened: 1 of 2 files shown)


backend/lib/use-cases/dataflows/import-zoom-csv.ts
```diff
@@ -308,6 +308,48 @@ function buildUnmatchedReportRow(row: ZoomMatchedRow, outcome: ProcessResult['ou
   ].join('\t');
 }
 
+/**
+ * Counts outcomes from report lines to ensure metrics match the actual written report.
+ * The outcome column is at index 11 in the main report format.
+ * Format: zoomName | zoomEmail | atsTruIds | matchedNames | matchCount | similarity |
+ *         activeStatus | statusCodes | strategy | matchedTrusteeId | matchedTrusteeName | outcome | error
+ */
+function countOutcomesFromReportLines(reportLines: string[]): ZoomImportResult {
```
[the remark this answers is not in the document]
btposey: Implemented all suggestions from this review in commit 127dc6081:

✅ **Restored in-loop outcome counting** as primary metrics source
- Removed double-parsing overhead
- Added `outcomeToKey` mapping for direct counter updates

✅ **Derived outcome column index from header**
- Changed from hardcoded index 11 to `headers.indexOf('outcome')`
- Throws error if outcome column not found

✅ **Converted countOutcomesFromReportLines to sanity check**
- Now validates in-loop metrics against report
- Logs warning on mismatch for debugging

✅ **Reduced logging noise**
- Downgraded deduplication logs from `info` to `debug`

✅ **Added malformed line warnings**
- Reports line number and expected vs actual column count

This maintains the "metrics must match report" guarantee while improving performance and maintainability.

jalba-flexion (approved):

cruzjone-flexion (approved):

backend/lib/use-cases/dataflows/import-zoom-csv.ts
```diff
@@ -333,18 +310,31 @@ export async function processZoomMatchedRow(
...
     const notFoundIds: string[] = [];
+    const deduplicatedTruIds: string[] = []; // Track which TRU_IDs mapped to existing trustee
 
     for (const atsTruId of atsTruIds) {
       const trustee = await repo.findTrusteeByLegacyTruId(atsTruId);
       if (trustee) {
-        camsTrustees.push(trustee);
+        if (camsTrusteesMap.has(trustee.trusteeId)) {
+          // This TRU_ID maps to a trustee we've already found - expected for deduplicated records
+          deduplicatedTruIds.push(atsTruId);
```
mattstankey: Is this used anywhere?  It appears to be unused.

backend/lib/use-cases/dataflows/import-zoom-csv.test.ts
```diff
@@ -271,70 +271,144 @@ describe('import-zoom-csv', () => {
...
+      expect(warnSpy).not.toHaveBeenCalledWith(
         expect.any(String),
-        'zoom-import-unmatched-report.tsv',
-        expect.stringContaining('Jane Smith'),
+        expect.stringContaining('Mismatch between in-loop metrics'),
       );
     });
 
-    test('should use computed outcome in unmatched report, not original TSV outcome', async () => {
+    test('should throw error if outcome column not found in generated report', async () => {
```
mattstankey: It seems like this is testing to see if a report exists at all, rather than testing for an error.  Is this test needed?  If so, it would make more sense to rename the test.

backend/lib/use-cases/dataflows/import-zoom-csv.ts
btposey: ✅ Fixed! Removed the unused `deduplicatedTruIds` array. It was being populated but never read.

The debug log message for deduplicated records is still present, just without tracking them in an array.

backend/lib/use-cases/dataflows/import-zoom-csv.test.ts
btposey: ✅ Fixed! Renamed the test to `"should validate generated report has outcome column"` to accurately reflect what it actually tests.

You're right - the test doesn't actually test an error scenario. It validates that a valid report succeeds. The test name was misleading.

PR #2338, opened 2026-05-01
CAMS-739 closed cases visibility

mattstankey:
# Purpose

Added visibility to closed cases on Case Search results screen.

As a USTP staff member, in order to avoid missing cases that may be closed, I want clear notification when closed cases match my search criteria.

Screenshot 2026-05-01 at 4 07 50 PM

# Definition of Done:

## Summary by Sourcery

Improve search results visibility for closed cases and support displaying open/closed status in case search results.

New Features:
- Show contextual alerts and hints when closed cases may match search filters but are excluded from current results, including count-specific messaging for case-number searches.
- Allow users to quickly include closed cases via an inline control in alerts that updates the search predicate and reruns the search.
- Add an optional Open/Closed column in the case search results table to indicate each case’s status.

Enhancements:
- Expose a closedCasesCount field in pagination responses when searching by case number with closed cases excluded, and surface it in the UI to drive closed-case messaging.
- Update search validators to accept Chapter 9 and adjust chapter validation messaging accordingly.
- Simplify alert styling by relying on the standard slim variant and removing the custom compact-with-title class.
- Refine mock search API behavior to better simulate closed/open cases and filter behavior for chapters and debtor name.

Tests:
- Extend backend controller tests to cover closedCasesCount population and edge cases.
- Add comprehensive UI tests for new closed-case alerts, Include Closed Cases interactions, Open/Closed column rendering, and search form ordering.
- Update Alert component tests to reflect the new slim alert behavior.


user-interface/src/search-results/SearchResults.tsx
```diff
@@ -32,37 +32,73 @@ export type SearchResultsHeaderProps = {
   labels: string[];
   phoneticSearchEnabled?: boolean;
   showDebtorNameColumn?: boolean;
+  showOpenClosedColumn?: boolean;
 };
 
 export type SearchResultsRowProps = TableRowProps & {
   idx: number;
-  bCase: CaseSummary;
+  bCase: DxtrCase;
```
btposey: Why did we use `DxtrCase`? Search results all come from Cosmos and we only store `CaseSummary` in Cosmos. `DxtrCase` for one, screams a dependency model that belongs in a gateway, not in the CAMS domain, and two if additional attributes are needed then they should be sync'ed into Cosmos and added to `CaseSummary` if included in search results. Needs some discussion here if the problem is larger or more complicated than this.

user-interface/src/index.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 import { BrowserRouter } from 'react-router-dom';
 import './index.css';
+import './App.scss';
 import App from './App';
```
btposey, line 6: Why was App.scss moved from App.tsx to the index? Was something added to App.scss that should have been added to index.css?

common/src/api/pagination.ts
```diff
@@ -13,4 +13,5 @@ export type Pagination = {
   currentPage: number;
   totalPages?: number;
   totalCount?: number;
+  closedCasesCount?: number;
```
btposey: There was a purity to the pagination model that this seems to polute. Pagination is not generic anymore as it contains a count specifically for the case domain.

btposey (commented):
I have questions....

user-interface/src/index.tsx
fmaddenflx: It has to do with the order in which the build process in the new vite adds the css files.  Apparently we can't control the order.  Our css was then being overridden by USWDS but USWDS should always go in first.  This seemed to be the way to gaurantee USWDS makes it into the output before anything else.  I didn't like moving it to index.tsx but claude recommended it and after some thought, it made sense.

common/src/api/pagination.ts
fmaddenflx: Oh yeah!  This isn't good!

user-interface/src/search-results/SearchResults.tsx
mattstankey: Good catch, we shouldn't be using `DxtrCase`.  SearchResultsRowProps.bCase should stay typed as `SyncedCase` (the actual domain type used throughout the stack), not `DxtrCase`. The original change to 
  `DxtrCase` was unnecessarily reaching down the type hierarchy when the actual data is `SyncedCase`. CaseSummary was too narrow (no closedDate), but `DxtrCase` is also wrong — SyncedCase is the right type.

common/src/api/pagination.ts
mattstankey: We're going to fix this by extending `Pagination` in the cases domain — Define a `CasesPagination` = Pagination & { closedCasesCount?: number } in the cases domain (e.g., common/src/cams/) and use it only in the cases controller and search response. `Pagination` stays pure.

btposey (approved):
Pending agreed upon changes.

PR #2339, opened 2026-05-01
CAMS-740 Expand ATS districts to DXTR divisions

btposey:
# Purpose

Enhanced trustee migration to expand single ATS district appointments into multiple CAMS appointments (one per DXTR division). Added district-to-divisions mapping that leverages existing DXTR offices data through use-case orchestration.

# Major Changes

- Added optional courtName and courtDivisionName fields to TrusteeAppointmentInput for division metadata enrichment
- Created expandAppointmentToDivisions in migrate-trustees use-case to handle 1:N district-to-divisions expansion
- Enhanced getTrusteeAppointments to accept districtToDivisionsMap and perform expansion when provided
- Updated deduplication key to include divisionCode (prevents duplicate appointments for same trustee/court/division/chapter)
- Moved getAppointmentKey from ATS adapter to use-case layer to support divisionCode in key generation
- Maintained strict domain separation: ATS adapter handles ATS cleansing only, use-case layer orchestrates cross-domain logic

This approach respects domain boundaries while enabling rich division-level appointment tracking in CAMS.

## Summary by Sourcery

Expand trustee migration to support DXTR division-level appointments and metadata enrichment while preserving domain boundaries between ATS cleansing and use-case orchestration.

New Features:
- Allow trustee appointments to carry optional courtName and courtDivisionName metadata for division-level enrichment.
- Enable expansion of single district-level ATS appointments into multiple division-specific CAMS appointments using DXTR office mappings.
- Introduce a district-to-divisions mapper that derives division metadata from DXTR office data for use in trustee migrations.

Enhancements:
- Move appointment deduplication key generation into the migrate-trustees use case and include divisionCode to prevent duplicate division-level appointments.
- Extend trustee migration flows to accept an optional district-to-divisions map and apply expansion consistently across fetching, aggregation, and processing of appointments.
- Enrich ATS cleansing transforms to optionally attach divisionCode, courtName, and courtDivisionName when division information is available.

Tests:
- Add unit and integration tests covering division enrichment in ATS cleansing transforms, trustee appointment validation, and district-to-division expansion in the migrate-trustees flow.


backend/lib/use-cases/dataflows/district-division-mapper.ts
```diff
@@ -0,0 +1,46 @@
+import { UstpOfficeDetails } from '@common/cams/offices';
```
btposey: This module is not represent a stand alone use case. This file and it's tests should be moved inside the migrate-trustees module and it's test suite.

backend/lib/adapters/gateways/ats/cleansing/ats-cleansing-transform.ts
```diff
@@ -57,12 +68,21 @@ export function transformAppointmentRecord(
...
   };
+
+  // Add division enrichment if provided
+  if (divisionInfo) {
+    appointment.divisionCode = divisionInfo.divisionCode;
+    appointment.courtName = divisionInfo.courtName;
+    appointment.courtDivisionName = divisionInfo.courtDivisionName;
+  }
+
+  return appointment;
```
btposey: There should be no division enrichment in the ATS domain.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -273,19 +290,73 @@ export async function getPageOfTrustees(
...
 export async function getTrusteeAppointments(
   context: ApplicationContext,
   trusteeId: number,
+  districtToDivisionsMap?: Map<string, DivisionInfo[]>,
 ): Promise<MaybeData<TrusteeAppointmentsResult>> {
   try {
     const atsGateway = factory.getAtsGateway(context);
     const result = await atsGateway.getTrusteeAppointments(context, trusteeId);
 
+    // If district-to-divisions map provided, expand district appointments into divisions
```
btposey: Is there a question of "if"? There should always be a district to division mapping from DXTR.

backend/lib/use-cases/dataflows/district-division-mapper.ts
btposey: ✅ Fixed! Moved `district-division-mapper.ts` and its tests into `migrate-trustees.ts`. The mapper function `buildDistrictToDivisionsMap` and `DivisionInfo` type are now part of the migrate-trustees module.

backend/lib/adapters/gateways/ats/cleansing/ats-cleansing-transform.ts
btposey: ✅ Fixed! Removed the `DivisionEnrichment` type and optional `divisionInfo` parameter from `transformAppointmentRecord`. Division enrichment is now only performed in the use-case layer via `expandAppointmentToDivisions`, keeping the ATS domain focused on cleansing only.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: ✅ Fixed! Made `districtToDivisionsMap` a required parameter (removed the `?` optional marker). The use case now fetches office data from DXTR via `factory.getOfficesGateway()` and builds the map at the start of `processPageOfTrustees`, ensuring the mapping is always available.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -10,13 +10,30 @@ import factory from '../../factory';
...
 const MODULE_NAME = 'MIGRATE-TRUSTEES-USE-CASE';
 
+/**
+ * Division information with metadata
+ */
+type DivisionInfo = {
+  divisionCode: string;
+  courtName: string;
+  courtDivisionName: string;
+};
```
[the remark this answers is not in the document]
btposey: ✅ Addressed! There is now only ONE `DivisionInfo` type defined in `migrate-trustees.ts` that includes `courtId`. The duplicate type definitions have been eliminated by merging the district-division-mapper into the migrate-trustees module.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -273,19 +290,73 @@ export async function getPageOfTrustees(
...
+  appointment: TrusteeAppointmentInput,
+  divisions: DivisionInfo[],
+): TrusteeAppointmentInput[] {
+  return divisions.map((division) => ({
+    ...appointment,
+    divisionCode: division.divisionCode,
+    courtName: division.courtName,
+    courtDivisionName: division.courtDivisionName,
+  }));
+}
```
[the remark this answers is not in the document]
btposey: Not applicable. The source appointments from ATS never have division data - enrichment always comes fresh from DXTR via the mapper. Nullish coalescing would be pointless since the source is always unenriched.

backend/lib/use-cases/dataflows/district-division-mapper.ts
```diff
@@ -0,0 +1,46 @@
...
+export function buildDistrictToDivisionsMap(
+  offices: UstpOfficeDetails[],
+): Map<string, DivisionInfo[]> {
+  const districtMap = new Map<string, DivisionInfo[]>();
+
+  for (const office of offices) {
+    for (const group of office.groups) {
+      for (const division of group.divisions) {
+        const courtId = division.court.courtId;
+        const divisionInfo: DivisionInfo = {
```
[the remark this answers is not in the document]
btposey: Not needed. The DXTR office data is an acyclic tree structure that guarantees no duplicate divisions per courtId. Deduplication would be defensive programming against a data integrity issue that cannot occur.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -10,13 +10,30 @@ import factory from '../../factory';
...
 import { Trustee } from '@common/cams/trustees';
 import { TrusteeAppointmentInput } from '@common/cams/trustee-appointments';
-import { getAppointmentKey } from '../../adapters/gateways/ats/cleansing/ats-cleansing-utils';
 import { CamsUserReference } from '@common/cams/users';
 import { LegacyAddress } from '@common/cams/parties';
 import { normalizeName } from './trustee-match.helpers';
-
 const MODULE_NAME = 'MIGRATE-TRUSTEES-USE-CASE';
 
+/**
```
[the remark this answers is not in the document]
btposey: ✅ Partially addressed! Consolidated to one `DivisionInfo` type. Expansion is centralized in `expandAppointmentToDivisions`. The use case fetches office data via `factory.getOfficesGateway()` at the boundary (`processPageOfTrustees`), keeping the map instantiation at the appropriate architectural layer per CAMS dependency rules.

backend/lib/use-cases/dataflows/district-division-mapper.ts
```diff
@@ -0,0 +1,46 @@
+import { UstpOfficeDetails } from '@common/cams/offices';
+
+/**
+ * Division information extracted from DXTR offices data
+ */
+export type DivisionInfo = {
```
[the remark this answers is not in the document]
btposey: ✅ Resolved! Merged everything into the migrate-trustees module. There is now only one `DivisionInfo` type, eliminating the need for a separate shared type file.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -755,6 +870,13 @@ export async function processPageOfTrustees(
     failedAppointments: FailedAppointment[];
   }>
 > {
+  // Fetch office data from DXTR via factory
+  const officesGateway = factory.getOfficesGateway(context);
+  const offices = await officesGateway.getOffices(context);
```
bityogi: `getOffices()` is called at the start of every `processPageOfTrustees` invocation. During a full migration with many pages, this results in repeated DXTR calls for office data that doesn't change between pages. Consider building the districtToDivisionsMap once at a higher level (e.g., the migration orchestrator/scheduler) and passing it in, or caching the result so it's only fetched once per migration run.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -371,19 +371,15 @@ async function handleRetry(event: TrusteeEvent, invocationContext: InvocationCon
     return;
   }
 
-  // Wrap single trustee into MergedTrusteeData structure for processing
-  const mergedData = MigrateTrusteesUseCase.mergeTrusteeRecords([event]);
-  const result = await MigrateTrusteesUseCase.processTrusteeWithAppointments(context, mergedData);
+  // Requeue to PAGE queue for normal processing
+  // PAGE handler will fetch this trustee and process it through the standard flow
+  const cursorMessage: CursorMessage = { lastId: (event.ID - 1).toString() };
```
bityogi: This changes retry behavior from directly processing the single failed trustee to requeueing a cursor message. The PAGE handler will then call `getPageOfTrustees` from ATS starting at `event.ID - 1`, potentially fetching a full page of trustees — not just the one that failed.                                                                                 
                  
  For each trustee on that page, the flow hits:
  - **ATS** — appointments fetch
  - **DXTR** — offices fetch
  - **ACMS** — professional ID lookup
  - **Cosmos** — trustee and appointment upserts

Already-completed trustees would get reprocessed through all of those external systems. Is this the intended behavior, or should the retry still target just the single failed trustee?

bityogi (changes requested):
Several error-handling test scenarios were removed along with the `processTrusteeWithAppointments` and `getTrusteeAppointments` test suites and don't appear to have replacements at the `processPageOfTrustees` level:

  - **Trustee upsert failure** — previously verified that a database error during `upsertTrustee` was handled gracefully
  - **Appointment fetch failure** — verified that an ATS timeout during `getTrusteeAppointments` didn't prevent the trustee
   from being saved
  - **Appointment creation failure** — verified that a database error during `createAppointments` didn't mark the whole
  trustee as failed
  - **Trustee with no appointments** — verified the zero-appointment path

The new `processPageOfTrustees` tests cover the happy path and the division expansion logic, but the error resilience behavior is no longer explicitly tested. Could we add error-scenario tests at the `processPageOfTrustees` level to maintain confidence in fault tolerance?

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Yes you have a pragmatic point about efficiency, but that would be an OEA violation. The use case is depending on less important code (the Azure Function infrastructure) to reliably provide it with a gateway that it itself should have authority over getting from the factory. Furthermore orchestration is not a well defined thing here as that would be the Azure Function infrastructure polling the page queue. That's less important code too that would go away if we changed cloud platforms, blah, blah blah. 😄

backend/function-apps/dataflows/migrations/migrate-trustees.ts
btposey: Will reconsider.

mattstankey (commented):
Three issues worth discussing before merge.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
mattstankey: **Issue: retry requeue may cause duplicate processing and bypasses the retry limit.**

Using `{ lastId: event.ID - 1 }` queues a full PAGE starting at that trustee, which will re-process any trustees that come after it in that page — not just the one that failed. The original retry processed only the single failed trustee, so this is a behavioral change with broader blast radius.

Also, the RETRY_LIMIT check (lines 370-374) guards only this invocation of `handleRetry`. Because a `CursorMessage` is placed on PAGE rather than a `TrusteeEvent` with a `retryCount`, the individual trustee that triggered the retry will never reach the DLQ via this path if it keeps failing — `retryCount` is never incremented on the new PAGE-sourced event.

Consider whether this is the intended retry semantics, and if so, documenting why duplicate re-processing of already-succeeded trustees in the same page is acceptable.

backend/lib/use-cases/dataflows/migrate-trustees.ts
mattstankey: **Issue: unguarded `getOffices` call will fail the entire page if the DXTR offices gateway throws.**

```ts
const offices = await officesGateway.getOffices(context);
```

If this call throws, `processPageOfTrustees` propagates the error and no trustees in the page are processed — no appointments are written and no failures are logged to DLQ. The existing pattern in `fetchAndAggregateAppointments` logs individual errors and continues processing other records.

A try/catch that falls back to an empty map would match that pattern:
```ts
let offices: UstpOfficeDetails[] = [];
try {
  offices = await officesGateway.getOffices(context);
} catch (e) {
  context.logger.error(MODULE_NAME, "Failed to load DXTR offices — appointments will not be expanded to divisions", e);
}
```
This degrades gracefully (no expansion) rather than failing the whole page.

backend/lib/use-cases/dataflows/migrate-trustees.test.ts
```diff
@@ -506,82 +461,26 @@ describe('Migrate Trustees Use Case', () => {
...
   });
 
-  describe('processTrusteeWithAppointments', () => {
-    beforeEach(() => {
-      // Stub ACMS gateway so upsertProfessionalIds does not reach the network
-      vi.spyOn(factory, 'getAcmsGateway').mockReturnValue({
-        getTrusteeProfessionalIds: vi.fn().mockResolvedValue([]),
-      } as unknown as AcmsGateway);
-    });
+  describe('processPageOfTrustees', () => {
```
mattstankey, line 954: **Issue: error-path test coverage was removed and not replaced.**

The previous `processTrusteeWithAppointments` tests covered three important failure modes:
1. Upsert failure → trustee processing fails
2. Appointment fetch failure → trustee still succeeds with 0 appointments
3. Appointment creation failure → trustee still succeeds with 0 appointments

These tests are gone and the new `processPageOfTrustees` tests only cover the happy path and division expansion. The error-handling logic in `processTrusteeWithAppointments` is still present but no longer exercised. If that logic changes, there is nothing to catch the regression. Consider adding error-path cases to the `processPageOfTrustees` describe block.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
btposey: ✅ Fixed in a086e5668! Removed all retry logic from the Azure Function entirely. Retry is now only handled in the use case via `processTrusteeWithRetry()` helper with exponential backoff (MAX_RETRY_ATTEMPTS = 2).

The retry requeue concern is no longer applicable - there is no handleRetry function, no RETRY queue, and no cursor-based requeue logic. Individual trustee failures are retried in-place within the use case, then logged as errors if they still fail after retries.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: ✅ Fixed in 037b4d56b! Wrapped `officesGateway.getOffices()` in try-catch that returns a `MaybeData` error instead of throwing. This ensures DXTR failures flow properly to DLQ for visibility rather than silently failing the entire page.

Added test `should return error when getOffices fails and not process any trustees` verifying the error-path behavior.

backend/lib/use-cases/dataflows/migrate-trustees.test.ts
btposey: ✅ Fixed in a086e5668! All error-path scenarios now have test coverage at the `processPageOfTrustees` level:

- Malformed trustee record (line 494)
- Trustee upsert failure (line 510) 
- Appointment creation failure (line 531)
- Appointment fetch failure (line 585)
- Professional ID lookup failure (line 611)
- Trustee with no appointments (line 650)
- Mixed success/failures (line 684)
- Failed appointments propagation (line 729)

Additionally, in 037b4d56b added test for `getOffices` failure scenario (line 761).

backend/function-apps/dataflows/migrations/migrate-trustees.ts
btposey: I had Claude refactor the error handling and retry logic in the use case. Retry logic is fully handled in the use case now and not thrown and rehandled by the controller. I also removed nested try catch logic and added tests coverage for the refactored use case.

bityogi (approved):

PR #2340, opened 2026-05-02
Add CamsTable component and migrate search/assignment tables

fmaddenflx:
# Purpose

Replace the USWDS HTML table components on the Search Results, My Cases, and Staff Assignment screens with a new div-based responsive table library (CamsTable) that follows the established Trustees screen pattern and USWDS borderless table styling.

# Major Changes

- **New `CamsTable` component library** (`lib/components/cams/CamsTable/`) — `CamsTable`, `CamsTableHeader`, `CamsTableHeaderCell`, `CamsTableBody`, `CamsTableRow`, `CamsTableCell` with built-in responsive CSS via `cams-table--responsive` class
- **Search Results, My Cases, Staff Assignment** — migrated from USWDS `Table`/`TableRow`/`TableRowData` to `CamsTable` components with mobile-stacked layout and USWDS borderless table styling
- **Responsive breakpoints** — tables stay stacked on mobile and at tablet width (when side filter panel is present); flex-row layout activates at 1025px on Search and Staff Assignment
- **Closed-cases hint alerts moved to SearchScreen** — removed screen-specific hint logic from `SearchResults`; `SearchScreen` now owns and renders all closed-cases hints including the Slice 3 count-specific variant
- **`ClosedCasesHintMessage`** exported from `SearchResults` for reuse by `SearchScreen`
- **`onResultsChanged` callback** extended to pass `closedCasesCount` so `SearchScreen` can render the count-specific hint without needing internal `SearchResults` state
- **Alert styling fixes** — removed `compact-with-title` custom alert variant; alerts with titles now render as standard USWDS full-size alerts; heading font size corrected to 1.375rem (USWDS scale step 9); `App.scss` import order fixed so component overrides take precedence over USWDS compiled styles

# Testing/Validation

- All 200 unit test files pass
- Updated unit tests to use CamsTable CSS selectors instead of HTML table element selectors
- Updated accessibility test selectors for Search and My Cases screens
- Manual testing across Search, My Cases, and Staff Assignment at mobile, tablet, and desktop breakpoints

# Notes

- The `CamsTable.scss` emits a `.cams-table--responsive` class automatically applied by `CamsTable.tsx` — consumers only need column flex values in their own SCSS
- Column proportions are screen-specific and defined in each screen's SCSS
- The `CAMS-739` alert styling fixes (compact-with-title removal, heading font size, App.scss load order) are included in this branch as they were developed alongside the table work

# Definition of Done:

## Summary by Sourcery

Introduce a reusable responsive CamsTable component and migrate search-related and staff assignment tables to use it while centralizing closed-case hint logic in SearchScreen.

New Features:
- Add CamsTable component library providing div-based, responsive table structure with accessible roles and captions.
- Expose ClosedCasesHintMessage for reuse and support a count-based closed-case hint variant driven by API pagination metadata.

Bug Fixes:
- Move closed-case hint alerts from SearchResults into SearchScreen and drive them via new results state (hasResults, closedCasesCount, isSearching) to avoid stale internal state.
- Fix alert styling by relying on standard USWDS alerts, correcting heading font size, and ensuring App.scss overrides load after USWDS styles.
- Align toolbar alignment and button positioning within search results rows to improve layout consistency.

Enhancements:
- Refactor SearchScreen layout to use a flexible search pane with separate filter and results panels and breakpoint-specific behavior.
- Update SearchResults, SearchResultsHeader, SearchResultsRow, and StaffAssignmentRow/Header to render via CamsTable instead of USWDS Table components, with screen-specific column layouts.
- Adjust responsive SCSS for Search, My Cases, and Staff Assignment screens to control stacked vs flex-row table layouts and column proportions.
- Tighten USWDS Table typings by making TableRowProps and TableProps internal to the module.

Tests:
- Update unit and accessibility tests for Search, My Cases, SearchResults, and StaffAssignmentRow to target CamsTable selectors and cover new closed-case hint behaviors, including count-specific variants.


fmaddenflx:
Addressed the Sourcery complexity suggestion: extracted `showNoResultsOpenOnlyHint`, `showGenericClosedHint`, and `showCountClosedHint` as named booleans, then collapsed the three duplicated Alert/ClosedCasesHintMessage blocks into two — preserving all existing IDs, roles, titles, and variants. Commit: a9bb8e900.

fmaddenflx:
@sourcery-ai

user-interface/src/lib/components/uswds/Table.tsx
```diff
@@ -53,7 +53,7 @@ export function TableRowData(props: TableRowDataProps) {
   );
 }
 
-export type TableRowProps = PropsWithChildren & JSX.IntrinsicElements['tr'];
+type TableRowProps = PropsWithChildren & JSX.IntrinsicElements['tr'];
```
fmaddenflx, line 56: This was just because Table isn't being used anywhere and Knip was complaining.  I don't want to get rid of it though because I'd like to have a set up uswds components that all of Flexion can use.  Perhaps I should put them in their own flexion wide repo?

user-interface/src/search/SearchScreen.test.tsx
```diff
@@ -321,10 +321,10 @@ describe('search screen', () => {
     await userEvent.click(searchButton);
 
     await waitFor(() => {
-      expect(document.querySelector('.search-results table')).toBeVisible();
+      expect(document.querySelector('.search-results .cams-table')).toBeVisible();
     });
 
-    const rows = document.querySelectorAll('#search-results-table-body > tr');
+    const rows = document.querySelectorAll('.search-results .cams-table__body .cams-table__row');
```
fmaddenflx: Looks like a lot of repetitive dom queries in this test file.  We should maybe replace those with some simple helper functions.

fmaddenflx:
The breakpoint overrides per screen are intentional — each screen has a unique layout context (e.g., SearchScreen has a side-by-side filter+results pane from 1025px+, while StaffAssignment stays stacked longer). Sharing breakpoints across screens would force unrelated layouts to couple, which would create more problems than it solves.

user-interface/src/lib/components/uswds/Table.tsx
mattstankey: Could you put a knip ignore comment to suppress the warning?

user-interface/src/search/SearchScreen.tsx
```diff
@@ -34,6 +34,37 @@ import useFeatureFlags, {
...
+import { CamsHttpError } from '@/lib/models/api';
+
+type ClosedCasesHintMessageProps = {
+  closedCasesCount?: number;
+  variant: 'generic' | 'count';
+  onIncludeClosedCases?: () => void;
+};
+
+function ClosedCasesHintMessage({
+  closedCasesCount = 0,
```
mattstankey: **Issue #7 — `ClosedCasesHintMessage` not actually shared**

This component is module-private. The PR description says it's exported from `SearchResults` for reuse, but it was removed from there and re-declared here without being exported. If `MyCasesScreen` or a future screen needs it, it'll be copy-pasted and you'll end up with diverging versions. Consider either exporting it from this file or extracting it to `lib/components/cams/`.

user-interface/src/search/SearchScreen.tsx
```diff
@@ -82,6 +113,56 @@ export default function SearchScreen() {
   const [temporarySearchPredicate, setTemporarySearchPredicate] =
     useState<CasesSearchPredicate>(defaultSearchPredicate);
   const [searchPredicate, setSearchPredicate] = useState<CasesSearchPredicate>({});
+  const [hasResults, setHasResults] = useState<boolean>(false);
```
mattstankey: **Issue #4 — `hasResults` initialization inconsistency**

`SearchScreen` initializes `hasResults` as `false`. `MyCasesScreen` initializes the equivalent as `boolean | null`, using `null` as a 'not yet searched' sentinel to prevent the no-results alert from rendering before any search has run. `SearchScreen` achieves the same effect via `isValidSearchPredicate` gating instead. Both work, but the inconsistency is a readability hazard. Worth aligning the pattern across both screens so the intent is obvious.

user-interface/src/my-cases/MyCasesScreen.tsx
```diff
@@ -42,6 +43,26 @@ export const MyCasesScreen = () => {
   const [doShowClosedCases, setDoShowClosedCases] = useState(false);
   const [draftNotesCaseIds, setDraftNotesCaseIds] = useState<string[]>([]);
   const [draftNotes, setDraftNotes] = useState<Cacheable<CaseNoteInput>[]>([]);
+  const [hasResults, setHasResults] = useState<boolean | null>(null);
```
mattstankey, line 46: **Issue #4 (cont.) — `hasResults` initialization inconsistency**

`null` is used here as a 'not yet searched' sentinel to prevent the no-results alert from appearing before the first fetch. `SearchScreen` uses `false` for the same state and relies on `isValidSearchPredicate` gating instead. The two screens solve the same problem differently. Worth picking one pattern and applying it consistently.

user-interface/src/my-cases/MyCasesScreen.tsx
```diff
@@ -42,6 +43,26 @@ export const MyCasesScreen = () => {
...
+  const [searchError, setSearchError] = useState<{ title: string; message: string } | null>(null);
+
+  function handleSearchError(error: unknown) {
```
mattstankey, line 49: **Issue #8 — No unit tests for new error-handling paths**

`handleSearchError` and `handleResultsChanged` are new code with no test coverage in `MyCasesScreen.test.tsx` — that file only updates CSS selectors. At minimum consider tests for: error alert renders on a generic fetch failure; 504 timeout produces the correct title/message; no-results alert renders when `hasResults` becomes `false`; alert clears when a subsequent search succeeds.

user-interface/src/search/SearchScreen.scss
```diff
@@ -17,37 +14,74 @@
...
+    }
+
+    .search-results-panel {
+      flex: 1 1 0;
+      min-width: 0;
+    }
+
+    // Undo the CamsTable flex-row layout — table stays stacked at this breakpoint
+    // because the filter panel leaves insufficient width for columns
+    .cams-table--responsive {
```
mattstankey: **Issue #1 — CSS override block duplicated across screen files**

This stacked-layout override block (plus the 1025px flex-row re-enable block) appears verbatim in `StaffAssignmentScreen.scss` as well. If `CamsTable`'s class names or breakpoints change, there are now three places to update. Consider extracting these as SCSS mixins in `CamsTable.scss` (`@mixin cams-table-stacked` / `@mixin cams-table-flex-row`) and `@include`-ing them in each screen, so the override logic stays co-located with the component.

user-interface/src/search/SearchScreen.tsx
```diff
@@ -321,158 +334,224 @@ export default function SearchScreen() {
...
+                    type={UswdsAlertStyle.Info}
+                    show={true}
+                    inline={true}
+                    role="alert"
+                  ></Alert>
+                </div>
+              )}
+              {isValidSearchPredicate(searchPredicate) && (
+                <>
+                  {!isSearching &&
```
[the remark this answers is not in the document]
fmaddenflx: Removed — `scope` was never passed at any call site, so it was dead API surface spreading silently onto a div where it has no semantic meaning.

user-interface/src/search/SearchScreen.test.tsx
fmaddenflx: Addressed — extracted `getResultsTable`, `getResultRows`, `getSearchButton`, and `getCaseNumberInput` as module-level helpers shared across all describe blocks. Commit: 266d78e0c.

user-interface/src/lib/components/uswds/Table.tsx
fmaddenflx: That was an option, but I felt that could go unnoticed and would be hard to track.  I felt this change was a little more obvious and easy and obvious to remedy.

user-interface/src/search/SearchScreen.tsx
fmaddenflx: It's not really a "component" in the sense of a react component.  We are using this pattern in 3 modules.  I suppose there is a way we can structure it, such that we can pass in the options for what to display in each variant, rather than defining the method in each module.

user-interface/src/search/SearchScreen.scss
fmaddenflx: You're right in that lines 55-62 in SearchScreen.scss is completely pointless.  As for lines 72-78 are concerned, Let me see if a mixin would make this easier.

user-interface/src/search/SearchScreen.test.tsx
fmaddenflx: Addressed — extracted `getResultsTable`, `getResultRows`, `getSearchButton`, and `getCaseNumberInput` as module-level helpers shared across all describe blocks. Commit: 266d78e0c.

user-interface/src/search/SearchScreen.tsx
fmaddenflx: Addressed — extracted `ClosedCasesHintMessage` to `lib/components/cams/ClosedCasesHintMessage/ClosedCasesHintMessage.tsx`. The component now accepts a `message` string prop so each screen supplies its own context-appropriate wording. Commit: 477527a9a.

user-interface/src/search/SearchScreen.tsx
fmaddenflx: Addressed — aligned `SearchScreen` to use `boolean | null` (matching `MyCasesScreen`), with strict `=== false` / `=== true` checks on the hint booleans so `null` (not yet searched) never triggers an alert. Commit: 477527a9a.

user-interface/src/my-cases/MyCasesScreen.tsx
fmaddenflx: Both screens now use `boolean | null` with the same sentinel pattern. Commit: 477527a9a.

user-interface/src/my-cases/MyCasesScreen.tsx
fmaddenflx: Added four tests to `MyCasesScreen.test.tsx`: generic fetch failure shows error alert, 504 shows timeout-specific title, empty results shows no-results alert, and error alert clears when a subsequent search succeeds. Commit: 477527a9a.

user-interface/src/search/SearchScreen.scss
fmaddenflx: Extracted `cams-table-stacked` and `cams-table-flex-row` as SCSS mixins in `_cams-table-mixins.scss`, co-located with the CamsTable component. `SearchScreen.scss` and `StaffAssignmentScreen.scss` both `@use` the partial and call the mixins. Commit: 2464b5a54.

mattstankey (approved):

PR #2359, opened 2026-05-06
Increase branch coverage to 100% on 5 backend files

btposey:
# Purpose

Increase backend test coverage. Our coverage metric is dangerously close to the threshold.

## Summary by Sourcery

Increase backend test coverage and branch coverage for repository, factory, gateway helper, ACMS gateway, and case-associated use case components.

Tests:
- Add singleton lifecycle and enhanced error-path coverage for CasesMongoRepository, including phonetic token and case history flows.
- Broaden factory tests to cover a wide range of gateway/repository getters and provider-specific behavior for authorization, user groups, and user sessions.
- Extend ACMS gateway tests to cover error wrapping and new migration/deleted case helper methods, including success and failure paths.
- Introduce tests for GatewayHelper mock extract methods, including file I/O success and failure scenarios.
- Add comprehensive tests for the CaseAssociated use case, covering lead/member consolidation paths and date selection logic.


backend/lib/factory.test.ts
```diff
@@ -126,11 +205,22 @@ describe('Factory function-apps', () => {
     expect(obj).toBeInstanceOf(CasesDxtrGateway);
   });
 
-  test('getAssignmentRepository', async () => {
```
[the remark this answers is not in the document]
btposey: Addressed in the follow-up commit. The factory tests have been rewritten: all 35 getters are now exercised via a single `test.each` table with `toBeDefined()` assertions, the `makeOktaUserGroupContext` helper was already extracted, and the `getUserGroupGateway` tests are isolated in their own `describe` block with `vi.resetModules()`. The file went from ~800 lines to ~160. No singleton-reset boilerplate remains in the main suite.

btposey:
@sourcery-ai review

backend/lib/factory.test.ts
```diff
@@ -108,153 +20,134 @@ describe('Factory function-apps', () => {
...
+    ctx.config.authConfig.provider = 'okta';
+    ctx.config.userGroupGatewayConfig.provider = 'okta';
+    ctx.config.userGroupGatewayConfig.clientId = 'mock';
+    ctx.config.userGroupGatewayConfig.keyId = 'mock';
+    ctx.config.userGroupGatewayConfig.url = 'https://fake.url';
+    ctx.config.userGroupGatewayConfig.privateKey = '{}';
+    return ctx;
+  }
+
+  test.each([
```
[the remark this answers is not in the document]
btposey: Thanks for the detailed feedback. Here's what was done and the reasoning behind the remaining choices:

**Point 1 — explicit type assertions:** Addressed in the latest commit. The `test.each` table now passes the expected concrete class as a third column and asserts `toBeInstanceOf(ExpectedType)` for all 34 class-returning getters. The mapping from factory method to implementation type is now explicit in the table.

**Point 2 — dynamic imports / `vi.resetModules`:** This is intentional and necessary. `factory.ts` holds several module-level singleton variables (e.g., `idpApiGateway`). Without `vi.resetModules()`, a cached instance from a prior test bleeds into subsequent tests, making the `getUserGroupGateway` tests order-dependent and unreliable. The dynamic import pattern is scoped to that one `describe` block precisely because it's the only function that requires it.

**Point 3 — `makeOktaUserGroupContext` consistency:** The helper is used for the okta-path tests; the mock-provider test intentionally sets only the two fields that differ (`authConfig.provider` and `userGroupGatewayConfig.provider`) rather than the full okta config, since it only exercises the mock branch. Applying the okta helper there and then overriding would be misleading about intent.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -175,6 +175,17 @@ describe('ACMS gateway tests', () => {
...
 
+  test('should wrap non-Error thrown values in CamsError', async () => {
+    vi.spyOn(AbstractMssqlClient.prototype, 'executeQuery').mockRejectedValue('plain string error');
+
+    const context = await createMockApplicationContext();
+    const gateway = new AcmsGatewayImpl(context);
+
+    await expect(
+      gateway.getLeadCaseIds(context, { chapter: '11', divisionCode: '010' }),
+    ).rejects.toThrow(expect.objectContaining({ status: 500, module: 'ACMS-GATEWAY' }));
```
fmaddenflx: The description says that this should be a CamsError, but what indication does this have that this is specifically a CamsError?  Is it just the fact that it has a module defined?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -232,6 +241,110 @@ describe('ACMS gateway tests', () => {
...
+    const context = await createMockApplicationContext();
+    const gateway = new AcmsGatewayImpl(context);
+
+    await expect(gateway.loadMigrationTable(context)).rejects.toThrow(
+      expect.objectContaining({ status: 500, module: 'ACMS-GATEWAY' }),
+    );
+  });
+
+  describe('getMigrationCaseIds', () => {
+    test('should return caseIds between start and end', async () => {
```
fmaddenflx: I don't understand the description of this test.  Nothing in the test indicates a "start" and "end" so I don't know what it's talking about.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -232,6 +241,110 @@ describe('ACMS gateway tests', () => {
...
+      });
+
+      const context = await createMockApplicationContext();
+      const gateway = new AcmsGatewayImpl(context);
+      const result = await gateway.getMigrationCaseIds(context, 1, 2);
+
+      expect(result).toEqual(['081-24-00001', '081-24-00002']);
+    });
+
+    test('should throw CamsError when executeQuery fails', async () => {
```
fmaddenflx: Duplicate test, verbatim I believe.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -232,6 +241,110 @@ describe('ACMS gateway tests', () => {
...
+        new Error('db error'),
+      );
+
+      const context = await createMockApplicationContext();
+      const gateway = new AcmsGatewayImpl(context);
+
+      await expect(gateway.emptyMigrationTable(context)).rejects.toThrow(
+        expect.objectContaining({ status: 500, module: 'ACMS-GATEWAY' }),
+      );
+    });
```
fmaddenflx: Is this a 3rd copy of the same test?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -232,6 +241,110 @@ describe('ACMS gateway tests', () => {
...
+        new Error('db error'),
+      );
+
+      const context = await createMockApplicationContext();
+      const gateway = new AcmsGatewayImpl(context);
+
+      await expect(gateway.getMigrationCaseCount(context)).rejects.toThrow(
+        expect.objectContaining({ status: 500, module: 'ACMS-GATEWAY' }),
+      );
+    });
```
fmaddenflx: And a 4th copy?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
fmaddenflx: I actually saw 4 or 5 of these but eventually realized they all have 1 line difference.  So maybe disregard or find a way to do this with a test.each, and maybe supplied the function to call for each, if that's possible?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
fmaddenflx: Maybe use expect(actualError.isCamsError).toBeTruthy(); ?

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Done. Consolidated all 7 error-path tests into a single `test.each` that passes the gateway method to invoke as a lambda, identical to the pattern you suggested.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Yes — all collapsed into the `test.each` table in the latest commit.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Done — both the non-Error wrapping test and the new `test.each` error table now assert `error.isCamsError` and `error.module`.

btposey: Good catch. Switched to `error.isCamsError` as Fritz suggested in the follow-up comment — that's the explicit indicator set by the `CamsError` constructor. Updated in the latest commit.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Fixed. Renamed to `should return caseIds from the migration table for the given range`.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Yes — all collapsed into the `test.each` table in the latest commit.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
btposey: Addressed. All 7 error-path tests (getLeadCaseIds, getConsolidationDetails, loadMigrationTable, getMigrationCaseIds, emptyMigrationTable, getMigrationCaseCount, getDeletedCaseIds) have been collapsed into a single `test.each` table.

jalba-flexion (approved):

backend/lib/factory.test.ts
```diff
@@ -1,260 +1,230 @@
...
-        COSMOS_ENDPOINT: 'https://cosmos-ustp-cams-dev.documents.azure.us:443/',
-        CAMS_USER_GROUP_GATEWAY_CONFIG:
-          'url=https://fake.url|clientId=mock|keyId=mock|privateKey={"foo": "bar"}}',
-      },
+    context = await createMockApplicationContext();
+  });
+
+  function makeOktaUserGroupContext(base: ApplicationContext): ApplicationContext {
+    const ctx = { ...base };
+    ctx.config.authConfig.provider = 'okta';
```
mattstankey: **Shallow-copy mutation bug**: `{ ...base }` only spreads the top-level object, so `ctx.config`, `ctx.config.authConfig`, and `ctx.config.userGroupGatewayConfig` are still shared references with `context`. Mutations here (e.g. `ctx.config.authConfig.provider = 'okta'`) will bleed into the shared `context` used by the `test.each` block and other tests, which can cause order-dependent failures.

Fix by deep-spreading each nested level you intend to mutate:
```ts
const ctx = {
  ...base,
  config: {
    ...base.config,
    authConfig: { ...base.config.authConfig, provider: 'okta' },
    userGroupGatewayConfig: { ...base.config.userGroupGatewayConfig, provider: 'okta', clientId: 'mock', keyId: 'mock', url: 'https://fake.url', privateKey: '{}' },
  },
};
```
The same issue applies to the shallow copies in the `getOfficesRepository`, `getAuthorizationGateway`, and `getUserSessionUseCase` tests (lines 154, 158, 164, 168, 172, 201, 215, 222, 226).

backend/lib/factory.test.ts
```diff
@@ -1,260 +1,230 @@
...
+      AzureBlobObjectStorageGateway,
+    ],
+    ['getAcmsGateway', (f, ctx) => f.getAcmsGateway(ctx), AcmsGatewayImpl],
+    ['getAtsGateway', (f, ctx) => f.getAtsGateway(ctx), MockAtsGateway],
+    [
+      'getApiToDataflowsGateway',
+      (f, ctx) => f.getApiToDataflowsGateway(ctx),
+      ApiToDataflowsGatewayImpl,
+    ],
+  ] as const)('%s returns an instance of the expected type', (_label, getter, ExpectedType) => {
```
mattstankey: **Coverage regression on non-mock paths**: The old test suite used `dbContext` (`DATABASE_MOCK=false`, `CAMS_LOGIN_PROVIDER=okta`) to verify the factory routes to real implementations: `CasesDxtrGateway`, `DxtrOrdersGateway`, `OfficesDxtrGateway`, `OrdersMongoRepository`, `ConsolidationOrdersMongoRepository`, `RuntimeStateMongoRepository`, `CaseNotesMongoRepository`, `UsersMongoRepository`, `TrusteeAppointmentsMongoRepository`, `UserSessionCacheMongoRepository`.

All of those tests are gone. The new suite only exercises the mock branch. If the `DATABASE_MOCK=false` or `CAMS_LOGIN_PROVIDER=okta` routing logic in the factory breaks, nothing here will catch it.

Is dropping non-mock path coverage intentional? If these paths require a live Cosmos connection, the tests could use `toBeInstanceOf` checks on the class itself (which doesn't instantiate a connection) similar to how `OktaGateway` is checked via `toEqual` rather than `toBeInstanceOf`.

backend/lib/factory.test.ts
btposey: Fixed. `makeOktaUserGroupContext` and all inline context mutations now use `Object.assign(Object.create(proto), base.config)` to clone the `ApplicationConfiguration` class instance (preserving prototype/methods), then spread only the plain `authConfig` and `userGroupGatewayConfig` sub-objects before reassigning them. This isolates mutations without touching the shared `context`.

backend/lib/factory.test.ts
btposey: Fixed. The suite now has two `describe` blocks — one for `DATABASE_MOCK=false` (real implementations: `CasesDxtrGateway`, `DxtrOrdersGateway`, `OfficesDxtrGateway`, real Mongo repos, etc.) and one for `DATABASE_MOCK=true` (mock implementations). Each block runs `vi.resetModules()` + dynamic factory import in `beforeEach` to clear singletons, with all types declared via `import type` aliases to satisfy the no-any rule. Real-path coverage is fully restored.

PR #2361, opened 2026-05-07
CAMS-691/737/738: Fix screen reader accessibility for trustee name, district and chapter filters

diabagatekelly:
## Summary
Fixes screen reader accessibility issues in trustee filter components, ensuring proper label announcements and preventing unwanted automatic announcements.

## Changes
- **Name filter label**: Removed `aria-hidden` from label so screen readers announce "Trustee Name" instead of the input value when accordion expands
- **Input announcements**: Added `aria-live="off"` and `aria-atomic="false"` to name filter input to suppress automatic value announcements
- **Filter expansion announcements**: Fixed to include all active filters (name, district/division, chapter) and only say "filtered by" when filters are applied
- **Singular filter names**: Changed from "districts" to "district" in announcements for consistency
- **Duplicate announcements**: Fixed by completing useEffect dependency array to include all referenced variables
- **Test coverage**: Added tests for name filter accessibility and consolidated ComboBoxAlt tests

## Test Plan
- [x] Screen reader announces "Trustee Name" (not the value) when accordion expands
- [x] Filter expansion announces correct count and active filters
- [x] No duplicate announcements when selecting chapter options
- [x] All tests passing
- [x] TypeScript compilation successful
- [x] Build successful

### DIFF
--- a/user-interface/src/lib/components/combobox/ComboBoxAlt.tsx
+++ b/user-interface/src/lib/components/combobox/ComboBoxAlt.tsx
@@ -14,7 +14,6 @@ import Button, { UswdsButtonStyle } from '../uswds/Button';
 import useOutsideClick from '@/lib/hooks/UseOutsideClick';
 import { ComboBoxRef } from '@/lib/type-declarations/input-fields';
 
-// Configuration constants
 const DROPDOWN_POSITION_THRESHOLD_RATIO = 0.5; // Show dropdown above input when element is in bottom half of viewport
 
 export type ComboOption<T = string> = {
@@ -77,13 +76,6 @@ function ComboBox_(props: ComboBoxProps, ref: React.Ref<ComboBoxRef>) {
     ...otherProps
   } = props;
 
-  // Validate required props
-  if (!comboBoxId || typeof comboBoxId !== 'string') {
-    throw new Error('ComboBox requires a valid id prop');
-  }
-
-  // ========== STATE ==========
-
   const [comboboxDisabled, setComboboxDisabled] = useState<boolean>(!!disabled);
   const [expanded, setExpanded] = useState<boolean>(false);
   const expandedRef = useRef<boolean>(false);
@@ -100,16 +92,12 @@ function ComboBox_(props: ComboBoxProps, ref: React.Ref<ComboBoxRef>) {
   );
   const [filter, setFilter] = useState<string | null>(null);
 
-  // ========== REFS ==========
-
   const comboBoxRef = useRef(null);
   const comboBoxListRef = useRef<HTMLUListElement>(null);
   const filterRef = useRef<HTMLInputElement>(null);
   const containerRef = useRef<HTMLDivElement>(null);
 
   useOutsideClick([comboBoxRef], isOutsideClick);
-
-  // ========== MISC FUNCTIONS ==========
 
   function clearFilter() {
     setFilter(null);
@@ -258,8 +246,6 @@ function ComboBox_(props: ComboBoxProps, ref: React.Ref<ComboBoxRef>) {
       containerRef.current?.focus();
     });
   }
-
-  // ========== HANDLERS ==========
 
   function handleClearAllClick() {
     clearSelections();
@@ -495,8 +481,6 @@ function ComboBox_(props: ComboBoxProps, ref: React.Ref<ComboBoxRef>) {
     }
   }
 
-  // ========== USE EFFECTS ==========
-
   useEffect(() => {
     expandedRef.current = expanded;
 
@@ -564,8 +548,6 @@ function ComboBox_(props: ComboBoxProps, ref: React.Ref<ComboBoxRef>) {
     focusInput,
     focus: focusInput,
   }));
-
-  // ========== JSX ==========
 
   return (
     <div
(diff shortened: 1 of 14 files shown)


mattstankey:
## Code Review — Accessibility & Correctness Issues

### High Priority

**1. Missing `aria-selected` on `role=\"option\"` (ARIA violation)**
`ComboBoxAlt.tsx` suppresses `jsx-a11y/role-has-required-aria-props` via an eslint-disable comment. The `role=\"option\"` ARIA spec requires `aria-selected`. The current workaround encodes selection state in `aria-label` (e.g., `\"option1, not selected\"`), but screen readers that programmatically query `aria-selected` (like VoiceOver's list navigation) will not reflect the correct state. Consider adding `aria-selected={selectedMap.has(option.value)}` to the `<li>` alongside the descriptive `aria-label`.

**2. `isOutsideClick` closes dropdown on inside clicks when nothing is selected**
The condition in `ComboBoxAlt.tsx` includes `|| selectedMap.size === 0`, which means any click inside the combobox closes it immediately when nothing is yet selected — including the initial click to open it and begin making a selection. This appears to break single-select and first-time multi-select interactions.

**3. Direct state mutation before `setSelectedMap`**
`handleDropdownItemSelection` calls `selectedMap.delete()`, `selectedMap.clear()`, and `selectedMap.set()` on the existing Map before wrapping it in `new Map(selectedMap)`. React state should be treated as immutable. The correct pattern is to build a new Map first:
```typescript
const next = new Map(selectedMap);
next.delete(option.value);
setSelectedMap(next);
```

---

### Medium Priority

**4. `setTimeout` inside async effect without cleanup**
`TrusteesList.tsx` uses a raw `setTimeout` inside the debounced fetch callback to post-announce filter results. If the component unmounts while the timer is pending, this will attempt to call `setLiveAnnouncement` on an unmounted component. The timer ID should be captured and cleared in the effect's cleanup, or a mounted ref should guard the call.

**5. `hasExpandedOnceRef` declared after the `useEffect` that uses it**
The ref is defined after the `useEffect` that references it. While this works at runtime due to hoisting, it breaks the top-down readability contract for hooks and will confuse readers. Move the ref declaration above its first use.

**6. `onDistrictFilterClose`, `onChapterFilterClose`, `onDivisionFilterClose` are dead props**
These props are added to `TrusteeDistrictFilterProps` and `TrusteeDistrictFilterViewModel` and threaded through the component chain, but `TrusteesList.tsx` never actually passes them. Either wire them up to do something or remove them to avoid dead API surface.

**7. `aria-live="off"` on `<input role="combobox">` is non-standard**
`aria-live` is specified for live region containers, not interactive form elements. Placing it on the combobox input may suppress announcements in some browser/screen reader combinations but is not reliable cross-platform. The goal (suppressing automatic value announcements) is worth achieving, but this attribute on an `<input>` is not the right mechanism — this behavior is typically controlled by the screen reader's interaction mode (forms/browse mode).

**8. `aria-hidden="true"` on visible filter label spans**
The "District", "Chapter", and "District (Division)" `filter-control-label` spans now have `aria-hidden="true"`, removing them from the accessibility tree. The `ComboBox` `label` prop already surfaces the label when the widget is focused, so this is arguably correct — but it creates an unlabeled visual region for users not focused on the widget. This should be validated with an actual screen reader to confirm no gaps in context.

---

### Minor

- **CSS selector coupling**: `TrusteeDistrictFilter.scss` uses `> div[aria-live="off"]` to position the clear button. Styling based on ARIA attributes creates fragile coupling between accessibility structure and layout — a class name would be more maintainable.
- **`handleClearAllKeyDown` is redundant**: `<Button>` elements natively handle `Enter`, making the `onKeyDown` duplicate unnecessary.
- **Mixed prop access style**: The component destructures most props at the top but directly accesses `props.onUpdateSelection`, `props.onFocus`, `props.options`, etc. in several places — inconsistent.

user-interface/src/lib/components/combobox/ComboBoxAlt.tsx
```diff
@@ -0,0 +1,720 @@
...
+  useRef,
+  useState,
+  type JSX,
+} from 'react';
+import Icon from '../uswds/Icon';
+import Button, { UswdsButtonStyle } from '../uswds/Button';
+import useOutsideClick from '@/lib/hooks/UseOutsideClick';
+import { ComboBoxRef } from '@/lib/type-declarations/input-fields';
+
+// Configuration constants
```
cruzjone-flexion: Comment necessary ?

cruzjone-flexion: And more in this file such as "STATE" "REFS" "MISC FUNCTIONS" "HANDLERS" "USE EFFECTS" "JSX"

user-interface/src/lib/components/combobox/ComboBoxAlt.tsx
```diff
@@ -0,0 +1,720 @@
...
+    overflowStrategy,
+    errorMessage,
+    hideClearAllButton,
+    placeholder,
+    scrollToSelected,
+    ...otherProps
+  } = props;
+
+  // Validate required props
+  if (!comboBoxId || typeof comboBoxId !== 'string') {
```
cruzjone-flexion: This check necessary? Typescript will catch this.

jamesobrooks (approved):

user-interface/src/lib/components/combobox/ComboBoxAlt.tsx
jamesobrooks: Removed — section header dividers and the `// Configuration constants` label are gone.

user-interface/src/lib/components/combobox/ComboBoxAlt.tsx
jamesobrooks: Removed — TypeScript enforces the `string` type at compile time so the runtime guard is unreachable.

mattstankey (approved):

PR #2367, opened 2026-05-07
CAMS-670 Implement bank list and add bank (Slice 1)

fmaddenflx:
# Purpose

OoO users need to manage which banks are available for trustees. This PR delivers the first deployable slice: viewing the bank list and creating new banks from the Administration panel.

# Major Changes

- New `BankProfile` and `BankAuditHistory` types in common — dedicated entity replacing the generic `lists/banks` key-value pattern
- Full backend stack: MongoDB repository (`banks` collection), use case, SuperUser-gated controller, and Azure Function route `GET/POST /banks`
- Every bank creation writes an audit record (`AUDIT_BANK`) with `before: null` and the created bank as `after`
- `Api2.getBanks` and `Api2.createBank` updated to call `/banks` instead of the old `/lists/banks` endpoint
- Banks nav item added to the Administration sidebar
- Banks list page shows bank name (linked to detail, Slice 2) and status columns
- Add Bank modal with client-side required-field validation; modal stays open on API failure

# Testing/Validation

Implemented using TDD (red-green-refactor) throughout. Test coverage:

- Repository: `getBanks`, `createBank`, `createBankAuditRecord` — happy paths and error wrapping
- Use case: delegation to repo, audit record written with `before: null`, user stamped from context
- Controller: 200/201 success, 403 for non-SuperUser, 400 for missing/blank name
- Azure Function: GET and POST routing, 405 for unsupported methods, error propagation
- Api2: new endpoint URLs asserted in existing test suite
- `Banks` component: table render, loading state, error alert, nav link, route registration
- `AddBankModal`: show/hide lifecycle, validation error stays modal open, success calls `onSuccess` and closes, API failure keeps modal open, cancel clears form

# Notes

Banks use a dedicated MongoDB collection (`banks`) rather than the generic `lists` collection. This was a deliberate architectural choice — banks need audit history, status lifecycle, and their own CRUD surface that the generic key-value lists pattern can't support. The old `lists/banks` endpoint remains intact for backward compatibility but is no longer used by the UI.

The `closeOnClick: false` flag is set on the Add Bank modal submit button so the modal stays open on validation errors or API failures — the modal closes only on explicit success or cancel.

Slice 2 (bank detail + edit) and Slice 3 (trustees tab) will follow in subsequent PRs.

# Definition of Done:

## Summary by Sourcery

Introduce a dedicated banks domain with backend and UI support for viewing and creating bank profiles in the administration panel.

New Features:
- Add BankProfile and BankAuditHistory domain types and expose them from the shared common package.
- Implement a Banks MongoDB repository, use case, controller, and Azure Function HTTP endpoint for listing and creating banks, including audit log records on creation.
- Add an Administration Banks screen that lists banks with status and supports creating new banks via an Add Bank modal with validation and analytics tracking.
- Expose new Api2.getBanks and Api2.createBank client methods backed by the new /banks endpoint and update the mock API accordingly.
- Extend the admin navigation and routing to include the new Banks section.

Enhancements:
- Introduce a dedicated BanksRepository gateway type and factory wiring for resolving concrete banks repositories in different environments.
- Update the mock Mongo repository to implement the new BanksRepository interface methods.
- Standardize bank-related testing across backend and frontend for repository, use case, controller, function routing, API client, admin navigation, and UI components.

Tests:
- Add comprehensive tests for the banks Mongo repository, banks use case, banks controller, and Azure Function routing and behavior.
- Add frontend tests covering Api2 bank endpoints, admin navigation to the Banks screen, the Banks list component, and the Add Bank modal’s lifecycle, validation, and error handling.

### DIFF
--- a/backend/function-apps/api/banks/banks.function.ts
+++ b/backend/function-apps/api/banks/banks.function.ts
@@ -1,0 +1,46 @@
+import { app, HttpRequest, HttpResponseInit, InvocationContext } from '@azure/functions';
+import ContextCreator from '../../azure/application-context-creator';
+import { toAzureError, toAzureSuccess } from '../../azure/functions';
+import { BanksController } from '../../../lib/controllers/banks/banks.controller';
+
+const MODULE_NAME = 'BANKS-FUNCTION';
+
+export default async function handler(
+  request: HttpRequest,
+  invocationContext: InvocationContext,
+): Promise<HttpResponseInit> {
+  const logger = ContextCreator.getLogger(invocationContext);
+
+  try {
+    const context = await ContextCreator.applicationContextCreator({
+      invocationContext,
+      logger,
+      request,
+    });
+
+    const controller = new BanksController(context);
+    context.session = await ContextCreator.getApplicationContextSession(context);
+
+    const method = request.method;
+    let responseBody;
+
+    if (method === 'GET') {
+      responseBody = await controller.handleGet(context);
+    } else if (method === 'POST') {
+      responseBody = await controller.handlePost(context);
+    } else {
+      return { status: 405, jsonBody: 'Method Not Allowed' };
+    }
+
+    return toAzureSuccess(responseBody);
+  } catch (error) {
+    return toAzureError(logger, MODULE_NAME, error);
+  }
+}
+
+app.http('banks', {
+  methods: ['GET', 'POST'],
+  authLevel: 'anonymous',
+  handler,
+  route: 'banks/{bankId?}',
+});
--- a/backend/function-apps/api/banks/banks.function.ts
+++ b/backend/function-apps/api/banks/banks.function.ts
@@ -18,8 +18,8 @@ export default async function handler(
       request,
     });
 
-    const controller = new BanksController(context);
     context.session = await ContextCreator.getApplicationContextSession(context);
+    const controller = new BanksController(context);
 
     const method = request.method;
     let responseBody;
--- a/backend/function-apps/api/banks/banks.function.ts
+++ b/backend/function-apps/api/banks/banks.function.ts
@@ -42,5 +42,5 @@ app.http('banks', {
   methods: ['GET', 'POST'],
   authLevel: 'anonymous',
   handler,
-  route: 'banks/{bankId?}',
+  route: 'banks',
 });
--- a/backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
@@ -1,0 +1,74 @@
+import { BankAuditHistory, BankProfile } from '@common/cams/banks';
+import { Creatable } from '@common/cams/creatable';
+import { getCamsError } from '../../../common-errors/error-utilities';
+import QueryBuilder from '../../../query/query-builder';
+import { BanksRepository } from '../../../use-cases/gateways.types';
+import { ApplicationContext } from '../../types/basic';
+import { BaseMongoRepository } from './utils/base-mongo-repository';
+
+const MODULE_NAME = 'BANKS-MONGO-REPOSITORY';
+const COLLECTION_NAME = 'banks';
+
+const { using, orderBy } = QueryBuilder;
+const doc = using<BankProfile>();
+
+export class BanksMongoRepository extends BaseMongoRepository implements BanksRepository {
+  private static referenceCount: number = 0;
+  private static instance: BanksMongoRepository;
+
+  constructor(context: ApplicationContext) {
+    super(context, MODULE_NAME, COLLECTION_NAME);
+  }
+
+  public static getInstance(context: ApplicationContext) {
+    if (!BanksMongoRepository.instance) {
+      BanksMongoRepository.instance = new BanksMongoRepository(context);
+    }
+    BanksMongoRepository.referenceCount++;
+    return BanksMongoRepository.instance;
+  }
+
+  public static dropInstance() {
+    if (BanksMongoRepository.referenceCount > 0) {
+      BanksMongoRepository.referenceCount--;
+    }
+    if (BanksMongoRepository.referenceCount < 1) {
+      BanksMongoRepository.instance?.client.close().then();
+      BanksMongoRepository.instance = null;
+    }
+  }
+
+  public release() {
+    BanksMongoRepository.dropInstance();
+  }
+
+  async getBanks(): Promise<BankProfile[]> {
+    const query = doc('documentType').equals('BANK_PROFILE');
+    try {
+      return await this.getAdapter<BankProfile>().find(
+        query,
+        orderBy<BankProfile>(['name', 'ASCENDING']),
+      );
+    } catch (originalError) {
+      throw getCamsError(originalError, MODULE_NAME, 'Unable to retrieve banks.');
+    }
+  }
+
+  async createBank(bank: Creatable<BankProfile>): Promise<BankProfile> {
+    try {
+      const newId = await this.getAdapter<BankProfile>().insertOne(bank as BankProfile);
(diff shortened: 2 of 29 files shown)


mattstankey (commented):
Three items worth addressing before merge.

backend/function-apps/api/banks/banks.function.ts
```diff
@@ -0,0 +1,46 @@
...
+    return toAzureError(logger, MODULE_NAME, error);
+  }
+}
+
+app.http('banks', {
+  methods: ['GET', 'POST'],
+  authLevel: 'anonymous',
+  handler,
+  route: 'banks/{bankId?}',
+});
```
mattstankey, line 46: The route `banks/{bankId?}` means `GET /banks/123` is routed here today. The controller ignores the bankId and returns all banks, so a caller expecting a single-bank response silently gets the full list. Until Slice 2 ships, consider either dropping `{bankId?}` from the route (just `route: 'banks'`) or returning 404/501 when `context.request.params.bankId` is present.

user-interface/src/lib/testing/mock-api2.ts
```diff
@@ -2809,35 +2810,39 @@ async function deleteBankruptcySoftware(_ignore: string) {
...
   };
 }
 
-async function postBank(_ignore: Creatable<BankListItem>) {
-  return '--id--';
-}
-
-async function deleteBank(_ignore: string) {
+async function createBank(_ignore: Pick<BankProfile, 'name'>) {
   return;
```
mattstankey: `createBank` returns `undefined`, but `AddBankModal` destructures `response.data` from the result. Any test that goes through the full mock API flow (rather than spying on `Api2.createBank` directly) will get a runtime error. Should return `{ data: <a mock BankProfile> }` to match the real function shape.

user-interface/src/admin/banks/AddBankModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    try {
+      const response = await Api2.createBank({ name: trimmed });
+      const created = (response as { data: BankProfile }).data;
+      onSuccess(created);
+      modalRef.current?.hide();
+      alert?.success('Bank added successfully.');
+    } catch (error) {
+      alert?.error(`Failed to add bank. ${(error as Error).message}`);
+    }
```
mattstankey, line 62: This surfaces the raw `Error.message` to the user, which can leak internal details. Consider a static user-facing message here and log the original error separately — e.g. `alert?.error('Failed to add bank. Please try again.')` with a `console.error(error)` or telemetry call.

user-interface/src/admin/banks/Banks.tsx
```diff
@@ -0,0 +1,95 @@
...
+    loadBanks().then(() => setIsLoaded(true));
+  }, []);
+
+  function handleBankAdded(bank: BankProfile) {
+    setBanks((prev) => [...prev, bank]);
+    getAppInsights().appInsights.trackEvent({
+      name: 'Bank Created',
+      properties: { bankId: bank.id, bankName: bank.name },
+    });
+  }
```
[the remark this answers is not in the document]
fmaddenflx: Fixed —  now sorts the updated list alphabetically: `[...prev, bank].sort((a, b) => a.name.localeCompare(b.name))`. The backend already returns results sorted by name via `orderBy`, so this keeps the table consistent after an insert without a round trip.

user-interface/src/admin/banks/AddBankModal.tsx
fmaddenflx: Fixed — the catch block now calls `console.error(error)` and shows a static message: `'Failed to add bank. Please try again.'`

backend/function-apps/api/banks/banks.function.ts
fmaddenflx: Fixed — removed `{bankId?}` from the route, now just `route: 'banks'`. The detail endpoint will be added in Slice 2.

user-interface/src/lib/testing/mock-api2.ts
fmaddenflx: Fixed — `createBank` in `mock-api2.ts` now returns `{ data: BankProfile }` matching the real API shape.

backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ApplicationContext } from '../../types/basic';
+import { BaseMongoRepository } from './utils/base-mongo-repository';
+
+const MODULE_NAME = 'BANKS-MONGO-REPOSITORY';
+const COLLECTION_NAME = 'banks';
+
+const { using, orderBy } = QueryBuilder;
+const doc = using<BankProfile>();
+
+export class BanksMongoRepository extends BaseMongoRepository implements BanksRepository {
```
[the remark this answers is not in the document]
fmaddenflx: The singleton + reference-counting pattern is consistent with all other repositories in this codebase (e.g. `ListsMongoRepository`, `TrusteesMongoRepository`). Deviating here would create inconsistency. The lifecycle management is handled the same way across the board, so we're keeping this as-is.

backend/function-apps/api/banks/banks.function.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  try {
+    const context = await ContextCreator.applicationContextCreator({
+      invocationContext,
+      logger,
+      request,
+    });
+
+    const controller = new BanksController(context);
+    context.session = await ContextCreator.getApplicationContextSession(context);
```
bityogi, line 21: The controller is constructed on line 21 **before** `context.session` is set here. The established pattern (see `trustees.function.ts`) sets the session first, then constructs the controller.

This works today only because `BanksUseCase` stores the `context` object reference and reads `context.session.user` later at call time. But it's inconsistent with the pattern and fragile — if the constructor or use case ever accesses `session` during initialization, it'll be `undefined`.

Suggested fix — swap lines 21 and 22:
```ts
context.session = await ContextCreator.getApplicationContextSession(context);
const controller = new BanksController(context);
```

bityogi (commented):
A few additional suggestions beyond what sourcery-ai and Matt already flagged:

**1. Session initialization order (potential bug)** — see inline comment on `banks.function.ts`

**2. Missing nav highlight for `/admin/banks`** — `getInitialNavState()` in `AdminScreen.tsx:21-31` handles `bankruptcy-software`, `privileged-identity`, and `case-reload` but is missing the banks path. Navigating directly to `/admin/banks` (e.g. browser refresh) will show `UNKNOWN` state with no nav link highlighted. Suggested fix:
```ts
if (location.pathname.includes('/admin/banks')) {
  return AdminNavState.BANKS;
}
```

**3. `banks/*` wildcard may be premature** — The route uses `path="banks/*"` but there are no nested routes yet (links were removed until Slice 2). This means paths like `/admin/banks/typo` render the Banks component rather than falling through to the catch-all. Consider using just `path="banks"` until Slice 2 adds nested detail routes.

**4. `MockMongoRepository` stubs still throw** — Already noted by sourcery-ai. Acknowledging this is a pre-existing codebase pattern, but calling it out for awareness in case `dbMock` is exercised against `/banks` in any environment.

**5. `console.error` in `AddBankModal`** — In production, `console.error` is only visible in DevTools. If diagnosing bank creation failures via telemetry matters, consider sending to AppInsights (`getAppInsights().appInsights.trackException`) alongside or instead of `console.error`.

backend/function-apps/api/banks/banks.function.ts
fmaddenflx: Fixed — swapped the two lines so `context.session` is set before constructing the controller, matching the established pattern.

bityogi (approved):

PR #2373, opened 2026-05-08
CAMS-670 Implement manage bank profiles (Slice 2)

fmaddenflx:
# Purpose

Introduce a dedicated bank management feature for Office of Oversight administrators, enabling them to view, create, and edit bank profiles used throughout CAMS.

# Major Changes

- **New `banks` entity** — dedicated MongoDB collection with `BankProfile` and `BankAuditHistory` types, replacing the legacy `/lists/banks` approach with a proper audited entity
- **Backend** — full vertical stack: repository, use case, controller, and Azure Function route (`GET/POST/PUT /banks/{bankId?}`), restricted to SuperUser role
- **Bank list page** (`/admin/banks`) — sortable table with Add Bank modal; newly added banks inserted in alphabetical order
- **Bank detail page** (`/admin/banks/:bankId`) — Overview tab with General Information card (name, status) and Edit Bank modal with pre-filled name input and Active/Inactive radio buttons; audit record written on each update
- **Navigation** — Banks nav item added to Admin sidebar; detail page has its own `BankDetailNavigation` shell ready for Slice 3 Trustees tab
- **Infrastructure** — `banks` Cosmos DB collection created with indexes on `documentType` and `name` to support ORDER BY; Snyk XXE vulnerability in transitive `fast-xml-builder` resolved

# Testing/Validation

Implemented using TDD throughout. 129 new tests across backend (40) and frontend (89) covering repository, use case, controller, Azure Function, API client, and all UI components.

Manual testing confirmed: bank list loads and sorts correctly, Add Bank modal validates and creates, bank detail page renders with correct layout, Edit Bank modal pre-fills and saves with audit trail.

# Notes

- Slice 3 (Trustees tab on bank detail) is deferred; `BankDetailNavigation` and `BankDetailOverview` are already separated as components to make that addition straightforward
- The bank detail page renders outside the Admin two-column layout (no sidebar nav) matching the Figma design
- Migration script for seeding banks from the legacy `lists` collection is at `.ustp-cams-fdp/ai/specs/CAMS-670_manage-bank-profiles/scripts/migrate-banks-from-lists.ts`

# Definition of Done:

## Summary by Sourcery

Add support for viewing and updating individual bank profiles, including a dedicated bank detail UI and corresponding backend endpoints.

New Features:
- Introduce a bank detail page with overview navigation, general information card, and an Edit Bank modal for administrators.
- Expose API client methods and Azure Function routing for fetching and updating a single bank profile by ID.

Enhancements:
- Extend the banks domain use case and Mongo repository to retrieve and update individual bank profiles with audited changes.
- Update the admin routing/layout so bank detail pages render outside the main admin navigation while linking from the banks list to each bank’s detail view.

Tests:
- Add backend tests for banks repository, use case, controller, and Azure Function covering get/update operations for single banks.
- Add frontend tests for admin routing, bank list linking, bank detail page behavior, and the Edit Bank modal including validation and error handling.


backend/lib/controllers/banks/banks.controller.ts
```diff
@@ -24,6 +24,38 @@ export class BanksController {
...
+    return httpSuccess({
+      statusCode: 200,
+      body: { meta: { self: context.request.url }, data: bank },
+    });
+  }
+
+  async handlePut(context: ApplicationContext): Promise<CamsHttpResponseInit<BankProfile>> {
+    this.requireSuperUser(context);
+    const { bankId } = context.request.params;
+    const body = context.request.body as { name?: string; status?: string } | null;
```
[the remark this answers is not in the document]
fmaddenflx: Fixed. Changed the inline cast from `{ name?: string; status?: string }` to `Partial<Pick<BankProfile, 'name' | 'status'>>`. This reuses the existing `BankProfile` type and gives `status` the `'active' | 'inactive'` union from the start, so after the runtime guard TypeScript narrows it correctly — no implicit cast needed.

user-interface/src/admin/banks/EditBankModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const trimmed = name.trim();
+      if (!trimmed) {
+        setNameError('Bank Name is required');
+        return;
+      }
+      setNameError(null);
+
+      try {
+        const response = await Api2.updateBank(bank.id, { name: trimmed, status });
+        const updated = (response as { data: BankProfile }).data;
```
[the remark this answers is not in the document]
fmaddenflx: Fixed. `Api2.updateBank` already returns `Promise<ResponseBody<BankProfile>>` and `ResponseBody<T>` has `data: T`, so `response.data` is typed as `BankProfile` directly — the cast was redundant. Changed to `response.data`.

user-interface/src/admin/banks/BankDetail.tsx
```diff
@@ -0,0 +1,95 @@
...
+    Api2.getBank(bankId)
+      .then((response) => {
+        setBank((response as { data: BankProfile }).data);
+        setLoadError(null);
+      })
+      .catch((error: Error) => {
+        setLoadError(error.message);
+      })
+      .finally(() => setIsLoaded(true));
+  }, [bankId]);
```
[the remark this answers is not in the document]
fmaddenflx: Fixed both issues here. Dropped the redundant `(response as { data: BankProfile }).data` cast — `Api2.getBank` returns `Promise<ResponseBody<BankProfile>>` so `response.data` is already `BankProfile`. Also added the `isCancelled` flag with a cleanup function so rapid `bankId` changes or unmounts bail out of the pending promise chain without triggering stale state updates.

mattstankey:
## Code Review — Issues 1–6

### High Priority

**1. Non-atomic `replaceOne` + `findOne` in `updateBank` repository**
`backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts` (lines 209–217)

The pattern of calling `replaceOne` followed by a separate `findOne` has a race condition: a concurrent update between those two calls means `findOne` could return a different state than what was just written, and the audit record's `after` value could reflect a different update than the one committed.

If `MongoCollectionAdapter` supports `findOneAndReplace` with `returnDocument: 'after'`, that would be the safer pattern. Otherwise this is worth a comment acknowledging the limitation.

---

**2. `bankId` param not validated in controller**
`backend/lib/controllers/banks/banks.controller.ts` (lines 313–343)

Both `handleGetOne` and `handlePut` read `bankId` from params without checking it's a non-empty string. The function handler uses `if (bankId)` to gate dispatch, but the controller doesn't re-validate. If a code path ever reaches these methods without a `bankId`, an `undefined` value propagates silently to the use case and repository. A guard consistent with how `name` and `status` are validated in `handlePut` would close this:

```ts
if (!bankId) throw new BadRequestError(MODULE_NAME, { message: 'Bank ID is required.' });
```

---

### Medium Priority

**3. Redundant `role="navigation"` on `<nav>`**
`user-interface/src/admin/banks/BankDetailNavigation.tsx` (line 8)

```tsx
<nav aria-label="Bank detail navigation" role="navigation">
```

The `<nav>` element carries the `navigation` landmark role implicitly. The explicit `role` attribute is redundant and can confuse screen reader tooling. Remove it.

---

**4. Submit button label repeats the modal heading**
`user-interface/src/admin/banks/EditBankModal.tsx` (line 1207)

The modal heading is "Edit Bank" and the action button also reads "Edit Bank." The button should describe what happens on click — "Save Changes" or "Update Bank" is clearer and more accessible.

---

**5. `DocumentTitle` always shows "Administration" after routing refactor**
`user-interface/src/admin/AdminScreen.tsx` (line 552)

`<DocumentTitle name="Administration" />` now sits outside the conditional branches, so it renders for the bank detail page too. Users navigating to `/admin/banks/:bankId` will see "Administration" in the browser tab rather than the bank name. `BankDetail` should set its own `DocumentTitle` once the bank loads.

---

**6. `getAppInsights()` call in error handler can throw**
`user-interface/src/admin/banks/EditBankModal.tsx` (line 1193)

```ts
getAppInsights().appInsights.trackException({ exception: error as Error });
```

If Application Insights isn't initialized (test or certain dev environments), this line throws and masks the original error. Optional chaining or a null-check would be safer:

```ts
getAppInsights()?.appInsights?.trackException({ exception: error as Error });
```

fmaddenflx:
Thanks Matt! Addressed in commit 2d4c0efa9:

**Comment 1** — Added a comment to `updateBank` in the repository acknowledging the non-atomic race condition between `replaceOne` and `findOne`. `findOneAndReplace` isn't currently available on `MongoCollectionAdapter`, so the comment serves as the documented acknowledgment.

**Comment 2** — Added `if (!bankId)` guards in both `handleGetOne` and `handlePut` that throw `BadRequestError`.

**Comment 3** — Removed the redundant `role="navigation"` from the `<nav>` in `BankDetailNavigation.tsx`.

**Comment 6** — Changed `getAppInsights().appInsights.trackException` to use optional chaining (`?.`) so it won't throw and mask the original error when AppInsights isn't initialized.

backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
```diff
@@ -64,6 +73,18 @@ export class BanksMongoRepository extends BaseMongoRepository implements BanksRe
     }
   }
 
+  async updateBank(id: string, bank: BankProfile): Promise<BankProfile> {
+    const query = doc('id').equals(id);
+    try {
+      // replaceOne + findOne is not atomic; a concurrent update between these
+      // two calls could cause the returned document to reflect a different write.
+      await this.getAdapter<BankProfile>().replaceOne(query, bank);
+      return await this.getAdapter<BankProfile>().findOne(query);
```
bityogi: **suggestion:** Consider eliminating the race condition entirely by returning the input instead of re-fetching.

The comment correctly notes that `replaceOne` + `findOne` is non-atomic. Since `bank` is the exact object that was just written (with all fields already set by the use case, including `updatedOn` and `updatedBy`), the `findOne` re-read adds latency and a race window with no benefit — unless there are Cosmos-generated fields (like `_ts` or `_etag`) that callers need.

If those server-side fields aren't needed downstream, this simplification removes the race entirely:

```suggestion
      await this.getAdapter<BankProfile>().replaceOne(query, bank);
      return bank;
```

user-interface/src/admin/banks/EditBankModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+        label: 'Update Bank',
+        onClick: handleSubmit,
+        closeOnClick: false,
+      },
+      cancelButton: {
+        label: 'Cancel',
+        onClick: handleCancel,
+      },
+    };
+
```
bityogi, line 87: **suggestion (ux):** Mixed terminology — heading says "Edit Bank" but the submit button (line 73) says "Update Bank".

Consider using consistent terminology throughout the modal. Options:
- Heading: "Edit Bank" + Button: "Save Changes" (common pattern: heading describes the action context, button describes what clicking does)
- Heading: "Update Bank" + Button: "Update Bank" (fully consistent)

The current mix of "Edit" and "Update" is minor but could confuse users about whether these are the same action.

backend/function-apps/api/banks/banks.function.ts
```diff
@@ -22,12 +22,17 @@ export default async function handler(
...
+      responseBody = await controller.handleGetOne(context);
+    } else if (method === 'GET') {
       responseBody = await controller.handleGet(context);
     } else if (method === 'POST') {
       responseBody = await controller.handlePost(context);
+    } else if (method === 'PUT' && bankId) {
+      responseBody = await controller.handlePut(context);
     } else {
       return { status: 405, jsonBody: 'Method Not Allowed' };
     }
```
jamesobrooks: We usually handle this logic in the controller, leaving the `*.function.ts` file to be an abstraction for Azure Functions. Also, we have `common/src/api/http-status-codes.ts` to manage response codes.

backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
```diff
@@ -64,6 +73,18 @@ export class BanksMongoRepository extends BaseMongoRepository implements BanksRe
     }
   }
 
+  async updateBank(id: string, bank: BankProfile): Promise<BankProfile> {
+    const query = doc('id').equals(id);
+    try {
+      // replaceOne + findOne is not atomic; a concurrent update between these
+      // two calls could cause the returned document to reflect a different write.
```
jamesobrooks: This seems unnecessary.

backend/lib/controllers/banks/banks.controller.test.ts
```diff
@@ -71,6 +71,78 @@ describe('BanksController', () => {
...
+    });
+
+    test('should throw BadRequestError when body is null', async () => {
+      context.request.params = { bankId: 'bank-1' };
+      context.request.body = null;
+
+      await expect(controller.handlePut(context)).rejects.toThrow(
+        expect.objectContaining({ status: 400 }),
+      );
+    });
```
jamesobrooks: These could all be in a `test.each()`.

backend/lib/use-cases/banks/banks.ts
```diff
@@ -17,6 +17,42 @@ export class BanksUseCase {
     return this.repository.getBanks();
   }
 
+  async getBank(id: string): Promise<BankProfile> {
+    return this.repository.getBank(id);
+  }
```
jamesobrooks, line 33: We usually catch and wrap errors here to ensure we're not exposing error messages from Cosmos (or elsewhere) and we set the message to something more meaningful to CAMS.

jamesobrooks: This would also apply to the rest of the file.

user-interface/src/admin/banks/EditBankModal.tsx
fmaddenflx: We just today went over this with Katie, and she recommended "Update Bank".  But I think you bring up a good point.  However, I think to be consistent with other modals for saving information in our app, "Update Bank" is probably better than "Save Changes".

user-interface/src/admin/banks/Banks.tsx
```diff
@@ -80,7 +81,9 @@ export function Banks() {
               <CamsTableBody>
                 {banks.map((bank) => (
                   <CamsTableRow key={bank.id}>
-                    <CamsTableCell>{bank.name}</CamsTableCell>
+                    <CamsTableCell>
+                      <Link to={`/admin/banks/${bank.id}`}>{bank.name}</Link>
```
jamesobrooks: I guess I didn't realize we don't have our own link component. This is missing the `usa-link` class and a title. Maybe we need to consider creating a component?

user-interface/src/admin/banks/EditBankModal.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+    });
+  });
+
+  test('should show validation error when name is cleared and submitted', async () => {
+    renderComponent();
+    openModal();
+    await waitFor(() => expect(screen.getByTestId(SUBMIT_BTN)).toBeVisible());
+
+    fireEvent.change(screen.getByLabelText(/bank name/i), { target: { value: '' } });
+    fireEvent.click(screen.getByTestId(SUBMIT_BTN));
```
jamesobrooks: Should we be using `userEvent` here instead?

backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
fmaddenflx: Fixed. Removed the  re-fetch and return  directly. Since the use case validates the bank exists before calling  and all fields are already set, the round-trip added latency with no benefit — and as you noted, introduced a race window.

user-interface/src/admin/banks/EditBankModal.tsx
fmaddenflx: Intentionally leaving as-is. The heading 'Edit Bank' describes the context of the modal, while 'Update Bank' describes the action of the submit button. We think this is clear enough and prefer not to change it.

backend/function-apps/api/banks/banks.function.ts
fmaddenflx: Fixed. Moved the method/bankId routing logic into BanksController.handleRequest() following the pattern used by CaseAssignmentController. The function now delegates to a single controller.handleRequest(context) call.

backend/lib/adapters/gateways/mongo/banks.mongo.repository.ts
fmaddenflx: Fixed. Removed the findOne re-fetch after insertOne in createBank. Since insertOne returns the generated id and the adapter stamps it onto the document before inserting, we can return { ...bank, id: newId } directly — no round-trip needed.

backend/lib/controllers/banks/banks.controller.test.ts
fmaddenflx: Fixed. Consolidated all validation/authorization failure cases for handlePut and handlePost into test.each() blocks.

backend/lib/use-cases/banks/banks.ts
fmaddenflx: Fixed. Added try/catch with getCamsError wrapping to all four BanksUseCase methods (getBanks, getBank, updateBank, createBank) with appropriate CAMS-facing error messages.

fmaddenflx: Done — applied to all methods in the file.

user-interface/src/admin/banks/Banks.tsx
fmaddenflx: Added className="usa-link" and a title attribute ("View details for {bank.name}") to the Link. Agreed a shared CamsLink component would be a good addition — filed as a follow-up to avoid scope creep here.

user-interface/src/admin/banks/EditBankModal.test.tsx
fmaddenflx: Fixed. Replaced all fireEvent calls throughout the file with userEvent via TestingUtilities.setupUserEvent(), consistent with the pattern in router.test.tsx and TransferOrderAccordion.test.tsx.

backend/lib/controllers/banks/banks.controller.ts
```diff
@@ -15,6 +15,26 @@ export class BanksController {
...
+    if (method === 'GET' && bankId) {
+      return this.handleGetOne(context);
+    } else if (method === 'GET') {
+      return this.handleGet(context);
+    } else if (method === 'POST') {
+      return this.handlePost(context);
+    } else if (method === 'PUT' && bankId) {
+      return this.handlePut(context);
+    }
+    return httpSuccess({ statusCode: 405 });
```
jamesobrooks: Looks like the agent missed the part of the comment that we usually use common/src/api/http-status-codes.ts to handle status codes. Below we use 200 also.

user-interface/src/admin/banks/BankDetail.tsx
```diff
@@ -0,0 +1,107 @@
...
+    };
+  }, [bankId]);
+
+  function handleBankUpdated(updated: BankProfile) {
+    setBank(updated);
+  }
+
+  return (
+    <div className="bank-detail" data-testid="bank-detail">
+      <Link to="/admin/banks" className="usa-link back-link">
```
jamesobrooks: It looks like I missed this link not having a title.

backend/lib/controllers/banks/banks.controller.ts
fmaddenflx: Fixed. Replaced all literal status codes in BanksController with HttpStatusCodes constants (OK, CREATED, METHOD_NOT_ALLOWED). Also added METHOD_NOT_ALLOWED: 405 to common/src/api/http-status-codes.ts since it was missing.

user-interface/src/admin/banks/BankDetail.tsx
fmaddenflx: Fixed. Added title="Back to Banks list" to the Back to Banks link in BankDetail.

bityogi (approved):

diabagatekelly (approved):

PR #2380, opened 2026-05-12
CAMS-669 Migrate bankruptcy software to dedicated entity (Slice 1)

fmaddenflx:
# Purpose

Replace the legacy bankruptcy software admin feature — which stored vendors as generic key/value entries in the shared `lists` collection — with a first-class `bankruptcy-software` entity following the same dedicated pattern established for banks in CAMS-670.

# Major Changes

- **Common types**: `BankruptcySoftwareProfile` and `BankruptcySoftwareAuditHistory` with `documentType` discriminators
- **Full backend stack**: dedicated MongoDB repository, use case, controller, and Azure Function at route `bankruptcy-software/`; all layers require `CamsRole.SuperUser`
- **Old lists endpoints removed**: `POST /lists/bankruptcy-software` and `DELETE /lists/bankruptcy-software/:id` removed from controller, use case, and repository; `GET` kept temporarily for migration verification
- **UI rewrite**: `BankruptcySoftware.tsx` replaced with a `CamsTable`-based list; new `AddSoftwareModal.tsx` using imperative ref pattern matching `AddBankModal`
- **Api2 updated**: old `getBankruptcySoftwareList`/`postBankruptcySoftware`/`deleteBankruptcySoftware` replaced with `getSoftwareList`/`createSoftware`
- **Feature flag**: Bankruptcy Software nav item gated behind `TRUSTEE_SOFTWARE_BANK_DISPLAY`
- **Data migration script**: `ops/migrations/CAMS-669-migrate-bankruptcy-software.ts` — idempotent, uses CAMS factory/repository pattern, writes audit records via `SYSTEM_USER_REFERENCE`

# Testing/Validation

Implemented using TDD. All unit tests pass across backend layers (repository, use case, controller, Azure Function) and frontend (component, modal, Api2). `TrusteeDetailScreen` updated to consume new API. No regressions to Banks, Privileged Identity, or Case Reload features.

# Notes

- Software name links (clickable rows → detail page) are intentionally plain text in this slice, matching the banks precedent. Links will be added in Slice 2 when the detail page exists.
- `GET /lists/bankruptcy-software` is left in place deliberately — it serves as a verification point after running the data migration script in each environment. It will be removed in a follow-up once production migration is confirmed.
- The migration script must be run before or at deploy time to copy existing vendor names from the `lists` collection into the new `bankruptcy-software` collection.

# Definition of Done:

## Summary by Sourcery

Introduce a dedicated bankruptcy software entity and API, replacing legacy list-based storage, and update the admin UI and trustee workflows to use the new endpoints with feature-flagged navigation and supporting migration tooling.

New Features:
- Add dedicated bankruptcy software domain types, repository, use case, controller, and Azure Function endpoint for managing software profiles with audit history.
- Introduce a new CamsTable-based admin UI and modal workflow for viewing and creating bankruptcy software entries via the new Api2 methods.
- Add a feature flag to control visibility of the Bankruptcy Software admin navigation entry and update trustee detail screens to consume the new software list API.

Enhancements:
- Refactor existing lists infrastructure and tests to remove create/delete bankruptcy software responsibilities now handled by the dedicated entity.
- Extend shared testing and mock API utilities to support the new bankruptcy software endpoints and data shapes.

Tests:
- Add comprehensive backend and frontend tests for the new bankruptcy software controller, repository, use case, Azure Function, admin UI, modal flow, and trustee software transformations.
- Add an idempotent migration script to populate the new bankruptcy software collection from legacy list entries while creating corresponding audit records.

### DIFF
--- a/backend/function-apps/api/bankruptcy-software/bankruptcy-software.function.ts
+++ b/backend/function-apps/api/bankruptcy-software/bankruptcy-software.function.ts
@@ -2,6 +2,7 @@ import { app, HttpRequest, HttpResponseInit, InvocationContext } from '@azure/fu
 import ContextCreator from '../../azure/application-context-creator';
 import { toAzureError, toAzureSuccess } from '../../azure/functions';
 import { BankruptcySoftwareController } from '../../../lib/controllers/bankruptcy-software/bankruptcy-software.controller';
+import { CamsHttpResponseInit } from '../../../lib/adapters/utils/http-response';
 
 const MODULE_NAME = 'BANKRUPTCY-SOFTWARE-FUNCTION';
 
@@ -20,19 +21,8 @@ export default async function handler(
 
     context.session = await ContextCreator.getApplicationContextSession(context);
     const controller = new BankruptcySoftwareController(context);
-
-    const method = request.method;
-    let responseBody;
-
-    if (method === 'GET') {
-      responseBody = await controller.handleGet(context);
-    } else if (method === 'POST') {
-      responseBody = await controller.handlePost(context);
-    } else {
-      return { status: 405, jsonBody: 'Method Not Allowed' };
-    }
-
-    return toAzureSuccess(responseBody);
+    const response = await controller.handleRequest(context);
+    return toAzureSuccess(response as CamsHttpResponseInit);
   } catch (error) {
     return toAzureError(logger, MODULE_NAME, error);
   }
--- a/backend/function-apps/api/bankruptcy-software/bankruptcy-software.function.ts
+++ b/backend/function-apps/api/bankruptcy-software/bankruptcy-software.function.ts
@@ -1,0 +1,46 @@
+import { app, HttpRequest, HttpResponseInit, InvocationContext } from '@azure/functions';
+import ContextCreator from '../../azure/application-context-creator';
+import { toAzureError, toAzureSuccess } from '../../azure/functions';
+import { BankruptcySoftwareController } from '../../../lib/controllers/bankruptcy-software/bankruptcy-software.controller';
+
+const MODULE_NAME = 'BANKRUPTCY-SOFTWARE-FUNCTION';
+
+export default async function handler(
+  request: HttpRequest,
+  invocationContext: InvocationContext,
+): Promise<HttpResponseInit> {
+  const logger = ContextCreator.getLogger(invocationContext);
+
+  try {
+    const context = await ContextCreator.applicationContextCreator({
+      invocationContext,
+      logger,
+      request,
+    });
+
+    context.session = await ContextCreator.getApplicationContextSession(context);
+    const controller = new BankruptcySoftwareController(context);
+
+    const method = request.method;
+    let responseBody;
+
+    if (method === 'GET') {
+      responseBody = await controller.handleGet(context);
+    } else if (method === 'POST') {
+      responseBody = await controller.handlePost(context);
+    } else {
+      return { status: 405, jsonBody: 'Method Not Allowed' };
+    }
+
+    return toAzureSuccess(responseBody);
+  } catch (error) {
+    return toAzureError(logger, MODULE_NAME, error);
+  }
+}
+
+app.http('bankruptcy-software', {
+  methods: ['GET', 'POST'],
+  authLevel: 'anonymous',
+  handler,
+  route: 'bankruptcy-software',
+});
--- a/backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.test.ts
+++ b/backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.test.ts
@@ -1,0 +1,122 @@
+import { vi } from 'vitest';
+import { ApplicationContext } from '../../adapters/types/basic';
+import { createMockApplicationContext } from '../../testing/testing-utilities';
+import { BankruptcySoftwareController } from './bankruptcy-software.controller';
+import { BankruptcySoftwareUseCase } from '../../use-cases/bankruptcy-software/bankruptcy-software';
+import { CamsRole } from '@common/cams/roles';
+import { BankruptcySoftwareProfile } from '@common/cams/bankruptcy-software';
+
+vi.mock('../../use-cases/bankruptcy-software/bankruptcy-software');
+
+describe('BankruptcySoftwareController', () => {
+  let context: ApplicationContext;
+  let controller: BankruptcySoftwareController;
+
+  const mockSoftware: BankruptcySoftwareProfile[] = [
+    {
+      id: 'sw-1',
+      documentType: 'BANKRUPTCY_SOFTWARE',
+      name: 'Axos',
+      status: 'active',
+      updatedOn: '2024-01-01T00:00:00.000Z',
+      updatedBy: { id: 'user-1', name: 'User One' },
+    },
+  ];
+
+  const createdSoftware: BankruptcySoftwareProfile = {
+    id: 'sw-new',
+    documentType: 'BANKRUPTCY_SOFTWARE',
+    name: 'New Software',
+    status: 'active',
+    updatedOn: '2024-01-01T00:00:00.000Z',
+    updatedBy: { id: 'user-1', name: 'User One' },
+    createdOn: '2024-01-01T00:00:00.000Z',
+    createdBy: { id: 'user-1', name: 'User One' },
+  };
+
+  beforeEach(async () => {
+    context = await createMockApplicationContext();
+    context.session.user.roles = [CamsRole.SuperUser];
+    controller = new BankruptcySoftwareController(context);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  describe('handleGet', () => {
+    test('should return 200 with software list for SuperUser', async () => {
+      vi.spyOn(BankruptcySoftwareUseCase.prototype, 'getSoftwareList').mockResolvedValue(
+        mockSoftware,
+      );
+
+      const result = await controller.handleGet(context);
+
+      expect(result.statusCode).toBe(200);
+      expect(result.body.data).toEqual(mockSoftware);
+    });
+
+    test('should throw ForbiddenError when user lacks SuperUser role', async () => {
+      context.session.user.roles = [CamsRole.TrialAttorney];
+
+      await expect(controller.handleGet(context)).rejects.toThrow(
+        expect.objectContaining({ status: 403 }),
+      );
+    });
+
+    test('should throw ForbiddenError when user has no roles', async () => {
+      context.session.user.roles = [];
+
+      await expect(controller.handleGet(context)).rejects.toThrow(
+        expect.objectContaining({ status: 403 }),
+      );
+    });
+  });
+
+  describe('handlePost', () => {
+    test('should return 201 with created software for SuperUser', async () => {
+      context.request.body = { name: 'New Software' };
+      vi.spyOn(BankruptcySoftwareUseCase.prototype, 'createSoftware').mockResolvedValue(
+        createdSoftware,
+      );
+
+      const result = await controller.handlePost(context);
+
+      expect(result.statusCode).toBe(201);
+      expect(result.body.data).toEqual(createdSoftware);
+    });
+
+    test('should throw ForbiddenError when user lacks SuperUser role', async () => {
+      context.session.user.roles = [CamsRole.TrialAttorney];
+      context.request.body = { name: 'New Software' };
+
+      await expect(controller.handlePost(context)).rejects.toThrow(
+        expect.objectContaining({ status: 403 }),
+      );
+    });
+
+    test('should throw BadRequestError when name is missing', async () => {
+      context.request.body = {};
+
+      await expect(controller.handlePost(context)).rejects.toThrow(
+        expect.objectContaining({ status: 400 }),
+      );
+    });
+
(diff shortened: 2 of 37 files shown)


backend/function-apps/api/bankruptcy-software/bankruptcy-software.function.ts
```diff
@@ -0,0 +1,46 @@
...
+    const method = request.method;
+    let responseBody;
+
+    if (method === 'GET') {
+      responseBody = await controller.handleGet(context);
+    } else if (method === 'POST') {
+      responseBody = await controller.handlePost(context);
+    } else {
+      return { status: 405, jsonBody: 'Method Not Allowed' };
+    }
```
jamesobrooks: Similar comment to one in the CAMS-670 PR, we typically keep this in the controller and use constants for http status codes.

common/src/index.ts
```diff
@@ -9,7 +9,9 @@ export * from './api/search';
 export * from './cams/actions';
 export * from './cams/assignments';
 export * from './cams/auditable';
+export * from './cams/bankruptcy-software';
 export * from './cams/banks';
+export * from './cams/bankruptcy-software';
```
mattstankey: Duplicate export — `bankruptcy-software` is exported twice here (also appears three lines above). This will cause a TypeScript duplicate identifier error. Remove one of the two lines.

user-interface/src/admin/AdminScreen.test.tsx
```diff
@@ -76,12 +81,8 @@ describe('Admin screen tests', () => {
   });
```
mattstankey, line 84: The original assertion verified that the nav link has `usa-current` when the path matches — a real behavioral contract. The replacement (`getByTestId("admin-screen") toBeInTheDocument()`) only checks the screen renders at all, which is much weaker. If `testFeatureFlags` includes `PRIVILEGED_IDENTITY_MANAGEMENT: true`, the original assertion should be recoverable without the spy.

mattstankey (commented):
Two issues worth flagging:

**1. Duplicate export in `common/src/index.ts` (line 13)** — `bankruptcy-software` is exported twice, which will cause TypeScript duplicate-identifier errors at compile time. One of the two lines needs to be removed.

**2. Weakened test in `user-interface/src/admin/AdminScreen.test.tsx`** — The `"should select privileged-identity nav when path matches"` test was updated to just verify the screen renders (`getByTestId("admin-screen")`), dropping the assertion that the nav link has the `usa-current` class when the path matches. That assertion was verifying a real behavioral contract (active nav state on route match). If `testFeatureFlags` includes `PRIVILEGED_IDENTITY_MANAGEMENT: true`, the original assertion should still be possible and worth preserving.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.test.ts
```diff
@@ -0,0 +1,122 @@
+import { vi } from 'vitest';
+import { ApplicationContext } from '../../adapters/types/basic';
+import { createMockApplicationContext } from '../../testing/testing-utilities';
+import { BankruptcySoftwareController } from './bankruptcy-software.controller';
+import { BankruptcySoftwareUseCase } from '../../use-cases/bankruptcy-software/bankruptcy-software';
+import { CamsRole } from '@common/cams/roles';
+import { BankruptcySoftwareProfile } from '@common/cams/bankruptcy-software';
+
+vi.mock('../../use-cases/bankruptcy-software/bankruptcy-software');
```
jamesobrooks: I don't think this is doing anything. We seem to be properly using `vi.spyOn()` within the actual tests.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.useCase = new BankruptcySoftwareUseCase(context);
+  }
+
+  async handleGet(
+    context: ApplicationContext,
+  ): Promise<CamsHttpResponseInit<BankruptcySoftwareProfile[]>> {
+    this.requireSuperUser(context);
+    const softwareList = await this.useCase.getSoftwareList();
+    return httpSuccess({
+      statusCode: 200,
```
jamesobrooks: We should be using common/src/api/http-status-codes.ts here and below as well.

backend/function-apps/api/bankruptcy-software/bankruptcy-software.function.ts
fmaddenflx: Fixed. Moved routing dispatch into BankruptcySoftwareController.handleRequest() following the same pattern. The function now calls controller.handleRequest(context) directly.

common/src/index.ts
fmaddenflx: Fixed. Removed the duplicate export of ./cams/bankruptcy-software from common/src/index.ts.

user-interface/src/admin/AdminScreen.test.tsx
fmaddenflx: Fixed. Restored the usa-current assertion on the privileged-identity nav link.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.test.ts
fmaddenflx: Fixed. Removed the vi.mock() call — spyOn() in the tests is sufficient and the mock was redundant.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
fmaddenflx: Fixed. Added METHOD_NOT_ALLOWED (405) to common/src/api/http-status-codes.ts and replaced all literal status codes in the controller with HttpStatusCodes constants.

bityogi (approved):

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+  > {
+    const { method } = context.request;
+    if (method === 'GET') {
+      return this.handleGet(context);
+    } else if (method === 'POST') {
+      return this.handlePost(context);
+    }
+    return httpSuccess({ statusCode: HttpStatusCodes.METHOD_NOT_ALLOWED }) as CamsHttpResponseInit;
+  }
+
```
bityogi, line 34: **issue (high):** `requireSuperUser` on the GET endpoint will break `TrusteeDetailScreen` for non-admin users.

`TrusteeDetailScreen` calls `Api2.getSoftwareList()` to populate the bankruptcy software dropdown, and it's accessible to non-SuperUser roles (Trial Attorneys, etc.). The old `/lists/bankruptcy-software` GET had no role check.

With this change, non-admin staff viewing trustee details will get a 403 and see an empty software combo box — a functional regression.

**Suggestion:** Either remove the role check on `handleGet` (read-only access is safe for any authenticated user), or introduce a separate unauthenticated read endpoint for the software list.

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -0,0 +1,48 @@
...
+export class BankruptcySoftwareUseCase {
+  private readonly repository;
+  private readonly context: ApplicationContext;
+
+  constructor(context: ApplicationContext) {
+    this.context = context;
+    this.repository = factory.getBankruptcySoftwareRepository(context);
+  }
+
+  async getSoftwareList(): Promise<BankruptcySoftwareProfile[]> {
```
bityogi, line 19: **suggestion:** Both `getSoftwareList()` and `createSoftware()` call the repository directly without try/catch + `getCamsError` wrapping.

The banks use case (the model this follows) wraps every method:

```ts
async getBanks(): Promise<BankProfile[]> {
  try {
    return await this.repository.getBanks();
  } catch (originalError) {
    throw getCamsError(originalError, MODULE_NAME, 'Unable to retrieve banks.');
  }
}
```

Without this, raw Cosmos/Mongo errors will propagate undecorated to the controller layer without CAMS-friendly messages or proper module attribution.

user-interface/src/admin/bankruptcy-software/AddSoftwareModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      try {
+        const response = await Api2.createSoftware({ name: trimmed });
+        onSuccess(response!.data);
+        modalRef.current?.hide();
+        alert?.success('Bankruptcy software added successfully.');
+      } catch (error) {
+        getAppInsights().appInsights.trackException({ exception: error as Error });
+        alert?.error('Failed to add bankruptcy software. Please try again.');
+      }
+    }
```
bityogi, line 60: **suggestion (bug_risk):** Two issues here:

1. `response!.data` uses a non-null assertion, but `api().post()` can return `void` (when response body is empty). This would throw a runtime error. Add a null check or type the `createSoftware` wrapper with a generic type parameter (`api().post<BankruptcySoftwareProfile>(...)`) to guarantee a typed response.

2. `getAppInsights().appInsights.trackException(...)` in the catch block (line 63) should use optional chaining (`getAppInsights()?.appInsights?.trackException(...)`) — same issue that was fixed in PR #2373 for EditBankModal. Without it, if AppInsights isn't initialized the error handler itself throws and masks the original error.

ops/migrations/CAMS-669-migrate-bankruptcy-software.ts
```diff
@@ -0,0 +1,118 @@
...
+    } catch (err) {
+      console.error(`  ✗ Failed:   ${name} — ${(err as Error).message}`);
+      fail++;
+    }
+  }
+
+  console.log(`\nDone. ${success} migrated, ${fail} failed.`);
+}
+
+run().catch(console.error).finally(() => process.exit(0));
```
bityogi: **suggestion:** The script always exits with code 0 even on total failure:

```ts
run().catch(console.error).finally(() => process.exit(0));
```

If `run()` throws (e.g., cannot connect to the database), it logs the error but reports success to the calling process. For operational scripts this makes failure invisible to CI/CD or manual verification.

**Suggested fix:**
```ts
run()
  .then(() => process.exit(0))
  .catch((err) => {
    console.error(err);
    process.exit(1);
  });
```

user-interface/src/admin/AdminScreen.tsx
```diff
@@ -71,7 +71,7 @@ export function AdminScreen() {
                     <Routes>
                       <Route path="privileged-identity" element={<PrivilegedIdentity />} />
                       <Route path="banks" element={<Banks />} />
-                      <Route path="bankruptcy-software" element={<BankruptcySoftware />} />
+                      <Route path="bankruptcy-software/*" element={<BankruptcySoftware />} />
```
bityogi: **nit:** The `/*` wildcard on this route implies nested routing, but `BankruptcySoftware` doesn't use `<Routes>` or `<Outlet>` internally. The `banks` route doesn't use a wildcard either.

Not a bug — just slightly misleading. If this is forward-looking for a future detail page, a brief comment would help; otherwise plain `"bankruptcy-software"` matches the pattern of the other routes.

bityogi (commented):
Code review — 5 suggestions (1 high priority)

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
fmaddenflx: WOW!  Good catch!

fmaddenflx: Fixed. Removed the  call from  — the software list is read-only and must be accessible to all authenticated users, including non-admin roles like Trial Attorneys who need it to populate the dropdown on TrusteeDetailScreen. Updated the controller tests to reflect open GET access.

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
fmaddenflx: Fixed. Wrapped both `getSoftwareList()` and `createSoftware()` in try/catch with `getCamsError` following the banks use case pattern. Also added `MODULE_NAME = 'BANKRUPTCY-SOFTWARE-USE-CASE'` for proper error attribution.

user-interface/src/admin/bankruptcy-software/AddSoftwareModal.tsx
fmaddenflx: Fixed both issues. Typed `createSoftware` with `post<BankruptcySoftwareProfile, { name: string }>` so the response is guaranteed typed and the non-null assertion is safe. Added optional chaining on `getAppInsights()?.appInsights?.trackException()` in the catch block.

ops/migrations/CAMS-669-migrate-bankruptcy-software.ts
fmaddenflx: Fixed. Changed to `.then(() => process.exit(0)).catch(err => { console.error(err); process.exit(1); })` so failures are visible to the calling process.

user-interface/src/admin/AdminScreen.tsx
fmaddenflx: Fixed. Removed the wildcard — `BankruptcySoftware` doesn't use nested routing so `path="bankruptcy-software"` matches the pattern of the other routes.

PR #2382, opened 2026-05-12
Complete ComboBox accessibility refactor

diabagatekelly:
# Purpose

Replace old ComboBox implementation with improved ComboBoxAlt version, keeping the ComboBox file name. This completes the accessibility improvements for screen reader support.

Key changes:
- Replace old ComboBox with ComboBoxAlt implementation, keeping ComboBox.tsx file name for compatibility
- Fix race condition where getSelections() returned stale data when called from onUpdateSelection callback by adding selectedMapRef
- Add aria-live wrapper to PillBox to prevent screen reader announcements on selection changes
- Remove external clear buttons from trustee filters, use internal ComboBox clear button with absolute positioning
- Prevent Enter key from submitting forms when expanding ComboBox
- Update all test selectors to match new ComboBox structure
- Add UTM networking documentation for screen reader testing

Jira ticket: N/A

# Major Changes - files updated with new ComboBox

- 3 in Admin -> Privileged Identity
- 1 in Admin -> Reload Case
- 1 in Case Detail Screen -> Court Docket filter
- 2 in Data Verification Screen
- 1 in Data Verification -> Consolidation -> Add Case modal
- 1 in Data Verification -> Transfer -> Select New Case
- 2 in Data Verification -> Trustee Match -> Trustee Search modal
- 2 in Case Search Screen
- 1 in Staff Assignment Screen
- 1 in Trustee Details Screen -> Assigned Staff -> Trustee Oversight Add/Edit modal
- 2 in Trustee Screen filters
- 6 in Trustee Appointments form
- 1 in Trustee Other Info form (ie. Software and Bank form)

# Testing/Validation

# Notes - suggested screen reader follow-up tickets

1. **Modals** - Anytime there's any change in dropdown selections, the screen reads re-reads the entire modal
2. **Data verification screen** - When a new task type or task status is selected, the screen reader reads the content of all the re-rendered accordions
3. **Trustee appointments form** - When selecting all the divisions available for the selected district, we go from displaying pills for all individual divisions to a single "All Divisions" pill. This causes the screen reader to make some confusing announcements.

# Definition of Done:


mattstankey:
**Issue 1: `setupUserEvent()` called per-toggle instead of shared instance**

`toggleDropdown()` creates a new `userEvent` instance on every call:

```ts
async function toggleDropdown(id: string = comboboxId) {
  await TestingUtilities.setupUserEvent().click(toggleButton!);
}
```

`@testing-library/user-event` is designed to share a single instance across a test so it can maintain internal state (pointer position, modifier key state, etc.). Creating a fresh instance per call is wasteful and can cause subtle interaction issues in tests that call `toggleDropdown()` multiple times.

**Fix:** Use the module-level `userEvent` variable that's already set up in `beforeEach`:

```ts
async function toggleDropdown(id: string = comboboxId) {
  const toggleButton = document.querySelector(`#${id}-expand`);
  expect(toggleButton).toBeInTheDocument();
  await userEvent.click(toggleButton!);
}
```

mattstankey:
**Issue 2: `aria-describedby` test asserts less than its name claims**

The test `'should reference both filter description and hint via aria-describedby when hint provided'` only checks for one of the two expected IDs:

```ts
test('should reference both filter description and hint via aria-describedby when hint provided', () => {
  renderWithProps({ ariaDescription: 'Custom hint text' });
  const input = screen.getByRole('combobox');
  expect(input).toHaveAttribute(
    'aria-describedby',
    `${comboboxId}-filter-input-aria-description`,  // only checking one ID
  );
});
```

The test name says "both filter description and hint" but the assertion doesn't verify the hint element ID is also present. This means the test passes whether or not the hint is correctly wired into `aria-describedby` — potentially masking a real accessibility gap.

**Fix:** Either update the assertion to check both IDs are present, or rename the test to accurately reflect what it actually verifies. If the implementation does include both IDs, the assertion should look something like:

```ts
expect(input.getAttribute('aria-describedby')).toContain(`${comboboxId}-filter-input-aria-description`);
expect(input.getAttribute('aria-describedby')).toContain(`${comboboxId}-hint`);
```

mattstankey:
**Issue 3: Three known follow-up accessibility problems have no tracking issues**

The PR Notes section documents three real screen reader regressions or rough edges:

1. **Modals** — any dropdown selection change causes the screen reader to re-read the entire modal
2. **Data verification screen** — selecting a task type/status causes the screen reader to read all re-rendered accordion content
3. **Trustee appointments form** — switching from individual division pills to the single "All Divisions" pill produces confusing screen reader announcements

These are observable accessibility problems for users of assistive technology. Documenting them in the PR description is a good start, but PR descriptions disappear from the team's active awareness once the PR merges. Without issue tickets, there's no mechanism to prioritize, assign, or track resolution.

**Fix:** Create a backlog issue for each of the three items before merging so they can be properly triaged and scheduled.

mattstankey:
**Issue 4: `onClose` callback coverage is hard to find after reorganization**

The old test suite had an explicit test for the `onClose` callback directly in the toggle behavior test. After reorganization, `onClose` coverage exists but is buried in the `Edge Cases` describe block (`'should call onClose with current selections when dropdown closes'`), which is not where a reviewer or future developer would look for core toggle behavior.

This is a low-severity structural issue, but `onClose` is an important contract — callers depend on it to react to dropdown dismissal — and its placement makes it easy to overlook when verifying toggle behavior or when writing future toggle-related tests.

**Fix:** Move the `onClose` test into the `Toggle & Dropdown Behavior` describe block where it logically belongs alongside the other open/close tests.

diabagatekelly:
✅ **Issues 1, 2, and 4 have been addressed** (commit 159d1f7a9)

**Issue 1:** Moved `toggleDropdown()` helper inside the `describe` block so it can access the shared `userEvent` instance from `beforeEach`.

**Issue 2:** Renamed test from "should reference both filter description and hint via aria-describedby when hint provided" to "should reference filter description via aria-describedby when hint provided" to accurately reflect what it tests.

**Issue 4:** Moved the `onClose` callback test from the "Edge Cases" section to the "Toggle & Dropdown Behavior" section where it logically belongs.

All tests passing (132/132).

**Issue 3** (accessibility tracking issues) will need separate discussion about whether to create tracking tickets for the three screen reader issues documented in the PR notes.

mattstankey:
## Code Review Notes

### 1. Logic bug in `isOutsideClick` — closes dropdown when no selections exist

In `ComboBox.tsx`, the `isOutsideClick` condition includes `selectedMap.size === 0` as an OR branch alongside the out-of-bounds checks:

```typescript
if (
  targetX < boundingRect.x ||
  targetX > containerRight ||
  targetY < boundingRect.y ||
  targetY > containerBottom ||
  selectedMap.size === 0   // <-- this closes dropdown on any click when empty
) {
  closeDropdown(false);
}
```

This means clicking **inside** the combobox will still close the dropdown if there are no selections. A user with an empty combobox who opens the dropdown and clicks the input field will immediately close it. The `selectedMap.size === 0` condition should be removed from this guard.

---

### 2. Redundant logic in the `Enter`/`Space` handler in `handleKeyDown`

The two branches in the new code both do the same thing:

```typescript
if (event.target instanceof HTMLInputElement && !event.target.classList.contains('combo-box-input')) {
  if (option) { handleDropdownItemSelection(option); }
  event.preventDefault();
} else if (!(event.target instanceof HTMLInputElement)) {
  if (option) { handleDropdownItemSelection(option); }
  event.preventDefault();
}
```

Both branches execute identically. The combined condition simplifies to `!event.target.classList.contains('combo-box-input')`, which is what the original code did. The extra `instanceof` check adds noise without changing behavior and makes the intent harder to see.

---

### 3. `console.warn` in production code paths

Several new defensive guards log to the console:

```typescript
console.warn('List ref not available for navigation');
console.warn('Invalid target index for list navigation');
console.warn(`Input container not found for combobox: ${comboBoxId}`);
console.warn('Failed to scroll to selected element', error);
```

If these guard impossible states, they should be silent. If they represent real recoverable errors, they should route through the project's Application Insights telemetry rather than logging raw to the console. Raw `console.warn` in production is noise.

---

### 4. `handleClearAllClick` focuses the container div instead of the input

After clearing, focus is sent to `containerRef` via `focusCombobox()`:

```typescript
function handleClearAllClick() {
  clearSelections();
  clearFilter();
  focusCombobox();  // focuses containerRef (a div)
  ...
}
```

But `containerRef` no longer has `tabIndex` or `role="combobox"` — those were moved to the `<input>` element. All other close/collapse paths call `focusInput()`. This leaves keyboard users stranded after clearing. `focusCombobox()` here should be `focusInput()`.

---

### 5. `toggleDropdownByKeystroke` creates two separate `userEvent` instances

```typescript
async function toggleDropdownByKeystroke() {
  const button1 = document.querySelector(`.button1`) as HTMLButtonElement;
  button1!.focus();
  await TestingUtilities.setupUserEvent().keyboard('{Tab}');    // new instance
  await TestingUtilities.setupUserEvent().keyboard('{ArrowDown}');  // new instance
}
```

Each `setupUserEvent()` call likely creates a fresh instance with independent pointer/keyboard state. Two separate keystrokes that depend on sequential state (Tab then ArrowDown) should use the same instance. This can cause intermittent failures depending on how the testing library tracks event sequences between calls.

cruzjone-flexion (approved):

mattstankey (approved):
Updates look good barring issues #3 followup.

PR #2398, opened 2026-05-15
CAMS-616 trustee appointment downstream

btposey:
# Purpose

Provide trustee appointments and staff assignments (CAMS-362) to downstream systems

## Summary by Sourcery

Add downstream integration for trustee appointments and staff assignments, emitting enriched events from CAMS to new Azure Functions that write into ACMS replica tables for use by existing downstream consumers.

New Features:
- Emit trustee appointment downstream events from SyncTrusteeAppointments when ACMS professional IDs are available, including open and soft-close events.
- Emit case assignment downstream events (behind a feature flag) carrying ACMS professional IDs (currently null) for staff assignments.
- Introduce a new downstream workspace with Azure Functions that transform CAMS events into CMMAP_CAMS rows and expose a unified CMMAP_ALL view for downstream reporting.
- Add an end-to-end integration harness and local container stack to exercise the trustee appointment downstream flow through to the ACMS replica SQL schema.

Enhancements:
- Harden trustee matching logic by covering previously untested branches for imperfect, multiple, and fuzzy match scenarios and re-verification tracking.
- Refine the API-to-dataflows gateway and dataflow event types to distinguish downstream event payloads and support trustee appointment queuing.
- Centralize Azure queue helper functions in the common library and share them between backend dataflows and downstream handlers.
- Adjust backend test configuration and case-assignment tests to use new repositories and downstream event shapes with feature-flagged emission.

Build:
- Add a new downstream workspace with its own TypeScript, Vitest, ESLint, and esbuild configuration, and wire it into root build and coverage scripts.

Tests:
- Add unit tests for new downstream gateway methods and queue helpers, extended trustee appointment sync tests, and case-assignment downstream emission behavior.
- Add SQL-based integration tests and seed data to validate the CMMAP_ALL view behavior and ACMS/CAMS merge logic, plus a TypeScript harness for running end-to-end integration flows.


backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -140,6 +136,18 @@ async function applyResolvedTrustee(
...
+  const professionalIds = await professionalIdsRepo.findByCamsTrusteeId(trusteeId);
+  const acmsProfessionalId = professionalIds[0]?.acmsProfessionalId ?? null;
+  const canEmitDownstream = !!acmsProfessionalId;
+  if (!canEmitDownstream) {
+    context.logger.warn(
+      MODULE_NAME,
+      `No acmsProfessionalId found for trustee ${trusteeId} on case ${event.caseId} — skipping downstream event`,
+    );
+  }
+  const apiToDataflows = factory.getApiToDataflowsGateway(context);
```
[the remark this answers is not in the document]
btposey: Fixed. The close event now performs a separate `findByCamsTrusteeId` lookup using `existingAppointment.trusteeId` (the old trustee being replaced) rather than reusing the new trustee's already-resolved ACMS ID. The new appointment's ID lookup follows in a separate block after the close is handled.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -0,0 +1,426 @@
...
+  queueName: buildQueueName(STAFF_MODULE_NAME, 'DLQ'),
+  connection: 'DataflowsStorage',
+});
+
+export function extractLastName(fullName: string): string {
+  const parts = fullName.trim().split(/\s+/);
+  return parts[parts.length - 1].toUpperCase();
+}
+
+export function transformStaffAssignmentToRow(event: CaseAssignmentDownstreamEvent): CmmapCamsRow {
```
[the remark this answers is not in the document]
btposey: Done. Extracted `buildBaseCmmapRow` (shared invariant fields for both appointment types — case/professional ID parsing, CDB/UPDATE dates, SOURCE, CAMS identifiers) and `handleQueueEvent` (try/catch, structured logging, DLQ routing). Both transform functions now spread `buildBaseCmmapRow` and only specify the fields unique to their appointment type. Both handlers delegate their error boundary to `handleQueueEvent`.

btposey (commented):
## Tech Lead Review — Remaining Issues

### 1. Stale comment on `handleQueueEvent`

**File:** `downstream/acms-cams-transition/acms-cams-transition.ts`, line 261

```ts
// Shared handler wrapper: skip on missing acmsProfessionalId, DLQ on unexpected errors
```

This comment is no longer accurate. Neither handler skips on a missing `acmsProfessionalId` — `buildBaseCmmapRow` now writes sentinel values instead. Please update the comment to reflect the actual behavior (e.g., _"writes sentinel values when `acmsProfessionalId` is absent; routes to DLQ on unexpected errors"_).

---

### 2. Trustee validation guard missing `chapter`

**File:** `downstream/acms-cams-transition/acms-cams-transition.ts`, line 413

```ts
if (!appointmentEvent.caseId || !appointmentEvent.trusteeId || !appointmentEvent.assignedOn) {
```

`chapter` is used unconditionally in the row transform and there is no sentinel fallback for it (unlike `acmsProfessionalId`). A malformed event missing `chapter` will produce a confusing generic error deep inside the transform. Please add `!appointmentEvent.chapter` to the guard so the explicit "missing required fields" error fires instead.

---

### 3. No sentinel-path test for `transformTrusteeAppointmentToRow`

**File:** `downstream/acms-cams-transition/acms-cams-transition.test.ts`

There is a test covering the sentinel path (`acmsProfessionalId: null`) for `transformStaffAssignmentToRow` (line 343), but no equivalent for `transformTrusteeAppointmentToRow`. Since `TrusteeAppointmentDownstreamEvent.acmsProfessionalId` is typed as non-nullable `string`, this path can only be reached via `buildBaseCmmapRow` directly — but it's still untested behavior from the trustee handler's perspective. Please add a test that passes `acmsProfessionalId: null` through `buildBaseCmmapRow` directly, or confirm the trustee sentinel path is intentionally excluded and document why.

---

### 4. Test mocks re-implement `buildQueueName` logic

**File:** `downstream/acms-cams-transition/acms-cams-transition.test.ts`, line ~37–41 (and the equivalent in any remaining separate handler test files)

```ts
vi.mock('@common/queues', () => ({
  buildQueueName: vi.fn((name: string, suffix?: string) =>
    suffix ? `${name}-${suffix}`.toLowerCase() : name.toLowerCase(),
  ),
}));
```

The mock re-implements the production function's logic. If `buildQueueName` semantics ever change (e.g., a separator change), these tests will pass against their own mock rather than catching the divergence. Since `buildQueueName` has no side effects, prefer `vi.importActual('@common/queues')` so the tests exercise the real implementation.

btposey (commented):
## Addressed

1. Stale comment on handleQueueEvent — Updated in 0126dda. Comment now reads: 'Shared handler wrapper: structured logging, DLQ routing on unexpected errors'.

2. Trustee validation guard missing chapter — Fixed in 0126dda. chapter added to the required-field guard alongside caseId, trusteeId, and assignedOn.

3. No sentinel-path test for transformTrusteeAppointmentToRow — Test added in 0126dda. Uses null as unknown as string to force the runtime null path and asserts PROF_CODE === SENTINEL_PROF_CODE and GROUP_DESIGNATOR === SENTINEL_GROUP_DESIGNATOR.

4. Test mocks re-implement buildQueueName logic — No action needed. @common/queues is not mocked in the test file. The vitest resolve.alias config maps @common to the real common workspace source, so tests already exercise the production buildQueueName implementation.

downstream/README.md
```diff
@@ -0,0 +1,144 @@
...
+```bash
+sqlcmd -S your-server -d ACMS_REP_SUB \
+  -i downstream/database/acms-cams-transition/migrations/001-initial-schema.sql
+```
+
+### Function App
+- Deployed as an Azure Functions app (Node.js v4 programming model)
+- Triggered by Azure Storage Queue messages
+- Dead-letter queue (DLQ) outputs for failed messages
+
```
btposey: I think we should not include this here. We will be adding deployment logic as bicep and github dowkrflow amendments.

backend/function-apps/dataflows/downstream/README.md
```diff
@@ -0,0 +1,144 @@
...
+      └─> Azure Storage Queue
+          └─> Azure Function Handler (downstream/)
+              └─> Write to CMMAP_CAMS table (ACMS_REP_SUB Azure SQL)
+                  └─> CMMAP_ALL View (UNION)
+                      ├─> SELECT * FROM CMMAP_CAMS (CAMS data)
+                      └─> UNION ALL
+                          └─> SELECT * FROM CMMAP WHERE no active CAMS row exists
+                              └─> BOBJ reads CMMAP_ALL view
+```
+
```
btposey, line 18: Not necessary here.

downstream/README.md
```diff
@@ -0,0 +1,144 @@
...
+                  └─> CMMAP_ALL View (UNION)
+                      ├─> SELECT * FROM CMMAP_CAMS (CAMS data)
+                      └─> UNION ALL
+                          └─> SELECT * FROM CMMAP WHERE no active CAMS row exists
+                              └─> BOBJ reads CMMAP_ALL view
+```
+
+## Purpose
+
+**Problem:** BOBJ reports query ACMS tables directly and cannot be easily modified. CAMS is replacing ACMS functionality incrementally.
```
btposey: I generally like how this is phrased, but I would remove the specific BOBJ reference and replace it with a more generic "downstream systems". Don't qualify that the downstream systems "cannot be easily modified".

downstream/database/acms-cams-transition/README.md
```diff
@@ -0,0 +1,104 @@
+# ACMS-CAMS Transition Database
```
btposey: We should probably delete this README and let the implementation speak for itself per the mantra "documentation lies".

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -157,11 +186,21 @@ async function applyResolvedTrustee(
...
+    const openEvent: TrusteeAppointmentDownstreamEvent = {
+      caseId: event.caseId,
+      trusteeId,
+      acmsProfessionalId,
+      assignedOn: now,
+      appointedDate: event.appointedDate,
+      chapter: syncedCase.chapter,
+    };
+    await apiToDataflows.queueTrusteeAppointmentEvent(openEvent);
+  }
```
btposey: Events should be published regardless of whether a professional ID is available. Not all trustees or staff have professional IDs.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -149,6 +148,36 @@ async function applyResolvedTrustee(
...
+
+  const newProfessionalIds = await professionalIdsRepo.findByCamsTrusteeId(trusteeId);
+  const acmsProfessionalId = newProfessionalIds[0]?.acmsProfessionalId ?? null;
+  if (!acmsProfessionalId) {
+    context.logger.warn(
+      MODULE_NAME,
+      `No acmsProfessionalId found for trustee ${trusteeId} on case ${event.caseId} — skipping downstream event`,
+    );
   }
```
btposey: No need to write a warning if a professional ID is not available. We are not skipping sending the event.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -149,6 +148,36 @@ async function applyResolvedTrustee(
...
+        chapter: syncedCase.chapter,
+        unassignedOn: now,
+      };
+      await apiToDataflows.queueTrusteeAppointmentEvent(closeEvent);
+    } else {
+      context.logger.warn(
+        MODULE_NAME,
+        `No acmsProfessionalId found for old trustee ${existingAppointment.trusteeId} on case ${event.caseId} — skipping close downstream event`,
+      );
+    }
```
btposey: This relates to the same issue around the professional ID. We should not suppress events because we can't find the professional ID. We'll need to review the CMMAP table in staging to get a better idea of how to address missing professional IDs when they appear.

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,433 @@
...
+  };
+}
+
+// Convert ISO date string to ACMS positional integer date format (YYYYMMDD)
+export function toAcmsDateNumeric(isoDateString: string): number {
+  const datePortion = isoDateString.split('T')[0];
+  return parseInt(datePortion.replace(/-/g, ''), 10);
+}
+
+// Parse "{GROUP_DESIGNATOR}-{PROF_CODE}" (e.g. "NY-00063") into components
```
btposey: Use JSDocs, not comment lines. Do this for all functions this comment line pattern has been applied to in this file.

backend/lib/use-cases/case-assignment/case-assignment.ts
```diff
@@ -199,8 +199,13 @@ export class CaseAssignmentUseCase {
     history.updatedOn = currentDate;
     await casesRepo.createCaseHistory(history);
 
-    for (const assignment of [...addedAssignments, ...removedAssignments]) {
-      await this.apiToDataflowsGateway.queueCaseAssignmentEvent(assignment);
+    if (context.featureFlags['downstream-staff-assignments-enabled']) {
+      for (const assignment of [...addedAssignments, ...removedAssignments]) {
+        await this.apiToDataflowsGateway.queueCaseAssignmentEvent({
+          ...assignment,
+          acmsProfessionalId: null, // Staff attorney ACMS IDs not yet available; tracked in CAMS-362
```
btposey: Get rid of the comment on this line. It keeps poisoning Claude Code contexts.

backend/function-apps/dataflows/dataflows-common.ts
```diff
@@ -37,31 +37,8 @@ export function isAuthorized(request: HttpRequest) {
...
- * by using the MODULE_NAME as a name space and abides by naming requirements.
- *
- * @param parts
- * @returns
- */
-export function buildQueueName(...parts): string {
-  return buildFunctionName(...parts).toLowerCase();
-}
+// Re-export shared queue helpers from common
+export { buildFunctionName, buildQueueName } from '@common/queues';
```
btposey: And dependent that imports from here should just import from @common/queues so we don't cause so much indirection.

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,433 @@
...
+export function parseProfessionalId(acmsProfessionalId: string): { group: string; code: number } {
```
jalba-flexion: **`parseProfessionalId` silently corrupts data when the input has no dash.**

```typescript
const dashIndex = acmsProfessionalId.indexOf('-'); // returns -1 if absent
const group = acmsProfessionalId.slice(0, dashIndex); // slice(0, -1) → drops last char
const code = parseInt(acmsProfessionalId.slice(dashIndex + 1), 10); // slice(0) → NaN
```

A malformed ID (missing dash) silently produces a truncated `group` and `NaN` for `code`. The caller in `buildBaseCmmapRow` has no guard — it will write a corrupt record to SQL without error. Add a guard:
```typescript
if (dashIndex === -1) throw new Error(`Invalid acmsProfessionalId format: "${acmsProfessionalId}"`);
```

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,433 @@
...
+  request.input('CDB_CREATE_DATE', sql.Numeric(8, 0), row.CDB_CREATE_DATE);
+  request.input('CDB_UPDATE_DATE', sql.Numeric(8, 0), row.CDB_UPDATE_DATE);
+  request.input('CDB_CREATE_DATE_DT', sql.DateTime2(3), row.CDB_CREATE_DATE_DT);
+  request.input('CDB_UPDATE_DATE_DT', sql.DateTime2(3), row.CDB_UPDATE_DATE_DT);
+  request.input('UPDATE_DATE', sql.DateTime2(3), row.UPDATE_DATE);
+  request.input('SOURCE', sql.VarChar(10), row.SOURCE);
+  request.input('CAMS_CASE_ID', sql.VarChar(50), row.CAMS_CASE_ID);
+  request.input('CAMS_USER_ID', sql.VarChar(50), row.CAMS_USER_ID);
+  request.input('CAMS_USER_NAME', sql.VarChar(100), row.CAMS_USER_NAME);
+  request.input('LAST_UPDATED', sql.DateTime2, row.LAST_UPDATED);
```
jalba-flexion: **`LAST_UPDATED` uses `sql.DateTime2` without a precision argument; every other DateTime2 column uses `sql.DateTime2(3)`.**

Without the precision specifier, mssql defaults to 7 decimal places, which may conflict with the column definition in `CMMAP_CAMS` and cause implicit rounding or a schema type mismatch. Use `sql.DateTime2(3)` here to match all other DateTime2 inputs.

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,433 @@
...
+      const event = typeof queueItem === 'string' ? JSON.parse(queueItem) : queueItem;
+      const appointmentEvent = event as TrusteeAppointmentDownstreamEvent;
+
+      if (
+        !appointmentEvent.caseId ||
+        !appointmentEvent.trusteeId ||
+        !appointmentEvent.assignedOn ||
+        !appointmentEvent.chapter
+      ) {
+        throw new Error('Invalid trustee appointment event: missing required fields');
```
jalba-flexion: **The trustee appointment handler validates `caseId`, `trusteeId`, `assignedOn`, and `chapter` — but not `acmsProfessionalId`.**

`TrusteeAppointmentDownstreamEvent` declares `acmsProfessionalId: string` (non-nullable), but the upstream emitter only queues events when the ID is present. If a message ever arrives without it (queue replay, manual injection, future change), `parseProfessionalId` receives `undefined`, hits the no-dash path (see separate comment), and writes a corrupt record silently.

Add `!appointmentEvent.acmsProfessionalId` to the existing validation block to give this a proper error path into the DLQ.

common/src/queues/queue-helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param parts - Parts to join into function name
+ * @returns Function name with underscores and spaces replaced by hyphens
+ *
+ * @example
+ * buildFunctionName('SYNC', 'CASES', 'handler')
+ * // Returns: 'SYNC-CASES-handler'
+ */
+export function buildFunctionName(...parts: string[]): string {
+  return parts.join('-').replace(/_/g, '-').replace(' ', '-');
```
jalba-flexion: **`replace(' ', '-')` only replaces the first space; use a regex to replace all occurrences.**

```typescript
return parts.join('-').replace(/_/g, '-').replace(' ', '-');
//                                            ^^^ stops after first match
```

`/_/g` correctly handles all underscores, but the string-literal form of `replace` stops at the first space. A module name with two spaces (e.g. `'MY MODULE NAME'`) would yield `'MY-MODULE NAME-...'`. Should be:
```typescript
.replace(/ /g, '-')
```
This bug exists in the original `dataflows-common.ts` and is being copied into this new shared helper.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -157,11 +186,21 @@ async function applyResolvedTrustee(
...
+  if (acmsProfessionalId) {
+    const openEvent: TrusteeAppointmentDownstreamEvent = {
+      caseId: event.caseId,
+      trusteeId,
+      acmsProfessionalId,
+      assignedOn: now,
+      appointedDate: event.appointedDate,
+      chapter: syncedCase.chapter,
+    };
+    await apiToDataflows.queueTrusteeAppointmentEvent(openEvent);
```
jalba-flexion: **`createCaseAppointment` is called before the downstream event is queued — if queuing fails, the appointment is already written with no recovery path.**

The current sequence:
1. Line 183: `appointmentsRepo.createCaseAppointment(...)` — writes to Cosmos
2. Line 202: `apiToDataflows.queueTrusteeAppointmentEvent(openEvent)` — enqueues downstream

If step 2 throws, the appointment exists in Cosmos but downstream never sees the open event, leaving the systems permanently out of sync with no retry mechanism. The same asymmetry applies to the soft-close path (line 165: queue after `updateCaseAppointment`).

Consider queuing first and persisting after, or at minimum logging a distinct, actionable error message that ops can use to replay the downstream event manually.

downstream/README.md
btposey: Removed deployment section. Deployment via bicep and GitHub workflow amendments will be addressed separately.

backend/function-apps/dataflows/downstream/README.md
btposey: Removed the `BOBJ reads CMMAP_ALL view` line from the architecture diagram.

downstream/README.md
btposey: Updated to "downstream systems"; removed BOBJ reference and the "cannot be easily modified" qualifier.

downstream/database/acms-cams-transition/README.md
btposey: Deleted this README.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Fixed. Events are now always emitted regardless of whether `acmsProfessionalId` is available. When absent, `null` is passed and `buildBaseCmmapRow` writes sentinel values (ZZ/99999) so the record lands in ACMS and can be corrected once the ID is available.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Removed the warn log. Events are now emitted unconditionally.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Fixed. The `canEmitDownstream` guard and the warn log are removed. Events always go out; we will review CMMAP staging to understand how to handle missing IDs when they appear.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Converted all comment lines to JSDoc style across the file.

backend/lib/use-cases/case-assignment/case-assignment.ts
btposey: Removed.

backend/function-apps/dataflows/dataflows-common.ts
btposey: Removed the re-export comment. New files introduced in this branch (`downstream/acms-cams-transition/acms-cams-transition.ts`) import directly from `@common/queues`. Existing consumers still go through `dataflows-common` via the re-export; full migration deferred.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Fixed. Added `if (dashIndex === -1) throw new Error(...)` guard before slicing.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Fixed. Changed to `sql.DateTime2(3)` to match all other DateTime2 inputs.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Fixed. Added `!appointmentEvent.acmsProfessionalId` to the validation block so malformed or replayed messages without an ID route to the DLQ instead of hitting `parseProfessionalId` with undefined.

common/src/queues/queue-helpers.ts
btposey: Fixed. Changed to `/ /g` regex so all spaces are replaced, not just the first.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Noted. Queuing after persisting means a queuing failure leaves Cosmos and downstream out of sync with no retry path. We will revisit this in a follow-up — the options are queuing first, a structured error log actionable by ops, or an outbox/saga pattern. Deferring the architectural decision to keep this PR scoped.

btposey: Addressed with a targeted log rather than reordering. Both queue calls (`closeEvent` and `openEvent`) are now wrapped in `try/catch`. On failure, `context.logger.error` emits a structured message identifying the case, trustee, and which direction (open/close) was missed — giving ops a searchable signal to trigger a resync or manual replay if the situation arises in production. The Cosmos write is intentionally kept first since it is the source of truth.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -148,20 +147,61 @@ async function applyResolvedTrustee(
       MODULE_NAME,
       `Soft-closed case appointment for case ${event.caseId}, old trustee ${existingAppointment.trusteeId}`,
     );
+    const oldProfessionalIds = await professionalIdsRepo.findByCamsTrusteeId(
+      existingAppointment.trusteeId,
+    );
+    const oldAcmsProfessionalId = oldProfessionalIds[0]?.acmsProfessionalId ?? null;
```
fmaddenflx: I'm curious about this.  You're grabbing possibly several professional ids but then only using the first one?  Why is only the first one relevant and how can we be certain it's always the first one?

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -148,20 +147,61 @@ async function applyResolvedTrustee(
...
+      context.logger.error(
+        MODULE_NAME,
+        `Failed to queue close event for case ${event.caseId}, trustee ${existingAppointment.trusteeId} — appointment updated in Cosmos but downstream not notified`,
+        queueError,
+      );
+    }
   }
 
+  const newProfessionalIds = await professionalIdsRepo.findByCamsTrusteeId(trusteeId);
+  const acmsProfessionalId = newProfessionalIds[0]?.acmsProfessionalId ?? null;
```
fmaddenflx: Isn't this the same thing we did earlier?

jamesobrooks (commented):
## CAMS-616 Completeness Assessment

This review was generated by AI review agents (software architect, DevSecOps, SRE, pragmatic, and product manager). All findings are documentation-only — nothing has been changed locally.

---

### Acceptance Criteria Status

| Criterion | Status |
|-----------|--------|
| Appointment data available to downstream SQL consumers | ✅ Met |
| Data includes current **and historical** appointments | ❌ **Not Met** |
| Compatible with ACMS CMMAP format | ✅ Met |
| Downstream systems need no code changes to consume | ❌ **Not Met** |
| Follows PR #2078 pattern | ✅ Met |
| Acceptable latency | ✅ Met |
| 100% accuracy between CAMS and downstream | ❌ **Not Met** |
| Testable in non-production environments | ❌ **Not Met** — integration README has broken file paths |
| Monitoring detects sync failures or data discrepancies | ❌ **Not Met** |

### Definition of Done Status

| DoD Item | Status |
|----------|--------|
| Feature toggles created | ✅ Met |
| Threat model updated | ❌ Not Met |
| Build process updated | ✅ Met |
| Technical documentation updated | ❌ Partial — no ADR, no ops docs |
| Deployment changes documented | ❌ Not Met — no Bicep, no pipeline step |
| User documentation updated | ❌ Not Met |

---

### Key Gaps Requiring Action Before Production

**Historical data:** The sync is forward-looking only. Appointments already stored in Cosmos before this PR will not appear in `CMMAP_ALL` until the dataflow re-encounters them. No backfill mechanism or plan is documented. The acceptance criterion explicitly requires "both current and historical appointments."

**S1 accuracy:** The staff assignment path always emits `acmsProfessionalId: null`, which causes `buildBaseCmmapRow` to write sentinel values (`PROF_CODE=99999`, `GROUP_DESIGNATOR=ZZ`) for every S1 row. Downstream consumers relying on these fields for staff assignment rows will receive systematically incorrect data.

**Deployment path:** There is no Bicep template, CI pipeline step, or runbook for deploying the `downstream/` function app. An operator cannot stand up this integration without direct engineering involvement.

**Consumer coordination:** TUFR and BOBJ must be reconfigured to query `CMMAP_ALL` instead of `CMMAP`. This is mentioned only in a `PRINT` statement inside the migration SQL. No coordination guide, runbook, or stakeholder communication exists.

**Threat model:** Not updated despite a new outbound SQL write path from CAMS to `ACMS_REP_SUB` being introduced.

See inline comments for individual findings.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+    password: process.env.ACMS_MSSQL_PASS,
+    options: {
+      encrypt: process.env.ACMS_MSSQL_ENCRYPT !== 'false',
+      trustServerCertificate: process.env.ACMS_MSSQL_TRUST_UNSIGNED_CERT === 'true',
+    },
+  };
+}
+
+export async function upsertCmmapCamsRow(row: CmmapCamsRow, sqlConfig: sql.config): Promise<void> {
+  const pool = await sql.connect(sqlConfig);
```
jamesobrooks: **[High — Reliability] SQL connection pool created per message**

`sql.connect()` creates (or reuses) `mssql`'s global connection pool on every queue invocation and then closes it in the `finally` block. Under concurrent invocations this is a race condition: one invocation closing the global pool while another is mid-query will produce a runtime error. It also bypasses connection pooling entirely, adding a TCP+TLS handshake cost to every message.

The backend's `AbstractMssqlClient` solves this with a static `Map<string, ConnectionPool>` keyed on server/database. The downstream handler should adopt the same pattern (or extend `AbstractMssqlClient` if the workspace dependency graph allows it).

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+      handlerName,
+      success: false,
+      durationMs: Date.now() - startTime,
+      error: error instanceof Error ? error.message : String(error),
+    });
+    context.extraOutputs.set(dlq, {
+      type: 'QUEUE_ERROR',
+      module: moduleName,
+      activityName: handlerName,
+      error,
```
jamesobrooks: **[High — Reliability + Security] DLQ payload loses error details and original event**

`Error` properties are non-enumerable; `JSON.stringify(error)` produces `{}`. The DLQ message will contain `"error": {}` — useless for recovery. Additionally, `mssql` errors can embed server hostname and credential context in the stack, which should not be stored in durable queue storage.

The original `queueItem` is also missing from the envelope, making manual replay impossible.

Suggested fix:
```ts
context.extraOutputs.set(dlq, {
  type: 'QUEUE_ERROR',
  module: moduleName,
  activityName: handlerName,
  error: error instanceof Error
    ? { message: error.message, name: error.name }
    : String(error),
  originalPayload: queueItem,
});
```
`handleQueueEvent` would need `queueItem` threaded through.

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+    });
+  }
+}
+
+// ─── Staff assignment handler ─────────────────────────────────────────────────
+
+const STAFF_MODULE_NAME = 'DOWNSTREAM-STAFF-ASSIGNMENTS';
+const STAFF_QUEUE_NAME = buildQueueName('CASE-ASSIGNMENT-EVENT');
```
jamesobrooks: **[Medium — Architecture] Queue name constants duplicated from backend — silent drift risk**

`'CASE-ASSIGNMENT-EVENT'` is a string literal here, but it is already a named constant in `backend/function-apps/dataflows/module-names.ts`. If the queue name is ever changed, the backend and downstream will silently diverge — the queue stops receiving messages with no compile-time error.

These string values should live in `common/src/queues/` (e.g., `queue-names.ts`) and be imported by both sides. Both packages already depend on `@common`.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+
+/** Transforms a staff assignment event into a CMMAP_CAMS row for upsert. */
+export function transformStaffAssignmentToRow(event: CaseAssignmentDownstreamEvent): CmmapCamsRow {
+  const isUnassigned = !!event.unassignedOn;
+  const apptDate = toAcmsDateNumeric(event.assignedOn);
+  const dispDate = isUnassigned ? toAcmsDateNumeric(event.unassignedOn!) : null;
+
+  return {
+    ...buildBaseCmmapRow(event.caseId, event.acmsProfessionalId, event.userId, event.name),
+    RECORD_SEQ_NBR: 1,
```
jamesobrooks, line 361: **[High — Data Correctness] Hardcoded `RECORD_SEQ_NBR: 1` with no last-writer-wins guard**

Both handlers hardcode sequence 1. The primary key is `(CASE_DIV, CASE_YEAR, CASE_NUMBER, APPT_TYPE, RECORD_SEQ_NBR)`, so the MERGE always targets the same row. Azure Storage Queues provide no ordering guarantee. `applyResolvedTrustee` enqueues a close event followed immediately by an open event for the same case. If the close event is processed after the open event, it overwrites the active appointment row with `APPTEE_ACTIVE: 'N'` — leaving ACMS believing no trustee is assigned.

Add a last-writer-wins guard to the MERGE:
```sql
WHEN MATCHED AND @LAST_UPDATED > target.LAST_UPDATED THEN
  UPDATE SET ...
```

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+  handler: staffAssignmentHandler,
+});
+
+// ─── Trustee appointment handler ──────────────────────────────────────────────
+
+const TRUSTEE_MODULE_NAME = 'TRUSTEE-APPOINTMENT-HANDLER';
+const TRUSTEE_QUEUE_NAME = buildQueueName('TRUSTEE-APPOINTMENT-EVENT');
+
+const TRUSTEE_DLQ = output.storageQueue({
+  queueName: buildQueueName('TRUSTEE-APPOINTMENT-EVENT', 'DLQ'),
```
jamesobrooks: **[Low — Consistency] DLQ naming convention differs between the two handlers**

The staff DLQ is named from `STAFF_MODULE_NAME` (line 309): `buildQueueName(STAFF_MODULE_NAME, 'DLQ')` → `downstream-staff-assignments-dlq`.

This trustee DLQ is named from the event queue name: `buildQueueName('TRUSTEE-APPOINTMENT-EVENT', 'DLQ')` → `trustee-appointment-event-dlq`.

These inconsistent conventions complicate monitoring alert patterns and queue-level access policies. Suggest aligning: `buildQueueName(TRUSTEE_MODULE_NAME, 'DLQ')`.

downstream/acms-cams-transition/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+    async () => {
+      const event = typeof queueItem === 'string' ? JSON.parse(queueItem) : queueItem;
+      const appointmentEvent = event as TrusteeAppointmentDownstreamEvent;
+
+      if (
+        !appointmentEvent.caseId ||
+        !appointmentEvent.trusteeId ||
+        !appointmentEvent.assignedOn ||
+        !appointmentEvent.chapter ||
+        !appointmentEvent.acmsProfessionalId
```
jamesobrooks, line 455: **[Critical — Data Correctness] Validation guard rejects null `acmsProfessionalId`, but the producer always sends null for unmatched trustees**

`sync-trustee-appointments.ts` builds the downstream event with `acmsProfessionalId: newProfessionalIds[0]?.acmsProfessionalId ?? null`. When no professional ID match exists, the emitted value is `null`. This guard then throws, routing every such event to the DLQ permanently — burning through all 5 Azure retry attempts before landing there.

`buildBaseCmmapRow` already handles `null` with sentinel values (`PROF_CODE=99999`, `GROUP_DESIGNATOR=ZZ`) and that path is tested. The validator should not require this field.

Remove `!appointmentEvent.acmsProfessionalId` from the validation check. The sentinel path in `buildBaseCmmapRow` is the correct handler for the null case.

common/src/cams/dataflow-events.ts
```diff
@@ -10,6 +10,30 @@ import { TrusteeAppointment } from './trustee-appointments';
...
+
+/**
+ * Downstream event for trustee case appointments. Carries ACMS-native field values
+ * so the downstream SQL handler requires no external lookups or translation.
+ * APPT_TYPE is always 'TR' for trustee appointments and is hardcoded by the handler.
+ */
+export type TrusteeAppointmentDownstreamEvent = {
+  caseId: string;
+  trusteeId: string;
+  acmsProfessionalId: string;
```
jamesobrooks: **[High — Architecture] Type contract lies — `acmsProfessionalId` is `string` but emitted as `null`**

`CaseAssignmentDownstreamEvent` (line 10) correctly declares `acmsProfessionalId: string | null`. This type declares `string` (non-nullable), but `sync-trustee-appointments.ts` constructs the event with `acmsProfessionalId: ?? null`.

The downstream handler's validation guard (line 425 of `acms-cams-transition.ts`) catches this mismatch at runtime and permanently DLQs the message. The type should reflect reality:
```ts
acmsProfessionalId: string | null;
```

backend/lib/use-cases/case-assignment/case-assignment.ts
```diff
@@ -199,8 +199,13 @@ export class CaseAssignmentUseCase {
     history.updatedOn = currentDate;
     await casesRepo.createCaseHistory(history);
 
-    for (const assignment of [...addedAssignments, ...removedAssignments]) {
-      await this.apiToDataflowsGateway.queueCaseAssignmentEvent(assignment);
+    if (context.featureFlags['downstream-staff-assignments-enabled']) {
+      for (const assignment of [...addedAssignments, ...removedAssignments]) {
+        await this.apiToDataflowsGateway.queueCaseAssignmentEvent({
```
jamesobrooks: **[High — Reliability] Queue failure here propagates as a 500, leaving Cosmos and downstream out of sync**

The queue call has no try/catch. If `queueCaseAssignmentEvent` throws (transient storage connectivity, storage key rotation, queue not yet created), the exception propagates up through `assignTrialAttorneys` to the HTTP controller — returning a 500 to the user **after** the assignment has already been committed to Cosmos at the previous step.

The user retries, potentially creating a duplicate assignment. Meanwhile, the downstream queue never received the event.

`sync-trustee-appointments.ts` wraps its equivalent calls in try/catch. Apply the same pattern here:
```ts
try {
  await this.apiToDataflowsGateway.queueCaseAssignmentEvent({ ...assignment, acmsProfessionalId: null });
} catch (queueError) {
  context.logger.error(MODULE_NAME, `Failed to queue downstream event — assignment written but downstream not notified`, queueError);
}
```

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -148,20 +147,61 @@ async function applyResolvedTrustee(
...
+      acmsProfessionalId: oldAcmsProfessionalId,
+      assignedOn: existingAppointment.assignedOn,
+      appointedDate: existingAppointment.appointedDate,
+      chapter: syncedCase.chapter,
+      unassignedOn: now,
+    };
+    try {
+      await apiToDataflows.queueTrusteeAppointmentEvent(closeEvent);
+    } catch (queueError) {
+      context.logger.error(
```
jamesobrooks: **[High — Observability] Queue failures here are swallowed with no alertable signal**

The `context.logger.error(...)` call logs the failure but produces no structured metric, no custom Application Insights event, and no counter. A monitoring system cannot alert on this — it requires someone to know to search the logs for this specific message.

The CAMS-616 acceptance criterion "monitoring detects sync failures" is not met by log statements alone. Consider emitting a structured event with a stable `type` field that a KQL alert can target:
```ts
context.logger.error(MODULE_NAME, 'DOWNSTREAM_QUEUE_FAILURE', {
  caseId: event.caseId,
  failureType: 'close-event',
});
```
Then define an Azure Monitor alert on this structured field.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -139,6 +135,9 @@ async function applyResolvedTrustee(
     return; // Same trustee already active — nothing to do
   }
 
+  const professionalIdsRepo = factory.getTrusteeProfessionalIdsRepository(context);
+  const apiToDataflows = factory.getApiToDataflowsGateway(context);
```
jamesobrooks: **[Medium — Architecture] Trustee appointment downstream emission has no feature flag**

`case-assignment.ts` gates its equivalent queue call behind `context.featureFlags['downstream-staff-assignments-enabled']`. The trustee appointment path here has no such guard — once this code is deployed, trustee events flow unconditionally.

If the downstream consumer is not yet production-ready or needs to be rolled back, there is no operational kill switch short of a code deployment. Consider adding a symmetrical feature flag for the trustee path during the rollout period.

downstream/host.json
```diff
@@ -0,0 +1,8 @@
+{
+  "version": "2.0",
+  "isDefaultHostConfig": true,
```
jamesobrooks: **[Critical — Reliability] `host.json` has no retry policy, timeouts, or concurrency limits**

This is a queue-triggered function writing to SQL. Without explicit configuration:
- No `functionTimeout` — a degraded ACMS SQL endpoint can hold workers for up to the platform default (5 min), exhausting scale-out under sustained failures
- No `extensions.queues.maxDequeueCount` — Azure defaults to 5 retries before DLQ; this should be explicit
- No `extensions.queues.visibilityTimeout` — determines how long a message is locked during processing
- No SQL connection/request timeouts in `getSqlConfig()`

Suggested addition:
```json
"functionTimeout": "00:05:00",
"extensions": {
  "queues": {
    "maxDequeueCount": 5,
    "visibilityTimeout": "00:01:00",
    "batchSize": 4,
    "newBatchThreshold": 2
  }
}
```

downstream/local.settings.json.template
```diff
@@ -0,0 +1,20 @@
...
+  "Values": {
+    "AzureWebJobsStorage": "DefaultEndpointsProtocol=https;AccountName=YOUR_ACCOUNT;AccountKey=YOUR_KEY;EndpointSuffix=core.usgovcloudapi.net",
+    "AzureWebJobsDataflowsStorage": "DefaultEndpointsProtocol=https;AccountName=YOUR_DATAFLOWS_ACCOUNT;AccountKey=YOUR_KEY;EndpointSuffix=core.usgovcloudapi.net",
+    "FUNCTIONS_WORKER_RUNTIME": "node",
+    "ACMS_MSSQL_HOST": "sql-ustp-cams.database.usgovcloudapi.net",
+    "ACMS_MSSQL_DATABASE": "ACMS_REP_SUB",
+    "ACMS_MSSQL_USER": "YOUR_SQL_USER",
+    "ACMS_MSSQL_PASS": "YOUR_SQL_PASS",
+    "ACMS_MSSQL_ENCRYPT": "true",
+    "ACMS_MSSQL_TRUST_UNSIGNED_CERT": "true"
```
jamesobrooks: **[High — Security] `ACMS_MSSQL_TRUST_UNSIGNED_CERT: true` should not be the template default**

This disables TLS certificate validation for the SQL connection. The Azure SQL Government endpoint (`*.database.usgovcloudapi.net`) has a valid Microsoft CA-issued certificate — there is no legitimate reason to bypass validation there.

The README's Azure lower-environment configuration block copies this value as-is, normalising TLS bypass for non-local environments.

Change to `"false"` and add a comment:
```json
"ACMS_MSSQL_TRUST_UNSIGNED_CERT": "false"
```
`true` is only appropriate for the local SQL Edge container.

test/integration/acms-cams-transition/README.md
```diff
@@ -0,0 +1,192 @@
...
+### Step 1 — Verify environment
+```bash
+$HARNESS check-env
+```
+All lines must show `✓ PASS`. Resolve any `✗ FAIL` before proceeding.
+
+### Step 2 — Apply schema
+Creates `CMMAP_STAGING` table and `CMMAP_TRANSITION` view in `ACMS_REP_SUB`.
+```bash
+$HARNESS run-sql downstream/database/acms-cams-transition/schema/cmmap-staging.sql ACMS_REP_SUB
```
jamesobrooks: **[High — Testability] Stale file paths and object names throughout this README**

This README consistently references objects and files that do not exist:
- `CMMAP_STAGING` → actual table is `CMMAP_CAMS`
- `CMMAP_TRANSITION` view → actual view is `CMMAP_ALL`
- `cmmap-staging.sql` → actual file is `cmmap-cams.sql`
- `cmmap-view.sql` → actual file is `cmmap-all.sql`
- `02-seed-cmmap-staging.sql` → actual file is `02-seed-cmmap-cams.sql`

An engineer following this README will hit file-not-found errors at Step 2. The integration test runner (`test-cmmap-view.sql`) uses the correct name `CMMAP_ALL`, but the README's narrative uses the stale names throughout. This makes the acceptance criterion "testable in non-production environments" unverifiable until the README is corrected.

test/integration/acms-cams-transition/scripts/start-services.sh
```diff
@@ -0,0 +1,91 @@
...
+#
+# After this script exits cleanly all three services are accepting connections:
+#   SQL Edge  → localhost:1433  (sa / YourStrong!Passw0rd)
+#   MongoDB   → localhost:27017
+#   Azurite   → localhost:10001 (queue endpoint)
+
+set -e
+
+POD_NAME="cams-integration-pod"
+SQLEDGE_PASS="${MSSQL_PASS:-YourStrong!Passw0rd}"
```
jamesobrooks: **[Medium — Security] Hardcoded SA password default; echoed in cleartext to stdout**

`YourStrong!Passw0rd` is a well-known SQL Server complexity-satisfying default that developers are unlikely to override. The password is then echoed at line 87: `echo "  SQL Edge  → localhost:1433  (user=sa, pass=${SQLEDGE_PASS})"` — if this script is ever invoked in a CI pipeline, the password appears in build logs.

Suggested fix: remove the default, fail fast if `MSSQL_PASS` is unset, and remove the cleartext echo:
```bash
if [ -z "${MSSQL_PASS}" ]; then
  echo "ERROR: MSSQL_PASS must be set" >&2
  exit 1
fi
SQLEDGE_PASS="${MSSQL_PASS}"
```

common/src/queues/queue-helpers.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import { buildFunctionName, buildQueueName } from './queue-helpers';
+
+describe('buildFunctionName', () => {
+  it('should return empty string for no arguments', () => {
+    expect(buildFunctionName()).toEqual('');
+  });
+
+  it('should return single argument unchanged', () => {
+    expect(buildFunctionName('ONE')).toEqual('ONE');
+  });
```
fmaddenflx: We usually prefer test() to it().  (I don't care too much but just to stick with convension)

test/integration/acms-cams-transition/scripts/test-trustee-appointment-downstream.ts
```diff
@@ -0,0 +1,839 @@
...
+async function getContext() {
+  const invocationContext = new InvocationContext();
+  const appContext = await ApplicationContextCreator.getApplicationContext({
+    invocationContext,
+    logger: ApplicationContextCreator.getLogger(invocationContext),
+  });
+  return { appContext, invocationContext };
+}
+
+async function flushExtraOutputsToAzurite(invocationContext: InvocationContext): Promise<number> {
```
fmaddenflx, line 139: Azurite?  :-)

jamesobrooks: It's a local implementation of storage, so we can use queues without actually putting messages in our real resources.

test/integration/acms-cams-transition/scripts/test-trustee-appointment-downstream.ts
```diff
@@ -0,0 +1,839 @@
...
+// ---------------------------------------------------------------------------
+// clean
+// ---------------------------------------------------------------------------
+
+async function clean() {
+  console.log('\nCleaning up test data...\n');
+
+  const { client, db } = await getMongoDb();
+  try {
+    // Find all integration trustee IDs via the professional ID mapping
```
fmaddenflx, line 689: So is the test data written to and deleted from a live production database?  Is that why we have to be so careful about deleting specific records when we clean?  That makes me a bit nervous ;-)

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -0,0 +1,441 @@
...
+  };
+}
+
+export async function upsertCmmapCamsRow(row: CmmapCamsRow, sqlConfig: sql.config): Promise<void> {
+  const pool = await sql.connect(sqlConfig);
+  const request = pool.request();
+
+  const query = `
+    MERGE INTO CMMAP_CAMS AS target
+    USING (VALUES (
```
fmaddenflx, line 177: Interesting.  I'm not familiar with sql MERGE.  Maybe you could give me a quick overview.

fmaddenflx (approved):

backend/lib/use-cases/case-assignment/case-assignment.test.ts
```diff
@@ -105,26 +106,29 @@ describe('Case assignment tests', () => {
...
+          return Promise.resolve([officeStaffJoeNobel]);
+        } else if (predicate.userId === attorneyJaneSmith.id) {
+          return Promise.resolve([officeStaffJaneSmith]);
+        } else {
+          return Promise.resolve([]);
+        }
+      };
+      vi.spyOn(factory, 'getOfficesRepository').mockReturnValue({
+        search: searchImpl,
+      } as unknown as OfficesRepository);
```
jamesobrooks, line 126: `MockMongoRepository` already implements `OfficesRepository` (including the `search` stub), so `as unknown as OfficesRepository` is not needed here or in the other six occurrences below.

The factory already returns `MockMongoRepository` in the test context, so the pattern that avoids the cast is the same one used elsewhere in this file (e.g. line 110):
```typescript
vi.spyOn(MockMongoRepository.prototype, 'search').mockImplementation(searchImpl);
```

backend/lib/use-cases/dataflows/sync-trustee-appointments.test.ts
```diff
@@ -109,6 +114,15 @@ describe('SyncTrusteeAppointments', () => {
       vi.spyOn(factory, 'getTrusteeMatchVerificationRepository').mockReturnValue(
         mockVerificationRepo as TrusteeMatchVerificationRepository,
       );
+      vi.spyOn(factory, 'getTrusteeProfessionalIdsRepository').mockReturnValue({
```
jamesobrooks, line 119: `vi.spyOn(factory, 'getTrusteeProfessionalIdsRepository').mockReturnValue({ findByCamsTrusteeId: ... })` follows the same pattern flagged in `case-assignment.test.ts`. `MockMongoRepository` already implements `TrusteeProfessionalIdsRepository` (including `findByCamsTrusteeId` and `findByAcmsProfessionalId`), so these factory spies should be replaced with:

```typescript
vi.spyOn(MockMongoRepository.prototype, 'findByCamsTrusteeId').mockResolvedValue([...]);
```

The same applies to the other repository factory spies added in this file (`getCasesRepository`, `getTrusteeAppointmentsRepository`, `getTrusteeMatchVerificationRepository`, `getTrusteesRepository`) — all of these are on `MockMongoRepository.prototype` and can be spied on directly without going through the factory.

downstream/host.json
btposey: Fixed. Added `functionTimeout: "00:05:00"` and `extensions.queues` config: `maxDequeueCount: 5`, `visibilityTimeout: "00:01:00"`, `batchSize: 1`. Transient errors in the handler now re-throw so Azure retries up to 5 times before moving the message to the Azure poison queue. Validation errors (malformed events that will never succeed) are caught and routed directly to the app-managed DLQ without consuming retry attempts.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: Acknowledged. The `sql.connect()` / `pool.close()` per-message pattern is inherited from the previous implementation — we are deferring connection pool lifecycle management (shared module-level pool or connection pooling middleware) to a follow-up. The immediate focus of this commit is retry/resilience behavior. Noted as a future improvement.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Fixed. Added a `serializeError(error)` utility that extracts `name`, `message`, and `stack` from `Error` instances into a plain object (falls back to `{ raw: String(error) }` for non-Error values). The DLQ payload now uses `serializeError(error)` and also includes `originalEvent: queueItem` so ops have everything needed to replay the message manually. Covered by unit tests for both Error and non-Error cases.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: The validation guard on `acmsProfessionalId` has been removed from the trustee handler. Events with a null/missing `acmsProfessionalId` now flow through — `buildBaseCmmapRow` writes sentinel values (`ZZ`/`99999`) so the record lands in ACMS and can be corrected once the ID becomes available. The `acmsProfessionalId` check in the staff assignment handler is intentionally kept because staff events always require a resolved professional ID before they can be written.

backend/lib/use-cases/case-assignment/case-assignment.ts
btposey: Fixed. The `queueCaseAssignmentEvent` call is now wrapped in a try/catch. On failure, a structured error is logged with `case ${caseId}, user ${userId}` identifiers and a "Manual replay required" note. The error does not propagate — the assignment write to Cosmos already succeeded, and rethrowing would result in a 500 with no way to recover the duplicate-write scenario on retry.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: The existing structured log call already captures `caseId`, `trusteeId`, and the error. The queue failure after a Cosmos write is intentionally caught rather than rethrown — rethrowing would retry the full `applyResolvedTrustee` call and attempt a duplicate Cosmos write. Structured logging here is the replay signal for ops. Application Insights alert configuration is infrastructure-level and outside the scope of this PR.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: This ordering is intentional. The current sequence — Cosmos write first, queue second — is by design: Cosmos is the system of record. If the queue call fails, we log a structured error with `caseId`/`trusteeId` and a "Manual replay required" note so ops can replay by re-running sync for that case. Reversing the order (queue first) would cause the downstream handler to attempt a SQL write for an appointment that does not yet exist in Cosmos, creating a different class of inconsistency.

common/src/queues/queue-helpers.test.ts
btposey: Fixed. Replaced all `it()` calls with `test()` and updated the import accordingly.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Fixed. The `[0]` index was non-deterministic when a trustee has professional IDs in multiple ACMS groups. The fix uses `syncedCase.courtDivisionCode` to resolve the correct `GROUP_DESIGNATOR` via the offices graph (`findGroupDesignatorForDivision`), then selects the `acmsProfessionalId` whose group prefix matches. When no match is found, a `TRUSTEE_APPOINTMENT_DOWNSTREAM_SYNC_ERROR` document is written to Cosmos (queryable and replayable) instead of passing a wrong ID downstream. Sentinel values removed entirely.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Good catch — yes, the second `findByCamsTrusteeId` was the same lookup repeated for the new trustee open event. Both lookups have been consolidated into a shared `resolveGroupMatchedProfessionalId` helper that handles both close (old trustee) and open (new trustee) paths. The helper also resolves the correct ID by group designator rather than blindly taking `[0]`.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Fixed. `acmsProfessionalId` is now included in the trustee handler validation block. An event missing this field throws a `ValidationError` and routes to the DLQ immediately rather than reaching `parseProfessionalId`. Sentinel fallback removed entirely — a missing professional ID now produces a `TRUSTEE_APPOINTMENT_DOWNSTREAM_SYNC_ERROR` document in Cosmos written upstream before the event is ever queued.

btposey:
Raising for team discussion: **close-event ordering when old trustee has no matching professional ID**.

When replacing an old trustee and the old trustee has no matching `acmsProfessionalId` for the case's group, the new implementation writes a `TRUSTEE_APPOINTMENT_DOWNSTREAM_SYNC_ERROR` doc with `replacedByTrusteeId` set to the incoming trustee. This field exists so a future replay knows the correct ordering: close (old trustee) must be replayed before open (new trustee).

However, the new trustee's open event is still queued and processed normally. This means ACMS can briefly see an open record for the new trustee without a preceding close for the old one — the CMMAP row for the old trustee remains in `APPTEE_ACTIVE=Y` state until the sync error is corrected and replayed.

Options to discuss:
1. **Current behavior (deferred close)** — new trustee open proceeds; old trustee close deferred to manual replay after professional ID is corrected. `replacedByTrusteeId` preserves ordering intent.
2. **Block new open on close failure** — hold the new trustee open event until the close event can also be queued. Requires a more stateful retry mechanism.
3. **Accept the gap** — CMMAP is append-merge keyed on `APPT_TYPE + RECORD_SEQ_NBR`; having two active trustee rows for the same case may be tolerable short-term in ACMS.

Option 1 is what is currently implemented. Happy to revisit if the team has a stronger preference.

test/integration/acms-cams-transition/scripts/test-trustee-appointment-downstream.ts
btposey: Azurite is Microsoft's local emulator for Azure Storage (blobs, queues, tables). It runs in Docker and lets the integration tests exercise real queue operations — enqueue, dequeue, poison queue routing — without touching actual Azure infrastructure. It's the standard local stand-in for Azure Storage Queue in Azure Functions dev/test workflows.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: `MERGE` is a SQL upsert — it atomically handles insert-or-update in a single statement. It matches rows in the target table against a source dataset on a key (here the composite `CASE_DIV + CASE_YEAR + CASE_NUMBER + RECORD_SEQ_NBR`). When a match is found it runs `WHEN MATCHED THEN UPDATE`, and when no match exists it runs `WHEN NOT MATCHED THEN INSERT`. This makes the handler idempotent — replaying the same queue message won't create duplicate rows.

downstream/acms-cams-transition/acms-cams-transition.ts
btposey: Fixed. Changed `buildQueueName('TRUSTEE-APPOINTMENT-EVENT', 'DLQ')` to `buildQueueName(TRUSTEE_MODULE_NAME, 'DLQ')` — now consistent with the staff handler pattern.

backend/lib/use-cases/dataflows/sync-trustee-appointments.test.ts
btposey: Removed the unnecessary casts. `TrusteeProfessionalIdsRepository` and `OfficesGateway` are small interfaces and the inline mocks satisfy them fully, so `as unknown as` was not needed. Left the casts on `CasesInterface` and `TrusteesRepository` where the inline mocks are genuinely partial.

backend/lib/use-cases/case-assignment/case-assignment.test.ts
btposey: The `{ search: searchImpl }` mock here is a partial implementation — `OfficesRepository` extends `Searches`, `Releasable`, and several other methods, so dropping the cast would be a type error. The cast is load-bearing. That said, if you'd prefer we switch to `MockMongoRepository` and spy on `search` instead, happy to do that.

downstream/local.settings.json.template
btposey: The USTP environment requires TLS certificate validation to be disabled — `true` is the correct default for this template. No change.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Added `downstream-trustee-appointments-enabled` feature flag following the same pattern as `downstream-staff-assignments-enabled` in `case-assignment.ts`. Both queue emissions (open and close events) are gated behind the flag. The `apiToDataflows` gateway instantiation is inside the flag check so no unnecessary gateway lookup occurs when the flag is off.

common/src/cams/dataflow-events.ts
btposey: Fixed. `acmsProfessionalId` is now `string | null` in `TrusteeAppointmentDownstreamEvent` — the type contract now accurately reflects what the producer emits.

test/integration/acms-cams-transition/scripts/start-services.sh
btposey: Fixed. `start-services.sh` now sources `scripts/.env` (gitignored) and fails fast with a clear message if `MSSQL_PASS` is unset. Added `scripts/.env.template` as a committed guide. The password is no longer hardcoded in the script or echoed to stdout.

test/integration/acms-cams-transition/README.md
btposey: Fixed. Updated throughout:
- `CMMAP_STAGING` → `CMMAP_CAMS`
- `CMMAP_TRANSITION` view → `CMMAP_ALL`
- `cmmap-staging.sql` → `cmmap-cams.sql`
- `cmmap-view.sql` → `cmmap-all.sql`
- `02-seed-cmmap-staging.sql` → `02-seed-cmmap-cams.sql`
- `downstream/functions` → `downstream` (no `functions` subdirectory)

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
jamesobrooks: After digging into this, Brian is right that `RECORD_SEQ_NBR: 1` is correct.

`CMMAP_CAMS` is not the same as `CMMAP`. The `CMMAP_CAMS` DDL includes:

```sql
CONSTRAINT UQ_CMMAP_CAMS_CAMS_CASE UNIQUE (CAMS_CASE_ID, APPT_TYPE)
```

This enforces exactly one row per case per appointment type at the DB level. `CMMAP_CAMS` is intentionally a *current-state store* — not an appointment history log. The hardcoded `1` correctly reflects that invariant. There is no scenario where a second row with a different `RECORD_SEQ_NBR` for the same case+type would be inserted; the unique constraint would reject it first.

---

That said, the ordering concern in this comment is real, but independent of `RECORD_SEQ_NBR`.

In `applyResolvedTrustee`, when a trustee changes, a close event (old trustee) and an open event (new trustee) are enqueued to the same Azure Storage Queue. Both events target the same PK row `(CASE_DIV, CASE_YEAR, CASE_NUMBER, 'TR', 1)`. If the close event is processed after the open event, the MERGE updates `APPTEE_ACTIVE` to `'N'`, leaving ACMS believing no trustee is assigned.

The `LAST_UPDATED` column already exists on `CMMAP_CAMS`. A last-writer-wins guard on the MERGE would address this:

```sql
WHEN MATCHED AND @LAST_UPDATED > target.LAST_UPDATED THEN
  UPDATE SET ...
```

Opening a follow-up issue for that guard — it's a separate concern from the sequence number.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -0,0 +1,458 @@
...
+    password: process.env.ACMS_MSSQL_PASS,
+    options: {
+      encrypt: process.env.ACMS_MSSQL_ENCRYPT !== 'false',
+      trustServerCertificate: process.env.ACMS_MSSQL_TRUST_UNSIGNED_CERT === 'true',
+    },
+  };
+}
+
+async function upsertCmmapCamsRow(row: CmmapCamsRow, sqlConfig: sql.config): Promise<void> {
+  const pool = await sql.connect(sqlConfig);
```
jamesobrooks: **SQL connection pool is destroyed on every invocation, breaking concurrent calls.**

`sql.connect(sqlConfig)` returns (or creates) mssql's module-level `globalConnection` singleton. The `pool.close()` in the `finally` block patches the global to `null`, so the next invocation must recreate the pool from scratch. Under concurrent Azure queue-trigger invocations, Invocation B may hold a reference to the same pool that Invocation A is closing in its `finally`, producing mid-flight "ConnectionPool is closed" errors and unnecessary retries.

Every other SQL gateway in the codebase uses `AbstractMssqlClient` (`backend/lib/adapters/gateways/abstract-mssql-client.ts`) which keeps a static `Map<string, ConnectionPool>` and registers cleanup via `deferClose()` instead of closing per-request. This file should follow that pattern.

backend/function-apps/dataflows/downstream/README.md
```diff
@@ -0,0 +1,128 @@
...
+downstream/
+├── database/
+│   └── acms-cams-transition/
+│       ├── README.md
+│       ├── schema/
+│       │   ├── cmmap-cams.sql          # CMMAP_CAMS table DDL
+│       │   └── cmmap-all.sql           # CMMAP_ALL view DDL
+│       └── migrations/
+│           └── 001-initial-schema.sql  # Initial deployment migration
+├── shared/
```
jamesobrooks: This Project Structure section describes directories and files that do not exist. The actual `downstream/` directory contains four flat files: `acms-cams-transition.ts`, `acms-cams-transition.test.ts`, `staff-assignment-downstream.ts`, and `trustee-appointment-downstream.ts`. There is no `shared/`, `staff-assignment-handler/`, or `trustee-appointment-handler/` subdirectory, and no standalone `index.ts`, `host.json`, `package.json`, or `tsconfig.json` at this level. Please update or remove this section.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
```diff
@@ -157,11 +233,48 @@ async function applyResolvedTrustee(
...
+        trusteeId,
+        acmsProfessionalId,
+        assignedOn: now,
+        appointedDate: event.appointedDate,
+        chapter: syncedCase.chapter,
+      };
+      const apiToDataflows = factory.getApiToDataflowsGateway(context);
+      try {
+        await apiToDataflows.queueTrusteeAppointmentEvent(openEvent);
+      } catch (queueError) {
```
jamesobrooks, line 281: Queue errors are caught and swallowed here (and in the close-event catch above). If this call fails after `createCaseAppointment` has already committed to Cosmos, the appointment exists in CAMS with no downstream notification sent — and no `TRUSTEE_APPOINTMENT_DOWNSTREAM_SYNC_ERROR` doc is written for this failure mode. The function continues and returns success, so Azure Functions will not retry, leaving CMMAP_CAMS permanently out of sync with no recovery path.

If reordering Cosmos-write-after-queue is out of scope for this PR, at minimum the catch block should write a sync error document so the failure is tracked and replayable.

jamesobrooks:
**Stale TODO comments still present in `case-assignment.ts`**

The TODO comments at lines 27 and 71 of `backend/lib/use-cases/case-assignment/case-assignment.ts` were not removed:

```
// TODO: createTrialAttorneyAssignments should not take a role, or should be renamed
// TODO: assignTrialAttorneys should not take a role, or should be renamed
```

These lines weren't in the diff so they can't be inline-commented, but they're still present in the file. Please remove them.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
jamesobrooks: The current code does not match this description. When `acmsProfessionalId` cannot be resolved, the event is suppressed entirely — a `TRUSTEE_APPOINTMENT_DOWNSTREAM_SYNC_ERROR` doc is written to Cosmos, but no downstream queue message is sent. There are no sentinel values (ZZ/99999) in `buildBaseCmmapRow`, and the `trusteeAppointmentHandler` throws `ValidationError` (routing to DLQ) if it receives an event with a null professional ID.

Please update this reply to accurately describe the implemented behavior: events are deferred (not emitted unconditionally) when no professional ID is found, and the failure is tracked via a sync error document.

btposey:
Removed both TODO comments from `case-assignment.ts`.

backend/function-apps/dataflows/downstream/README.md
btposey: Fixed. Updated the Project Structure section to reflect the actual flat layout — four .ts files plus the `database/` subtree. Removed the stale Development section (no standalone npm project exists at this level) and the BOBJ reference.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: Fixed. Replaced `sql.connect()` + `pool.close()` in the `finally` block with a module-level `ConnectionPool` singleton. The pool is initialized once via `getConnectionPool()`, registered with `deferClose()` for process shutdown cleanup, and reused across invocations — matching the pattern in `AbstractMssqlClient`.

backend/lib/use-cases/dataflows/sync-trustee-appointments.ts
btposey: Fixed. Both catch blocks now call `appointmentsRepo.upsertDownstreamSyncError()` after logging, writing a `TRUSTEE_APPOINTMENT_DOWNSTREAM_SYNC_ERROR` doc so the failure is tracked and replayable. Reordering the Cosmos write after queuing is deferred to a follow-up.

PR #2401, opened 2026-05-15
CAMS-669 Add associated banks table and edit status flow

fmaddenflx:
# Purpose

Allow OoO administrators to view, add, and manage the status of banks associated with a bankruptcy software vendor from the vendor detail page.

# Major Changes

- Added `SoftwareBankAssociation` type and `associatedBanks` field to `BankruptcySoftwareProfile`
- Extended the backend use case and controller to handle `addBank` and `updateBankAssociation` PUT body variants, with audit history
- Added `AssociatedBanksTable` component displaying bank name, status, and Edit Status action, with an inline combobox to add new banks (filtered to exclude already-associated banks)
- Added `AddAssociatedBankConfirmModal` — confirmation dialog before adding a bank
- Added `EditBankAssociationStatusModal` — Active/Inactive radio modal for updating a bank association's status
- Wired all modals and API calls into `BankruptcySoftwareDetail`
- Extended `Api2` and `mock-api2` with `addAssociatedBank` and `updateBankAssociationStatus`

# Testing/Validation

Implemented using TDD. Unit tests cover all new components and modal interactions:

- `AssociatedBanksTable` — renders associations, filters available banks, opens confirm modal, clears combobox after add
- `AddAssociatedBankConfirmModal` — heading shows bank name, warning text, confirm/cancel behavior
- `EditBankAssociationStatusModal` — pre-populates status, radio selection, save/cancel/error scenarios
- Backend use case and controller tests cover both `addBank` and `updateBankAssociation` paths including validation and audit history

# Notes

The `updateBankAssociationStatus` API call re-uses the existing `PUT /bankruptcy-software/:id` endpoint with a discriminated union body (`addBank` vs `updateBankAssociation`), consistent with the pattern already established for `updateSoftware`.

The Edit Status modal heading uses the pattern `Edit [Bank Name] Bank Status` and warning copy matches the spec exactly. The Add Bank confirm modal uses the confirmation question as the heading (no separate title) per design.

Slice 4 (Trustees tab) is on hold pending CAMS-671.

# Definition of Done:

## Summary by Sourcery

Add support for managing banks associated with a bankruptcy software vendor, including viewing associations, adding new banks, and updating association status from the vendor detail page.

New Features:
- Introduce associated bank metadata on bankruptcy software profiles to track linked banks and their active/inactive status.
- Expose backend endpoints and client API methods to add a bank association and update an existing association's status using the existing bankruptcy software PUT endpoint.
- Add an Associated Banks section to the bankruptcy software detail overview with a table, bank-selection control, and confirmation/modal flows for adding banks and editing association status.

Enhancements:
- Extend the bankruptcy software update use case and controller to validate and process addBank and updateBankAssociation operations with audit history for before/after changes.
- Integrate global alerts in the bankruptcy software detail page to surface success and failure messages when modifying bank associations.

Tests:
- Add comprehensive unit tests for the new use-case branches, controller validation paths, associated banks table behavior, and the add/edit association modals, including error handling and UI state.


user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  function handleConfirm(bankId: string, bankName: string) {
+    onAddBank(bankId, bankName);
+    setSelectedBank(null);
+  }
+
+  return (
+    <div className="associated-banks-section">
```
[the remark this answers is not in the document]
bityogi: This is already in place — the div at line 57 has `data-testid="associated-banks-section"` and all tests pass using that selector. No change needed.

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -34,17 +38,56 @@ export class BankruptcySoftwareUseCase {
 
   async updateSoftware(
```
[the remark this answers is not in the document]
bityogi: Implemented. Refactored `updateSoftware` into a dispatcher pattern:
- `applyUpdate()` routes to the correct private helper based on the update type
- `addBankAssociation()` handles the add-bank domain logic
- `updateBankAssociationStatus()` handles the status-change domain logic
- `saveWithAudit()` centralizes the persist + audit record flow

All 41 backend tests pass.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -51,13 +51,51 @@ export class BankruptcySoftwareController {
     softwareId: string,
   ): Promise<CamsHttpResponseInit<BankruptcySoftwareProfile>> {
     this.requireSuperUser(context);
```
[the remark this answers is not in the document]
bityogi: Implemented. Extracted `handlePut` into orchestration-only logic:
- `parseAddBankBody()` validates and returns typed addBank payload or null
- `parseUpdateBankAssociationBody()` validates and returns typed payload or null
- `parseProfileUpdateBody()` handles the generic profile update case
- `buildSuccessResponse()` eliminates the repeated httpSuccess call

`handlePut` is now a simple dispatcher that delegates to parsers and the use case.

bityogi (approved):

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -89,6 +95,66 @@ export class BankruptcySoftwareController {
...
+    return {
+      updateBankAssociation: { bankId: raw.bankId.trim(), status: raw.status },
+    };
+  }
+
+  private parseProfileUpdateBody(
+    body: Record<string, unknown> | null,
+  ): Partial<Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>> {
+    const profileUpdate =
+      (body as Partial<Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>> | null) ??
```
jamesobrooks: **[Security – High] Arbitrary field injection: cast-only type guard performs no runtime key filtering**

`body as Partial<Pick<...>>` is a TypeScript compile-time assertion only. At runtime, the entire request body object is returned and spread into the saved document via `{ ...base, ...update }` in `applyUpdate`. A SuperUser (or attacker with a SuperUser token) can send:

```json
{ "associatedBanks": [{"bankId": "injected", "bankName": "Evil", "status": "active"}] }
```

This bypasses all of the `addBankAssociation` guards (duplicate check, `status: 'active'` default) and overwrites the banks array directly. They can also inject `id`, `updatedBy`, `documentType`, etc.

Fix: explicitly construct the output object from allowed keys rather than casting:

```typescript
private parseProfileUpdateBody(
  body: Record<string, unknown> | null,
): Partial<Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>> {
  const update: Partial<Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>> = {};
  if (!body) return update;
  if ('name' in body) {
    if (typeof body.name !== 'string' || !body.name.trim())
      throw new BadRequestError(MODULE_NAME, { message: 'Software name is required.' });
    update.name = (body.name as string).trim();
  }
  if ('status' in body) {
    if (body.status !== 'active' && body.status !== 'inactive')
      throw new BadRequestError(MODULE_NAME, { message: "status must be 'active' or 'inactive'." });
    update.status = body.status as BankruptcySoftwareProfile['status'];
  }
  if ('contact' in body) update.contact = body.contact as SoftwareContactInfo;
  return update;
}
```

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -89,6 +95,66 @@ export class BankruptcySoftwareController {
...
+  private parseAddBankBody(
+    body: Record<string, unknown> | null,
+  ): { addBank: { bankId: string; bankName: string } } | null {
+    if (!body || !('addBank' in body)) return null;
+
+    const raw = body.addBank as Record<string, unknown> | null;
+    if (!raw || typeof raw.bankId !== 'string' || !raw.bankId.trim()) {
+      throw new BadRequestError(MODULE_NAME, { message: 'bankId is required.' });
+    }
+    if (typeof raw.bankName !== 'string' || !raw.bankName.trim()) {
```
jamesobrooks, line 111: **[Security – High] No max-length bounds on `bankId` or `bankName`**

Both fields are only validated for presence. There is no upper-length guard. A value of 100,000 characters will be stored in Cosmos DB without rejection, inflating document size and potentially hitting the 2 MB per-document limit when combined with existing banks. The linear scan `existingBanks.some((b) => b.bankId === bankId)` also operates on an unbounded string.

The established project pattern (trustees-validators) bounds similar fields at 50–255 characters. Apply the same here:

```typescript
const MAX_BANK_ID_LENGTH = 50;
const MAX_BANK_NAME_LENGTH = 100;
if (raw.bankId.trim().length > MAX_BANK_ID_LENGTH)
  throw new BadRequestError(MODULE_NAME, { message: `bankId must not exceed ${MAX_BANK_ID_LENGTH} characters.` });
if (raw.bankName.trim().length > MAX_BANK_NAME_LENGTH)
  throw new BadRequestError(MODULE_NAME, { message: `bankName must not exceed ${MAX_BANK_NAME_LENGTH} characters.` });
```

Apply the same `bankId` length guard in `parseUpdateBankAssociationBody`.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -89,6 +95,66 @@ export class BankruptcySoftwareController {
...
+  private parseProfileUpdateBody(
+    body: Record<string, unknown> | null,
+  ): Partial<Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>> {
+    const profileUpdate =
+      (body as Partial<Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>> | null) ??
+      {};
+    if (profileUpdate.name !== undefined && !profileUpdate.name.trim()) {
+      throw new BadRequestError(MODULE_NAME, { message: 'Software name is required.' });
+    }
+    return profileUpdate;
```
jamesobrooks: **[Security – Medium] No server-side validation of `contact.emails` / `contact.website`**

The frontend validates email and website via `trustees-validators`, but `parseProfileUpdateBody` passes the `contact` field through to Cosmos DB without any server-side validation. A SuperUser making a direct API call (curl, Postman) can store malformed emails and unconstrained URLs.

Frontends are not a trust boundary. All other PUT endpoints in this codebase validate at the API layer. Apply the same `email` and `website` validators from `trustees-validators` here when `contact` is present in the body.

jamesobrooks (commented):
> **AI agent review** — findings generated by an automated AI coding agent. Please treat these as a second set of eyes, not ground truth. Verify each finding before acting on it.

## Summary

29 files reviewed. The core feature logic — discriminated union type, bank association methods, modal components, ARIA live region, audit trail — is solid and correctly scoped. No speculative abstractions. The issues below range from security-critical to low-priority polish.

### Must-address before merge (security + reliability)
1. Arbitrary field injection via untyped profile update body (security)
2. No length bounds on `bankId` / `bankName` (security)
3. No server-side validation of `contact.emails` / `contact.website` (security)
4. Double-submit guard missing on both modals (reliability)
5. `Promise.all` couples software load to banks load (reliability)
6. `saveWithAudit` re-derives `userRef` redundantly (correctness)

Inline comments below for all findings.

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -95,4 +85,83 @@ export class BankruptcySoftwareUseCase {
...
+    updatedBanks[index] = { ...updatedBanks[index], status };
+    return { ...software, associatedBanks: updatedBanks };
+  }
+
+  private async saveWithAudit(
+    id: string,
+    before: BankruptcySoftwareProfile,
+    after: BankruptcySoftwareProfile,
+  ): Promise<BankruptcySoftwareProfile> {
+    const userRef = getCamsUserReference(this.context.session.user);
```
jamesobrooks: **[Correctness – High] `saveWithAudit` re-derives `userRef` redundantly**

`updateSoftware` fetches `userRef` at line 47 and passes it to `applyUpdate` (which stamps `updatedBy` on the merged record). Then `saveWithAudit` independently calls `getCamsUserReference(this.context.session.user)` again here solely to pass to `createAuditRecord`.

The two derivations are redundant — it's the same session user in the same request. Thread `userRef` from `updateSoftware` into `saveWithAudit` as a parameter instead:

```typescript
private async saveWithAudit(
  id: string,
  before: BankruptcySoftwareProfile,
  after: BankruptcySoftwareProfile,
  userRef: CamsUserReference,
): Promise<BankruptcySoftwareProfile>
```

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -95,4 +85,83 @@ export class BankruptcySoftwareUseCase {
...
+    const updated = await this.repository.updateSoftware(id, after);
+    await this.repository.createSoftwareAuditRecord(
```
jamesobrooks: **[Reliability – High] Audit record creation failure is indistinguishable from data write failure**

`updateSoftware` and `createSoftwareAuditRecord` are two sequential, non-transactional writes. If the data write succeeds but the audit write throws (network hiccup, quota exhaustion, serialisation error), the mutation is committed with no audit record. The outer `getCamsError` wrapper in `updateSoftware` will catch both errors identically — there is no way for ops to distinguish "data written, audit missing" from "write failed entirely".

In a compliance-sensitive bankruptcy case management system, silent audit loss is a significant operational risk.

Suggested fix: wrap `createSoftwareAuditRecord` in its own try/catch so audit failures can be logged and alerted on independently:

```typescript
const updated = await this.repository.updateSoftware(id, after);
try {
  await this.repository.createSoftwareAuditRecord(createAuditRecord({ ... }, userRef));
} catch (auditError) {
  // Data already committed — log prominently so ops can replay.
  this.context.logger.error('AUDIT_WRITE_FAILED', {
    softwareId: id,
    error: (auditError as Error).message,
  });
  throw getCamsError(auditError as Error, MODULE_NAME, 'Audit record creation failed after successful data write.');
}
return updated;
```

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -95,4 +85,83 @@ export class BankruptcySoftwareUseCase {
...
+
+  private addBankAssociation(
+    software: BankruptcySoftwareProfile,
+    bankId: string,
+    bankName: string,
+  ): BankruptcySoftwareProfile {
+    const existingBanks = software.associatedBanks ?? [];
+    if (existingBanks.some((b) => b.bankId === bankId)) {
+      throw new BadRequestError(MODULE_NAME, {
+        message: `Bank ${bankId} is already associated with this software.`,
```
jamesobrooks: **[Security – Medium] Error messages echo caller-supplied `bankId` — enumeration signal**

```
Bank ${bankId} is already associated with this software.
Bank ${bankId} is not associated with this software.
```

These messages confirm to a probing caller whether a specific `bankId` exists in the system for this software profile. In a FedRAMP-scoped application, echoing user input in error messages that distinguish between "exists" and "does not exist" is an enumeration risk.

Use generic messages that don't echo the input:
- `'This bank is already associated with this software.'`
- `'The specified bank is not associated with this software.'`

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -1,12 +1,22 @@
...
+import { CamsUserReference } from '@common/cams/users';
 import { getCamsUserReference } from '@common/cams/session';
 import { ApplicationContext } from '../../adapters/types/basic';
 import factory from '../../factory';
 import { getCamsError } from '../../common-errors/error-utilities';
+import { BadRequestError } from '../../common-errors/bad-request';
 
 const MODULE_NAME = 'BANKRUPTCY-SOFTWARE-USE-CASE';
 
+type SoftwareUpdate =
```
jamesobrooks: **[Architecture – High] `SoftwareUpdate` is not exported — controller has no type-safe contract**

`SoftwareUpdate` is declared as a module-private type. The controller independently reconstructs identical shapes in `parseAddBankBody` and `parseUpdateBankAssociationBody` with no compiler enforcement that they match the union arms. If a new arm is added to `SoftwareUpdate` (e.g., `{ removeBank: ... }`), TypeScript will not warn that the controller needs a new parse branch.

Export `SoftwareUpdate` and use it as the return type of the controller's parse helpers to make the controller-to-use-case boundary type-safe:

```typescript
// bankruptcy-software.ts
export type SoftwareUpdate = ...;

// bankruptcy-software.controller.ts
import { SoftwareUpdate } from '../../use-cases/bankruptcy-software/bankruptcy-software';

private parseAddBankBody(body: ...): Extract<SoftwareUpdate, { addBank: unknown }> | null
```

This keeps the union owned by the strategic layer (correct direction) while making the boundary compiler-enforced.

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  function handleCancel() {
+    modalRef.current?.hide();
+  }
+
+  const actionButtonGroup = {
+    modalId,
+    modalRef,
+    submitButton: {
+      label: 'Save',
```
jamesobrooks, line 50: **[Reliability – High] No in-flight guard — double-submit possible**

The Save button is never disabled while the PUT request is in-flight. A user who clicks Save twice before the first response returns fires two concurrent requests, producing duplicate audit records.

The established pattern in this codebase (e.g., `TrusteeInternalContactForm`) uses an `isSubmitting` state flag:

```typescript
const [isSubmitting, setIsSubmitting] = useState(false);

async function handleSubmit() {
  if (isSubmitting) return;
  setIsSubmitting(true);
  try {
    const response = await Api2.updateBankAssociationStatus(softwareId, bankId, status);
    onSuccess(response.data);
    modalRef.current?.hide();
    alert?.success(`${bankName} status has been updated.`);
  } catch {
    alert?.error('Failed to update bank status. Please try again.');
  } finally {
    setIsSubmitting(false);
  }
}

// In actionButtonGroup:
submitButton: {
  label: isSubmitting ? 'Saving…' : 'Save',
  onClick: handleSubmit,
  closeOnClick: false,
  disabled: isSubmitting,
},
```

The same issue applies to `handleAddBank` in `BankruptcySoftwareDetail.tsx`.

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    },
+  }));
+
+  async function handleSubmit() {
+    try {
+      const response = await Api2.updateBankAssociationStatus(softwareId, bankId, status);
+      onSuccess(response.data);
+      modalRef.current?.hide();
+      alert?.success(`${bankName} status has been updated.`);
+    } catch {
```
jamesobrooks: **[Reliability – Medium] Broad catch swallows actionable 400 errors**

All failure modes — 400 (e.g., "bank not associated"), 403, 500 — show the same generic message. 400 errors from the API are directly actionable by the user and should be surfaced as-is rather than replaced with "please try again".

```typescript
} catch (e) {
  const err = e as { status?: number; message?: string };
  if (err.status && err.status < 500 && err.message) {
    alert?.error(err.message);
  } else {
    alert?.error('Failed to update bank status. Please try again.');
  }
}
```

The same pattern applies to the `handleAddBank` catch in `BankruptcySoftwareDetail.tsx`.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetail.tsx
```diff
@@ -7,29 +8,39 @@ import Alert, { UswdsAlertStyle } from '@/lib/components/uswds/Alert';
...
 
   useEffect(() => {
     if (!softwareId) return;
 
     let isCancelled = false;
 
     setIsLoaded(false);
-    Api2.getSoftware(softwareId)
-      .then((response) => {
+    Promise.all([Api2.getSoftware(softwareId), Api2.getBanks()])
```
jamesobrooks, line 41: **[Reliability – High] `Promise.all` couples software load to banks load — partial failure breaks the entire page**

If `getBanks()` fails transiently (the banks endpoint is a separate service path), the entire detail page becomes unavailable even though the software record loaded fine. Users lose access to all detail page functionality due to a failure in a non-critical data source.

Fetch independently so the banks combobox degrades gracefully:

```typescript
Api2.getSoftware(softwareId)
  .then((r) => { if (!isCancelled) setSoftware(r.data); })
  .catch((e: Error) => { if (!isCancelled) setLoadError(e.message); })
  .finally(() => { if (!isCancelled) setIsLoaded(true); });

Api2.getBanks()
  .then((r) => { if (!isCancelled) setBanks(r.data); })
  .catch(() => {
    if (!isCancelled) {
      setBanks([]);
      alert?.error('Bank list could not be loaded. The Add Bank option may be unavailable.');
    }
  });
```

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+      modalRef.current?.show({});
+    },
+    hide() {
+      modalRef.current?.hide();
+    },
+  }));
+
+  async function handleSubmit() {
+    try {
+      const response = await Api2.updateBankAssociationStatus(softwareId, bankId, status);
```
jamesobrooks: **[Architecture – High] Asymmetric API call ownership between the two modals**

`AddAssociatedBankConfirmModal` is a pure presenter — it calls `onConfirm` and the parent (`BankruptcySoftwareDetail`) owns the API call, global alert, and state update.

This modal does the opposite: it imports and calls `Api2.updateBankAssociationStatus` directly, manages its own `useGlobalAlert`, and hands back the result via `onSuccess`. This creates an inconsistency where:
- Confirm modal tests require no `Api2` setup
- Edit modal tests require `vi.spyOn(Api2, 'updateBankAssociationStatus')` in every test

It also removes the parent's ability to control async flow (e.g., optimistic updates, refreshing related state).

Suggested fix: lift the API call to the parent to match the confirm modal pattern. The modal receives `onSave(bankId: string, status: 'active' | 'inactive')`; the parent calls the API, handles the alert, and updates state — mirroring `handleAddBank`. The modal becomes a pure presenter.

user-interface/src/lib/components/uswds/Input.tsx
```diff
@@ -94,6 +131,20 @@ function Input_(props: InputProps, ref: React.Ref<InputRef>) {
...
+  useEffect(() => {
+    debounceIdRef.current = setTimeout(() => {
+      if (errorMessage) {
+        announceInvalid(errorMessage);
+      } else {
+        announceValid();
+      }
+    }, 400);
+    return () => clearTimeout(debounceIdRef.current);
+  }, [errorMessage, inputValue]);
```
jamesobrooks, line 146: **[Accessibility – Medium] `inputValue` in `useEffect` deps causes premature announcements on mount for pre-errored fields**

Two issues with the current dependency array `[errorMessage, inputValue]`:

1. **Premature mount announcement**: When `errorMessage` is set as an initial prop (e.g., the contact form loaded with existing invalid data), the 400ms debounce fires `announceInvalid` immediately on mount before the user has touched the field. Screen reader users hear error announcements on page load.

2. **Unnecessary timer churn**: `inputValue` in deps resets the 400ms timer on every keystroke. Since `announceInvalid`/`announceValid` are already transition-guarded, the re-fires are no-ops — just wasted timer allocation.

Suggested fix — gate on first interaction and remove `inputValue` from deps:

```typescript
const hasInteracted = useRef(false);

// In handleBlur:
function handleBlur() {
  hasInteracted.current = true;
  // ... existing logic
}

// In useEffect:
useEffect(() => {
  if (!hasInteracted.current) return;
  debounceIdRef.current = setTimeout(() => {
    if (errorMessage) announceInvalid(errorMessage);
    else announceValid();
  }, 400);
  return () => clearTimeout(debounceIdRef.current);
}, [errorMessage]); // inputValue removed
```

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <CamsTableHeaderCell>Status</CamsTableHeaderCell>
+          <CamsTableHeaderCell className="text-right"></CamsTableHeaderCell>
+        </CamsTableHeader>
+        <CamsTableBody>
+          {associations.map((association) => (
+            <CamsTableRow key={association.bankId} data-testid={`bank-row-${association.bankId}`}>
+              <CamsTableCell data-cell="Associated Bank">
+                <Link
+                  to={`/admin/banks/${association.bankId}`}
+                  target="bank-detail"
```
jamesobrooks: **[Accessibility – Medium] Named browsing context `target="bank-detail"` with misleading `aria-label`**

`target="bank-detail"` is a named browsing context — it opens in the same tab if a window named `bank-detail` already exists, or a new window otherwise. The `aria-label` says `"(opens in new tab)"` which is inaccurate.

Every other external link in this codebase uses `target="_blank" rel="noopener noreferrer"` (e.g., `NewTabLink`, `CaseNumber`). Use the project's existing `NewTabLink` component or match the standard pattern:

```tsx
<Link
  to={`/admin/banks/${association.bankId}`}
  target="_blank"
  rel="noopener noreferrer"
  aria-label={`${association.bankName} (opens in new tab)`}
>
```

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -0,0 +1,123 @@
...
+    <div className="associated-banks-section">
+      <h3>Associated Banks</h3>
+      <CamsTable id="associated-banks-table" aria-label="Associated banks">
+        <CamsTableHeader>
+          <CamsTableHeaderCell>Associated Bank</CamsTableHeaderCell>
+          <CamsTableHeaderCell>Status</CamsTableHeaderCell>
+          <CamsTableHeaderCell className="text-right"></CamsTableHeaderCell>
+        </CamsTableHeader>
+        <CamsTableBody>
+          {associations.map((association) => (
```
jamesobrooks, line 74: **[Accessibility – Medium] Empty table state not announced to screen readers**

When `associations` is empty, `CamsTableBody` renders with a header row and no body rows and no message. Screen reader users encountering `role="table"` with no rows get no indication that the list is intentionally empty.

Add an explicit empty-state row:

```tsx
<CamsTableBody>
  {associations.length === 0 ? (
    <CamsTableRow>
      <CamsTableCell colSpan={3} data-cell="">
        <span data-testid="no-associated-banks">No banks associated yet.</span>
      </CamsTableCell>
    </CamsTableRow>
  ) : (
    associations.map((association) => ( /* existing rows */ ))
  )}
</CamsTableBody>
```

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -12,6 +13,25 @@ import { useGlobalAlert } from '@/lib/hooks/UseGlobalAlert';
 import { getAppInsights } from '@/lib/hooks/UseApplicationInsights';
 import Api2 from '@/lib/models/api2';
 import { ComboOption } from '@/lib/components/combobox/ComboBox';
+import {
+  email as emailValidator,
```
jamesobrooks, line 17: **[Architecture – Medium] Validators imported from the wrong domain module**

`email` and `website` validators are imported from `@common/cams/trustees-validators`. These were written for the trustee domain, not the bankruptcy software vendor domain. Importing from a trustee-domain module into a bankruptcy-software form creates a cross-domain coupling that makes the architecture harder to read, and means the two domains cannot diverge their validation rules independently.

Suggested fix: move `email` and `website` validators to a domain-neutral location (e.g., `@common/cams/contact-validators` or `@common/cams/validators`) and import from there.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -84,14 +125,22 @@ export function SoftwareVendorContactInfoForm({
   }
 
   function handleExtensionChange(e: React.ChangeEvent<HTMLInputElement>) {
-    setExtension(e.target.value);
+    const digitsOnly = (e.target.value.match(/\d/g) ?? []).slice(0, 6).join('');
+    extensionRef.current?.setValue(digitsOnly);
```
jamesobrooks, line 138: **[Complexity – Medium] Triple-write for one state change**

```typescript
extensionRef.current?.setValue(digitsOnly);  // write to DOM via ref
setExtension(digitsOnly);                    // write to React state
e.target.value = digitsOnly;                 // mutate the native event
```

Three writes to express one value change. `setExtension(digitsOnly)` is the React state update that matters — the controlled input will display the correct value on re-render. Mutating `e.target.value` after the fact is a sign the normal React controlled-input contract isn't being relied on.

The `extensionRef` was introduced to replace `maxLength={6}`, but the two can coexist. Restoring `maxLength` collapses this back to one line:

```typescript
function handleExtensionChange(e: React.ChangeEvent<HTMLInputElement>) {
  setExtension((e.target.value.match(/\d/g) ?? []).slice(0, 6).join(''));
}
```

With `maxLength={6}` restored on the `<Input>` and `extensionRef` removed.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -84,14 +125,22 @@ export function SoftwareVendorContactInfoForm({
...
   function handleWebsiteChange(e: React.ChangeEvent<HTMLInputElement>) {
-    setWebsite(e.target.value);
+    const value = e.target.value;
+    setWebsite(value);
+    setWebsiteError(validateWebsiteValue(value));
   }
 
   async function handleSave() {
+    const hasEmailErrors = Object.values(emailErrors).some(Boolean);
+    if (hasEmailErrors || websiteError) return;
```
jamesobrooks: **[Low] Dead code — `handleSave` guard is unreachable**

The Save button has `disabled={Object.values(emailErrors).some(Boolean) || !!websiteError}`. If the button is disabled, `handleSave` can never be called via a click, making this early-return guard dead code. Delete it.

.snyk
```diff
@@ -30,3 +30,10 @@ ignore:
           Transitive dependency via launchdarkly-js-sdk-common, @azure/msal-node,
           and njwt. No fixed version available from upstream packages.
         expires: 2026-05-23T00:00:00.000Z
+  SNYK-JS-QS-16721866:
+    - '*':
+        reason: >-
+          Transitive dependency via express@5.2.1 and typed-rest-client@2.3.1.
```
jamesobrooks: **[Snyk policy – Medium] Rationale incorrectly names `typed-rest-client` as a production attack path**

`typed-rest-client` is a dev dependency of `@stryker-mutator/core` (mutation testing tooling) — it is not present in the production deployment artifact. Listing it alongside `express@5.2.1` misrepresents the actual production attack surface and may cause confusion at the next policy review.

Suggested correction:

```yaml
reason: >-
  Transitive dependency via express@5.2.1, which depends on qs@6.15.1.
  Fix available in qs@6.15.2 but express@5.2.1 has not released an update
  that pulls in the patched qs. No direct upgrade path available.
```

Also worth checking: the backend `package.json` already uses `overrides` for `@azure/identity`. It may be possible to add a targeted `qs` override within the express dependency tree to pull in the patch before the expiry date.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -51,17 +51,23 @@ export class BankruptcySoftwareController {
...
   ): Promise<CamsHttpResponseInit<BankruptcySoftwareProfile>> {
     this.requireSuperUser(context);
-    const body = context.request.body as Partial<
-      Pick<BankruptcySoftwareProfile, 'name' | 'status' | 'contact'>
-    > | null;
-    if (body?.name !== undefined && !body.name.trim()) {
-      throw new BadRequestError(MODULE_NAME, { message: 'Software name is required.' });
+    const body = context.request.body as Record<string, unknown> | null;
+
+    const addBankUpdate = this.parseAddBankBody(body);
```
jamesobrooks: **[Complexity – Medium] Three parse methods that could be one**

The three parse methods (`parseAddBankBody`, `parseUpdateBankAssociationBody`, `parseProfileUpdateBody`) each validate and return different shapes, then all dispatch to the same `this.useCase.updateSoftware` call. Merging them into a single `parseUpdate(body): SoftwareUpdate` method would reduce indirection, give the compiler a single return type to enforce (see the `SoftwareUpdate` export comment), and make `buildSuccessResponse` needed at exactly one call site:

```typescript
async handlePut(context, softwareId) {
  this.requireSuperUser(context);
  const update = this.parseUpdate(context.request.body as Record<string, unknown> | null);
  const software = await this.useCase.updateSoftware(softwareId, update);
  return this.buildSuccessResponse(context, software);
}
```

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.test.ts
```diff
@@ -176,6 +176,70 @@ describe('BankruptcySoftwareController', () => {
     });
   });
 
+  describe('handlePut with addBank', () => {
+    test('should return 200 when addBank body is valid', async () => {
+      context.request.params = { softwareId: 'sw-1' };
+      context.request.body = { addBank: { bankId: 'bank-1', bankName: 'Chase' } };
+      vi.spyOn(BankruptcySoftwareUseCase.prototype, 'updateSoftware').mockResolvedValue(
+        createdSoftware,
```
jamesobrooks, line 252: **[Low] Missing `ForbiddenError` test for `addBank` path**

The existing `handlePut` test suite has one `ForbiddenError` test using a profile-update body. The new `addBank` and `updateBankAssociation` describe blocks contain no authorization test. While `requireSuperUser` is correctly called at the top of `handlePut` before any branch, the absence of per-path tests means a future refactor moving the check inside each branch would introduce a regression with no failing test.

Add to each new describe block:
```typescript
test('should throw ForbiddenError for non-SuperUser', async () => {
  context.session.user.roles = [CamsRole.TrialAttorney];
  context.request.body = { addBank: { bankId: 'bank-1', bankName: 'Chase' } };
  await expect(controller.handlePut(context, 'sw-1')).rejects.toThrow(
    expect.objectContaining({ status: 403 }),
  );
});
```

user-interface/src/lib/components/uswds/Input.tsx
fmaddenflx: This is going to require some conversation.  What I was trying to achieve was covering a situation where there was already bad data in the database, which was then placed in the input fields, and if the user goes ahead and saves the form without changing the bad data, then it was getting resaved.  Perhaps there ought to be an additional check on save, that catches the errors, updates the notices on the fields, reads to those who can't see (voice over app) and prevents the save when the data is bad.  That ought to be in addition to the checks that happen as the user is typing (with a debounce).

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
fmaddenflx: Alright, let's go ahead with the common pattern used throughout the software.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetail.tsx
fmaddenflx: The software profile and banks list are both fetched from the same Cosmos DB instance. If one query fails, the other is very likely to fail as well. And if the software load fails, the page is useless regardless — there's nothing to display. Keeping `Promise.all` is the simpler approach for a scenario that would already result in a broken page either way.

user-interface/src/lib/components/uswds/Input.tsx
fmaddenflx: Addressed in 4c59d6424. Added save-time revalidation in the contact form: when Save is clicked, all emails and the website are re-validated. If anything fails, the error state is updated (triggering the Input component's useEffect screen reader announcements) and a global alert tells the user the save was blocked. This catches the pre-loaded bad data case without removing the mount-time announcement that already helps screen reader users.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -89,6 +91,137 @@ export class BankruptcySoftwareController {
...
+    if (!contact) return contact;
+
+    if (contact.emails) {
+      for (const addr of contact.emails) {
+        if (!EMAIL_REGEX.test(addr)) {
+          throw new BadRequestError(MODULE_NAME, {
+            message: 'One or more email addresses are invalid.',
+          });
+        }
+        if (addr.length > 254) {
```
jalba-flexion, line 193: **Bug: regex checked before length — oversized-but-valid address gets wrong error message**

The `EMAIL_REGEX.test(addr)` check at line 188 runs first. An address that is syntactically valid but longer than 254 characters passes the regex and never reaches the length check, so the client receives `'One or more email addresses are invalid.'` instead of the accurate `'Email address must not exceed 254 characters.'` message. Length should be checked before format:

```typescript
if (addr.length > 254) {
  throw new BadRequestError(MODULE_NAME, { message: 'Email address must not exceed 254 characters.' });
}
if (!EMAIL_REGEX.test(addr)) {
  throw new BadRequestError(MODULE_NAME, { message: 'One or more email addresses are invalid.' });
}
```

The same ordering issue exists for `website` (lines 201 vs 206).

backend/lib/use-cases/bankruptcy-software/bankruptcy-software.ts
```diff
@@ -95,4 +94,99 @@ export class BankruptcySoftwareUseCase {
...
+    } catch (auditError) {
+      this.context.logger.error(
+        MODULE_NAME,
+        'Audit record creation failed after successful data write.',
+        {
+          softwareId: id,
+          error: (auditError as Error).message,
+        },
+      );
+      throw getCamsError(
```
jalba-flexion, line 184: **Bug: throws after a successful write — client receives 500 for a change that already persisted**

Line 162 writes to Cosmos successfully. If audit record creation then fails, this `throw` causes the controller to return a 500. The client sees an error and may re-submit, resulting in a duplicate write and a second (stale) audit entry.

If audit integrity is mandatory, consider writing the audit record first and rolling back on failure rather than the current write-then-audit order. If audit failure should be non-fatal, swallow the error here (after logging) and return `updated` — the log is already there.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -51,26 +57,22 @@ export class BankruptcySoftwareController {
...
   }
 
   async handleGet(
     context: ApplicationContext,
   ): Promise<CamsHttpResponseInit<BankruptcySoftwareProfile[]>> {
     const softwareList = await this.useCase.getSoftwareList();
+    const isSuperUser = context.session.user.roles?.includes(CamsRole.SuperUser);
+    const data = isSuperUser
+      ? softwareList
+      : softwareList.map(({ contact: _contact, ...safe }) => safe);
```
jalba-flexion, line 72: **Possible data exposure: `associatedBanks` not stripped for non-superusers**

`contact` is correctly omitted for non-superusers, but `associatedBanks` (bank IDs and names) remains in the response. If bank association data is also intended to be superuser-only, add it to the destructure:

```typescript
softwareList.map(({ contact: _contact, associatedBanks: _banks, ...safe }) => safe)
```

If non-superusers are allowed to see bank associations, this is fine — just worth confirming.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetail.tsx
```diff
@@ -58,6 +71,54 @@ export function BankruptcySoftwareDetail() {
...
+
+  function handleEditBankStatus(
+    bankId: string,
+    bankName: string,
+    currentStatus: 'active' | 'inactive',
+  ) {
+    editBankStatusModalRef.current?.show(bankId, bankName, currentStatus);
+  }
+
+  async function handleSaveBankStatus(
```
jalba-flexion, line 101: **Missing loading guard — rapid double-click fires duplicate PUT and writes a duplicate audit record**

`handleAddBank` (line 75) guards with `if (!softwareId || isAddingBank) return` and flips `setIsAddingBank(true)` before the async call. `handleSaveBankStatus` only checks `if (!softwareId)` — there is no in-flight guard. A slow response or rapid double-click will fire two identical PUTs. The second PUT is idempotent for the status value but still creates a duplicate audit entry.

Add an `isSavingBankStatus` state the same way `isAddingBank` is used:

```typescript
const [isSavingBankStatus, setIsSavingBankStatus] = useState(false);

async function handleSaveBankStatus(bankId, bankName, status) {
  if (!softwareId || isSavingBankStatus) return;
  setIsSavingBankStatus(true);
  try { ... } finally { setIsSavingBankStatus(false); }
}
```

PR #2409, opened 2026-05-19
CAMS-751 Assign public/internal addresses using DISP_ON_WEB flags

btposey:
# Purpose

Assign public/internal addresses using DISP_ON_WEB flags

# Major Changes

The trustee migration was assigning public vs internal addresses based solely on whether A2 fields were present, ignoring the DISP_ON_WEB and DISP_ON_WEB_A2 display flags from ATS. When DISP_ON_WEB_A2 is 'y' the A2 address set should be the public one. When both flags share the same value, the trustee is ambiguous and cannot be resolved automatically — those records are accumulated and written as JSONL to the migration output container for manual review.

## Summary by Sourcery

Assign trustee public vs internal addresses based on ATS DISP_ON_WEB flags and surface ambiguous cases for manual review.

New Features:
- Use DISP_ON_WEB and DISP_ON_WEB_A2 flags to decide whether primary or A2 address fields are treated as the public trustee address, with case-insensitive handling.
- Detect trustees whose DISP_ON_WEB flags are ambiguous (both y or both N) and emit their details as JSONL files for manual review during migration.

Enhancements:
- Extend trustee transformation and ATS gateway mappings to include DISP_ON_WEB flags and derive internal address data from the non-public address fields.

Tests:
- Add unit tests covering flag-driven public/internal address assignment, ambiguous flag detection logic, and JSONL output behavior in trustee migration.


backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -17,6 +17,41 @@ import { UstpOfficeDetails } from '@common/cams/offices';
 import { ObjectStorageGateway } from '../gateways.types';
 const MODULE_NAME = 'MIGRATE-TRUSTEES-USE-CASE';
 
+export type AmbiguousFlagTrustee = {
+  trusteeId: number;
+  name: string;
+  condition: 'both-y' | 'both-n';
+  address: { street?: string; city?: string; state?: string; zip?: string };
+  addressA2: { street?: string; city?: string; state?: string; zip?: string };
```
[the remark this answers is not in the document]
btposey: ✅ Done. Added `dispOnWeb` and `dispOnWebA2` (raw flag values from `AtsTrusteeRecord`) to `AmbiguousFlagTrustee`. Also extracted `normalizeDispFlag` to `ats-mappings.ts` as a shared exported helper used by both `transformTrusteeRecord` and `detectAmbiguousFlagTrustees` — resolves the duplication concern.

Re the streaming/chunking suggestion: page size is hardcoded at 50 trustees in the dataflow handler, so the in-memory JSONL string is bounded and the overhead is negligible. Deferring that optimization.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -1007,6 +1059,29 @@ export async function processPageOfTrustees(
   };
 }
 
+async function writeAmbiguousFlagTrustees(
+  context: ApplicationContext,
+  ambiguous: AmbiguousFlagTrustee[],
+  outputContainerName: string,
+): Promise<void> {
+  const objectStorage: ObjectStorageGateway = factory.getObjectStorageGateway(context);
+  const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
```
btposey: Check `common` as I'm certain we have a date helper library for this.

btposey:
Scope increase required:

1. Remove the guard to import all trustees, not just ones with active appointments.
2. We need to write a list of all trustees that CANNOT be matched to a professional ID to a file in the storage container.

jalba-flexion (approved):

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -4,7 +4,10 @@ import {
   TrusteeAppointmentsResult,
   FailedAppointment,
 } from '../../adapters/types/ats.types';
-import { transformTrusteeRecord } from '../../adapters/gateways/ats/cleansing/ats-mappings';
+import {
+  transformTrusteeRecord,
+  normalizeDispFlag,
```
jamesobrooks: **[Architecture — High] OeSA Dependency Rule violation: use case imports from gateway implementation**

`migrate-trustees.ts` imports `normalizeDispFlag` directly from `adapters/gateways/ats/cleansing/ats-mappings.ts`. This is the only place in the codebase where a use-case module reaches into a gateway implementation file. Every other use-case-to-adapter relationship is mediated through `adapters/types/` (types) or `factory.ts` (composition root). The strategic layer now has a compile-time structural dependency on a tactical implementation module — if the ATS gateway is refactored or replaced, this use case breaks.

The same pre-existing import of `transformTrusteeRecord` on the same line has the same issue; this PR widens the breach by adding a second symbol.

**Recommended fix**: Move `normalizeDispFlag` to `adapters/types/ats.types.ts` (or a sibling `ats-flag-utils.ts` in that directory). Both `ats-mappings.ts` and `migrate-trustees.ts` can import it from the types layer, which use cases are already permitted to depend on. See also the related comment on `detectAmbiguousFlagTrustees` placement.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: CAMS follows Option-Enabling Software Architecture (OeSA). Use cases (strategic layer) must not
import from gateway implementation files (tactical layer). The current code in
backend/lib/use-cases/dataflows/migrate-trustees.ts imports normalizeDispFlag from
backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts — a Dependency Rule violation.

Task:
1. Move the normalizeDispFlag function (currently at ats-mappings.ts:180) to
   backend/lib/adapters/types/ats.types.ts. Keep it exported.
2. Update ats-mappings.ts to import normalizeDispFlag from ats.types.ts instead of defining it.
3. Update migrate-trustees.ts to import normalizeDispFlag from adapters/types/ats.types.ts
   instead of from ats-mappings.ts.
4. Remove the normalizeDispFlag import from the existing ats-mappings import line in
   migrate-trustees.ts (keep the transformTrusteeRecord import — that violation is pre-existing
   and out of scope for this task).
5. Run the backend tests to confirm nothing is broken: cd backend && npm test

Do not change any function logic or test assertions.
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -17,6 +20,45 @@ import { UstpOfficeDetails } from '@common/cams/offices';
 import { ObjectStorageGateway } from '../gateways.types';
 const MODULE_NAME = 'MIGRATE-TRUSTEES-USE-CASE';
 
+type AmbiguousFlagTrustee = {
```
jamesobrooks: **[Architecture — High] `detectAmbiguousFlagTrustees` and `AmbiguousFlagTrustee` belong in the adapter layer**

Both `detectAmbiguousFlagTrustees` and the `AmbiguousFlagTrustee` type operate entirely on raw ATS SQL column names (`DISP_ON_WEB`, `STREET_A2`, etc.) — these are tactical, source-system details with no CAMS business logic. They belong in `ats-mappings.ts` alongside the other ATS-specific transformations (`transformTrusteeRecord`, `normalizeDispFlag`), not in the use-case layer.

The signal: if the ATS schema changes column names, a function in the use-case layer breaks. Use cases should be insulated from that.

`AmbiguousFlagTrustee` being unexported from the module is the right intent but doesn't fix the cohesion problem — the type shouldn't exist in this file at all. It could move to `adapters/types/ats.types.ts` alongside `FailedAppointment`.

Additionally, `AmbiguousFlagTrustee` currently carries both a derived `condition: 'both-y' | 'both-n'` field **and** the raw `dispOnWeb`/`dispOnWebA2` values. `condition` is fully computable from the raw flags — the two are redundant. Recommend dropping `dispOnWeb`/`dispOnWebA2` from the type and keeping only `condition` (human-readable) alongside the address sets, which are the genuinely useful diagnostic data.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: CAMS uses OeSA. detectAmbiguousFlagTrustees in migrate-trustees.ts operates on raw ATS
record fields and should live in the adapter/cleansing layer, not the use-case layer.
AmbiguousFlagTrustee carries redundant fields (condition + raw flags).

Tasks:
1. Move the AmbiguousFlagTrustee type to backend/lib/adapters/types/ats.types.ts. Export it.
   Remove the redundant dispOnWeb and dispOnWebA2 fields — keep trusteeId, name, condition,
   address, and addressA2.

2. Move detectAmbiguousFlagTrustees to
   backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts. Keep it exported (it is tested).
   Update its import of normalizeDispFlag accordingly (it will now be a local call since both
   will be in the same file, or imported from ats.types.ts if normalizeDispFlag was already moved
   there per the other review comment).

3. In migrate-trustees.ts, update the import of detectAmbiguousFlagTrustees to come from
   ats-mappings.ts (or ats.types.ts if that is where it lands).

4. Remove the AmbiguousFlagTrustee type definition from migrate-trustees.ts.

5. Update migrate-trustees.test.ts: move the detectAmbiguousFlagTrustees describe block import
   to match the new location. Update any test assertions referencing dispOnWeb/dispOnWebA2
   (those fields are removed from the type in step 1).

6. Run all backend tests: cd backend && npm test
```
</details>

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts
```diff
@@ -177,6 +177,10 @@ export function formatZipCode(
   return cleanedZip;
 }
 
+export function normalizeDispFlag(flag: string | undefined): string | undefined {
```
jamesobrooks: **[YAGNI — Medium] `normalizeDispFlag` should not be exported; [Data quality — Medium] unexpected flag values silently produce wrong address assignment**

**Export**: `normalizeDispFlag` is a one-liner helper for `transformTrusteeRecord`. It was exported solely so `migrate-trustees.ts` could import it — which is itself a Dependency Rule violation (see the comment on line 9 of `migrate-trustees.ts`). If that violation is fixed by moving `normalizeDispFlag` to the types layer, this export can be dropped from `ats-mappings.ts`.

**Unexpected flag values**: After `trim().toLowerCase()`, any value that is not exactly `'y'` causes `a2IsPublic` to be `false`, silently treating non-A2 as public. Empty strings, `'yes'`, `'1'`, or other ATS data-quality issues will be silently swallowed — the trustee will not appear in the ambiguous-flag output (which only captures `'y'` and `'n'`) and will not trigger any warning. Given the extensive ATS data cleansing in this file, this silent default deserves an explicit warning log.

**Recommended fix**: Narrow the return type to `'y' | 'n' | undefined` and log a warning when a non-empty, non-y/n value is encountered.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: normalizeDispFlag in ats-mappings.ts silently treats any non-'y' value as 'not public',
including unexpected ATS data-quality values. This means corrupt/unexpected flag values produce
wrong public address assignment with no warning.

Tasks:
1. Narrow the return type of normalizeDispFlag to 'y' | 'n' | undefined:
   export function normalizeDispFlag(flag: string | undefined): 'y' | 'n' | undefined {
     if (flag === undefined || flag === null) return undefined;
     const normalized = flag.trim().toLowerCase();
     if (normalized === 'y') return 'y';
     if (normalized === 'n') return 'n';
     return undefined;
   }

2. In transformTrusteeRecord, after calling normalizeDispFlag for both flags, add a warning log
   (via context.logger if available, or console.warn as fallback) when a flag is present
   (non-undefined in the raw record) but normalizes to undefined (unexpected value). Include the
   raw value and field name in the log. If context is not available in this function signature,
   return the unexpected-flag case in the function's return value or use a separate validation
   step in the call site in migrate-trustees.ts.

3. Update the return type annotation in ats-mappings.ts and fix any TypeScript errors downstream
   (the consumers compare against 'y' and 'n' which still works with the narrowed type).

4. Add a test in ats-mappings.test.ts for normalizeDispFlag returning undefined for an unexpected
   value like 'yes'.

5. Run: cd backend && npm test
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -940,6 +982,16 @@ export async function processPageOfTrustees(
   // Build district-to-divisions map from office data
   const districtToDivisionsMap = buildDistrictToDivisionsMap(offices);
 
+  // Detect and log trustees with ambiguous display flags before deduplication
+  const ambiguousTrustees = detectAmbiguousFlagTrustees(trustees);
+  if (ambiguousTrustees.length > 0) {
+    context.logger.warn(
+      MODULE_NAME,
+      `${ambiguousTrustees.length} trustees have ambiguous DISP_ON_WEB flags and require manual review`,
```
jamesobrooks, line 963: **[Observability — High] No run-level count of ambiguous trustees; warn log missing structured fields**

Two issues on this code path:

**1. No aggregate count persisted.** Each page emits a per-page `warn` with the count, but nothing accumulates into `TrusteeMigrationState`. After log retention expires, there is no way to answer "how many trustees in this migration required manual review?" from the state document alone.

**2. Warn log missing structured context.** An operator triaging the next morning needs to know where the review file is and what the conditions mean. The current message omits the container name, file name, and a breakdown of `both-y` vs `both-n` (which have different operational meanings — one may expose a private address, the other may hide a public one). Without structured fields, this is not filterable in KQL.

Suggested structured fields:
```typescript
context.logger.warn(
  MODULE_NAME,
  `${ambiguousTrustees.length} trustees have ambiguous DISP_ON_WEB flags and require manual review`,
  {
    count: ambiguousTrustees.length,
    bothY: ambiguousTrustees.filter((t) => t.condition === 'both-y').length,
    bothN: ambiguousTrustees.filter((t) => t.condition === 'both-n').length,
    outputContainer: outputContainerName,
    actionRequired: 'MANUAL_REVIEW',
  },
);
```

The `actionRequired` field makes these findable in Application Insights with a simple KQL filter without parsing the message string.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: The migrate-trustees dataflow processes trustee records in pages. Each page logs a warn
when ambiguous-flag trustees are found, but (a) no aggregate count is persisted to migration
state, and (b) the warn log lacks structured fields needed for operator triage.

Tasks:
1. Update the warn log at migrate-trustees.ts:990 to include structured metadata:
   {
     count: ambiguousTrustees.length,
     bothY: ambiguousTrustees.filter((t) => t.condition === 'both-y').length,
     bothN: ambiguousTrustees.filter((t) => t.condition === 'both-n').length,
     outputContainer: outputContainerName,
     actionRequired: 'MANUAL_REVIEW',
   }

2. Add ambiguousCount to the processPageOfTrustees return object (alongside processed,
   appointments, errors). Set it to ambiguousTrustees.length. Search for MaybeData in this file
   to find the return type shape.

3. In the Azure Function entry point
   (backend/function-apps/dataflows/migrations/migrate-trustees.ts), accumulate ambiguousCount
   across pages alongside the other counters, and include the total in the migration completion
   log.

4. Update processPageOfTrustees tests in migrate-trustees.test.ts to assert the new
   ambiguousCount field in the return value for the ambiguous-flag test cases.

5. Run: cd backend && npm test
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -1007,6 +1059,29 @@ export async function processPageOfTrustees(
...
+  const content = ambiguous.map((r) => JSON.stringify(r)).join('\n');
+
+  try {
+    await objectStorage.writeObject(outputContainerName, fileName, content);
+    context.logger.info(
+      MODULE_NAME,
+      `Wrote ${ambiguous.length} ambiguous-flag trustees to ${outputContainerName}/${fileName}`,
+    );
+  } catch (originalError) {
+    context.logger.warn(MODULE_NAME, `Failed to write ambiguous-flag trustees file — continuing`, {
```
jamesobrooks: **[Reliability — High] Blob write failure silently loses ambiguous trustees with no durable signal**

If `writeObject` throws, the catch logs a warning and returns. The ambiguous trustees proceed through migration and no record of their existence survives the log retention window. For a one-time migration, this window does not reopen.

The current failure log (`Failed to write ambiguous-flag trustees file — continuing`) does not include which trustees were affected, making post-hoc reconstruction impossible.

**Minimum fix**: Include trustee IDs in the failure log:
```typescript
context.logger.warn(MODULE_NAME, `Failed to write ambiguous-flag trustees file — ${ambiguous.length} records may require manual review`, {
  error: getCamsError(originalError, MODULE_NAME).message,
  count: ambiguous.length,
  trusteeIds: ambiguous.map((r) => r.trusteeId),
  fileName,
});
```

Trustee IDs are integers (not names or addresses), so logging them does not expose PII but gives an operator enough information to look up the records manually.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: In writeAmbiguousFlagTrustees (migrate-trustees.ts), the catch block logs a warning when
the blob write fails but does not include which trustees were affected. In a one-time migration,
lost records have no recovery path once the migration run ends.

Tasks:
1. Update the catch block warn log in writeAmbiguousFlagTrustees to include:
   - count: ambiguous.length
   - trusteeIds: ambiguous.map((r) => r.trusteeId)  // integers only — no PII
   - fileName  (so the operator knows what file failed)

2. Add a test in migrate-trustees.test.ts that verifies: when writeObject rejects, (a)
   processPageOfTrustees does NOT throw, (b) the warn log is called with the trusteeIds array,
   and (c) the returned processed count is correct. Model it after the existing test
   'should continue and not throw when blob write fails' for writeFailedAppointments.

3. Run: cd backend && npm test
```
</details>

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts
```diff
@@ -195,14 +199,34 @@ export function transformTrusteeRecord(
   const computedName = computeTrusteeName(firstName, middleName, lastName);
   const fullName = computedName || 'Unknown';
 
+  const dispOnWeb = normalizeDispFlag(atsTrustee.DISP_ON_WEB);
+  const dispOnWebA2 = normalizeDispFlag(atsTrustee.DISP_ON_WEB_A2);
+  const a2IsPublic = dispOnWebA2 === 'y' && dispOnWeb !== 'y';
```
jamesobrooks, line 251: **[Design clarity — Low] `a2IsPublic` silent fallback for `both-y` is unexplained; [Test coverage — Low] missing tests for ambiguous-flag address routing**

**Clarity**: When both flags are `'y'`, `a2IsPublic` evaluates to `false` (`'y' === 'y' && 'y' !== 'y'`). This is a defensible default — fall back to non-A2 as public — but the logic is not self-evident. A reader has to work through the boolean to understand that the `both-y` case defaults to non-A2. Given that ambiguous-flag handling is the novel behavior this PR introduces, it deserves an explicit comment.

**Missing tests**: The `DISP_ON_WEB flag-driven address assignment` test block does not cover the `DISP_ON_WEB='y', DISP_ON_WEB_A2='y'` case. If someone refactors the `a2IsPublic` expression, this regression would be invisible. A `both-n` test (`DISP_ON_WEB='N', DISP_ON_WEB_A2='N'`) is also worth adding — `a2IsPublic` is `false` there too.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: The a2IsPublic expression in transformTrusteeRecord (ats-mappings.ts:204) silently
evaluates to false for both-y and both-n ambiguous cases, defaulting to non-A2 as public. There
is no test asserting this behavior.

Tasks:
1. Add a comment above the a2IsPublic line explaining the both-y fallback:
   // When both flags are 'y' (ambiguous), a2IsPublic is false: default to non-A2 as public.
   // Ambiguous records are separately captured by detectAmbiguousFlagTrustees for manual review.

2. Add two new test cases to the 'DISP_ON_WEB flag-driven address assignment' describe block in
   ats-mappings.test.ts:

   a) DISP_ON_WEB='y', DISP_ON_WEB_A2='y' (both-y): assert result.public.address.address1
      is the non-A2 street value ('1 Public St' from baseRecord).

   b) DISP_ON_WEB='N', DISP_ON_WEB_A2='N' (both-n): assert result.public.address.address1
      is the non-A2 street value ('1 Public St' from baseRecord).

3. Run: cd backend && npm test -- backend/lib/adapters/gateways/ats/cleansing/ats-mappings.test.ts
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -940,6 +982,16 @@ export async function processPageOfTrustees(
   // Build district-to-divisions map from office data
   const districtToDivisionsMap = buildDistrictToDivisionsMap(offices);
 
+  // Detect and log trustees with ambiguous display flags before deduplication
+  const ambiguousTrustees = detectAmbiguousFlagTrustees(trustees);
```
jamesobrooks: **[Data quality — Medium] Ambiguous detection runs pre-dedup — same trustee may appear in multiple output files**

`detectAmbiguousFlagTrustees` is called on the raw `trustees` array before `deduplicateTrusteesInPage` (line ~996). If a trustee appears multiple times within a page (the exact scenario page-level dedup handles), they will be written to the ambiguous-flag file multiple times. The same logical trustee can also appear in ambiguous-flag output from different pages if their records span a page boundary.

An operator reviewing the JSONL output cannot easily deduplicate without knowing this, and the output file gives no indication that duplicates may exist.

**Options**:
1. Move detection to after `deduplicateTrusteesInPage`, iterating over the deduplicated groups instead of the raw page. This ensures each logical trustee appears at most once per page's output.
2. If pre-dedup detection is intentional (to capture all raw flag combinations), add a comment warning consumers that records are not deduplicated and the same logical trustee may appear multiple times.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: In processPageOfTrustees (migrate-trustees.ts:986), detectAmbiguousFlagTrustees runs on
the raw trustees array BEFORE deduplicateTrusteesInPage. The same logical trustee can appear
multiple times in the JSONL output within a page or across pages.

Task:
1. Move the detectAmbiguousFlagTrustees call and the conditional writeAmbiguousFlagTrustees call
   to AFTER the deduplicateTrusteesInPage call (~line 996). Pass the deduplicated values:
   const deduplicatedTrustees = [...deduplicatedMap.values()].map(group => group[0]);
   const ambiguousTrustees = detectAmbiguousFlagTrustees(deduplicatedTrustees);

2. Verify the existing 'should write ambiguous-flag trustees to JSONL' tests still pass — the
   test data does not have duplicates so behavior should be unchanged.

3. Update the comment from '// Detect and log trustees with ambiguous display flags before
   deduplication' to '// Detect ambiguous-flag trustees from the deduplicated page'

4. Run: cd backend && npm test -- backend/lib/use-cases/dataflows/migrate-trustees.test.ts
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.ts
jamesobrooks: **[Reliability — Low] Timestamp filename has millisecond collision risk; [Code quality — Low] pre-existing dead alias in `writeFailedAppointments`**

**Filename collision**: `new Date().toISOString()` has millisecond resolution. If two pages are processed in rapid succession (both having ambiguous trustees), they produce the same file name and the second write silently overwrites the first. This same pattern exists in `writeFailedAppointments` at line ~1091. Using a stable, per-page value as the suffix eliminates the risk:
```typescript
// Use the last trustee ID from the page — unique per page, human-readable for correlation
const fileName = `ambiguous-flags-page-${ambiguous[ambiguous.length - 1].trusteeId}.jsonl`;
```

**Dead alias** (pre-existing, line ~1091 of `writeFailedAppointments`): `const outputContainer = outputContainerName;` assigns a parameter to a local variable with a shorter name, then uses the local name throughout. The new `writeAmbiguousFlagTrustees` written in this PR correctly uses the parameter directly — the contrast makes the alias in the older function more visible. Consider cleaning it up.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
Context: writeAmbiguousFlagTrustees and writeFailedAppointments in migrate-trustees.ts both use
new Date().toISOString() for blob file names, which has millisecond collision risk. The older
writeFailedAppointments also has a dead outputContainer alias.

Tasks:
1. In writeAmbiguousFlagTrustees (around line 1068), replace the timestamp-based file name
   with a page-keyed name. The function signature has the ambiguous array; use the first
   trusteeId as the page key:
   const fileName = `ambiguous-flags-page-${ambiguous[0].trusteeId}.jsonl`;

2. In writeFailedAppointments (~line 1091), remove the line:
   const outputContainer = outputContainerName;
   Replace all uses of outputContainer with outputContainerName in that function body.

3. Run: cd backend && npm test
```
</details>

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.test.ts
```diff
@@ -269,6 +270,18 @@ describe('ATS Mappings', () => {
     });
```
jamesobrooks, line 241: **[Spec quality — Medium] `parseTodStatus` hand-written tests duplicate the `test.each` tables**

The hand-written tests at lines 63–90 assert on codes (`P`, `PA`, `O`, `C`, `S`, `PI`, `PS`, `NP`, `VR`, `OD`, `1`, `3`, `8`, `9`) that are already exhaustively covered by the `test.each` tables at lines 103–160. When the lookup map changes, both sets must be updated — a maintenance burden that adds no coverage.

The tests worth keeping from that block:
- The case-insensitive test (lines 92–96) — the `test.each` tables only use uppercase
- The unknown-status defaults test (lines 98–101) — unique behavior not in the tables

**Recommendation:** Remove lines 63–90. Keep lines 92–101 and the whitespace test at line 157.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
In ats-mappings.test.ts, the parseTodStatus describe block has hand-written tests (lines ~63-90)
that duplicate coverage already provided by the test.each tables at lines ~103-160.

Task: Remove the duplicated hand-written test cases (lines ~63-90) from the parseTodStatus
describe block. Keep:
- The case-insensitive test (~lines 92-96): tests lowercase input, not covered by the tables
- The unknown-status defaults test (~lines 98-101): tests an unknown code, not in the tables
- The whitespace test (~line 157): tests trimming, not in the main tables

Run: cd backend && npm test -- lib/adapters/gateways/ats/cleansing/ats-mappings.test.ts
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.test.ts
```diff
@@ -1613,4 +1614,186 @@ describe('Migrate Trustees Use Case', () => {
...
+      DISP_ON_WEB: dispOnWeb,
+      DISP_ON_WEB_A2: dispOnWebA2,
+    });
+
+    test('should detect trustees where both flags are y', () => {
+      const trustees = [makeTrustee(1, 'y', 'y'), makeTrustee(2, 'y', 'N')];
+      const result = detectAmbiguousFlagTrustees(trustees);
+      expect(result).toHaveLength(1);
+      expect(result[0].trusteeId).toBe(1);
+      expect(result[0].condition).toBe('both-y');
```
jamesobrooks: **[Spec quality — Medium] `mockTrustee` fixture duplicated verbatim in 4 describe blocks**

The same `mockTrustee` object (identical shape: id, trusteeId, firstName, lastName, name, status, public.address, createdOn, updatedOn, updatedBy) is copy-pasted into four separate describe blocks. If the `Trustee` type gains a required field, all four copies must be updated.

**Recommendation:** Hoist to a module-level constant above the outer `describe`, or extract to a `createMockTrustee()` factory alongside the other test utilities.

---
<details>
<summary>💡 Coding agent prompt</summary>

```
In migrate-trustees.test.ts, the same mockTrustee object is defined verbatim in four separate
describe blocks (approximately lines 243, 467, 799, and 1373). This is pure duplication.

Task: Extract the mockTrustee constant to module scope (above the outer describe block) or to a
createMockTrustee() factory function. Replace all four inline definitions with references to the
shared constant/factory.

Run: cd backend && npm test -- lib/use-cases/dataflows/migrate-trustees.test.ts
All 63 tests should still pass.
```
</details>

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Fixed. Moved `normalizeDispFlag` to `@common/string-helper` as the more generic `parseYesNo`, which both layers are permitted to import. The OeSA violation for `transformTrusteeRecord` is pre-existing and noted as out of scope.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Fixed. Moved both `AmbiguousFlagTrustee` and `detectAmbiguousFlagTrustees` to `ats-mappings.ts` in the adapter/cleansing layer. Dropped the redundant `dispOnWeb`/`dispOnWebA2` raw fields from the original type — though in a subsequent change we replaced `condition` with those raw fields directly (see the Data quality comment below).

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts
btposey: Fixed both. `normalizeDispFlag` is gone — replaced by `parseYesNo` in `@common` with return type `'y' | 'n' | undefined`. For unexpected flag values: rather than logging at the call site, we extended `detectAmbiguousFlagTrustees` to also capture trustees with non-empty, non-y/n flag values and write them to the review JSONL — durable record rather than log-only.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Fixed both. Added `ambiguousCount` to `TrusteeMigrationState` and accumulated it across pages in the Azure Function. Updated the warn log with structured fields: `count`, `outputContainer`, and `actionRequired: 'MANUAL_REVIEW'`. Dropped `bothY`/`bothN` breakdown when we replaced `condition` with raw flag values.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Fixed. The catch block now logs `count`, `trusteeIds` (integers only), and `fileName` so an operator can identify and recover affected records. Test added to verify these fields appear in the warn log when `writeObject` rejects.

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts
btposey: Fixed both. Added the comment explaining the `both-y` fallback. Added `both-y` and `both-n` test cases to the `DISP_ON_WEB flag-driven address assignment` block asserting non-A2-as-public in both cases.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Fixed. Moved `detectAmbiguousFlagTrustees` to after `deduplicateTrusteesInPage`, iterating over deduplicated records. Each logical trustee now appears at most once per page's JSONL output.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: Fixed both. Replaced the timestamp-based filename with `ambiguous-flags-page-${ambiguous[0].trusteeId}.jsonl` — unique per page, no collision risk. Removed the dead `outputContainer` alias in `writeFailedAppointments`; `outputContainerName` is used directly throughout.

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.test.ts
btposey: Fixed. Removed the three hand-written test blocks (single-letter, two-letter, numeric codes) — all covered by the `test.each` tables. Kept the case-insensitive, unknown-status, and whitespace tests.

backend/lib/use-cases/dataflows/migrate-trustees.test.ts
btposey: Fixed. Extracted the four identical `mockTrustee` definitions to a module-level `MOCK_TRUSTEE` constant. The distinct one at line ~1497 (different trusteeId and name) was left in place.

PR #2410, opened 2026-05-19
CAMS-750 Add 429 handling to all dataflows and handle missed cases

jamesobrooks:
# Purpose

Dataflows were losing cases when Cosmos DB returned 429 (Too Many Requests) errors. The `sync-cases` dataflow was silently dropping syncs, and several other dataflows had no retry logic at all. This PR adds structured exponential-backoff 429 handling to every affected dataflow and adds the `handle-missed-division-changes` migration to recover cases that moved to a new Cosmos ID due to a division change.

# Major Changes

* **Shared `handleRateLimitRetry` utility** (`dataflows-rate-limit.ts`) — extracts the exponential backoff pattern (`2^n * 30s`, capped at 600s, 10-retry limit) into a single reusable function with DLQ routing on exhaustion. All callers supply an explicit `connectionString` and use `context.extraOutputs` (not `InvocationContext`) to satisfy the Good-Fences Rule.
* **429 handling wired into all import and migration dataflows** — `sync-cases`, `sync-deleted-cases`, `sync-trustee-appointments`, `resync-remaining-cases`, and `handle-missed-division-changes` all use the shared utility. The `ExportAndLoadCase.exportAndLoad` call in `resync-remaining-cases` was previously outside any try/catch, silently bypassing retry logic on 429 from the export path; this is now fixed.
* **Poison queue handlers** added to `sync-cases` and `sync-trustee-appointments` so Azure's automatic poison-queue delivery writes a structured `QUEUE_ERROR` to the DLQ rather than silently dropping the message.
* **Timer metrics 429 handling** — `sync-trustee-due-date-metrics` and `sync-trustee-notes-metrics` skip the current tick with a warning log when Cosmos returns 429 (timers self-heal on the next scheduled run).
* **`handle-missed-division-changes` migration** — new use case and Azure Function handler that reads a blob of case IDs, checks each one for a DXTR ID that has moved to a different Cosmos case, and queues division-change cleanup messages.
* **Bug template** added to `.github/ISSUE_TEMPLATE/`.

# Testing/Validation

Implemented using TDD. All 2478 backend tests pass. TypeScript type check is clean. Branch is rebased on main with CI passing.

Test coverage includes: exponential backoff formula, retry-limit exhaustion routing to DLQ, non-transient error re-throw, poison handler behavior, timer 429 skip, and the full case-check logic for the new use case.

# Notes

`function-apps/dataflows/` is excluded from coverage thresholds in `vitest.config.ts` by project convention. The new use case under `lib/use-cases/dataflows/` is within the covered path and meets the 90% threshold.

The `handle-missed-division-changes` migration is a one-time dataflow triggered manually by uploading a blob of case IDs. It is registered in `dataflows.ts` and wired to the existing `division-change-cleanup` FIX queue.

# Definition of Done:

## Summary by Sourcery

Add shared rate-limit retry handling across dataflows, improve error/poison handling and telemetry, and introduce a migration to recover missed division-change cases.

New Features:
- Introduce a shared exponential-backoff rate limit utility for Cosmos 429 handling used by multiple dataflows.
- Add the handle-missed-division-changes migration flow to detect and queue division-change cleanup for affected cases.
- Add a standard bug report issue template for GitHub.

Bug Fixes:
- Ensure sync-cases, resync-remaining-cases, and related flows no longer drop work when Cosmos returns 429 responses.
- Prevent silent loss of messages by wiring poison-queue handlers that send structured errors to DLQs.

Enhancements:
- Extend sync-cases, sync-deleted-cases, sync-trustee-appointments, and resync-remaining-cases to use consistent retry, DLQ routing, and telemetry behavior.
- Update trustee metrics timer functions to treat 429s as transient and skip the current tick with structured logging and tracing.
- Add targeted tests covering rate-limit behavior, DLQ routing, poison handling, and the new division-change migration logic.


jamesobrooks:
@sourcery-ai resolve

backend/function-apps/dataflows/migrations/handle-missed-division-changes.ts
```diff
@@ -0,0 +1,251 @@
...
+});
+
+const HANDLE_START = buildFunctionName(MODULE_NAME, 'handleStart');
+const HANDLE_CHECK = buildFunctionName(MODULE_NAME, 'handleCheck');
+const HANDLE_CHECK_POISON = buildFunctionName(MODULE_NAME, 'handleCheckPoison');
+const HTTP_TRIGGER = buildFunctionName(MODULE_NAME, 'httpTrigger');
+
+const BLOB_CONTAINER = process.env.CAMS_OBJECT_CONTAINER;
+if (!BLOB_CONTAINER) {
+  throw new Error('Missing required environment variable: CAMS_OBJECT_CONTAINER');
```
jalba-flexion: **Bug: module-level `throw` crashes the entire dataflows app if `CAMS_OBJECT_CONTAINER` is unset.**

This runs at import time. Since `dataflows.ts` imports this module, a missing env var causes the whole Azure Functions app registration to fail — every dataflow, not just this migration. Every other handler in this PR guards env vars inside the handler function at invocation time instead. Move this check inside `handleStart`.

backend/function-apps/dataflows/migrations/resync-cases-by-date.ts
```diff
@@ -0,0 +1,319 @@
...
+const HANDLE_ERROR = buildFunctionName(MODULE_NAME, 'handleError');
+const HANDLE_RETRY = buildFunctionName(MODULE_NAME, 'handleRetry');
+
+export async function handleStart(
+  message: ResyncCasesByDateStartMessage,
+  invocationContext: InvocationContext,
+) {
+  const connectionString = process.env.AzureWebJobsDataflowsStorage;
+  if (!connectionString) {
+    throw new Error('Missing required environment variable: AzureWebJobsDataflowsStorage');
```
jalba-flexion, line 68: **Dead code / misleading guard: `connectionString` is read and validated here but never used in `handleStart`.**

`handleStart` does not call `handleRateLimitRetry` and does not enqueue messages directly, so it has no use for this connection string. This guard appears to be a copy-paste from `handlePage`. If `AzureWebJobsDataflowsStorage` is absent, this will throw and suppress the actual work of `handleStart` even though that work would succeed without the variable. Remove the check, or move it to the point where the value is actually needed.

backend/function-apps/dataflows/migrations/resync-cases-by-date.ts
```diff
@@ -0,0 +1,319 @@
...
+    logger.info(MODULE_NAME, `Queued ${divisionChanges.length} division changes to FIX`);
+  }
+
+  const failedEvents = processedEvents.filter((event) => !!event.error);
+  if (failedEvents.length > 0) {
+    logger.warn(MODULE_NAME, `${failedEvents.length} events failed, sending to DLQ`);
+    invocationContext.extraOutputs.set(DLQ, failedEvents);
+  }
+
+  const successCount = processedEvents.length - failedEvents.length - divisionChanges.length;
```
jalba-flexion, line 196: **`successCount` can go negative if an event has both a `divisionChange` and an `error`.**

An event that satisfies both `event.divisionChange !== undefined` and `!!event.error` is counted in both `divisionChanges` and `failedEvents`, so it is subtracted twice. `successCount` goes negative and `documentsWritten` is reported as a negative number.

`sync-cases.ts` uses `processedEvents.length - failedEvents.length` only (division-change events count as successes). Consider whether that convention should apply here too, or add `Math.max(0, ...)` if the double-subtract is intentional.

backend/function-apps/dataflows/dataflows-rate-limit.ts
```diff
@@ -0,0 +1,95 @@
...
+    ...message,
+    retryCount: nextRetryCount,
+  };
+
+  logger.warn(
+    moduleName,
+    `Rate limited (429). Retrying in ${visibilityTimeout}s (attempt ${nextRetryCount}/${RATE_LIMIT_RETRY_LIMIT}). correlationId=${correlationId ?? 'n/a'} module=${moduleName}`,
+  );
+
+  if (!connectionString) {
```
jalba-flexion: **`connectionString` is validated after the warn log that says a retry was queued — the log fires even when the function is about to throw.**

If `connectionString` is empty, `logger.warn` on line 79 has already told operators "retrying in Xs", then this throws instead. That leaves a misleading log entry for a retry that was never enqueued. Move the `connectionString` guard (and the throw) above the `computeBackoffSeconds` / `logger.warn` block.

backend/function-apps/dataflows/migrations/resync-cases-by-date.ts
```diff
@@ -0,0 +1,319 @@
...
+  logger.info(MODULE_NAME, `Successfully retried to sync ${event.caseId}`);
+  completeDataflowTrace(context.observability, trace, MODULE_NAME, 'handleRetry', logger, {
+    documentsWritten: 1,
+    documentsFailed: 0,
+    success: true,
+    details: { disposition: 'retry-succeeded', retryCount: String(event.retryCount) },
+  });
+}
+
+function setup() {
```
jalba-flexion, line 294: **No HTTP trigger registered for this migration.**

The comparable `handle-missed-division-changes` migration exposes an HTTP POST endpoint via `buildStartQueueHttpTrigger` so it can be started by hitting a route. `resync-cases-by-date` takes a `fromDate` parameter in `handleStart` but registers no HTTP trigger in `setup()` — there is no way to invoke it without manually enqueuing a message to the `start` queue. If an HTTP trigger is intentional (as the `fromDate` parameter suggests), it is missing here.

jamesobrooks: I'm not in the habit of adding HTTP triggers to dataflows because they are unusable in USTP environments without exposing to the entire internet. Is there some skill indicating that we should be adding them?

backend/function-apps/dataflows/migrations/resync-cases-by-date.ts
jamesobrooks: I'm curious whether we're doing this elsewhere. The API has a healthcheck but we haven't implemented that for the dataflows app, so there isn't a natural check of the environment. Handling it here is a fail-fast approach, but maybe we should actually be performing validation through a healthcheck during deployment just like we do with the other apps.

backend/function-apps/dataflows/migrations/resync-cases-by-date.ts
btposey: Ditto this sentiment. I do NOT want to see HTTP triggers on dataflows anymore.

backend/function-apps/dataflows/migrations/resync-cases-by-date.ts
btposey: The implicit health check is the lack of activity from the dataflows. Better to have something monitoring AppInsights for abnormal logging.

btposey: MAYBE run a sanity check on env vars in the index.tsx for the dataflow on startup.

btposey (approved):
I think I'm good as is. There are some good suggestions that can be implemented at a later time, like:

Sourcery complexity comments (3–7 — withRateLimitAndTelemetry helper, createPoisonHandler factory, etc.)
   
Each handler open-codes a try/catch + handleRateLimitRetry branch + two completeDataflowTrace calls. Sourcery's proposed abstractions   (createRateLimitedRunner, makePoisonHandler) are reasonable.

PR #2434, opened 2026-05-27
CAMS-752 Add bank trustees tab with paginated API

bityogi:
# Purpose

Allow administrators to view which trustees use a given bank, directly from the bank detail page. This is the first slice of CAMS-752 (View Trustee usage of banks and software vendors).

# Major Changes

* **Backend:** New `GET /banks/{bankId}/trustees` endpoint with `BankTrusteesController`, QueryPipeline-based repository query using `contains` for array field lookup, and paginated response
* **Frontend:** New `BankDetailTrustees` component with loading/error/empty states, paginated trustee list with links, and "Trustees" nav link in the bank detail sidebar
* **Registration fix:** Added missing import in `index.ts` entry point so esbuild bundles the new function

# Testing/Validation

Implemented using TDD. Controller tests cover happy path, pagination params, NaN/negative fallback, error wrapping, and role authorization. Frontend tests cover loading spinner, error alert, trustee links, singular/plural count, empty state, and pagination rendering.

# Notes

Follows the same pattern as the existing software trustees tab (CAMS-669). Uses `QueryPipeline` `contains` condition which maps to MongoDB `$in` for querying trustee documents whose `banks` array includes the given bank ID.

# Definition of Done:

- [x] Code refactored for clarity: Developers can understand the work simply by reviewing the code
- [x] Dependency rule followed: More important code doesn't directly depend on less important code
- [x] Development debt eliminated: UX and code aligns to the team's latest understanding of the domain
- [x] No regressions: Changes do not cause regression in related or unrelated areas of the application

## Summary by Sourcery

Add a paginated API and admin UI tab to view trustees associated with a specific bank from the bank detail page.

New Features:
- Expose a GET /banks/{bankId}/trustees endpoint that returns paginated trustee summaries for a given bank, restricted to super users.
- Add a Trustees tab on the bank detail page showing a paginated, linkable list of trustees using the selected bank.

Enhancements:
- Extend bank and trustee use cases and repositories to support querying trustees by associated bank with pagination metadata.
- Wire the new bank trustees API into the shared API client layer and mocks for use in the frontend and tests.

Tests:
- Introduce backend controller tests covering authorization, pagination parameters, error handling, and happy-path responses for the new bank trustees endpoint.
- Add frontend tests for the new bank trustees tab covering loading, error, empty states, trustee links, count messaging, and pagination display.


backend/lib/controllers/bank-trustees/bank-trustees.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+      const limit =
+        Number.isFinite(parsedLimit) && parsedLimit > 0 ? parsedLimit : DEFAULT_SEARCH_LIMIT;
+      const parsedOffset = parseInt(context.request.query.offset as string);
+      const offset =
+        Number.isFinite(parsedOffset) && parsedOffset >= 0 ? parsedOffset : DEFAULT_SEARCH_OFFSET;
+
+      const result = await this.useCase.getTrusteesByBank(bankId, limit, offset);
+      const totalCount = result.metadata?.total ?? 0;
+      const currentPage = Math.floor(offset / limit) + 1;
+      const totalPages = Math.ceil(totalCount / limit);
```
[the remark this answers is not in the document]
bityogi: ✅ Fixed\! Extracted pagination calculation into a shared `calculatePagination` utility that clamps `totalPages` to a minimum of 1 when there are zero results. This also addresses the "centralize pagination math" suggestion.

user-interface/src/admin/banks/BankDetailTrustees.tsx
```diff
@@ -0,0 +1,123 @@
...
+        if (isCancelled) return;
+        setTrustees(response.data);
+        if (response.pagination) {
+          setPaginationValues(response.pagination);
+        }
+      })
+      .catch((err: Error) => {
+        if (isCancelled) return;
+        setError(err.message);
+      })
```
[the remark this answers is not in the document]
bityogi: Not implementing this one. All existing components in the codebase (BankruptcySoftwareDetailTrustees, BankDetailAuditHistory, etc.) follow the same pattern of surfacing `err.message` directly. The backend already sanitizes error messages before they reach the client via `getCamsError`, so internal details are not leaked. Changing only this component would introduce inconsistency.

bityogi:
Regarding the Sourcery feedback on `getBankTrustees` typing in `api2.ts`:

The generic parameter in `api().get<TrusteeSummary[]>(...)` specifies the *data payload* type, not the full response envelope. The `get` function returns `ResponseBody<TrusteeSummary[]>` which already includes `.data` (typed as `TrusteeSummary[]`) and `.pagination`. TypeScript correctly infers the full shape — no change needed here.

Addressed the other two suggestions:
- ✅ **Centralized pagination math** — Extracted into a shared `calculatePagination` utility in `backend/lib/controllers/pagination.ts`
- ✅ **totalPages clamp** — Returns `1` instead of `0` when there are zero results, keeping pagination semantics consistent for consumers

backend/function-apps/api/bank-trustees/bank-trustees.function.ts
fmaddenflx, line 1: I think for the bankruptcy-software trustees list we added it to the existing bankruptcy-software endpoint.  Maybe we can do the same with this, instead of creating a whole other endpoint?

fmaddenflx: I just noticed we have a separate one for bankruptcy-software, bankruptcy-software-history, banks, bank-history, bank-trustees.

Maybe this is worth a discussion with the whole group, but I feel like that is separating things out too much.  The azure functions are pretty simple boiler plate at this point, which is great.  But we should look at these in detail and see if they are all pretty much doing the same exact thing.

user-interface/src/lib/testing/mock-api2.ts
```diff
@@ -2979,6 +2979,13 @@ async function getBankHistory(_bankId: string) {
   return { data: [] };
 }
 
+async function getBankTrustees(_bankId: string, _limit?: number, _offset?: number) {
+  return {
+    data: [],
+    pagination: { count: 0, totalCount: 0, currentPage: 1, totalPages: 0, limit: 25 },
```
cruzjone-flexion: The mock returns `totalPages: 0` for the empty case, but `calculatePagination` explicitly returns `1` when `totalCount === 0` (verified
  by the new `pagination.test.ts`). The mock and the contract are out of sync.
  
  It doesn't cause a visible failure today because `BankDetailTrustees` only checks `totalPages > 1`, but any future test asserting on the
   empty-state `totalPages` value will get `0` from the mock and `1` from the real API.
   
  Suggest changing to `totalPages: 1` to keep the mock consistent with the server contract.

backend/function-apps/api/bank-trustees/bank-trustees.function.ts
fmaddenflx: I just verified that the handler for banks, bank-history, bank-trustees, bankruptcy-software, bankruptcy-software-history are all EXACTLY the same code with 1 exception -- bankruptcy-software.function.ts has 3 handlers in it and nameHandler() calls a different controller method and passes it the softwareId which comes off of the request.params.  However, since we pass context to the handler, and the request is in the context, we should be able to handle that in the controller.

So nameHandler could actually be identical to the others.

The only thing that differs in these modules is the app.http() calls, but since we can have multiple of those in each module (as seen in bankruptcy-software.function.ts which has 3), we don't need to have so many endpoint modules.

user-interface/src/admin/banks/BankDetailTrustees.tsx
```diff
@@ -0,0 +1,123 @@
...
+        if (isCancelled) return;
+        setIsLoading(false);
+      });
+
+    return () => {
+      isCancelled = true;
+    };
+  }, [bankId, searchPredicate]);
+
+  function retrievePage(predicate: SearchPredicate) {
```
cruzjone-flexion: `retrievePage` is a pure pass-through to `setSearchPredicate` with no additional logic:

  ```ts
  function retrievePage(predicate: SearchPredicate) {
    setSearchPredicate(predicate);
  }
  
  You can simplify by passing the setter directly:

  <Pagination ... retrievePage={setSearchPredicate} />

backend/function-apps/api/bank-trustees/bank-trustees.function.ts
bityogi: ✅ Done\! Consolidated the bank-trustees handler into `banks.function.ts` following the same pattern as `bankruptcy-software.function.ts` (multiple `app.http()` registrations in one file). Removed the standalone `bank-trustees/` directory and its import from `index.ts`. Agree this is cleaner — happy to discuss consolidating the other similar ones (bank-history, bankruptcy-software-history) as a follow-up.

user-interface/src/lib/testing/mock-api2.ts
bityogi: ✅ Fixed\! Updated the mock to return `totalPages: 1` to match the server contract established by `calculatePagination`.

user-interface/src/admin/banks/BankDetailTrustees.tsx
bityogi: ✅ Good call — removed the pass-through and passing `setSearchPredicate` directly to `<Pagination>`.

backend/function-apps/api/bank-trustees/bank-trustees.function.ts
fmaddenflx: The controller being used in each case is of course unique to the handler function, but maybe we can have a shared generic handler and we can pass it the controller?

fmaddenflx (approved):

cruzjone-flexion (approved):

PR #2436, opened 2026-05-27
CAMS-752 Styling and UI updates

cruzjone-flexion:
# Purpose

When an administrator changes a bank's status to inactive, warn them if trustees are currently using that bank so they understand the impact before proceeding.

# Major Changes

- Updated `EditBankModal` to intercept the submit when changing `active → inactive`: checks trustee count via the existing `GET /banks/{bankId}/trustees` API (with `limit=1` for minimal payload), shows the warning modal if count > 0, and proceeds directly if count is 0 or the check fails.
- "Proceed Anyway" completes the update and closes both modals. "Cancel" dismisses only the warning modal, leaving the edit modal open.
- Added requested design updates
- Added some additional missing UI features
- Improved test coverage front to back

# Testing/Validation

Implemented using TDD. 25 tests passing across both new/modified files:
- `InactivationWarningModal.test.tsx` — 8 unit tests: hidden by default, plural/singular/zero count messages, entity label, Proceed/Cancel callbacks, imperative hide
- `EditBankModal.test.tsx` — 17 tests: all pre-existing tests unchanged plus 7 new warning-flow tests covering: warning shown with count, no warning for zero trustees, no API call when already inactive, missing pagination fallback, Proceed Anyway closes both modals and shows success alert, Cancel keeps edit modal open, graceful fallback when count check throws

# Notes

The trustee count check calls `getBankTrustees(id, 1, 0)` — limit=1 is intentional since only `pagination.totalCount` is needed, not the actual records. If the count check throws (network error, etc.), the component silently proceeds with the update rather than blocking the user.

`InactivationWarningModal` is placed in `src/admin/shared/` because slices 3 and 4 will reuse it from different feature directories (`bankruptcy-software/` and `bankruptcy-software/`).

This is slice 2 of 4 for CAMS-752. Slice 1 (bank trustees tab + API) was merged in the previous PR.

# Definition of Done:

## Summary by Sourcery

Add a reusable inactivation warning modal and integrate it into the Edit Bank workflow to warn admins when deactivating banks in use by trustees.

New Features:
- Introduce a shared InactivationWarningModal component that displays trustee usage warnings for entities being inactivated.
- Add a bank inactivation flow that checks trustee usage before deactivating a bank and allows proceeding or cancelling from a warning dialog.

Enhancements:
- Refactor EditBankModal submission logic to separate the update operation from validation and pre-checks, and to surface success and error alerts.
- Extend EditBankModal tests and add dedicated tests for the InactivationWarningModal to cover the new warning behaviors and edge cases.


fmaddenflx:
When I left off last night there were 2 remaining design issues that hadn't beeen addressed:

1. The Add New Bankruptcy Software modal, when viewed using VoiceOver, is reading "dialog, with 4
      items, Software Name \*, required, edit text". We don't understand what "4 items" is
      referring to or why. It also doesn't read the modal title or the buttons. What does WCAG 2.1
      require or recommend? Let's get this working properly if possible.
2. Currently we can save the edit software modal without makeing changes. The result is that we
      get an entry in the audit logs that is blank (Update - - Flexion Team date). We should just
      disable the save button unless a change is made.
      
For (1) This looks like a more involved investigation and I made some headway doing that but decided to put it off last night.  We can address it today if we like but must keep in mind that it will affect probably all modals in the app.

For (2) I originally thought this was referring to a modal, but I believe it's referring to the full page form for entering "Vendor" information.

fmaddenflx:
We need to make sure we address the bug that we found yesterday where inactive software or banks were showing up as options for trustees in the comboboxes.  Any inactive Software or Bank should not be showing up as an option in any form anywhere outside of the admin control panel.

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -9,30 +10,69 @@ import {
...
+
+  useEffect(() => {
+    if (associations.length === 0) {
+      setCountsLoaded(true);
+      return;
+    }
+
+    setCountsLoaded(false);
+    const fetchCounts = async () => {
+      const counts: Record<string, number> = {};
```
fmaddenflx, line 57: I noticed that this useEffect looks to be getting fired every time any bank is activated or inactivated by the modal.  This is time consuming and throws up a loading state.  We probably shouldn't fetch the entire list every time the user updates something, or if we do, we should merge the changes so that we don't blank out the list every time.  We could do 1 of 2 things -- either just update the item that changed on the client-side, or do a full fetch but only update the items on the screen that actually change.

fmaddenflx: I'm looking into this now.

fmaddenflx:
Bug found in Add Software Vendor Contact Information with the Software Contact Phone field.  Deleted the first number in the area code and tried to type a new digit, but cursor jumped to the end of the number and put it at the end.
Screenshot 2026-05-28 at 1 46 03 PM

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -9,30 +10,68 @@ import {
...
+          } catch {
+            counts[association.bankId] = 0;
+          }
+        }),
+      );
+      setTrusteeCounts((prev) => ({ ...prev, ...counts }));
+      setCountsLoaded(true);
+    };
+
+    void fetchCounts();
```
mattstankey, line 83: Missing `isCancelled` (or `AbortController`) guard in `fetchCounts`. If the component unmounts while any of the `getSoftwareBankTrustees` calls are in flight, `setTrusteeCounts` and `setCountsLoaded` will fire on an unmounted component — producing a React warning and potentially updating stale state.

`SoftwareBankTrustees` handles this correctly with an `isCancelled` flag. The same pattern would work here:

```tsx
useEffect(() => {
  if (associations.length === 0) {
    setCountsLoaded(true);
    return;
  }

  let isCancelled = false;

  const fetchCounts = async () => {
    const counts: Record<string, number> = {};
    await Promise.all(
      associations.map(async (association) => {
        try {
          const response = await Api2.getSoftwareBankTrustees(softwareId, association.bankId, 1, 0);
          if (!isCancelled) counts[association.bankId] = response.pagination?.totalCount ?? 0;
        } catch {
          if (!isCancelled) counts[association.bankId] = 0;
        }
      }),
    );
    if (!isCancelled) {
      setTrusteeCounts((prev) => ({ ...prev, ...counts }));
      setCountsLoaded(true);
    }
  };

  void fetchCounts();
  return () => { isCancelled = true; };
}, [softwareId, associations]);
```

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -44,22 +83,29 @@ export function AssociatedBanksTable({
   }
 
   function handleAddBankClick() {
-    if (selectedBank) {
-      confirmModalRef.current?.show(selectedBank.value, selectedBank.label);
-    }
+    confirmModalRef.current?.show(selectedBank!.value, selectedBank!.label);
```
mattstankey: Non-null assertion replaces what was previously a guard. The old code was:

```tsx
if (selectedBank) {
  confirmModalRef.current?.show(selectedBank.value, selectedBank.label);
}
```

The new code relies on `!` — meaning if `handleAddBankClick` is somehow called when `selectedBank` is `null` (e.g. a future refactor mis-wires the `disabled` state, or a keyboard event fires before selection is confirmed), this will throw at runtime rather than silently no-op.

Suggest restoring the guard:
```tsx
function handleAddBankClick() {
  if (!selectedBank) return;
  confirmModalRef.current?.show(selectedBank.value, selectedBank.label);
}
```

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -147,26 +147,6 @@ export function SoftwareVendorContactInfoForm({
   }
 
   async function handleSave() {
```
mattstankey, line 149: The pre-submit re-validation block was removed entirely. Previously `handleSave` re-ran `validateEmailValue`, `validateWebsiteValue`, and `validateExtension` before calling the API, so a user who bypassed blur/change events (e.g. auto-fill, pasting, or navigating quickly) could not submit invalid data.

The current code relies solely on field-level validation firing on user interaction. If there is a scenario where those handlers are skipped, invalid values will reach the API.

Was this removal intentional? If field-level validation is now considered sufficient, a comment or test covering that guarantee would help future readers understand the intent.

user-interface/src/admin/bankruptcy-software/BankruptcySoftware.tsx
```diff
@@ -54,7 +54,6 @@ export function BankruptcySoftware() {
 
   return (
     <div className="bankruptcy-software-admin-panel" data-testid="bankruptcy-software-panel">
-      <h2 className="screen-reader-only">Bankruptcy Software</h2>
       {!isLoaded && <LoadingSpinner caption="Loading..." />}
```
mattstankey, line 58: The visually-hidden `<h2 className="screen-reader-only">Bankruptcy Software</h2>` was removed. This heading provided a landmark for screen reader users navigating the admin panel by heading structure.

Was this intentional? If so, please confirm there is an equivalent accessible heading or landmark that screen readers can use to orient within this section. The WCAG 2.1 AA requirement for heading hierarchy (2.4.6) and section labeling should still be satisfied.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
fmaddenflx: Yeah I was trying to improve test coverage and this was a block that seemed impossible to test. I evaluated the likelihood that this would ever get fired and decided it was an unnecessary dead block.  So I removed it.  But I was hesitant.  So if you disagree and think we should leave it in, then I'll reconsider.

user-interface/src/admin/bankruptcy-software/BankruptcySoftware.tsx
fmaddenflx: This was intentional and was agreed upon by the designer and a couple of us.  But you make a good point about landmark navigation.  We should bring it up again tomorrow and discuss.

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
fmaddenflx: This was addressed earlier — the fix is already in place.

user-interface/src/admin/banks/EditBankModal.tsx
```diff
@@ -24,11 +28,13 @@ type EditBankModalProps = {
 export const EditBankModal = forwardRef<EditBankModalRef, EditBankModalProps>(
   function EditBankModal({ modalId, bank, onSuccess }, ref) {
     const modalRef = useRef<ModalRefType>(null);
+    const warningModalRef = useRef<InactivationWarningModalRef>(null);
     const alert = useGlobalAlert();
 
     const [name, setName] = useState('');
     const [status, setStatus] = useState<'active' | 'inactive'>('active');
     const [nameError, setNameError] = useState<string | null>(null);
+    const pendingName = useRef('');
```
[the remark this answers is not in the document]
fmaddenflx: @sourcery if this is outdated and no longer needed, then close this thread.

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -9,30 +10,69 @@ import {
...
+        }),
+      );
+      setTrusteeCounts(counts);
+      setCountsLoaded(true);
+    };
+
+    void fetchCounts();
+  }, [softwareId, associations]);
 
   const associatedBankIds = new Set(associations.map((a) => a.bankId));
```
bityogi, line 89: The component fires one API call per associated bank to fetch trustee counts (`Promise.all` over `associations.map`). For a software with many associated banks, this becomes chatty (N+1 pattern). Consider adding a batch endpoint (e.g., `GET /bankruptcy-software/:id/banks/trustee-counts`) that returns all counts in a single round-trip, or at minimum let's discuss whether the expected data volumes make this acceptable as-is.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
mattstankey: I think we can leave it out, mostly just want to make sure it was intentionally removed.

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
bityogi: This issue will be addressed in a separate PR. An issue has been created for it here: https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/issues/2440

mattstankey (approved):

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.test.tsx
```diff
@@ -185,10 +217,112 @@ describe('AssociatedBanksTable', () => {
...
+
+  test('should display zero trustee count when API call fails', async () => {
+    vi.spyOn(Api2, 'getSoftwareBankTrustees').mockRejectedValue(new Error('Network error'));
+
+    renderTable([mockAssociations[0]]);
+
+    await waitFor(() => {
+      expect(screen.getByTestId('trustee-count-bank-1')).toHaveTextContent('0');
+    });
+  });
```
fmaddenflx: This is curious.  I'm not sure we shoudl display a count of trustees as 0 if an error occurs fetching the count.  We should probably indicate an error occured either in an alert, or by displaying "Error" or something in the count column.

user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
```diff
@@ -9,30 +10,75 @@ import {
...
+            const response = await Api2.getSoftwareBankTrustees(
+              softwareId,
+              association.bankId,
+              1,
+              0,
+            );
+            if (!isCancelled) counts[association.bankId] = response.pagination?.totalCount ?? 0;
+          } catch {
+            if (!isCancelled) counts[association.bankId] = 0;
+          }
```
fmaddenflx, line 71: This is probably a mis-representation of the fact that we couldn't retrieve the info.

PR #2439, opened 2026-05-29
CAMS-752 Sync bank status to software associations

cruzjone-flexion:
# Bug

When a bank profile was marked inactive, software profiles that referenced that bank retained active association statuses. There was also no guard preventing an API caller from manually reactivating a bank association while the bank remained inactive, and the UI showed the Edit Status button regardless of the bank's profile status.

# Purpose

Ensure that bank inactivation is consistently reflected across associated software profiles — automatically at the data layer, enforced at the API layer, and communicated clearly in the UI.

# Major Changes

- **Cascade (backend):** When a bank is set to inactive, all `SoftwareBankAssociation` entries referencing that bank are automatically set to inactive across every `BankruptcySoftwareProfile`, and an audit record is created per affected profile. No cascade occurs when a bank is set to active.
- **Guard (backend):** `BankruptcySoftwareUseCase.updateSoftware` now rejects any attempt to set a bank association status to `active` when the referenced bank profile is inactive, returning a 400 Bad Request. Setting an association to `inactive` is always permitted.
- **UI:** The Edit Status button in `AssociatedBanksTable` is hidden when the associated bank's profile status is inactive, preventing admins from attempting an action the API would reject.

# Testing/Validation

Implemented using TDD. All three changes are covered by unit tests:

- Cascade: 4 tests covering multi-profile fan-out, no cascade on active, empty result no-op, and `release()` called even when cascade throws
- Guard: 4 tests covering inactive bank rejects with correct message, active bank allows, setting to inactive skips bank lookup, and `release()` called on error
- UI: 3 tests covering hide when profile inactive, show when profile active, and mixed-row rendering

# Notes

- Cross-domain lookup pattern (use case → factory → repository) is consistent with `getTrusteesByBank` in the same codebase
- The `findSoftwareByBankId` query uses a `QueryBuilder` dot-notation cast (`'associatedBanks.bankId' as keyof BankruptcySoftwareProfile`) — a known limitation of the QueryBuilder for embedded array fields
- UI uses `bankProfileMap` (a `Map` keyed by bank id) derived from the existing `allBanks` prop — no new props or API calls needed
- Safe default: if a `bankId` is not found in `allBanks`, the Edit Status button is hidden

# Definition of Done:

## Summary by Sourcery

Cascade bank inactivation to associated bankruptcy software and enforce consistency across backend and UI when editing bank and association statuses.

Bug Fixes:
- Ensure software bank associations cannot be reactivated when the corresponding bank profile is inactive.
- Prevent users from editing bank association status in the UI when the underlying bank profile is inactive.
- Fix modals so cancel actions are ignored while a save is in progress, avoiding accidental closure during pending operations.

Enhancements:
- Cascade bank inactivation to all related bankruptcy software profiles and record software-level audit entries.
- Add repository support for querying bankruptcy software profiles by associated bank ID, including error handling.
- Improve bank and association edit modals by disabling submit and cancel buttons during async saves and re-enabling them on completion.

Tests:
- Add unit tests covering bank inactivation cascade behavior, including multi-profile updates, no-op scenarios, and error logging.
- Add unit tests validating guards on reactivating bank associations based on bank profile status.
- Add unit tests for the new repository query by bank ID and UI behavior for button disabling and Edit Status visibility.

### DIFF
--- a/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
@@ -14,6 +14,8 @@ const COLLECTION_NAME = 'bankruptcy-software';
 
 const { using, orderBy } = QueryBuilder;
 const doc = using<BankruptcySoftwareProfile>();
+// QueryBuilder does not support dot-notation into embedded arrays with typed keys;
+// use Record<string, unknown> to bypass the type constraint
 const docNested = using<Record<string, unknown>>();
 
 export class BankruptcySoftwareMongoRepository
--- a/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
@@ -14,6 +14,7 @@ const COLLECTION_NAME = 'bankruptcy-software';
 
 const { using, orderBy } = QueryBuilder;
 const doc = using<BankruptcySoftwareProfile>();
+const docNested = using<Record<string, unknown>>();
 
 export class BankruptcySoftwareMongoRepository
   extends BaseMongoRepository
@@ -72,7 +73,7 @@ export class BankruptcySoftwareMongoRepository
   async findSoftwareByBankId(bankId: string): Promise<BankruptcySoftwareProfile[]> {
     const query = QueryBuilder.and(
       doc('documentType').equals('BANKRUPTCY_SOFTWARE'),
-      (doc('associatedBanks.bankId') as unknown as ReturnType<typeof doc>).equals(bankId),
+      docNested('associatedBanks.bankId').equals(bankId),
     );
     try {
       return await this.getAdapter<BankruptcySoftwareProfile>().find(query);
--- a/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
@@ -69,6 +69,22 @@ export class BankruptcySoftwareMongoRepository
     }
   }
 
+  async findSoftwareByBankId(bankId: string): Promise<BankruptcySoftwareProfile[]> {
+    const query = QueryBuilder.and(
+      doc('documentType').equals('BANKRUPTCY_SOFTWARE'),
+      doc('associatedBanks.bankId' as keyof BankruptcySoftwareProfile).equals(bankId),
+    );
+    try {
+      return await this.getAdapter<BankruptcySoftwareProfile>().find(query);
+    } catch (originalError) {
+      throw getCamsError(
+        originalError,
+        MODULE_NAME,
+        'Unable to retrieve bankruptcy software by bank.',
+      );
+    }
+  }
+
   async updateSoftware(
     id: string,
     update: BankruptcySoftwareProfile,
--- a/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
@@ -72,7 +72,7 @@ export class BankruptcySoftwareMongoRepository
   async findSoftwareByBankId(bankId: string): Promise<BankruptcySoftwareProfile[]> {
     const query = QueryBuilder.and(
       doc('documentType').equals('BANKRUPTCY_SOFTWARE'),
-      doc('associatedBanks.bankId' as keyof BankruptcySoftwareProfile).equals(bankId),
+      (doc('associatedBanks.bankId') as unknown as ReturnType<typeof doc>).equals(bankId),
     );
     try {
       return await this.getAdapter<BankruptcySoftwareProfile>().find(query);
--- a/backend/lib/use-cases/banks/banks.ts
+++ b/backend/lib/use-cases/banks/banks.ts
@@ -70,41 +70,55 @@ export class BanksUseCase {
     }
 
     if (input.status === 'inactive') {
-      const softwareRepo = factory.getBankruptcySoftwareRepository(this.context);
-      try {
-        const affectedProfiles = await softwareRepo.findSoftwareByBankId(id);
-        for (const profile of affectedProfiles) {
-          const updatedProfile: BankruptcySoftwareProfile = {
-            ...profile,
-            associatedBanks: profile.associatedBanks.map((b) =>
-              b.bankId === id ? { ...b, status: 'inactive' } : b,
-            ),
-          };
-          await softwareRepo.updateSoftware(profile.id, updatedProfile);
-          await softwareRepo.createSoftwareAuditRecord(
-            createAuditRecord(
-              {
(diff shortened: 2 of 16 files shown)


jalba-flexion (approved):

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
```diff
@@ -35,11 +36,13 @@ export const EditBankAssociationStatusModal = forwardRef<
     },
   }));
 
-  function handleSubmit() {
-    onSave(bankId, bankName, status);
+  async function handleSubmit() {
+    setIsPending(true);
+    await onSave(bankId, bankName, status).finally(() => setIsPending(false));
   }
```
mattstankey, line 42: `onSave` is now typed as returning `Promise<void>`, and `handleSubmit` calls `.finally()` on the return value. If any caller passes a synchronous handler (returning `void`), this will throw at runtime. Can you confirm the parent component that wires up `onSave` has been updated (or was already returning a Promise)?

```ts
await onSave(bankId, bankName, status).finally(() => setIsPending(false));
//                                     ^^^^^^^^ runtime error if onSave returns void
```

backend/lib/adapters/gateways/mongo/bankruptcy-software.mongo.repository.ts
```diff
@@ -14,6 +14,7 @@ const COLLECTION_NAME = 'bankruptcy-software';
 
 const { using, orderBy } = QueryBuilder;
 const doc = using<BankruptcySoftwareProfile>();
+const docNested = using<Record<string, unknown>>();
```
mattstankey, line 19: `docNested` uses `Record<string, unknown>` as a workaround because `QueryBuilder` does not support dot-notation paths into embedded arrays via the typed `using<BankruptcySoftwareProfile>()` binding. Worth a one-line comment here so the next reader does not have to hunt through PR history to understand why there are two separate `using` bindings:

```ts
// QueryBuilder does not support dot-notation into embedded arrays with typed keys;
// use Record<string, unknown> to bypass the type constraint for nested field queries.
const docNested = using<Record<string, unknown>>();
```

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
mattstankey: It doesn't look like this got addressed in the latest commits still.

mattstankey:
There's still one comment that wasn't addressed.

We'll need to make sure we remove the CC co commit before merging, or make sure the commits get squashed.

backend/lib/use-cases/banks/banks.ts
```diff
@@ -63,11 +65,72 @@ export class BanksUseCase {
...
+              error: (profileError as Error).message,
+            },
+          );
+        }
+      }
+    } catch (originalError) {
+      throw getCamsError(
+        originalError,
+        MODULE_NAME,
+        'Unable to cascade bank inactivation to software.',
```
fmaddenflx: If `findSoftwareByBankId` throws here, the caller gets a 500 — but the bank update and its audit record already committed successfully. On retry the user would find the bank is already inactive, and since `existing.status != inactive` would now be false, the cascade wont re-run.

Consider logging and continuing (like the per-profile errors below) rather than throwing, so the primary operations success is reported to the caller. The cascade can be retried separately or noticed via monitoring.

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
```diff
@@ -35,11 +36,13 @@ export const EditBankAssociationStatusModal = forwardRef<
     },
   }));
 
-  function handleSubmit() {
-    onSave(bankId, bankName, status);
+  async function handleSubmit() {
```
fmaddenflx, line 39: The `onSave` type changed from `() => void` to `() => Promise<void>`. Sync callbacks are assignable to the async signature so existing callers compile fine, but if any caller wraps this in fire-and-forget (e.g. `void onSave(...)` without awaiting), the `isPending` state won't behave as expected. Worth confirming all callers of `EditBankAssociationStatusModal` pass an async-compatible handler that the modal can meaningfully await.

backend/lib/use-cases/banks/banks.ts
```diff
@@ -63,11 +65,72 @@ export class BanksUseCase {
...
+    userRef: ReturnType<typeof getCamsUserReference>,
+  ): Promise<void> {
+    const softwareRepo = factory.getBankruptcySoftwareRepository(this.context);
+    try {
+      const affectedProfiles = await softwareRepo.findSoftwareByBankId(bankId);
+      for (const profile of affectedProfiles) {
+        try {
+          const updatedProfile = this.inactivateBankAssociation(profile, bankId);
+          await softwareRepo.updateSoftware(profile.id, updatedProfile);
+          await softwareRepo.createSoftwareAuditRecord(
```
fmaddenflx, line 92: `inactivateBankAssociation` unconditionally sets the matched association to inactive, even if it already is. This means for an already-inactive association you still call `updateSoftware` + `createSoftwareAuditRecord`, producing a no-op audit entry (before: inactive → after: inactive).

Consider filtering to only process profiles where the association is currently active:

```ts
const affectedProfiles = (await softwareRepo.findSoftwareByBankId(bankId))
  .filter(p => p.associatedBanks.some(b => b.bankId === bankId && b.status === 'active'));
```

This avoids unnecessary writes and keeps the audit log meaningful.

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
cruzjone-flexion: The only caller is `handleSaveBankStatus` in `BankruptcySoftwareDetail.tsx`, which is already `async`. The prop type `() => Promise<void>` also enforces this at compile time, so any future sync caller would be a type error. No runtime risk here.

backend/lib/use-cases/banks/banks.ts
cruzjone-flexion: Good catch. Changed the outer catch in cascadeBankInactivationToSoftware to log-and-return instead of re-throwing, consistent with how per-profile errors are handled. A findSoftwareByBankId failure now logs and exits gracefully without surfacing a 500 or blocking the caller from seeing the successful bank update. Also added a test covering this scenario.

user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
cruzjone-flexion: The only caller is `handleSaveBankStatus` in `BankruptcySoftwareDetail.tsx:103`, which is `async` and properly awaited. No fire-and-forget usage exists.

backend/lib/use-cases/banks/banks.ts
cruzjone-flexion: Added a filter before the loop so only profiles where the target association is currently active are processed. Already-inactive associations are skipped, avoiding no-op writes and meaningless audit entries. Added a test covering this case.

mattstankey (approved):

PR #2446, opened 2026-06-01
CAMS-760 Migrate GHA workflows to OIDC least-privilege

jamesobrooks:
# Purpose

Eliminate long-lived \`AZURE_CREDENTIALS\` service principal secret by migrating all GitHub Actions Azure authentication to OIDC Workload Identity Federation. Each workflow that authenticates to Azure now uses a dedicated federated identity scoped to its GitHub environment, with Key Vault Secrets User RBAC granted only on the specific secrets it needs.

# Major Changes

* Applied \`permissions: read-all\` top-level default to all 27 workflows with scoped per-job blocks
* Migrated 11 Azure-authenticated workflows from \`creds: \${{ secrets.AZURE_CREDENTIALS }}\` to OIDC (\`client-id\` / \`tenant-id\` / \`subscription-id\`) with \`environment:\` anchors on every OIDC job
* Removed PGP-encrypted resource group name passing (\`cloudposse/github-action-secret-outputs\`); each job fetches secrets directly from Azure Key Vault after OIDC login
* Removed \`azResourceGrpApp\`, \`azResourceGrpNetwork\`, and \`slotName\` from workflow outputs — GitHub strips masked values from \`\$GITHUB_OUTPUT\`, making per-job KV fetch the correct pattern
* Migrated \`azure-remove-branch.yml\` RG names (\`AZ_APP_RG\`, \`AZ_NETWORK_RG\`, \`AZ_ANALYTICS_RG\`, \`AZURE_RG\`) from GitHub secrets to Key Vault fetch, matching the pattern used in all other workflows
* Replaced \`secrets: inherit\` with explicit secret declarations throughout the \`continuous-deployment.yml\` reusable workflow chain (11 files), eliminating over-sharing of unrelated secrets to callee workflows
* Added shell injection hardening: all \`\${{ inputs.* }}\`, \`\${{ github.* }}\`, and \`\${{ secrets.* }}\` values used in \`run:\` shell blocks are now routed through step-level \`env:\` blocks across 15+ workflow files
* Added 9 idempotent runbook scripts in \`ops/scripts/utility/\` for provisioning all OIDC identities with least-privilege RBAC (one per logical workflow identity, following \`setup-security-scan-federated-credential.sh\` from CAMS-730)
* Added \`ensure_role_assignment()\` helper to \`_oidc-helpers.sh\` for idempotent role assignment
* Granted subscription-scope \`Contributor\` to the deploy identity for both the main and branch environments, applied identically. (An earlier iteration used a custom \`CAMS Deploy Subscription Role\` plus \`Contributor\` scoped to the four known resource groups, but that is incompatible with branch deployments, whose resource groups are created dynamically per-hash at deploy time and cannot be pre-scoped — Azure RBAC has no wildcard scoping. Rather than diverge main from branch, both share the same grant. Tightening this to true least privilege is tracked as focused follow-up cams-y8s2.)
* Replaced \`User Access Administrator\` on the KV resource group with a custom role (\`CAMS KV Role Assignment Operator\`) granting only \`Microsoft.Authorization/roleAssignments/write/read/delete\`, scoped to the KV resource itself — eliminates privilege escalation surface while satisfying the Bicep \`kv-setup-module\` requirement
* Made \`az-delete-branch-resources.sh\` resilient to partial prior cleanup — warns on missing resources, skips already-deleted items, fails only if deletion leaves resources behind

# Testing/Validation

Full CD pipeline validated through slot deployment against a branch environment. Standalone DAST, E2E, and cleanup workflows exercised manually. Each RBAC gap surfaced by CI was fixed in the runbook scripts and re-validated by re-running them.

Ran workflows:

- [x] [Continuous deployment](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/27294420372)
- [x] [Standalone E2E](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/27306503393)
- [x] [Standalone DAST](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/27306340079)
- [x] [Branch resource cleanup](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/27308419225)
- [x] PR Validation — runs automatically

# Notes

**Masked-output architecture**: The root cause of the hardest-to-diagnose failures was that GitHub Actions strips \`::add-mask::\`'d values from \`\$GITHUB_OUTPUT\`, so downstream jobs received empty strings. Removing RG names and slot names from job outputs and having each job fetch directly from Key Vault is both more secure and architecturally correct.

**After merge, once all branches have rebased onto main, delete the following GitHub Actions secrets and variables** (no longer referenced in any workflow — either superseded by OIDC, migrated to Key Vault, or otherwise removed):

| Name | Type | Reason |
|---|---|---|
| \`AZURE_CREDENTIALS\` | Secret | Superseded by OIDC — zero workflow references remain |
| \`ANALYTICS_WORKSPACE_ID\` | Secret | Migrated to Key Vault as \`AZ-ANALYTICS-WORKSPACE-ID\` |
| \`AZ_APP_RG\` | Secret | Migrated to Key Vault as \`AZ-APP-RG\` |
| \`AZ_NETWORK_RG\` | Secret | Migrated to Key Vault as \`AZ-NETWORK-RG\` |
| \`AZURE_RG\` | Secret | Migrated to Key Vault as \`AZURE-RG\` |
| \`AZ_ANALYTICS_RG\` | Secret | Migrated to Key Vault as \`AZ-ANALYTICS-RG\` |
| \`AZ_PRIVATE_DNS_ZONE_ID\` | Secret | No longer referenced in any workflow |
| \`AZ_PRIVATE_DNS_ZONE_RG\` | Secret | No longer referenced in any workflow |
| \`AZ_STOR_VERACODE_KEY\` | Secret | No longer referenced in any workflow |
| \`AZ_STOR_VERACODE_NAME\` | Secret | No longer referenced in any workflow |
| \`CAMS_REACT_SELECT_HASH\` | Secret | No longer referenced in any workflow |
| \`MSSQL_DATABASE_DXTR\` | Secret | No longer referenced in any workflow |
| \`MSSQL_HOST\` | Secret | No longer referenced in any workflow |
| \`MSSQL_TRUST_UNSIGNED_CERT\` | Secret | No longer referenced in any workflow |
| \`AZ_HOSTNAME_SUFFIX\` | Variable | No longer referenced in any workflow |
| \`AZ_PRIVATE_DNS_ZONE\` | Variable | No longer referenced in any workflow |

**\`e2e\` dual-caller**: \`reusable-e2e.yml\` is called from both \`e2e-test.yml\` (standalone) and \`continuous-deployment.yml\` (CD gate). The runbook provisions two federated credentials per app registration.

**\`build-info\` triple-caller**: \`reusable-build-info.yml\` is called from \`continuous-deployment.yml\`, \`dast-scan.yml\`, and \`e2e-test.yml\`. The runbook provisions three credentials per app registration.

# Definition of Done:


diabagatekelly (approved):

.github/workflows/azure-remove-branch.yml
```diff
@@ -9,13 +9,19 @@ on:
         default: ''
         type: string
 
+permissions: read-all
+
 jobs:
   list:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
mattstankey, line 19: The `list` job uses OIDC-style parameters (`client-id`, `tenant-id`, `subscription-id`) on the `azure/login` step (line 20), but is missing `id-token: write` in its permissions block. Without that permission, the runner cannot request an OIDC token and the login step will fail.

```yaml
permissions:
  contents: read
  id-token: write   # required for OIDC login
```

The PR notes describe the `list` job as intentionally using `AZURE_CREDENTIALS` without OIDC — but the code was updated to OIDC-style parameters, so either the intent or the implementation needs to be reconciled.

mattstankey:
**Correction to my earlier review:** I incorrectly flagged `build-azure-cli-image.yml` as missing `packages: write`. On closer inspection, the `build-and-push` job already has `packages: write` in its per-job permissions block. That issue does not exist — only the `azure-remove-branch.yml` `list` job issue (noted inline above) is real.

.github/workflows/azure-remove-branch.yml
jamesobrooks: ✅ Already addressed — `id-token: write` is present on the `list` job (and all three jobs) in the current branch. The PR notes from the early draft were stale; the OIDC migration is fully complete in this PR.

ops/scripts/utility/_oidc-helpers.sh
```diff
@@ -0,0 +1,141 @@
...
+}
+
+# Set a secret in a GitHub Actions environment using the gh CLI.
+# Requires gh CLI authenticated with secrets:write scope.
+set_github_environment_secret() {
+  local ENVIRONMENT="$1"
+  local SECRET_NAME="$2"
+  local VALUE="$3"
+  echo "==> Setting GitHub secret $SECRET_NAME in environment '$ENVIRONMENT'..."
+  if gh secret set "$SECRET_NAME" \
```
diabagatekelly, line 113: **TECH DEBT: Secret Exposure via Command-Line**

Secret value passed via `--body "$VALUE"` is visible in process listings (`ps aux`) during execution.

**Fix:** Use stdin instead:
```bash
printf '%s' "$VALUE" | gh secret set "$SECRET_NAME" \
  --env "$ENVIRONMENT" \
  --repo "${GITHUB_ORG}/${GITHUB_REPO}"
```

Low risk in practice (scripts run locally by authorized admins), but violates security best practices.

.github/workflows/reusable-deploy.yml
```diff
@@ -25,41 +19,88 @@ on:
...
-          op: decode
-          in: ${{ inputs.azResourceGrpAppEncrypted }}
-
-      - uses: cloudposse/github-action-secret-outputs@67ae2f7ee1890986ba0012bacd079a5a36ecb449 # main
-        id: rgNetwork
-        with:
-          secret: ${{ secrets.PGP_SIGNING_PASSPHRASE }}
-          op: decode
-          in: ${{ inputs.azResourceGrpNetworkEncrypted }}
+      - name: Fetch config from Key Vault
```
diabagatekelly, line 55: **TECH DEBT: DRY Violation - Key Vault Fetching Pattern**

This "Fetch config from Key Vault" block is duplicated across 20 workflow files, creating maintenance burden and inconsistency risk.

**Recommendation:** Extract to composite action (`.github/actions/fetch-kv-secrets/action.yml`) or shared shell script to:
- Centralize KV fetching logic
- Ensure consistent secret masking across all workflows
- Parameterize hardcoded KV names (`kv-ustp-cams`, `kv-ustp-cams-dev`) via GitHub variables

Not blocking, but should be prioritized for next iteration to reduce future maintenance burden.

ops/scripts/utility/setup-deploy-federated-credential.sh
```diff
@@ -0,0 +1,163 @@
...
+  local SP_ID
+  SP_ID=$(lookup_or_create_sp "$APP_ID")
+
+  echo "==> Updating federated identity credential..."
+  upsert_federated_credential "$APP_ID" "$CREDENTIAL_NAME" "$SUBJECT"
+
+  # ---------------------------------------------------------------------------
+  # Role assignments
+  #
+  # Contributor at subscription scope: this identity deploys Bicep infrastructure
```
diabagatekelly: **TECH DEBT: Subscription-scoped Contributor RBAC**

Good documentation here on why this needs subscription scope (chicken-and-egg with RG names in KV). The justification is sound for this PR.

**Suggested refactoring path for next iteration:**
1. Pre-create resource groups (manual or separate privileged workflow)
2. Store RG names as GitHub variables instead of KV secrets
3. Grant Contributor only at RG scope: `/subscriptions/{sub}/resourceGroups/{rg}`
4. Optionally create custom role with minimal permissions

**Mitigations already in place:**
- GitHub environment protection rules
- Short token lifetime (10 minutes)
- Subject claim restricts to specific repo/workflow/environment
- Azure AD audit logs

Not blocking - compensating controls are solid and the trade-off is well-documented.

diabagatekelly:
## Deep PR Review Summary

### Inline Comments Posted

I've added **5 inline review comments** on specific lines where improvements can be made:

1. **POSSIBLE BLOCKER** (line 81 of `_oidc-helpers.sh`): JSON injection vulnerability in `upsert_federated_credential`
2. **TECH DEBT** (line 108 of `_oidc-helpers.sh`): Secret exposure via command-line args
3. **TECH DEBT** (line 36 of `reusable-deploy.yml`): DRY violation - KV fetching duplicated 20 times
4. **TECH DEBT** (line 99 of `setup-deploy-federated-credential.sh`): Subscription-scoped Contributor (well-documented trade-off)

### Additional Items (not in diff, can't add inline comments)

5. **TECH DEBT**: Missing idempotency in VNET disconnection at `az-delete-branch-resources.sh:150-160` - script will fail if resources were partially deleted in previous cleanup run
6. **FUTURE IMPROVEMENT**: Duplicate echo statement at `az-delete-branch-resources.sh:156,159`

.github/workflows/reusable-build-frontend.yml
```diff
@@ -33,27 +33,51 @@ on:
...
+            echo "AZ_APP_RG=${AZ_APP_RG_BASE}" >> "$GITHUB_ENV"
+          else
+            echo "AZ_APP_RG=${AZ_APP_RG_BASE}-${{ inputs.environmentHash }}" >> "$GITHUB_ENV"
+          fi
+          _val=$(az keyvault secret show --vault-name "$KV" --name SLOT-NAME --query value -o tsv 2>/dev/null || true)
+          if [[ -n "${_val}" ]]; then echo "::add-mask::${_val}"; fi
+          echo "SLOT_NAME=${_val}" >> "$GITHUB_ENV"
+          _val=$(az keyvault secret show --vault-name "$KV" --name CAMS-LOGIN-PROVIDER --query value -o tsv)
+          echo "::add-mask::${_val}"
+          echo "CAMS_LOGIN_PROVIDER=${_val}" >> "$GITHUB_ENV"
```
fmaddenflx, line 90: Are we echoing a secret here to the terminal?

fmaddenflx: AI answered my question to my satisfaction, but if you have more info, it's appreciated.

AI response:

No, this doesn't leak in its intended environment. This is the standard GitHub Actions pattern for dynamic secret masking:
   
  1. The secret is captured into a shell variable (command substitution doesn't echo to stdout).
  2. ::add-mask:: is a workflow command — the Actions runner intercepts this line, strips it from visible logs, and masks all future occurrences of that value with ***.

  Caveat: If someone ran this script outside of GitHub Actions (e.g., in a local terminal or a CI system that doesn't understand :: commands), it would print the secret in cleartext. But within
  GHA, the runner processes the command before it hits the log output.

ops/scripts/utility/_oidc-helpers.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
+# Shared helpers for OIDC federated credential runbook scripts.
+# Source this file from individual runbooks; do not execute it directly.
+#
```
fmaddenflx, line 4: This is interesting and makes me wonder, if this script isn't going to be executed directly, then perhaps we don't need the hashbang on line 1.  I imagine if it's going to be read and processed by another script, then maybe it doesn't need a hash-bang?  I'm not sure about that... just curious.  Point being that it might help prevent someone from running it directly without specifying to have bash run the file directly.

.github/workflows/continuous-deployment.yml
```diff
@@ -14,59 +14,80 @@ on:
...
+permissions: read-all
+
 jobs:
   setup:
     name: Setup
     uses: ./.github/workflows/reusable-build-info.yml
+    permissions:
+      contents: read
+      id-token: write
     secrets: inherit # pragma: allowlist secret
```
diabagatekelly: **TECH DEBT: `secrets: inherit` Over-Sharing**

This workflow passes ALL secrets to the reusable workflow via `secrets: inherit`, regardless of what it actually needs. This violates the principle of least privilege for secrets.

**Why it matters:**
- If the reusable workflow is compromised, it can access secrets it doesn't need
- Makes secret usage opaque (can't tell what secrets are actually used)
- Increases blast radius of secret-related security issues

**Recommendation:** Replace with explicit secret declarations:
```yaml
uses: ./.github/workflows/reusable-build-info.yml
secrets:
  AZ_CLIENT_ID: ${{ secrets.AZ_CLIENT_ID }}
  AZ_TENANT_ID: ${{ secrets.AZ_TENANT_ID }}
  AZ_SUBSCRIPTION_ID: ${{ secrets.AZ_SUBSCRIPTION_ID }}
  # Only list what this workflow actually uses
```

And add explicit `secrets:` block to the `workflow_call` definition.

**Note:** 17 workflows currently use this pattern. Consider tracking as a follow-up beads issue for systematic fix.

.github/workflows/reusable-endpoint-test.yml
```diff
@@ -39,23 +41,48 @@ jobs:
...
+          _val=$(az keyvault secret show --vault-name "$KV" --name AZ-APP-RG --query value -o tsv)
+          echo "::add-mask::${_val}"
+          AZ_APP_RG_BASE="${_val}"
+          if [[ "${GHA_ENVIRONMENT}" == "Main-Gov" ]]; then
+            echo "AZ_APP_RG=${AZ_APP_RG_BASE}" >> "$GITHUB_ENV"
+          else
+            echo "AZ_APP_RG=${AZ_APP_RG_BASE}-${{ inputs.branchHashId }}" >> "$GITHUB_ENV"
+          fi
+          if [[ -n "${{ inputs.slotNameOverride }}" ]]; then
+            echo "SLOT_NAME=${{ inputs.slotNameOverride }}" >> "$GITHUB_ENV"
```
mattstankey: `${{ inputs.slotNameOverride }}` is interpolated directly into the shell command here, which is inconsistent with the injection hardening applied elsewhere in this PR (where `${{ inputs.* }}` values are assigned to `env:` vars first and referenced as `${VAR}` in bash).

The callers today pass literal strings (`'self'`, `'initial'`), so there is no current exploitability — but it is worth fixing for consistency and to prevent a future caller from introducing a risk.

Suggested fix:

```yaml
- name: Fetch config from Key Vault
  env:
    GHA_ENVIRONMENT: ${{ inputs.ghaEnvironment }}
    SLOT_NAME_OVERRIDE: ${{ inputs.slotNameOverride }}
  run: |
    ...
    if [[ -n "${SLOT_NAME_OVERRIDE}" ]]; then
      echo "SLOT_NAME=${SLOT_NAME_OVERRIDE}" >> "$GITHUB_ENV"
    else
      ...
    fi
```

ops/scripts/utility/_oidc-helpers.sh
jamesobrooks: Risk accepted. These runbook scripts require both Azure CLI credentials and GitHub CLI credentials configured for the USTP org — only developers on the team can execute them, so the blast radius of a process listing exposure is limited to our controlled environment.

.github/workflows/reusable-deploy.yml
jamesobrooks: Good catch. Each workflow fetches a different set of secrets with per-secret conditional logic (optional vs. required, `-$hash` suffix handling), and GitHub Actions does not support array inputs cleanly, so a composite action abstraction would be awkward. Tracking the investigation in cams-59k1 — if the abstraction is more trouble than it is worth we will close it out. Not holding up this PR.

.github/workflows/reusable-endpoint-test.yml
fmaddenflx: Yeah makes sense.

fmaddenflx (approved):

ops/scripts/utility/setup-deploy-federated-credential.sh
jamesobrooks: Good call. After deeper analysis, subscription-scope Contributor is broader than necessary. The minimum required is a custom role granting only `Microsoft.Resources/deployments/*` + `Microsoft.Resources/resourceGroups/write` + `Microsoft.Resources/resourceGroups/read` at subscription scope (covering `az deployment sub create` and RG creation), plus `Contributor` scoped to each of the four known RGs for `az deployment group create`.

The chicken-and-egg concern (RG names fetched from KV at runtime) does not block this approach — a subscription-scope custom role by definition does not reference specific RG names, so there is nothing to bootstrap.

One important constraint: these workflows only apply to Flexion environments, not USTP/FedRAMP, so this is an engineering quality concern rather than a compliance requirement. Tracking in cams-o1bk to complete before this PR merges.

ops/scripts/utility/_oidc-helpers.sh
jamesobrooks: Good instinct, but omitting the shebang would not actually prevent direct execution — `bash _oidc-helpers.sh` still runs it, and without a shebang a `chmod +x` + direct invocation would just use the default shell, which could be worse. The shebang is worth keeping for shellcheck and editor tooling.

The real enforcement should be in the script body itself. A guard like this at the top actually prevents direct execution rather than just signaling it:

```bash
if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
  echo "ERROR: This script must be sourced, not executed directly." >&2
  exit 1
fi
```

Will add that.

.github/workflows/continuous-deployment.yml
jamesobrooks: Confirmed — `secrets: inherit` is a pass-all mechanism. GitHub docs are explicit: "pass all of the calling workflow's secrets to a directly called workflow." There is no filtering based on what the downstream workflow references.

The highest-risk over-shared secrets are `AZURE_CREDENTIALS` (legacy service principal JSON with a client_secret — redundant now that OIDC is in place), `BOT_PRIVATE_KEY`, `BOT_PASSPHRASE`, `ADMIN_KEY`, and `AZ_STOR_VERACODE_KEY`. None of these are referenced in any of the five reusable workflows called here.

Tracking in cams-j4qc to complete before this PR merges. The work includes auditing whether `AZURE_CREDENTIALS` can be deleted entirely (it should be superseded by OIDC), and replacing `secrets: inherit` with explicit declarations once the broader secrets cleanup lands.

jamesobrooks:
Thanks for the thorough review. Responding to each item:

1. **JSON injection in `upsert_federated_credential`**: Risk accepted — these runbook scripts require both Azure CLI and GitHub CLI credentials configured for the USTP org; only developers on the team can execute them.

2. **Secret exposure via command-line (`--body "$VALUE"`)**: Risk accepted for the same reason as above — local execution by authorized team members only.

3. **DRY violation — KV fetching pattern**: Tracked in cams-59k1 for investigation. Each workflow fetches a different set of secrets with per-secret conditional logic, and GitHub Actions does not support array inputs cleanly, so the abstraction may be more trouble than it is worth. Will close the bead if that turns out to be the case.

4. **Subscription-scoped Contributor**: After deeper analysis, the hypothesis that this is unnecessary was disproved. A custom role granting only `Microsoft.Resources/deployments/*` + `Microsoft.Resources/resourceGroups/write` at subscription scope is the correct minimum. Tracked in cams-o1bk to complete before merge.

5. **Missing idempotency in VNET disconnection**: Already addressed — the VNET disconnection block is guarded by `if [[ "${rgAppExists}" == "true" ]]`, so it is skipped if the app RG was already deleted in a prior partial run.

6. **Duplicate echo at lines 156/159**: Good catch — line 159 should read `"Completed disconnecting VNET integration for dataflows"`. Will fix.

.github/workflows/reusable-endpoint-test.yml
jamesobrooks: Fixed — moved `slotNameOverride` to an `env:` block and referenced it as `${SLOT_NAME_OVERRIDE}` in bash, consistent with the injection hardening applied elsewhere in this PR.

jamesobrooks: Good catch — and there are more. A full audit found 15 instances across 10+ files where `${{ inputs.* }}` or `${{ github.* }}` values are still interpolated directly into `run:` blocks rather than routed through `env:` blocks first. High-risk cases include `sub-build.yml` (`OUT=${{ inputs.apiFunctionName }} npm run` — no quoting at all), `reusable-infrastructure-deploy.yml` (inputs inside a JMESPath `--query` expression), `azure-remove-branch.yml` (unquoted ternary expression fed into `az group list --query`), `reusable-deploy.yml` (inputs inside a single-quoted `--parameters` string), `reusable-build-frontend.yml` (unquoted `export` assignments and `az` CLI args), `reusable-accessibility.yml` (`pushd ${{ inputs.path }}`), and `reusable-e2e.yml` (inline `TARGET_HOST` assignment). Tracking in cams-pf1w to fix all instances before this PR merges.

jalba-flexion (approved):

PR #2448, opened 2026-06-01
CAMS-752 Sync bank status to software associations

cruzjone-flexion:
# Bug

When a bank profile was marked inactive, software profiles that referenced that bank retained active association statuses. There was also no guard preventing an API caller from manually reactivating a bank association while the bank remained inactive, and the UI showed the Edit Status button regardless of the bank's profile status.

# Purpose

Ensure that bank inactivation is consistently reflected across associated software profiles — automatically at the data layer, enforced at the API layer, and communicated clearly in the UI.

# Major Changes

- **Cascade (backend):** When a bank is set to inactive, all `SoftwareBankAssociation` entries referencing that bank are automatically set to inactive across every `BankruptcySoftwareProfile`, and an audit record is created per affected profile. No cascade occurs when a bank is set to active.
- **Guard (backend):** `BankruptcySoftwareUseCase.updateSoftware` now rejects any attempt to set a bank association status to `active` when the referenced bank profile is inactive, returning a 400 Bad Request. Setting an association to `inactive` is always permitted.
- **UI:** The Edit Status button in `AssociatedBanksTable` is hidden when the associated bank's profile status is inactive, preventing admins from attempting an action the API would reject.

# Testing/Validation

Implemented using TDD. All three changes are covered by unit tests:

- Cascade: 4 tests covering multi-profile fan-out, no cascade on active, empty result no-op, and `release()` called even when cascade throws
- Guard: 4 tests covering inactive bank rejects with correct message, active bank allows, setting to inactive skips bank lookup, and `release()` called on error
- UI: 3 tests covering hide when profile inactive, show when profile active, and mixed-row rendering

# Notes

- Cross-domain lookup pattern (use case → factory → repository) is consistent with `getTrusteesByBank` in the same codebase
- The `findSoftwareByBankId` query uses a `QueryBuilder` dot-notation cast (`'associatedBanks.bankId' as keyof BankruptcySoftwareProfile`) — a known limitation of the QueryBuilder for embedded array fields
- UI uses `bankProfileMap` (a `Map` keyed by bank id) derived from the existing `allBanks` prop — no new props or API calls needed
- Safe default: if a `bankId` is not found in `allBanks`, the Edit Status button is hidden

# Definition of Done:

## Summary by Sourcery

Ensure bank inactivation is consistently enforced and communicated across bankruptcy software associations and admin UI workflows.

Bug Fixes:
- Prevent reactivation of software–bank associations when the underlying bank profile is inactive.
- Ensure software associations are automatically inactivated when a bank profile is set to inactive, avoiding stale active links.

Enhancements:
- Add repository support and use-case logic to fan out bank inactivation to all affected bankruptcy software profiles with auditing and error-tolerant logging.
- Update admin UI modals to handle in-progress saves by disabling submit/cancel actions and preventing accidental closure during pending requests.
- Hide the Edit Status control for software–bank associations whose bank profiles are inactive and refine trustee count display behavior in the associations table.

Tests:
- Expand backend unit tests to cover bank inactivation cascading, association reactivation guards, and repository queries by bank ID.
- Add frontend tests for association table filtering, trustee count link rendering, and button state behavior in bank and association status modals.

### DIFF
--- a/user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
+++ b/user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
@@ -11,7 +11,7 @@ export type EditBankAssociationStatusModalRef = {
 
 export type EditBankAssociationStatusModalProps = {
   modalId: string;
-  onSave: (bankId: string, bankName: string, status: 'active' | 'inactive') => void;
+  onSave: (bankId: string, bankName: string, status: 'active' | 'inactive') => Promise<void>;
 };
 
 export const EditBankAssociationStatusModal = forwardRef<
@@ -22,6 +22,7 @@ export const EditBankAssociationStatusModal = forwardRef<
   const [bankId, setBankId] = useState('');
   const [bankName, setBankName] = useState('');
   const [status, setStatus] = useState<'active' | 'inactive'>('active');
+  const [isPending, setIsPending] = useState(false);
 
   useImperativeHandle(ref, () => ({
     show(id: string, name: string, currentStatus: 'active' | 'inactive') {
@@ -35,11 +36,13 @@ export const EditBankAssociationStatusModal = forwardRef<
     },
   }));
 
-  function handleSubmit() {
-    onSave(bankId, bankName, status);
+  async function handleSubmit() {
+    setIsPending(true);
+    await onSave(bankId, bankName, status).finally(() => setIsPending(false));
   }
 
   function handleCancel() {
+    if (isPending) return;
     modalRef.current?.hide();
   }
 
@@ -50,10 +53,12 @@ export const EditBankAssociationStatusModal = forwardRef<
       label: 'Save',
       onClick: handleSubmit,
       closeOnClick: false,
+      disabled: isPending,
     },
     cancelButton: {
       label: 'Cancel',
       onClick: handleCancel,
+      disabled: isPending,
     },
   };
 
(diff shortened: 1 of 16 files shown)


user-interface/src/admin/bankruptcy-software/EditBankAssociationStatusModal.tsx
```diff
@@ -35,11 +36,13 @@ export const EditBankAssociationStatusModal = forwardRef<
     },
   }));
 
-  function handleSubmit() {
-    onSave(bankId, bankName, status);
+  async function handleSubmit() {
+    setIsPending(true);
+    await onSave(bankId, bankName, status).finally(() => setIsPending(false));
```
mattstankey: **BLOCKER: Unhandled promise rejection in `handleSubmit`**

`onSave(bankId, bankName, status).finally(...)` does not catch a rejection from `onSave`. Because `handleSubmit` is `async` and React treats `onClick` as fire-and-forget (the returned promise is discarded), a rejection becomes an unhandled promise rejection at runtime — `isPending` is reset correctly via `finally`, but the error is silently swallowed and never surfaced to the user.

`EditBankModal.tsx` already handles this correctly. Recommend following the same pattern here:

```typescript
async function handleSubmit() {
  setIsPending(true);
  try {
    await onSave(bankId, bankName, status);
  } catch (error) {
    // surface via alert or logging — see EditBankModal pattern
    alert?.error("Failed to save bank association status. Please try again.");
  } finally {
    setIsPending(false);
  }
}
```

(Sourcery also flagged this with a similar suggestion.)

mattstankey (approved):

diabagatekelly (approved):

PR #2458, opened 2026-06-02
Fix React act() warnings and knip CI warnings

mattstankey:
# Bug

Several unit test files produced `not wrapped in act(...)` warnings in CI logs due to React state updates occurring outside the testing library's act() scope. Two AppointmentCard tests were failing silently because a feature flag was never mocked to true, preventing the components under test from rendering. The knip dependency checker also flagged unused exports, unlisted dependencies, and missing binary entries.

# Purpose

Eliminate noisy act() warnings from CI test output and fix two broken unit tests, so CI logs reflect only genuine failures and all tests pass cleanly.

# Major Changes

* Wrapped imperative modal ref calls (`.show()` / `.hide()`), `window.dispatchEvent()` calls, and async handler invocations in `act()` across 9 test files
* Fixed two AppointmentCard tests by mocking `DISPLAY_CHPT7_PANEL_UPCOMING_KEY_DATES` to `true` so `UpcomingKeyDates` and `PastKeyDates` components actually render
* Removed `export` from 4 locally-scoped types in `dxtr-acms.mock.ts` that were not used outside the file
* Switched bare `common/src/` imports in `DatePicker` and `DateRangePicker` to the `@common` alias used everywhere else in the codebase
* Added `stryker` binary and `@stryker-mutator/core` to knip ignore lists; scoped `vitest` ignore to the root workspace

# Testing/Validation

All 3084 tests pass (10 pre-existing failures on main confirmed unrelated). Knip, lint, and typecheck all pass clean after changes.

# Notes

The act() warnings fall into three patterns:
1. **Imperative ref calls** (`.show()` / `.hide()`) — wrapped in synchronous `act()`
2. **`window.dispatchEvent()`** — wrapped in synchronous `act()` since event handlers synchronously update React state
3. **Async handler calls that trigger re-fetch** — wrapped in `await act(async () => { await handler(...) })`

The AppointmentCard fix was non-obvious: the feature flag defaults to `false` when `FEATURE_FLAG_SDK_KEY` is empty (as in CI), so the child components are never mounted and `waitFor` times out looking for their test IDs.

# Definition of Done:

## Summary by Sourcery

Resolve React testing warnings and CI noise by updating tests, feature flag usage, and shared config.

Bug Fixes:
- Ensure AppointmentCard key date tests render gated components by enabling the relevant feature flag in test setup.

Enhancements:
- Align DatePicker and DateRangePicker imports to use the shared @common alias instead of direct common/src paths.
- Limit type visibility in dxtr-acms test utilities by making internal row types file-local instead of exported.

Build:
- Update knip configuration to ignore specific Stryker-related entries and scope vitest ignores at the workspace level to eliminate false-positive dependency warnings.

Tests:
- Wrap imperative modal calls, window events, and async handlers in React testing act() across multiple modal, hook, and notes adapter tests to prevent unscoped state update warnings.
- Adjust RemovalModal tests to perform submit interactions within act() to better reflect React update semantics.

### DIFF
--- a/common/src/cams/test-utilities/dxtr-acms.mock.ts
+++ b/common/src/cams/test-utilities/dxtr-acms.mock.ts
@@ -1,7 +1,7 @@
 import { faker } from '@faker-js/faker';
 import { CourtDivisionDetails } from '../courts.js';
 
-export type AoDxtrCsRow = {
+type AoDxtrCsRow = {
   CS_CASEID: string;
   COURT_ID: string;
   CS_DIV: string;
@@ -12,7 +12,7 @@ export type AoDxtrCsRow = {
   CS_DATE_FILED: string;
 };
 
-export type AoDxtrPyRow = {
+type AoDxtrPyRow = {
   CS_CASEID: string;
   COURT_ID: string;
   PY_ROLE: string;
@@ -24,7 +24,7 @@ export type AoDxtrPyRow = {
   PY_ZIP: string;
 };
 
-export type AcmsCmmprRow = {
+type AcmsCmmprRow = {
   PROF_CODE: number;
   GROUP_DESIGNATOR: string;
   PROF_LAST_NAME: string;
@@ -34,7 +34,7 @@ export type AcmsCmmprRow = {
   EIN: string | null;
 };
 
-export type AcmsCmmapRow = {
+type AcmsCmmapRow = {
   RECORD_SEQ_NBR: number;
   CASE_DIV: string;
   CASE_YEAR: number;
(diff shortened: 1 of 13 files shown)


btposey (approved):

mattstankey:
Both suggestions implemented:

1. **NoteFormModal.test.tsx** — Extracted `openNoteModal(overrides?)` and `closeNoteModal()` helpers. `openNoteModal` merges against `DEFAULT_SHOW_OPTIONS` so each test only specifies what differs from the baseline.

2. **AppointmentCard.test.tsx** — Moved the `vi.spyOn(featureFlagsHook, 'default')` setup into a `beforeEach` inside a nested `describe('when DISPLAY_CHPT7_PANEL_UPCOMING_KEY_DATES flag is enabled')` block scoping it to only the two key-dates tests.

btposey (approved):

PR #2463, opened 2026-06-02
CAMS-593 Add trustee case list with seed data for E2E testing

jalba-flexion:
# Purpose

Allow TrusteeAdmin users to view a paginated list of cases assigned to a trustee directly from the trustee detail view, supporting the oversight workflow.

# Major Changes

* Added Case List tab to the Trustee detail screen, gated behind the `trustee-case-list` feature flag
* New `GET /trustees/:trusteeId/cases` backend endpoint returning paginated `TrusteeCaseListItem` records
* `TrusteeCasesUseCase` fetches active `CASE_APPOINTMENT` docs, bulk-loads matching `SyncedCase` records, merges and sorts by `dateFiled` descending, then applies in-memory pagination
* Table renders 5 columns: Case Number (Division), Case Title, Chapter, Case Filed, Appt. Date — with case number linking to the case detail page
* `appointedDate` mapped from `CaseAppointment.appointedDate` (court date) rather than `assignedOn` (system timestamp)
* `DEFAULT_SEARCH_LIMIT`/`DEFAULT_SEARCH_OFFSET` constants used throughout; reset effect uses functional update to avoid double API calls on mount
* `createNavStateMapper` guarded against `undefined` path input
* Seed scenario (`trustee-case-list.ts`) creates 60 cases in both DXTR (AO_CS, AO_PY, AO_DE) and Cosmos (SYNCED_CASE + CASE_APPOINTMENT) so case detail and docket tabs render completely from links

# Testing/Validation

Implemented using TDD across all slices. Unit tests cover:
- Controller feature flag gating, authorization, pagination, and error propagation
- Use case sort order, partial SyncedCase match, `appointedDate` field mapping, and pagination metadata
- Component loading/error/empty states, all 5 column headers, cell rendering, pagination navigation, double-API-call prevention on mount, and reset-to-page-1 on trusteeId change
- Nav state mapper `undefined` input guard and both-flags-disabled scenario

Manual E2E testing performed against local environment using the new seed scenario.

# Notes

- The `trustee-case-list` feature flag must be enabled to see the Case List nav link and route
- Seed trustee IDs for manual testing: `cams-593-paginated` (60 cases, 3 pages) and `cams-593-empty` (no cases, empty state)
- `appointedDate` is optional on `TrusteeCaseListItem` — the UI renders a blank cell when absent, matching source data semantics
- The stash `true flags` on this branch contains the temporary hardcode used during manual testing (not committed)

# Definition of Done:

## Summary by Sourcery

Expand trustee case list test data to seed cases in both DXTR and CAMS and document the scenario for E2E and manual testing.

Enhancements:
- Generate 60 trustee cases with richer debtor metadata, appointed dates, and DXTR docket entries to fully support trustee case list and case detail flows.
- Refactor the trustee-case-list seed script to build operations incrementally and use shared DXTR case creation utilities.

Documentation:
- Document the trustee-case-list scenario in the test data catalog, including trustee IDs, case characteristics, and seeding command.


mattstankey:
## Code Review Feedback

A few things worth addressing before merge:

### Blocker 1: Missing `toBeDefined()` guards before non-null assertions

In the new `trustee-case-list` test suite, `casesOp!.data` and `apptOp!.data` are dereferenced with `!` without a preceding `toBeDefined()` guard:

```ts
// scenarios.test.ts — cases batch test (~366) and appointments batch test (~382, ~393)
for (const c of casesOp!.data) { ... }
for (const a of apptOp!.data) { ... }
const caseMap = new Map(casesOp!.data.map(...));
```

If `find` returns `undefined`, this throws a `TypeError: Cannot read properties of undefined` rather than a clear Vitest assertion failure. The existing pattern in this file is better:

```ts
expect(casesOp).toBeDefined();
expect(apptOp).toBeDefined();
```

Add those guards before the loops so failures produce readable output.

---

### Blocker 2: Consolidation scenario chapter distribution changed to `['11', '11', '12']`

```ts
// Before
expect(chapters).toEqual(['11', '13', '7']);
// After
expect(chapters).toEqual(['11', '11', '12']);
```

The original three cases had distinct chapters — that variety was likely intentional for testing consolidation across chapter types. The new distribution has duplicate `11`s and introduces chapter `12` (family farmers/fishermen), which is an unusual choice for a general consolidation test scenario. If this is an unintended side effect of the ID-generation refactor (chapters being assigned by formula based on new IDs), the consolidation scenario should restore the `7`/`11`/`13` distribution explicitly rather than accepting whatever the formula produces.

---

### Note: Each test re-generates 60 cases independently

`multiIdContext()` is called fresh in every `operations()` invocation, and all 7 tests in the `trustee-case-list` suite call `operations()` independently. With the mocked SQL this is harmless today, but if the scenario grows or generation becomes expensive this will slow the suite noticeably. Worth extracting to a `beforeAll`-scoped fixture when convenient.

mattstankey (approved):

PR #2464, opened 2026-06-03
CAMS-757: Upgrade seed phonetic tokens to production + logging

diabagatekelly:
# Purpose

Ensure test data uses the same phonetic token generation as production for accurate fuzzy search testing. The lightweight token generator was producing incorrect token formats that didn't match the backend search algorithm.

# Major Changes

* Upgraded `dev-tools/db_scripts/lib/phonetic-tokens.ts` from lightweight version to full production implementation
* Now uses Soundex + Metaphone + bigrams (matches `backend/lib/adapters/utils/phonetic-helper.ts`)
* Removed 3 accent-related test cases (José García, Jose Garcia, François Müller) 
* Updated test-data-catalog.md to reflect 41 cases (down from 43)
* Added duplication warning noting need to keep dev-tools in sync with backend
* **Fixed HIGH severity telemetry routing and added p95 monitoring**
* **Fixed MEDIUM severity apostrophe normalization and removed dead props**

# Testing/Validation

* All tests pass (2,861 backend + 1,073 frontend)
* Reseeded local and main databases successfully
* Verified token format: uppercase phonetics (Soundex/Metaphone) + lowercase bigrams
* Test data now has proper coverage for homophones (John/Jon), misspellings (Smith/Smyth), and nickname variations

# Notes

The dev-tools phonetic token generator must be kept in sync with the backend implementation manually due to ESM module resolution issues preventing direct imports. Added warning comment in the duplicated code.

# Code Review Discussion Items

**MEDIUM Severity - Resolved:**

1. ~~**scoreBreakdown exposed in API response**~~ - **Decision: Not important**
   - Kept for observability - workbook "Individual Session Deep Dive" parses breakdown for debugging
   - Only exposed to authenticated internal users
   - No action needed
   
2. **Cardinality observability for phonetic pre-filter** - **Tracked as follow-up: cams-157v**
   - Need to log how many documents pass pre-filter per query
   - Requires metrics infrastructure design
   - Follow-up beads issue created (CAMS-757 label)

# Definition of Done:

## Summary by Sourcery

Add search metadata to debtor-name search results and track search result clicks, while tightening fuzzy-search test data and documentation.

New Features:
- Attach structured search metadata (match score, contributing match types, and score breakdown) to phonetic debtor-name search results in the aggregation pipeline and API models.
- Track searchResultClick telemetry events when users click search results, including rank, score details, filters, and higher-ranked alternatives.

Enhancements:
- Compute match-type flags and score breakdowns in the Mongo aggregation score pipeline using a dedicated searchMetadata stage that respects single- and multi-target scoring.
- Extend UI components to propagate click handling from case-number links through rows to the results list, enabling rank-aware click handling.
- Refine fuzzy-search seed data by removing accent-heavy examples while keeping coverage of phonetic, nickname, and misspelling scenarios.

Documentation:
- Update fuzzy-search test data catalog entry to describe the new dataset size and coverage focus for phonetic search scenarios.

Tests:
- Add aggregation renderer tests covering the new searchMetadata stage, including max-target selection, matchTypes construction, and scoreBreakdown behavior.
- Add common-model tests for the new SearchMetadata types and optional SyncedCase.searchMetadata field.
- Add UI tests for case-number click callbacks, rank handling, and searchResultClick telemetry payloads and edge cases.


user-interface/src/search-results/SearchResults.tsx
```diff
@@ -140,6 +142,37 @@ function SearchResults(props: SearchResultsProps) {
       });
   }
 
+  function handleCaseClick(bCase: SyncedCase, rank: number) {
+    if (!searchPredicate.debtorName || !bCase.searchMetadata) return;
+
+    const { matchScore, matchTypes, scoreBreakdown } = bCase.searchMetadata;
+    const cap = Math.min(rank - 1, 5);
+    const higherRankedResults = (searchResults?.data ?? [])
+      .slice(0, cap)
```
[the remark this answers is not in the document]
mattstankey: Good catch. Renamed the local variable to `higherRankedOnPage` and added a comment clarifying that earlier pages are not retained after pagination navigation, so this intentionally reflects the current page only.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.test.ts
```diff
@@ -8,6 +8,49 @@ type MongoScoreStage = {
   $project?: Record<string, number>;
 };
 
+type MaxTargetIdxExpr = {
```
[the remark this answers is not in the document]
mattstankey: Good suggestion. Replaced the six deeply nested helper types (`MaxTargetIdxExpr`, `MatchTypeBranch`, `ConcatArraysMatchTypesExpr`, `MatchTypesCondExpr`, `ScoreBreakdownExpr`, `SearchMetadataExpr`) with a single shallow `MongoSearchMetadataStage` type that uses `any` for `matchTypes` and `scoreBreakdown`. Tests now navigate the structure directly without intermediate casts, keeping all behavioral assertions intact.

user-interface/src/search-results/SearchResults.tsx
```diff
@@ -140,6 +142,38 @@ function SearchResults(props: SearchResultsProps) {
...
+    // Only includes results from the current page; earlier pages are not retained after navigation.
+    const higherRankedOnPage = (searchResults?.data ?? [])
+      .slice(0, cap)
+      .filter((r) => r.searchMetadata !== undefined)
+      .map((r, i) => ({
+        rank: (searchPredicate.offset ?? 0) + i + 1,
+        matchScore: r.searchMetadata!.matchScore,
+        primaryMatchType: r.searchMetadata!.primaryMatchType,
+      }));
+
```
jamesobrooks, line 164: **[High] Workbook KQL reads `rank` and `matchScore` from `customMeasurements`, but the App Insights JS SDK routes numeric values passed in the `properties` bag to `customDimensions`.**

The workbook's `match-rank-distribution` and `Individual Session Deep Dive` panels both do:
```kql
| extend e = customMeasurements
| extend rank = toint(e.rank)
| extend matchScore = todouble(e.matchScore)
```
But because `rank` and `matchScore` are passed as plain properties here (not as a separate `measurements` argument), the SDK puts them in `customDimensions`, not `customMeasurements`. Those KQL columns will be null for every event.

Fix: Either move numeric fields to the `measurements` argument:
```typescript
getAppInsights().appInsights.trackEvent(
  { name: 'searchResultClick' },
  { primaryMatchType, scoreBreakdown: JSON.stringify(scoreBreakdown), ... },  // strings → customDimensions
  { rank, matchScore }  // numbers → customMeasurements
);
```
Or update the workbook KQL to read from `customDimensions` instead.

user-interface/src/search-results/SearchResults.tsx
```diff
@@ -140,6 +142,38 @@ function SearchResults(props: SearchResultsProps) {
...
+  function handleCaseClick(bCase: SyncedCase, rank: number) {
+    if (!searchPredicate.debtorName || !bCase.searchMetadata) return;
+
+    const { matchScore, primaryMatchType, scoreBreakdown } = bCase.searchMetadata;
+    const cap = Math.min(rank - 1, 5);
+    // Only includes results from the current page; earlier pages are not retained after navigation.
+    const higherRankedOnPage = (searchResults?.data ?? [])
+      .slice(0, cap)
+      .filter((r) => r.searchMetadata !== undefined)
+      .map((r, i) => ({
```
jamesobrooks: **[Low] Magic number — consider naming this constant.**

`5` is the max number of higher-ranked results captured in the analytics payload. The comment above explains the page-boundary limitation but not why 5 specifically. A name makes it self-documenting and trivially tunable if the analytics team asks for more context:
```typescript
const MAX_HIGHER_RANKED_CONTEXT = 5;
const cap = Math.min(rank - 1, MAX_HIGHER_RANKED_CONTEXT);
```

jamesobrooks (commented):
Multi-reviewer code review of CAMS-757 debtor name search improvements. Findings are adversarially verified — each was challenged and only those that survived are included below.

**High:** Workbook KQL routing mismatch (rank/matchScore panels return null); no p95 alert rule
**Medium:** scoreBreakdown exposed in API response; apostrophe normalization asymmetry; phoneticSearchEnabled dead prop; charPrefix cardinality unobservable
**Low:** error message logs debtor name; maxTargetIdx guard; result count caption format; magic number

user-interface/src/search/SearchResultsRow.tsx
```diff
@@ -21,16 +21,24 @@ function formatDebtorNames(debtorName: string, jointDebtorName?: string): JSX.El
 export function SearchResultsRow(props: SearchResultsRowProps) {
   const {
     bCase,
+    idx,
+    rank,
     labels,
     phoneticSearchEnabled: _phoneticSearchEnabled,
```
jamesobrooks: **[Medium] `phoneticSearchEnabled` is wired here but consumed by nobody.**

It's destructured with an underscore prefix (`_phoneticSearchEnabled`), which signals it's intentionally ignored. The prop is threaded all the way from the feature-flag detection at the screen level through `SearchResults` into this component, then discarded. No row or header implementation acts on it.

If there's a future plan to show phonetic-specific UI in the row, this is premature — add it when that rendering decision exists. For now, removing it from `SearchResultsRowProps` eliminates dead wiring in three places.

common/src/cams/cases.ts
```diff
@@ -175,13 +175,32 @@ type ClosedDismissedReopened = {
...
+  nicknameScore: number;
+  phoneticScore: number;
+  charPrefixScore: number;
+}
+
+/** Search-specific metadata attached to results from phonetic debtor name searches. */
+export interface SearchMetadata {
+  matchScore: number;
+  primaryMatchType: MatchType;
+  scoreBreakdown: ScoreBreakdown;
```
jamesobrooks, line 193: **[Medium] `scoreBreakdown` flows through to the HTTP response body without being stripped.**

`searchMetadata` (including the full `ScoreBreakdown` with named sub-scores) is serialized into every search API response — the use-case layer and controller apply no projection. Authenticated users can read the scoring model's internal structure from network responses, which discloses more than necessary.

Consider either:
1. Stripping `searchMetadata` before the controller serializes, or
2. Projecting only `matchScore` and `primaryMatchType` to a separate API DTO, omitting `scoreBreakdown`

The frontend only needs `matchScore` and `primaryMatchType` for analytics — `scoreBreakdown` is logged to App Insights as an opaque JSON string anyway.

common/src/cams/cases.ts
```diff
@@ -175,13 +175,32 @@ type ClosedDismissedReopened = {
...
+}
+
 export type SyncedCase = DxtrCase &
   Auditable & {
     documentType: 'SYNCED_CASE';
     id?: string;
     trusteeId?: string;
     movedToCaseId?: string;
     movedOn?: string;
+    searchMetadata?: SearchMetadata;
```
jamesobrooks, line 203: **[Medium] Apostrophe/diacritic normalization is asymmetric between query and document words.**

`normalizeText` strips apostrophes from search queries (`O'Brien` → `obrien`). However, `buildParseWordsExpression` in the renderer applies only `$toLower` and split-on-hyphen to document names at query time — no apostrophe stripping. A stored name of `Patrick O'Brien` produces the document word `o'brien`; the query produces `obrien`. These don't intersect in exact matching.

The result partially recovers via phonetic + similar-length qualification, but only by coincidence of word lengths. Names where the length difference after stripping exceeds ±1 (the similar-length tolerance for >4-char words) will silently miss exact-match scoring.

Fix: Apply the same character normalization inside `buildParseWordsExpression` so document words are compared on the same normalized basis as search words:
```typescript
// inside buildParseWordsExpression, after :
{ : { input: lowerName, find: "'", replacement: '' } }
```
Or use `$regexReplace` to cover all diacritics consistently.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
```diff
@@ -532,6 +532,147 @@ function toMongoScore(stage: Score): object[] {
     },
   });
 
+  // Stage 4a: Compute searchMetadata from temp fields before cleanup
+  // When there are multiple targets, determine which scored highest.
+  // Single-target case skips this and always uses idx 0.
+  const isSingleTarget = targetNameFields.length === 1;
+  const maxTargetIdx = isSingleTarget
```
jamesobrooks, line 564: **[Medium] No cardinality observability for the pre-filter — performance risk is real but invisible.**

The indexed phonetic pre-filter meaningfully bounds the document set before the scoring pipeline runs. However, the number of documents that pass this pre-filter is never logged. The nested `$map`/`$anyElementTrue`/`$indexOfCP` expressions in `buildCharacterPrefixMatchExpression` run O(D_filtered × W_doc × S) per query — for a common surname phonetic cluster that matches thousands of documents, a latency spike would be invisible until it shows up in App Insights.

Suggestion: log pre-filter cardinality alongside `SearchDuration` in `completeSearchTrace` (e.g., `SearchPreFilterCount` from `searchResult.metadata.total` before pagination). This makes the risk observable without changing the algorithm.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
```diff
@@ -532,6 +532,147 @@ function toMongoScore(stage: Score): object[] {
...
+    ? 0
+    : {
+        $cond: {
+          if: { $gte: ['$_score_0', '$_score_1'] },
+          then: 0,
+          else: 1,
+        },
+      };
```
jamesobrooks, line 572: **[Low] `maxTargetIdx` only handles 1–2 targets — silently wrong for 3+.**

The outer scoring loop (stages 1–3) handles N targets generically, but `maxTargetIdx` hardcodes a binary comparison between `_score_0` and `_score_1`. If a caller ever passes a third target field, `searchMetadata` will pick the winner from only the first two targets, silently producing wrong `primaryMatchType` and `scoreBreakdown`.

All current callers pass 1 or 2 fields, so this isn't a present bug. A runtime guard converts it from a silent data corruption to a loud failure at build time:
```typescript
if (targetNameFields.length > 2) {
  throw new CamsError(MODULE_NAME, {
    message: `searchMetadata computation only supports ≤2 target fields. Got ${targetNameFields.length}.`
  });
}
```

ops/cloud-deployment/lib/workbooks/debtor-name-search.json
```diff
@@ -259,11 +259,52 @@
...
...Calculate the raw percentile ranks globally first\nlet percentiles = toscalar(\n    data \n    | summarize p = pack_array(percentile(rank, 90), percentile(rank, 50), percentile(rank, 10))\n);\nlet p90_val = toint(percentiles[0]);\nlet p50_val = toint(percentiles[1]);\nlet p10_val = toint(percentiles[2]);\n// 2. Capture the match types occurring at those exact rank boundaries\ndata\n| summarize \n    p90_Label = anyif(PrimaryMatchType, rank == p90_val),\n    p50_Label = anyif(PrimaryMatchType, rank == p50_val),\n    p10_Label = anyif(PrimaryMatchType, rank == p10_val)\n// 3. Package the full descriptive name and the rank value into pairs\n| project p90_Arr = pack_array(strcat(\"90th Percentile (\", coalesce(p90_Label, \"unknown\"), \")\"), p90_val),\n          p50_Arr = pack_array(strcat(\"Median (\", coalesce(p50_Label, \"unknown\"), \")\"), p50_val),\n          p10_Arr = pack_array(strcat(\"10th Percentile (\", coalesce(p10_Label, \"unknown\"), \")\"), p10_val)\n// 4. Expand out the rows so the chart can digest them\n| project packed = pack_array(p90_Arr, p50_Arr, p10_Arr)\n| mv-expand packed to typeof(dynamic)\n| project Percentile = tostring(packed[0]), Rank = toint(packed[1])",
```
jamesobrooks: **[High] `rank` is read from `customMeasurements` here, but the App Insights SDK routes it to `customDimensions`.**

The JS SDK's `trackEvent(event, properties)` routes values by type: strings/booleans go to `customDimensions`, numbers go to `customMeasurements` **only** when passed as the third `measurements` argument. Because `rank` and `matchScore` are passed in the `properties` bag in `SearchResults.tsx`, they land in `customDimensions`. This `customMeasurements.rank` path returns null for every event, making the rank distribution visualization silently empty.

Fix on the frontend side:
```typescript
getAppInsights().appInsights.trackEvent(
  { name: 'searchResultClick' },
  { primaryMatchType, scoreBreakdown: JSON.stringify(scoreBreakdown), ... },  // strings
  { rank, matchScore }  // numbers → goes to customMeasurements
);
```
Or fix on the KQL side by reading from `customDimensions` instead.

ops/cloud-deployment/lib/workbooks/debtor-name-search.json
```diff
@@ -259,11 +259,52 @@
...
+              "resourceType": "microsoft.insights/components",
+              "visualization": "barchart",
+              "chartSettings": {
+                "xAxis": "Percentile",
+                "yAxis": [
+                  "Rank"
+                ]
+              }
+            },
+            "name": "match-rank-distribution"
```
jamesobrooks, line 319: **[High] Same SDK routing issue — `rank` and `matchScore` will be null from `customMeasurements`.**

Same root cause as the rank-distribution panel. The `Rank` and `MatchScore` columns in this grid will be null for all rows.

Additionally: no p95 latency panel exists in this workbook for the 3-second acceptance criterion. The `Case Search Performance` frontend event already includes a `debtorNameUsed` dimension — adding a panel here that segments p95 by phonetic vs. standard would close that observability gap:
```kql
customEvents
| where name == "Case Search Performance"
| where timestamp {TimeRange}
| extend durationMs = todouble(customMeasurements.durationMs)
| extend searchType = iif(tostring(customDimensions.debtorNameUsed) =~ "true", "Phonetic", "Standard")
| summarize P50Ms = round(percentile(durationMs, 50), 0), P95Ms = round(percentile(durationMs, 95), 0) by searchType
```

jamesobrooks:
## Remaining findings (lines not in this diff)

**[Low] Error message in `cases.mongo.repository.ts:458` logs debtor name to server-side audit trail**

```typescript
message: `Failed to search cases with word-level scoring for "${predicate.debtorName}".`,
```
The debtor name does not appear in the HTTP response body (confirmed — it lands in `camsStack`, not in `error.message`). However, it does appear in server-side logs via `logger.camsError(error)`. In a FedRAMP context, PII in audit log entries is a consideration under AU-2/AU-3 controls. Low severity since it's log-only, but worth removing the name from the message and logging it separately at a controlled severity if needed for diagnostics.

---

**[High — no changed file to annotate] No scheduled query alert for the 3-second p95 SLO**

The `monitoring-alerts/` directory has no scheduled query alert targeting `SearchDuration` where `searchType == 'phonetic'`. The existing alerts cover only infrastructure metrics (`Http5xx`, `ServiceAvailability`, `storage_percent`). The 3-second p95 acceptance criterion has monitoring data flowing (the `Case Search Performance` event with `debtorNameUsed` dimension) but no alert will fire if it's breached. A log-based alert (Bicep `scheduledqueryrules`) would close this gap.

---

**[Low] Result count caption missing search term echo**

`SearchResults.tsx:209` renders `caption={`${displayCount} ${totalCount === 1 ? 'case' : 'cases'}`}`. The acceptance criterion specifies the format `"Found 15 cases matching 'John Smith'"`. The count is present but the echoed search term is not. This line is unchanged on this branch so can't be annotated inline, but it's worth tracking.

user-interface/src/search-results/SearchResults.tsx
diabagatekelly: ✅ Fixed! Moved `rank` and `matchScore` to the `measurements` parameter (third argument of `trackEvent`):

```typescript
getAppInsights().appInsights.trackEvent(
  { name: 'searchResultClick' },
  {
    primaryMatchType,
    scoreBreakdown: JSON.stringify(scoreBreakdown),
    chapters: searchPredicate.chapters ? JSON.stringify(searchPredicate.chapters) : undefined,
    divisionCodes: searchPredicate.divisionCodes
      ? JSON.stringify(searchPredicate.divisionCodes)
      : undefined,
    excludeClosedCases: searchPredicate.excludeClosedCases,
    higherRankedResults: JSON.stringify(higherRankedOnPage),
  },
  {
    rank,
    matchScore,
  },
);
```

This routes numeric values to `customMeasurements` where the workbook KQL queries expect them. The workbook queries already use `customMeasurements.rank` and `customMeasurements.matchScore`, so they will now receive the data correctly.

ops/cloud-deployment/lib/workbooks/debtor-name-search.json
diabagatekelly: ✅ Fixed! Added several improvements:

**1. Fixed SDK routing:** Moved `rank` and `matchScore` to measurements parameter in `SearchResults.tsx`

**2. Added p95 latency panel:** Inserted new "Search Performance (p95 Latency)" panel in the workbook that:
- Queries `Case Search Performance` events
- Segments by search type (Phonetic vs. Standard) using `debtorNameUsed` dimension
- Displays p95 and p50 latency in seconds
- Shows SLO status (⚠️ if phonetic p95 > 3s, ✅ otherwise)
- Uses `customMeasurements.durationMs` correctly

**3. Enhanced Case Search Performance event:** Added `debtorNameUsed` dimension and moved `durationMs` to measurements parameter:

```typescript
getAppInsights().appInsights.trackEvent(
  { name: 'Case Search Performance' },
  {
    debtorNameUsed: !!searchPredicate.debtorName,
    excludeClosedCases: excludeClosedCases ?? true,
  },
  {
    durationMs: Math.round(performance.now() - searchStart),
  },
);
```

**4. Created alert infrastructure:** Added `scheduled-query-alert-rule.bicep` module for creating KQL-based alerts on Application Insights data. The team can now deploy a scheduled query alert for the 3-second SLO using this module.

The workbook's existing `match-rank-distribution` and `Individual Session Deep Dive` panels will now receive data correctly since we fixed the SDK routing.

diabagatekelly:
✅ **Fixed - HIGH severity telemetry routing issues**

Both workbook KQL routing issues have been resolved:

1. **SearchResults.tsx**: Moved `rank` and `matchScore` to measurements parameter
2. **Added p95 monitoring**: New panel in workbook with SLO status
3. **Enhanced telemetry**: Added `debtorNameUsed` dimension to Case Search Performance event

The workbook's `match-rank-distribution` and `Individual Session Deep Dive` panels will now receive data correctly since numeric values are properly routed to `customMeasurements`.

See commit 44b7759a6 for full implementation.

common/src/cams/cases.ts
diabagatekelly: ✅ Fixed! Added apostrophe stripping to `buildParseWordsExpression` to match `normalizeText` behavior:

```typescript
// Strip apostrophes to match normalizeText behavior in phonetic-helper.ts
const noApostrophes = { $replaceAll: { input: lowerName, find: "'", replacement: '' } };
```

**Before:** Query `obrien` vs document `o'brien` → no match ❌
**After:** Query `obrien` vs document `obrien` → match ✅

This fixes exact matching for names with apostrophes like Patrick O'Brien.

All tests pass: 43/43 backend, 15/15 frontend.

user-interface/src/search/SearchResultsRow.tsx
diabagatekelly: ✅ Fixed! Removed dead `phoneticSearchEnabled` prop that was threaded through components but never consumed.

**Removed from:**
- `SearchResultsRowProps` interface
- Component destructuring (was using `_phoneticSearchEnabled` prefix indicating unused)
- All parent components passing the prop

The prop was premature - can be added back when/if we actually need phonetic-specific UI rendering in the row component.

diabagatekelly:
## Code Review Status Update

**✅ HIGH Severity (3/3 complete)**
1. Telemetry routing mismatch - Fixed
2. p95 latency monitoring - Added
3. No p95 alert rule - Infrastructure created

**✅ MEDIUM Severity (2/4 complete)**
1. Apostrophe normalization asymmetry - Fixed
2. Dead `phoneticSearchEnabled` prop - Removed

**⏸️ MEDIUM Severity (2/4 deferred - need discussion)**

3. **`scoreBreakdown` exposed in API response** - Kept for now
   - Workbook queries parse breakdown for observability (Individual Session Deep Dive panel)
   - Shows exactScore, nicknameScore, charPrefixScore, phoneticScore to operators
   - Only exposed to authenticated internal users
   - **Question:** Should we still strip this, or is it valuable observability data?

4. **Cardinality observability for pre-filter** - Would require architectural changes
   - Need to log documents passing phonetic pre-filter
   - Requires adding metrics tracking to aggregation pipeline execution
   - Larger scope change
   - **Suggestion:** Create follow-up task?

---

**All tests passing:** 43/43 backend, 15/15 frontend
**Commit:** b80b70aa4

user-interface/src/search-results/SearchResults.tsx
diabagatekelly: ✅ Fixed! Added named constant for clarity:

```typescript
// Maximum number of higher-ranked results to include in analytics payload
const MAX_HIGHER_RANKED_CONTEXT = 5;
const cap = Math.min(rank - 1, MAX_HIGHER_RANKED_CONTEXT);
```

Now self-documenting and easily tunable if analytics team needs more/less context.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
diabagatekelly: ✅ Fixed! Added runtime guard to prevent silent failure:

```typescript
// Guard: maxTargetIdx only handles 1-2 targets
if (targetNameFields.length > 2) {
  throw new CamsError(MODULE_NAME, {
    message: `maxTargetIdx only supports 1-2 target fields. Got ${targetNameFields.length}. Generalize the comparison logic to handle ${targetNameFields.length} targets.`,
  });
}
```

Current callers pass 2 targets: `['debtor.name', 'jointDebtor.name']`. If someone adds a third target (e.g., `trustee.name`), the guard will throw immediately with a clear error message instead of silently producing incorrect searchMetadata.

All tests pass: 43/43 backend.

diabagatekelly:
**RE: scoreBreakdown exposed in API (comment on cases.ts:193)**

This MEDIUM severity issue is **documented in PR description for team discussion**.

**Current state:** Kept for observability - workbook "Individual Session Deep Dive" panel parses scoreBreakdown to display exactScore, nicknameScore, charPrefixScore, phoneticScore for debugging.

**Question for team:** Should we mask internal scoring weights while preserving observability?

**Options:**
- Send percentages instead of raw scores
- Normalized 0-100 scale
- Keep as-is for internal users

Waiting for team input before implementing a change.

diabagatekelly:
**RE: Cardinality observability for pre-filter (comment on mongo-aggregate-renderer.ts:551)**

This MEDIUM severity issue is **documented in PR description for team discussion**.

**Issue:** No logging of how many documents pass the phonetic pre-filter per query. Can't observe if pre-filter is too tight (missing results) or too loose (performance degradation).

**Why not fixed:** Requires architectural design for metrics infrastructure in aggregation pipeline execution. Larger scope than other review findings.

**Suggestion:** Create follow-up task to discuss:
- Metrics approach (logging, Application Insights custom metrics, etc.)
- Where in pipeline to capture counts
- What thresholds to alert on

This is valuable observability but needs design discussion first.

common/src/cams/cases.ts
jamesobrooks: Unresolving because the fix broke something else. It might not be super important to address this right now though.

jalba-flexion (approved):

user-interface/src/search-results/SearchResults.tsx
```diff
@@ -140,6 +144,42 @@ function SearchResults(props: SearchResultsProps) {
...
+    const { matchScore, primaryMatchType, scoreBreakdown } = bCase.searchMetadata;
+    // Maximum number of higher-ranked results to include in analytics payload
+    const MAX_HIGHER_RANKED_CONTEXT = 5;
+    const cap = Math.min(rank - 1, MAX_HIGHER_RANKED_CONTEXT);
+    // Only includes results from the current page; earlier pages are not retained after navigation.
+    const higherRankedOnPage = (searchResults?.data ?? [])
+      .slice(0, cap)
+      .filter((r) => r.searchMetadata !== undefined)
+      .map((r, i) => ({
+        rank: (searchPredicate.offset ?? 0) + i + 1,
```
mattstankey, line 160: **Bug: rank index is wrong after `.filter()`**

After `.filter(r => r.searchMetadata !== undefined)`, the index `i` in `.map((r, i) => ...)` restarts from 0 over the *filtered* array, not the original sliced positions. If any result in the top-N lacks metadata, ranks will be off.

Fix by preserving the original index before filtering:

```ts
const higherRankedOnPage = (searchResults?.data ?? [])
  .slice(0, cap)
  .map((r, i) => ({ r, i }))
  .filter(({ r }) => r.searchMetadata !== undefined)
  .map(({ r, i }) => ({
    rank: (searchPredicate.offset ?? 0) + i + 1,
    matchScore: r.searchMetadata!.matchScore,
    primaryMatchType: r.searchMetadata!.primaryMatchType,
  }));
```

dev-tools/db_scripts/scenarios/cases-fuzzy-search.ts
```diff
@@ -436,7 +406,7 @@ export async function generate(_ctx: SeedContext): Promise<SeedOperation[]> {
       regionName: 'NEW YORK',
       courtId: '0209',
     }),
-    // 25. Montgomery Bartholomew Wellington-Fitzpatrick III
+    // 27. Montgomery Bartholomew Wellington-Fitzpatrick III
```
mattstankey, line 409: **Cosmetic: comment numbers jump from 22 to 27 here**

After removing cases 22 (José García) and 23 (François Müller), the former case 24 correctly renumbered to 22. But the next entry jumps to `// 27` and continues off by 5 through the rest of the file (ending at 45 instead of 41).

The PR body says 41 cases; the comments imply 45. This doesn't affect functionality but breaks the sequence's usefulness as a quick index. The comments from this point through the end of the array need to be decremented by 5.

user-interface/src/search-results/SearchResults.tsx
```diff
@@ -122,13 +123,16 @@ function SearchResults(props: SearchResultsProps) {
...
-            durationMs: Math.round(performance.now() - searchStart),
+        getAppInsights().appInsights.trackEvent({
+          name: 'Case Search Performance',
+          properties: {
+            debtorNameUsed: !!searchPredicate.debtorName,
             excludeClosedCases: excludeClosedCases ?? true,
           },
-        );
+          measurements: {
+            durationMs: Math.round(performance.now() - searchStart),
```
mattstankey, line 133: **Risk: historical telemetry data gap after deploy**

This changes how `durationMs` is stored. Before this PR it was passed as the second argument to `trackEvent()` (Application Insights stores that as `customDimensions`, a string field). After this PR it's in `measurements` (stored as `customMeasurements`, a numeric field).

The new workbook p95 latency query reads:
```kql
| extend durationMs = todouble(m.durationMs)  // only hits customMeasurements
```
Historical events won't have data in `customMeasurements.durationMs`, so the chart will appear empty or show a hard cutoff at deploy time.

Suggestion — make the KQL resilient to both formats during the transition:
```kql
| extend durationMs = coalesce(todouble(customMeasurements.durationMs), todouble(customDimensions.durationMs))
```

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.test.ts
```diff
@@ -8,6 +8,18 @@ type MongoScoreStage = {
   $project?: Record<string, number>;
 };
 
+type MongoSearchMetadataStage = {
+  $addFields: {
+    searchMetadata: {
+      matchScore: string;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      primaryMatchType: any;
```
mattstankey: **Minor: prefer `unknown` over `any` here**

Using `any` (with eslint suppressions) loses type safety for the assertions below. `unknown` with type narrowing in assertions would be safer and wouldn't need the suppressions:

```ts
type MongoSearchMetadataStage = {
  $addFields: {
    searchMetadata: {
      matchScore: string;
      primaryMatchType: unknown;
      scoreBreakdown: unknown;
    };
  };
};
```

The assertions that navigate into these (e.g. `stage.$addFields.searchMetadata.primaryMatchType.$cond.if`) would need `as Record<string, unknown>` casts, but that's more honest about what's happening than suppressing the linter.

user-interface/src/search-results/SearchResults.tsx
diabagatekelly: ✅ Fixed! Preserved original index before filtering to ensure correct rank calculation even when results lack searchMetadata:

```ts
const higherRankedOnPage = (searchResults?.data ?? [])
  .slice(0, cap)
  .map((r, i) => ({ r, i }))  // Preserve original index
  .filter(({ r }) => r.searchMetadata !== undefined)
  .map(({ r, i }) => ({
    rank: (searchPredicate.offset ?? 0) + i + 1,  // Use original index
    matchScore: r.searchMetadata!.matchScore,
    primaryMatchType: r.searchMetadata!.primaryMatchType,
  }));
```

dev-tools/db_scripts/scenarios/cases-fuzzy-search.ts
diabagatekelly: ✅ Fixed in other PR on testing data improvement

user-interface/src/search-results/SearchResults.tsx
diabagatekelly: ✅ Fixed! Updated the workbook KQL query to use `coalesce()` for backward compatibility:

```kql
| extend durationMs = coalesce(todouble(m.durationMs), todouble(d.durationMs))
```

This ensures the p95 latency chart will display both historical data (from `customDimensions`) and new data (from `customMeasurements`) without a hard cutoff at deploy time.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.test.ts
diabagatekelly: ✅ Fixed! Replaced `any` with proper MongoDB aggregation expression types:

```ts
type MongoCondExpr = {
  $cond: {
    if: Record<string, unknown>;
    then: unknown;
    else: unknown;
  };
};

type MongoArrayElemAtExpr = {
  $arrayElemAt: [Record<string, unknown>, number];
};

type MongoSearchMetadataStage = {
  $addFields: {
    searchMetadata: {
      matchScore: string;
      primaryMatchType: MongoCondExpr | MongoArrayElemAtExpr;
      scoreBreakdown: MongoCondExpr;
    };
  };
};
```

This provides proper type safety for MongoDB expressions while keeping test assertions straightforward. All 44 tests pass.

mattstankey (approved):

PR #10, opened 2023-03-10
111 java pre commit

jamesobrooks:

PR #2474, opened 2026-06-04
CAMS-746: Task date sorting (2 of 3)

diabagatekelly:
# Purpose

Fix inconsistent task ordering on the Data Verification screen by introducing a dedicated `taskDate` field for all verification items (transfer orders, consolidation orders, and trustee-match verifications).

**Problem:** Previously, the Data Verification screen sorted items using different date fields depending on the item type (transfer/consolidation orders used `orderDate`, trustee-match verifications used `createdOn` or `updatedOn`). This caused inconsistent ordering because these dates represent different events and can be set at different times.

**Solution:** Add a `taskDate` field to all three verification item types that consistently represents when the verification task should be actioned. Backfill existing records via Azure dataflow migrations.

# Major Changes

- Added `taskDate: string` field to `TransferOrder`, `ConsolidationOrder`, and `TrusteeMatchVerification` types
- Added `computeTaskDate()` helper function to derive taskDate from existing date fields  
- Updated data verification sorting to use `taskDate` consistently across all item types
- Added repository methods to query and update records missing `taskDate` (3 methods per repository: get/update for each item type)
- Created three backfill use cases for migrating existing data with queue-based paging
- Wired three Azure Function dataflow migrations (one per item type) to handle backfilling at scale
- Updated `sync-trustee-appointments.ts` to set `taskDate` when creating new verification records
- Updated `DataVerificationScreen` and all accordion components to sort/display `taskDate`
- Updated mock data, test fixtures, and seed data to include `taskDate`

# Testing/Validation

**Unit Tests:**
- All three backfill use cases have comprehensive unit tests (237+ lines each)
- Repository methods tested for querying/updating taskDate
- `computeTaskDate()` helper function tested for all item types  
- Frontend sorting tested with taskDate values

**Manual Testing Steps:**
1. Deploy backend with new repository methods and use cases
2. Deploy dataflow functions
3. Trigger each backfill migration via queue messages
4. Monitor logs for completion and any warnings about skipped records
5. Deploy frontend with taskDate sorting
6. Verify Data Verification screen shows consistent chronological ordering
7. Verify new records created after deployment have taskDate set automatically

**Acceptance Criteria:**
- All existing transfer orders, consolidation orders, and trustee-match verifications have `taskDate` backfilled
- Data Verification screen sorts items consistently by `taskDate`
- New records created after deployment automatically include `taskDate`
- No runtime errors when sorting verification items

# Notes

**Deployment Order (Critical):** Backfill migrations must complete before frontend deployment to avoid sorting undefined values.
1. Deploy backend + dataflows
2. Run backfill migrations (monitor completion)
3. Deploy frontend

**Future Cleanup:** After backfill completes and `taskDate` is guaranteed present on all records, consider removing UI fallbacks to `orderDate`/`createdOn` and migration-specific repository methods.

**Architectural Decisions:**
- Three separate dataflows (not one combined): Each item type has isolated backfill logic for observability and failure isolation
- Queue-based paging: Handles large datasets without timeouts
- Stored vs computed: `taskDate` is stored (not computed on-read) for consistent sorting and query performance

# Definition of Done:


mattstankey:
## Code Review Feedback

### Potential Bugs

**#1 — `taskDate` is required in the type but optional at runtime between deploy and backfill**

`taskDate: string` (non-optional) is added to all three types. Before the backfill completes, existing records returned from the API will be missing this field. The UI sort — `.sort((a, b) => sortByDate(a.taskDate, b.taskDate))` — will silently sort `undefined` values or throw depending on `sortByDate`'s behavior. Consider either making the type `taskDate?: string` with a fallback in the sort, or documenting that the backfill must complete before the UI deploy.

**#3 — Records that can't be backfilled are silently skipped as `success: true`**

In all three backfill use cases, when `computeTaskDate` returns a falsy value, the code pushes `{ success: true }` and continues. Those records will never receive a `taskDate`, meaning they'll remain broken in production with no visible signal. At minimum, use a `warn` log instead of `debug` so operators know records were skipped.

**#4 — Dev seed data uses date-only strings where production uses ISO 8601**

The seed `taskDate` values (e.g., `'2012-06-14'`) are deliberately different from `orderDate` values (which use ISO 8601 like `'2025-02-01T...'`). The format inconsistency could cause subtle string-sort differences in dev environments if `sortByDate` relies on string comparison.

---

### Code Quality

**#5 — Near-total duplication across three backfill use cases**

`backfill-transfer-order-task-date.ts`, `backfill-consolidation-order-task-date.ts`, and `backfill-trustee-verification-task-date.ts` are structurally identical (~140 lines each, ~420 lines total). The only differences are the type, factory method, field name in `CursorPageResult`, and log strings. Tolerable for migration code that will be deleted post-backfill, but worth noting if this pattern is reused.

**#6 — `newLastId` in the `ok` result variant is redundant**

The `ok` branch of `ProcessBackfillPageResult` includes both `newLastId: string | null` and `nextCursor: { lastId: string | null } | null`. The dataflow handlers only consume `nextCursor`, so `newLastId` is an unused field on the result. Remove it from the type or the dataflow handlers, whichever is authoritative.

**#7 — Unexported type changes may break consumers**

`DataVerificationItem` in `data-verification.ts` was changed from `export type` to an unexported `type`. Same for `VerificationListParams` in `trustee-match-verification.use-case.ts`. These are appropriate if the types are internal-only, but worth a quick confirmation that no other files import them (a `grep` across the monorepo).

---

### Missing Coverage

**#8 — No tests for the dataflow handler layer**

`handleStart` and `handlePage` in each of the three `backfill-*.ts` dataflow files are untested. Only the use cases beneath them are covered. If queue wiring is wrong (wrong output binding, wrong queue name), it won't be caught by the test suite.

**#9 — `ConsolidationOrderAccordion` display change is untested**

The accordion now renders `taskDate` instead of `orderDate`. No test was updated to verify this rendering change.

---

### Minor Notes

- **`taskDate: new Date().toISOString()` in `sync-trustee-appointments.ts`** sets `taskDate` to wall-clock time at record creation, which is semantically different from using `createdOn` (which `computeTaskDate` uses during backfill). These will diverge by milliseconds in practice, but they represent different things. Using `createdOn` at the upsert site would be more consistent with the backfill logic.

- **PR description is entirely placeholder text** — Purpose, Major Changes, Testing/Validation, and Notes are all unfilled. Makes the PR harder to review and harder to understand in git history.

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
```diff
@@ -163,4 +163,35 @@ export default class ConsolidationOrdersMongoRepository<
...
+  }
+
+  public async updateConsolidationOrderTaskDate(mongoId: string, taskDate: string): Promise<void> {
+    try {
+      type ConsolidationQueryable = ConsolidationOrder & { _id: string };
+      const query = using<ConsolidationQueryable>()('_id').equals(mongoId);
+      await this.getAdapter<ConsolidationQueryable>().updateOne(query, {
+        taskDate,
+      } as Partial<ConsolidationQueryable>);
+    } catch (originalError) {
```
fmaddenflx, line 193: It seems to me that these may be methods only used for migration correct?  If so, they probably don't need to live here but in the migration dataflows.

backend/lib/adapters/gateways/mongo/orders.mongo.repository.ts
```diff
@@ -130,4 +130,35 @@ export class OrdersMongoRepository extends BaseMongoRepository implements Orders
       throw getCamsError(originalError, MODULE_NAME);
     }
   }
+
+  async findTransferOrdersMissingTaskDate(
+    lastId: string | null,
+    limit: number,
```
fmaddenflx, line 136: Same comment in this module -- aren't these methods only used for migration?  I would think that any data coming from dxtr in the future would just get a taskDate added when it gets added to mongo, so this wouldn't be needed moving forward, would it?

backend/lib/adapters/gateways/mongo/trustee-match-verification.mongo.repository.ts
```diff
@@ -143,4 +143,42 @@ export class TrusteeMatchVerificationMongoRepository
...
     }
   }
+
+  async findVerificationsMissingTaskDate(
+    lastId: string | null,
+    limit: number,
+  ): Promise<Array<TrusteeMatchVerification & { _id: string }>> {
+    try {
+      type VerificationQueryable = TrusteeMatchVerification & { _id: string };
+      const doc = using<VerificationQueryable>();
```
fmaddenflx, line 153: Same thing I mentioned earlier.

user-interface/src/data-verification/consolidation/ConsolidationOrderAccordion.tsx
```diff
@@ -64,7 +64,9 @@ export function ConsolidationOrderAccordion(props: ConsolidationOrderAccordionPr
     filteredOfficeRecords: getDivisionComboOptions(
       consolidationStore.filteredOfficesList ?? props.courts,
     ),
-    formattedOrderFiledDate: formatDate(consolidationStore.order.orderDate),
+    formattedOrderFiledDate: formatDate(
+      consolidationStore.order.taskDate ?? consolidationStore.order.orderDate,
```
fmaddenflx: I think this should just always be using taskDate shouldn't it?

user-interface/src/data-verification/consolidation/ConsolidationOrderAccordion.test.tsx
```diff
@@ -191,7 +191,7 @@ describe('ConsolidationOrderAccordion tests', () => {
     renderWithProps();
     const heading = findAccordionHeading(order.id!);
     expect(heading?.textContent).toContain(order.courtName);
-    expect(heading?.textContent).toContain(formatDate(order.orderDate));
+    expect(heading?.textContent).toContain(formatDate(order.taskDate ?? order.orderDate));
```
fmaddenflx: shouldn't this just be using taskDate all the time?

user-interface/src/data-verification/trustee-verification/TrusteeMatchVerificationAccordion.tsx
```diff
@@ -459,10 +459,10 @@ export function TrusteeMatchVerificationAccordion(props: TrusteeMatchVerificatio
           <div
             className="accordion-header-field grid-col-2 text-no-wrap"
             title="Event date"
-            aria-label={`${fieldHeaders[1]} on ${formatDate(order.createdOn ?? order.updatedOn)}.`}
+            aria-label={`${fieldHeaders[1]} on ${formatDate(order.taskDate ?? order.createdOn ?? '')}.`}
             data-cell={fieldHeaders[1]}
           >
-            {formatDate(order.createdOn ?? order.updatedOn)}
+            {formatDate(order.taskDate ?? order.createdOn ?? '')}
```
fmaddenflx: Shouldn't this just all use taskDate all the time?

user-interface/src/data-verification/DataVerificationScreen.tsx
```diff
@@ -232,11 +232,12 @@ export default function DataVerificationScreen() {
...
-    .sort((a, b) => {
-      const dateA = (a as TransferOrder).orderDate ?? (a as TrusteeMatchVerification).createdOn;
-      const dateB = (b as TransferOrder).orderDate ?? (b as TrusteeMatchVerification).createdOn;
-      return sortByDate(dateA, dateB);
-    })
+    .sort((a, b) =>
+      sortByDate(
+        a.taskDate ?? (a as TransferOrder).orderDate,
+        b.taskDate ?? (b as TransferOrder).orderDate,
+      ),
```
fmaddenflx: As long as we've done migration properly, I don't know if we need to continue to use orderDate as a fallback, but I can maybe be convinced otherwise.

user-interface/src/data-verification/TransferOrderAccordion.tsx
```diff
@@ -68,10 +68,10 @@ export function TransferOrderAccordion(props: TransferOrderAccordionProps) {
         <div
           className="accordion-header-field grid-col-2 text-no-wrap"
           title="Event date"
-          aria-label={`${fieldHeaders[1]} on ${formatDate(order.orderDate)}.`}
+          aria-label={`${fieldHeaders[1]} on ${formatDate(order.taskDate ?? order.orderDate)}.`}
           data-cell={fieldHeaders[1]}
         >
-          {formatDate(order.orderDate)}
+          {formatDate(order.taskDate ?? order.orderDate)}
```
fmaddenflx: rely on taskDate

backend/function-apps/dataflows/dataflows.ts
```diff
@@ -112,6 +115,9 @@ dataflows.register(
   StaffAssignmentDownstream,
   TrusteeAppointmentDownstream,
   BackfillTrusteeAppointmentsDownstreamDataflow,
+  BackfillTransferOrderTaskDate,
+  BackfillConsolidationOrderTaskDate,
+  BackfillTrusteeVerificationTaskDate,
```
fmaddenflx, line 120: I kind of wonder if we need to do this in 3 separate dataflows and why we can't simplify this and do it in one dataflow since this should be a 1 time thing, shouldn't it?

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
diabagatekelly: ✅ Refactored - moved migration methods into use cases (following phonetic tokens pattern). Query building is now inline in backfill use cases.

backend/lib/adapters/gateways/mongo/orders.mongo.repository.ts
diabagatekelly: ✅ Refactored - moved migration methods into use cases (following phonetic tokens pattern). Query building is now inline in backfill use cases.

backend/lib/adapters/gateways/mongo/trustee-match-verification.mongo.repository.ts
diabagatekelly: ✅ Refactored - moved migration methods into use cases (following phonetic tokens pattern). Query building is now inline in backfill use cases.

user-interface/src/data-verification/consolidation/ConsolidationOrderAccordion.tsx
diabagatekelly: ✅ Fixed - removed fallback, now using taskDate directly.

user-interface/src/data-verification/consolidation/ConsolidationOrderAccordion.test.tsx
diabagatekelly: ✅ Fixed - removed fallback, now using taskDate directly.

user-interface/src/data-verification/trustee-verification/TrusteeMatchVerificationAccordion.tsx
diabagatekelly: ✅ Fixed - removed fallback, now using taskDate directly.

user-interface/src/data-verification/DataVerificationScreen.tsx
diabagatekelly: ✅ Fixed - removed fallback, now using taskDate directly. Deployment order ensures taskDate is always present.

user-interface/src/data-verification/TransferOrderAccordion.tsx
diabagatekelly: ✅ Fixed - removed fallback, now using taskDate directly.

diabagatekelly:
## Code Review Responses

Thank you for the thorough review! I've addressed the feedback as follows:

### ✅ Addressed (Commits: b637167a0, 8beb664a2)

**Comment #3 - Records that can't be backfilled marked as success: true**
- Changed to `success: false` with error message for failed records
- Log level already at `warn` (from earlier commit)
- Failed records can be retried since `getPageNeedingBackfill()` queries for missing taskDate

**Comment #4 - Dev seed data format mismatch**
- Updated all seed data taskDate values to ISO 8601 format with timestamps
- Now matches production format: `'2012-06-14T00:00:00.000Z'`

**Comment #6 - Redundant newLastId field**
- Already clean - these backfill types only have `nextCursor`, no `newLastId`

**Comment #7 - Unexported type changes**
- Verified safe - no external imports found via grep
- `DataVerificationItem` and `VerificationListParams` are internal-only

**Comment #9 - Missing test for ConsolidationOrderAccordion**
- Added test case verifying taskDate rendering

**Comment #10 - Repository methods only for migration**
- Refactored to follow phonetic tokens pattern
- Moved query building inline to backfill use cases using QueryBuilder
- Removed 6 methods from repositories/interfaces/mocks

**Comment #11 - Remove orderDate fallbacks in UI**
- Removed all `?? orderDate` fallbacks from 5 files
- Relying on deployment order guarantee that taskDate exists

---

### 🤔 Left open for team discussion

**Comment #1 - taskDate required vs optional**
- **Decision:** Keep `taskDate: string` (required) in types
- **Approach:** Strict deployment order (backfill before frontend)
- **Question:** Does this approach work, or prefer optional with runtime fallbacks?

**Comment #5 - Duplication across backfill use cases**
- **Decision:** Leaving duplication as-is (~420 lines total)
- **Rationale:** Migration code has short lifespan, will be deleted post-backfill
- **Question:** Team comfortable with duplication for temporary migration code?

**Comment #8 - No tests for dataflow handler layer**
- **Observation:** Consistent with all other migrations in codebase
- **Current pattern:** Use-case tests exist, handler tests don't
- **Question:** Should we establish handler testing as a new standard?

**Comment #12 - Three separate dataflows vs one**
- **Rationale:** Granular re-run capability per entity type
- If one backfill fails, re-run only that one without re-processing successful ones
- **Question:** Prefer to keep 3 separate for operational flexibility, or consolidate?

@fmaddenflx @mattstankey - feedback welcome on the open items!

backend/function-apps/dataflows/dataflows.ts
diabagatekelly: Good question! We chose 3 separate dataflows for operational flexibility.

**Rationale:** If one entity type encounters issues during backfill (e.g., transfer orders fail due to data anomalies but consolidations succeed), we can re-run just the failed backfill without re-processing the successful ones.

**Example scenario:**
- Transfer orders: 5,000 records ✅
- Consolidation orders: 3,000 records ✅
- Trustee verifications: 2,000 records ❌ (fails due to missing data)

With 3 dataflows: Fix issue, re-run only trustee verification backfill
With 1 dataflow: Re-run all 10,000 records

**Tradeoff:** 3 dataflows = more queues/functions but granular control vs 1 dataflow = simpler but all-or-nothing re-runs.

Left open for discussion - prefer to keep 3 for flexibility or consolidate for simplicity?

mattstankey:
We still have a few TS errors that are showing up after the latest rebase.  These need to be addressed before merging.

mattstankey (approved):

jalba-flexion (approved):

PR #2478, opened 2026-06-05
CAMS-746: Replace orderType with taskType (3 of 3)

diabagatekelly:
## Summary by Sourcery

Rename domain-level orderType to taskType across orders and trustee match verification flows while preserving the existing MongoDB field name via repository mappers.

Enhancements:
- Introduce fromDb/toDb mapping helpers in Mongo repositories to translate between taskType in domain models and orderType in MongoDB documents.
- Update orders, consolidations, and trustee-match verification repositories, controllers, and use cases to operate on taskType instead of orderType.
- Adjust DXTR gateway queries and dataflows to emit and handle taskType values for transfer and consolidation tasks.
- Refine data verification and UI components (accordions, labels, filters) to display and filter by taskType rather than orderType.
- Add shared test utilities for simulating DB transformations and extend unit/e2e tests to validate the taskType/orderType mapping and new field naming.

Tests:
- Update backend, UI, Playwright, and migration tests to use taskType and to assert correct mapping between domain and database field names.

# Definition of Done:


diabagatekelly:
@sourcery-ai review

mattstankey (commented):
Two issues worth flagging before merge:

**Issue 1 —  filters against the wrong field name (potential silent data loss)**

 is a raw MongoDB dump, so documents have `orderType`, not `taskType`. After this change, `o.taskType !== 'transfer'` will always be `true` (the field is `undefined`), so `transferOrders` will silently be empty. Similarly, `buildSyntheticConsolidation()` now writes `taskType: 'consolidation'` directly into the fixture document — if this is stored as a raw MongoDB document it should be `orderType`, not `taskType`, since the repository mappers won't run over fixture data.

Worth verifying whether this data reaches the DB through a repository (in which case `taskType` is correct) or is written directly (in which case `orderType` is needed).

**Issue 4 — `findMissingTaskDate` queries reference `orderType` without explanation**

In `orders.mongo.repository.ts` and `consolidations.mongo.repository.ts`, the backfill queries still use `doc('orderType').equals(...)`. This is *correct* since these queries hit MongoDB directly and `orderType` is the actual stored field name — but after the rename it will look wrong to future readers. A short inline comment (e.g. `// orderType is the MongoDB field name; domain model uses taskType`) would prevent confusion, consistent with the comment added to the migration type in `migrate-childcases-to-membercases.ts`.

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
```diff
@@ -10,12 +10,20 @@ import { escapeRegExCharacters } from '@common/cams/regex';
...
+// ConsolidationOrderDbBase uses the concrete ConsolidationOrder so TypeScript can enumerate
+// keys for query building. The generic ConsolidationOrderDb<T> is used only for insert/replace
+// operations where the actual T subtype matters.
+type ConsolidationOrderDbBase = Omit<ConsolidationOrder, 'taskType'> & {
+  orderType: ConsolidationOrder['taskType'];
+};
+type ConsolidationOrderDbBaseQueryable = ConsolidationOrderDbBase & { _id: string };
+type ConsolidationOrderDb<T extends ConsolidationOrder> = Omit<T, 'taskType'> & {
+  orderType: T['taskType'];
+};
```
fmaddenflx: Wait a second... Why do we need this translation for ConsolidationOrderDbBase ???

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
```diff
@@ -25,7 +33,20 @@ export default class ConsolidationOrdersMongoRepository<
...
   private static instance: ConsolidationOrdersMongoRepository;
 
-  private doc = using<T>();
+  private dbDoc = using<ConsolidationOrderDbBase>();
+
+  private fromDb(doc: ConsolidationOrderDbBase): ConsolidationOrder {
+    const { orderType, ...rest } = doc;
+    if (orderType !== undefined) {
+      return { ...rest, taskType: orderType };
+    }
```
fmaddenflx: No.  Why are we doing this?  If we've migrated we don't need to do any translation here.  ConsolidationOrders are going to have a taskType only right?  Actually, does DXTR have an orderType or was that something we did on the CAMS side just for classification?  My gut tells me it's a CAMS thing and therefore we don't need to do this.  There shouldn't be an orderType any longer.

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
```diff
@@ -25,7 +33,20 @@ export default class ConsolidationOrdersMongoRepository<
...
+    return rest as ConsolidationOrder;
+  }
+
+  private toDb(item: T): ConsolidationOrderDb<T> {
+    const { taskType, ...rest } = item as T & { taskType: T['taskType'] };
+    return { ...rest, orderType: taskType } as ConsolidationOrderDb<T>;
+  }
```
fmaddenflx: No.  Just no.

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
```diff
@@ -77,42 +99,49 @@ export default class ConsolidationOrdersMongoRepository<
       return;
     }
     try {
-      await this.getAdapter<T>().insertMany(list);
+      await this.getAdapter<ConsolidationOrderDb<T>>().insertMany(
+        list.map((item) => this.toDb(item)),
```
fmaddenflx: no.

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
```diff
@@ -156,9 +188,13 @@ export default class ConsolidationOrdersMongoRepository<
...
     try {
-      const doc = using<ConsolidationOrderQueryable>();
+      type ConsolidationOrderDbQueryable = ConsolidationOrderDbBase & {
+        'memberCases.caseId': string;
+      };
+      const doc = using<ConsolidationOrderDbQueryable>();
       const query = doc('memberCases.caseId').equals(caseId);
-      return await this.getAdapter<T>().find(query as unknown as Query<T>);
+      const results = await this.getAdapter<ConsolidationOrderDbQueryable>().find(query);
+      return results.map((d) => this.fromDb(d) as T);
```
fmaddenflx: 👎

backend/lib/adapters/gateways/mongo/orders.mongo.repository.ts
```diff
@@ -11,6 +11,13 @@ const COLLECTION_NAME = 'orders';
 
 const { and, orderBy, using } = QueryBuilder;
 
+// MongoDB document shape: orderType replaces taskType in the domain model
+type OrderDb = Omit<Order, 'taskType'> & { orderType: Order['taskType'] };
+type TransferOrderDbQueryable = Omit<TransferOrder, 'taskType'> & {
+  orderType: TransferOrder['taskType'];
```
fmaddenflx: 👎

backend/lib/adapters/gateways/mongo/orders.mongo.repository.ts
```diff
@@ -41,45 +48,62 @@ export class OrdersMongoRepository extends BaseMongoRepository implements Orders
...
+    if (orderType !== undefined) {
+      return { ...rest, taskType: orderType } as Order;
+    }
+    return rest as Order;
+  }
+
+  private toDb(item: Order | TransferOrderAction): OrderDb {
+    const { taskType, ...rest } = item as Order & { taskType: Order['taskType'] };
+    return { ...rest, orderType: taskType } as OrderDb;
+  }
```
fmaddenflx: Nu'ah

backend/lib/adapters/gateways/mongo/consolidations.mongo.repository.ts
mattstankey: We'll still need to run this migration for staging data too.

mattstankey (approved):

jalba-flexion (approved):

backend/lib/adapters/gateways/mongo/orders.mongo.repository.ts
```diff
@@ -45,41 +45,44 @@ export class OrdersMongoRepository extends BaseMongoRepository implements Orders
...
 
   async update(data: TransferOrderAction) {
     try {
-      const existingQuery = using<TransferOrder>()('id').equals(data.id);
-
-      const adapter = this.getAdapter<TransferOrder>();
-      const foundOrder = await adapter.findOne(existingQuery);
+      const existingQuery = using<Order>()('id').equals(data.id);
+      const adapter = this.getAdapter<Order>();
+      const foundOrder = (await adapter.findOne(existingQuery)) as TransferOrder;
```
fmaddenflx: I'm curious to know why this is necessary.  How is Order different from TransferOrder and why use Order if we need to cast line 72 as TransferOrder?

fmaddenflx: I see how Order is TransferOrder | ConsolidationOrder.  But I don't know if this change of these 3 lines in necessary.

fmaddenflx (approved):

backend/lib/adapters/gateways/mongo/orders.mongo.repository.ts
mattstankey: Updated

PR #2483, opened 2026-06-05
CAMS-616 - Refactor downstream CMMAP_ALL

btposey:
# Purpose

Replace CMMAP_ALL view with unified appointment table

# Major Changes

The existing CMMAP_ALL UNION view was fragile — any gap in the exclusion logic silently produced duplicate
rows, and ACMS records for cases CAMS had superseded retained stale disposition status.

This PR replaces it with a physical table kept current by two write paths:

- CAMS event handler — writes to CMMAP_CAMS and CMMAP_ALL in a single SQL transaction; queue retry handles
failures automatically
- Daily timer trigger (ACMS-CAMS-TRANSITION-DAILY-SYNC) — polls CMMAP for active appointments newer than a
watermark stored in a new CMMAP_SYNC_CONTROL table; a SOURCE != 'CAMS' guard ensures CAMS-owned rows are
never overwritten

Also hardens downstream event emission: the staff assignment path previously swallowed queue failures with
only a log message; it now writes a STAFF_ASSIGNMENT_DOWNSTREAM_SYNC_ERROR document to Cosmos on failure,
matching the compensation pattern already used by the trustee appointment path.

Integration tests were updated to call trusteeAppointmentHandler directly (no function host required) and
assert the dual-write no-duplicate invariant against a local SQL Edge container.

## Summary by Sourcery

Replace the legacy CMMAP_ALL UNION view with a physical unified appointment table kept in sync by a new daily ACMS sync job and dual-write event handlers, while hardening downstream staff-assignment error handling and expanding integration and unit test coverage.

New Features:
- Introduce a CMMAP_ALL physical table in ACMS_REP_SUB, plus CMMAP_SYNC_CONTROL, to serve as the authoritative unified appointment store for ACMS and CAMS appointments.
- Add a daily timer-triggered AcmsDailySync function that merges ACMS CMMAP changes into CMMAP_ALL using a watermark-based incremental or full-load strategy.
- Expose multiple one-shot integration harness commands to exercise trustee and staff handlers, daily sync behaviour, source guards, and lifecycle scenarios end to end.

Bug Fixes:
- Ensure staff assignment downstream failures persist a STAFF_ASSIGNMENT_DOWNSTREAM_SYNC_ERROR document for later replay instead of only logging the error.
- Prevent ACMS daily sync from overwriting CAMS-owned rows in CMMAP_ALL using a SOURCE-based ownership guard and last-writer-wins semantics.

Enhancements:
- Refactor trustee and staff downstream handlers to use ApplicationContext and AbstractMssqlClient.withTransaction for dual-writing CMMAP_CAMS and CMMAP_ALL atomically.
- Augment observability for downstream handlers and the daily sync with structured traces and custom metrics for success, failure, and rows merged.
- Extend AbstractMssqlClient with transaction support and update tests to validate transactional behaviour and connection reuse.
- Document the ACMS-CAMS appointment sync architecture and assumptions in an ADR and update downstream and integration testing READMEs to match the new table- and sync-based design.

Build:
- Add an initial database migration that creates CMMAP_CAMS, CMMAP_ALL, CMMAP_SYNC_CONTROL, seeds CMMAP_ALL from CMMAP, and initializes the sync watermark for the daily job.

Documentation:
- Add an ADR describing the ACMS-CAMS appointment sync architecture and rationale for replacing the CMMAP_ALL view with a physical table and dual-write + sync model.
- Revise downstream and integration harness documentation to describe the new CMMAP_ALL table, CMMAP_SYNC_CONTROL, migration workflow, and the expanded set of integration commands.

Tests:
- Substantially expand unit tests for downstream handlers and AcmsDailySync to cover transactional semantics, SQL shape (MERGE guards), logging, and observability traces.
- Broaden integration tests to validate dual-writes to CMMAP_CAMS and CMMAP_ALL, daily sync watermark behaviour, CAMS source guards, full-load and incremental paths, and ACMS↔CAMS lifecycle override scenarios.


backend/function-apps/dataflows/downstream/acms-daily-sync.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+  return connectionPool;
+}
+
+async function readWatermark(pool: sql.ConnectionPool): Promise<Date> {
+  const request = pool.request();
+  const result = await request.query(`
+    SELECT LAST_SYNC_DATE
+    FROM CMMAP_SYNC_CONTROL
+    WHERE PROCESS_NAME = 'ACMS_DAILY'
```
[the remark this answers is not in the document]
btposey: ✅ Fixed. `updateWatermark` now checks `rowsAffected[0]` and throws a descriptive error if the `ACMS_DAILY` row is missing, rather than silently no-oping with a stale watermark. Added a unit test covering this path.

backend/function-apps/dataflows/downstream/acms-daily-sync.ts
```diff
@@ -0,0 +1,231 @@
...
+        APPT_DISP        = source.APPT_DISP,
+        DISP_DATE        = source.DISP_DATE,
+        DISP_DATE_DT     = source.DISP_DATE_DT,
+        APPTEE_ACTIVE    = source.APPTEE_ACTIVE,
+        ALPHA_SEARCH     = source.ALPHA_SEARCH,
+        UPDATE_DATE      = source.UPDATE_DATE,
+        CDB_UPDATE_DATE  = source.CDB_UPDATE_DATE,
+        CDB_UPDATE_DATE_DT = source.CDB_UPDATE_DATE_DT,
+        LAST_UPDATED     = @RUN_AT
+    WHEN NOT MATCHED BY TARGET THEN
```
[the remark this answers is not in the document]
btposey: ✅ Fixed. The `WHEN MATCHED` UPDATE now refreshes all ACMS-owned columns — `COMMENTS`, `USER_ID`, `HEARING_SEQUENCE`, `REGION_CODE`, all RGN/CDB date fields, and `REPLICATED_DATE` — preventing `CMMAP_ALL` from drifting from the ACMS source once a row exists.

backend/function-apps/dataflows/downstream/acms-daily-sync.ts
```diff
@@ -0,0 +1,231 @@
...
+    WHERE PROCESS_NAME = 'ACMS_DAILY'
+  `);
+}
+
+/**
+ * Merges active ACMS appointments (and their immediate predecessors) into CMMAP_ALL.
+ * Rows already owned by CAMS (SOURCE='CAMS') are never overwritten.
+ * Runs as a single MERGE statement so the DB engine handles the upsert atomically.
+ */
+async function mergeCmmapRows(pool: sql.ConnectionPool, watermark: Date): Promise<number> {
```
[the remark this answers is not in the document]
btposey: Skipping this one intentionally. The named-fragment approach adds indirection without reducing ambiguity — a reader still needs to mentally splice the fragments to understand what SQL is being sent. The SQL is dense but self-contained, and the inline comments (`-- Active appointments newer than watermark`, `-- Immediate predecessor per (case, APPT_TYPE)`) already mark the two logical sections. The `withConnectedPool` helper would save two lines at the cost of an extra layer to trace through. We will keep the current structure.

backend/function-apps/dataflows/downstream/database/acms-cams-transition/migrations/001-initial-schema.sql
```diff
@@ -22,14 +27,116 @@ GO
...
+-- "Immediate predecessor" = the highest RECORD_SEQ_NBR row per (case, APPT_TYPE)
+-- that is NOT the active row. We include it so CMMAP_ALL reflects the full
+-- transition context when CAMS takes over a case.
+INSERT INTO CMMAP_ALL (
+    DELETE_CODE,
+    CASE_DIV, CASE_YEAR, CASE_NUMBER, RECORD_SEQ_NBR,
+    APPT_TYPE,
+    PROF_CODE, GROUP_DESIGNATOR,
+    APPT_DATE, APPT_DATE_DT,
+    APPT_DISP, DISP_DATE, DISP_DATE_DT,
```
fmaddenflx, line 63: What is PROF in PROF_CODE?  What is the DT in APPT_DATE_DT?  Do we have a dictionary somewhere for all of these ACMS fields?

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -1,10 +1,11 @@
-import { InvocationContext, StorageQueueOutput } from '@azure/functions';
```
jamesobrooks, line 1: **[Security / High]** The new `getConnectionPool()` only supports SQL username/password auth (`user`/`password` from env vars), with no fallback to Azure AD managed identity. The established pattern in `application-configuration.ts` (lines 92–140) branches: if `ACMS_MSSQL_USER`/`ACMS_MSSQL_PASS` are absent it configures `authentication.type = 'azure-active-directory-default'`. On Flexion/AAD deployments the Bicep template intentionally omits those credentials and sets `ACMS_MSSQL_CLIENT_ID` instead — so the new pool will receive `undefined` for both fields and fail to connect on those environments.

Suggested fix: mirror the existing pattern from `application-configuration.ts`:

```typescript
const user = process.env.ACMS_MSSQL_USER;
const password = process.env.ACMS_MSSQL_PASS;
const config: sql.config = {
  server: process.env.ACMS_MSSQL_HOST || '',
  database: process.env.ACMS_MSSQL_DATABASE || '',
  options: {
    encrypt: ...,
    trustServerCertificate: ...,
  },
};
if (user && password) {
  config.user = user;
  config.password = password;
} else {
  const clientId = process.env.ACMS_MSSQL_CLIENT_ID;
  config.authentication = {
    type: 'azure-active-directory-default',
    ...(clientId ? { options: { clientId } } : {}),
  };
}
```

Alternatively, consolidate pool construction into the existing `application-configuration.ts` helper so this is a single source of truth.

jamesobrooks (commented):
Automated multi-agent review (software-architect, devsecops, SRE, pragmatic). All findings were treated as hypotheses and actively challenged before being reported. Disproved hypotheses are listed at the bottom.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -262,8 +271,30 @@ async function upsertCmmapCamsRow(row: CmmapCamsRow): Promise<void> {
   request.input('CAMS_USER_ID', sql.VarChar(50), row.CAMS_USER_ID);
   request.input('CAMS_USER_NAME', sql.VarChar(100), row.CAMS_USER_NAME);
   request.input('LAST_UPDATED', sql.DateTime2(3), row.LAST_UPDATED);
+}
+
+async function upsertCmmapCamsRow(row: CmmapCamsRow): Promise<void> {
+  const pool = getConnectionPool();
```
jamesobrooks: **[Security]** `getConnectionPool()` (used here and at line 707) only supports SQL username/password auth — there is no fallback to Azure AD managed identity. The established pattern in `application-configuration.ts` (lines 92–140) branches: if `ACMS_MSSQL_USER`/`ACMS_MSSQL_PASS` are absent it configures `authentication.type = 'azure-active-directory-default'` (with optional `ACMS_MSSQL_CLIENT_ID`). On Flexion/AAD deployments the Bicep template intentionally omits those credentials, so the pool will receive `undefined` for both fields and fail to connect.

Also in `getConnectionPool()`: the `encrypt` default is `ACMS_MSSQL_ENCRYPT !== 'false'` (default `true`), while every other ACMS connection uses `ACMS_MSSQL_ENCRYPT?.toLowerCase() === 'true'` (default `false`). Same env var, opposite defaults — operators controlling TLS via Key Vault will get inconsistent behaviour across write paths.

Suggested fix: consolidate pool construction into the shared config helper, or mirror the auth-branching pattern from `application-configuration.ts`.

jamesobrooks (commented):
Automated multi-agent review (software-architect, devsecops, SRE, pragmatic reviewers from ustp-cams-fdp). Each finding was treated as a hypothesis and actively challenged before being reported.

**Disproved / not flagged:** SQL injection via `buildMergeQuery` (tableName is TypeScript-constrained to two literals); watermark race condition (day-granular NUMERIC comparison is safe); MERGE source ambiguity in UNION ALL (active and predecessor rows have distinct RECORD_SEQ_NBR by definition); timer concurrency (Azure Functions timer mutex prevents overlap); `updateWatermark` rowsAffected guard spurious throw (MERGE upsert with constant source always affects ≥1 row — guard is harmless defensive code, not dead code); EPOCH_WATERMARK vs schema seed mismatch (intentional fallback for deleted control row); PII exposure via DLQ (no new surface introduced).

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -262,8 +271,30 @@ async function upsertCmmapCamsRow(row: CmmapCamsRow): Promise<void> {
...
+    const allRequest = transaction.request();
+    bindRowParams(allRequest, row);
+    await allRequest.query(buildMergeQuery('CMMAP_ALL'));
 
-  await request.query(query);
+    await transaction.commit();
+  } catch (error) {
+    await transaction.rollback();
+    throw error;
+  }
```
jamesobrooks: **[Architecture]** `getConnectionPool()` reimplements connection pool lifecycle management (lazy init, module-level singleton, `deferClose` registration) independently of `AbstractMssqlClient`, which already centralises this for all other ACMS/DXTR connections. This creates two independent `mssql` footholds in the codebase.

The deviation is partially justified — `AbstractMssqlClient.executeQuery()` does not expose transaction support, and the dual-write here requires a transaction. That explains bypassing the *query* abstraction, but not duplicating pool construction and lifecycle management. Consider extending `AbstractMssqlClient` with a `withTransaction` method, or at minimum extracting pool configuration into the existing `getAcmsDbConfig` helper so pool creation is a single source of truth.

backend/lib/use-cases/case-assignment/case-assignment.ts
```diff
@@ -198,18 +198,27 @@ export class CaseAssignmentUseCase {
     await casesRepo.createCaseHistory(history);
 
     if (context.featureFlags['downstream-staff-assignments-enabled']) {
+      const assignmentRepo = factory.getAssignmentRepository(context);
```
jamesobrooks: **[Design]** A second `assignmentRepo` is obtained here via `factory.getAssignmentRepository(context)`, but an identical one is already in scope from line 77 in the same method. The factory returns a singleton so both variables point to the same object — the extra acquisition is harmless but confusing. The outer repo already exposes `upsertDownstreamSyncError` (it's on the interface), so this line can be removed.

backend/lib/use-cases/case-assignment/case-assignment.ts
```diff
@@ -198,18 +198,27 @@ export class CaseAssignmentUseCase {
...
-          });
+          await this.apiToDataflowsGateway.queueCaseAssignmentEvent(downstreamEvent);
         } catch (queueError) {
           context.logger.error(
             MODULE_NAME,
-            `Failed to enqueue staff assignment event for case ${assignment.caseId}, user ${assignment.userId} — assignment written but downstream not notified. Manual replay required.`,
+            `Failed to enqueue staff assignment event for case ${assignment.caseId}, user ${assignment.userId} — writing sync error doc for replay.`,
             queueError,
           );
+          await assignmentRepo.upsertDownstreamSyncError({
```
jamesobrooks: **[Reliability]** `upsertDownstreamSyncError` is called inside the `catch (queueError)` block but is not wrapped in its own try/catch. If Cosmos is transiently unavailable here the exception propagates out of the `for` loop, aborting all remaining assignments — no compensation doc is written for them and the outer call returns HTTP 500, even though all primary assignment writes already succeeded.

The trustee path handles this correctly (see `backfill-trustee-appointments-downstream.test.ts` — "should swallow upsertDownstreamSyncError failure"). Suggested fix: wrap this call in its own try/catch, log the secondary failure, and continue the loop — matching the trustee pattern. A test covering the double-failure path would also confirm the behaviour.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -462,3 +493,256 @@ export async function trusteeAppointmentHandler(
...
+        m.RGN_CREATE_DATE, m.RGN_UPDATE_DATE, m.RGN_CREATE_DATE_DT, m.RGN_UPDATE_DATE_DT,
+        m.CDB_CREATE_DATE, m.CDB_UPDATE_DATE, m.CDB_CREATE_DATE_DT, m.CDB_UPDATE_DATE_DT,
+        m.UPDATE_DATE, m.REPLICATED_DATE, m.id, m.RRN
+      FROM dbo.CMMAP m
+      WHERE m.APPTEE_ACTIVE = 'Y'
+        AND m.DELETE_CODE = ' '
+        AND m.CDB_UPDATE_DATE > CONVERT(NUMERIC(8,0),
+              CONVERT(VARCHAR(8), @WATERMARK, 112))
+
+      UNION ALL
```
jamesobrooks: **[Complexity]** The predecessor `UNION ALL` branch fetches inactive rows whose active counterpart was updated after the watermark. After the full load, every `APPTEE_ACTIVE='N' AND DELETE_CODE=' '` predecessor row is already in `CMMAP_ALL`. The MERGE has no `WHEN NOT MATCHED BY SOURCE DELETE` clause, so predecessors already in `CMMAP_ALL` cannot be evicted. The incremental branch therefore re-upserts already-present, unchanged rows on every daily run.

Before merge: which specific downstream consumer reads inactive predecessor rows from `CMMAP_ALL` and would break without this daily re-sync? If none is identified, the branch adds query complexity and load without a current use case.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -462,3 +493,256 @@ export async function trusteeAppointmentHandler(
...
+    const rowsAffected = await mergeCmmapRows(pool, watermark, fullLoad);
+    const runAt = new Date();
+    await updateWatermark(pool, runAt);
+
+    context.log({
+      moduleName: DAILY_SYNC_MODULE,
+      success: true,
+      durationMs: Date.now() - startTime,
+      rowsAffected,
+      watermark: watermark.toISOString(),
```
jamesobrooks: **[Observability]** The success log emits `watermark` (the *previous* watermark) but not `runAt` (the new watermark just written). After a `rowsAffected: 0` run an operator cannot distinguish "quiet night, sync healthy" from "watermark stuck, merge repeating with no new rows". Suggested: add `newWatermark: runAt.toISOString()` to the log object.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -462,3 +493,256 @@ export async function trusteeAppointmentHandler(
...
+  }
+
+  try {
+    const watermark = await readWatermark(pool);
+    const fullLoad = watermark.getTime() === EPOCH_WATERMARK.getTime();
+    const rowsAffected = await mergeCmmapRows(pool, watermark, fullLoad);
+    const runAt = new Date();
+    await updateWatermark(pool, runAt);
+
+    context.log({
```
jamesobrooks: **[Observability]** Sync health signals land in `traces` (via `context.log`) rather than `customMetrics`. There is no way to create a metric alert on "zero rows merged for N consecutive days" (potential indicator of a frozen ACMS replica or a stuck watermark) or on merge duration exceeding a threshold without parsing raw trace JSON. The codebase has `observability.ts` with `trackMetric`. Consider emitting at minimum `acms_daily_sync_rows_merged` and `acms_daily_sync_duration_ms` as custom metrics to enable straightforward KQL-based metric alerts.

backend/function-apps/dataflows/downstream/database/acms-cams-transition/migrations/001-initial-schema.sql
btposey: The PROF is short for profession, and DT is likewise short for DATE. These are in the ACMS data dictionary for the CMMAP table which can be reasoned with the /cams-gateways skill by Claude.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed. `getConnectionPool()` and its inline SQL auth config (which lacked Azure AD fallback and had an inverted `encrypt` default) have been removed entirely. Both the dual-write upsert and the daily sync now construct an `AcmsRepSubClient` — a subclass of `AbstractMssqlClient` — from `context.config.acmsDbConfig`. This inherits the established auth-branching pattern (SQL auth if `ACMS_MSSQL_USER`/`PASS` are set, otherwise `azure-active-directory-default` with optional `ACMS_MSSQL_CLIENT_ID`) and the correct `encrypt` default. Commit: `Consolidate pool construction in downstream`.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed — same resolution as the Security/High comment above. `getConnectionPool()` is gone; `AcmsRepSubClient` inherits auth branching and the correct `encrypt` default from `AbstractMssqlClient` via `context.config.acmsDbConfig`. Both threads are resolved by the same commit: `Consolidate pool construction in downstream`.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed — and then some. We agreed that adding transaction support to `AbstractMssqlClient` is a generically useful capability, not a downstream-specific concern. `withTransaction<T>` was added to the shared base class (commit: `Add withTransaction to AbstractMssqlClient`) so any future CAMS use case that needs atomic SQL can use it without reimplementing pool or transaction lifecycle. `getConnectionPool()` and all its manual lifecycle management was then removed from the downstream module (commit: `Consolidate pool construction in downstream`). The downstream is now a clean consumer of shared infrastructure, not a reimplementor of it.

backend/lib/use-cases/case-assignment/case-assignment.ts
btposey: ✅ Fixed. The redundant `factory.getAssignmentRepository(context)` call inside the feature-flag block has been removed. The `assignmentRepo` already in scope from the outer function is used directly — it exposes `upsertDownstreamSyncError` and the factory returns a singleton anyway. Commit: `Fix assignmentRepo dup and sync error guard`.

backend/lib/use-cases/case-assignment/case-assignment.ts
btposey: ✅ Fixed. `upsertDownstreamSyncError` is now wrapped in its own `try/catch`. A Cosmos failure during compensation logs the secondary failure and continues the loop — remaining assignments are not aborted. This matches the trustee appointment pattern exactly. A test covering the double-failure path (queue unavailable + Cosmos unavailable) was added to confirm the loop continues. Commit: `Fix assignmentRepo dup and sync error guard`.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed — predecessor branch removed from the incremental path. The reasoning: `CMMAP_ALL` is required to be a drop-in replacement for `CMMAP`, including predecessor rows, so predecessors must exist in the table. They are seeded correctly at full-load. The incremental sync can then omit them because ACMS does not mutate predecessor rows after CAMS takes over appointments for a division — once a row is seeded, it is correct indefinitely. The MERGE has no DELETE clause so seeded rows remain permanently. This assumption is documented explicitly in the downstream README. Commit: `Drop predecessor re-sync from incremental sync`.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed. `newWatermark` is now emitted in the trace details alongside the previous `watermark`, so an operator can distinguish "quiet night, sync healthy" from "watermark stuck." Both values appear in the `completeDataflowTrace` call under `details.watermark` (previous) and `details.newWatermark` (just written). Commit: `Add sync health metrics to downstream handlers`.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed. All handler and sync telemetry now flows through `completeDataflowTrace`, which calls `observability.completeTrace` with named metrics. The following custom App Insights metrics are emitted on every invocation: `acms_cmmap_handler_write_success`, `acms_cmmap_handler_write_failure` (per queue handler invocation), and `acms_cmmap_sync_rows_merged` (per daily sync run). These can be used directly in KQL metric alerts without parsing trace JSON. This was enabled by first bridging the handlers to `ApplicationContext` (commit: `Bridge downstream handlers to ApplicationContext`), which gave the downstream layer access to the shared `AppInsightsObservability` gateway. Commit: `Add sync health metrics to downstream handlers`.

btposey:
@sourcery-ai re-review please.

backend/function-apps/dataflows/downstream/acms-cams-transition.test.ts
```diff
@@ -10,23 +10,64 @@ import {
   transformTrusteeAppointmentToRow,
   serializeError,
   ValidationError,
+  AcmsDailySync,
 } from './acms-cams-transition';
 import StaffAssignmentDownstream from './staff-assignment-downstream';
 import TrusteeAppointmentDownstream from './trustee-appointment-downstream';
 
+const { mockLogger, mockExtraOutputs, mockObservability } = vi.hoisted(() => ({
```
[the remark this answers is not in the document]
btposey: Declining. These handlers are transition-period code — they exist to bridge ACMS and CAMS during migration and are candidates for deletion once CAMS fully takes over. The abstraction overhead (`resetSqlMocks`, `makeHandlerHarness`, `mockDailySyncSequence`) adds indirection that increases cognitive load for future readers without providing any tangible benefit: it does not catch new bugs, does not clarify invariants, and does not simplify behavior. The `beforeEach` blocks are already short, explicit, and readable as-is. Gold-plating a disposable layer is not a worthwhile investment at this stage.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -146,34 +150,28 @@ function buildBaseCmmapRow(
...
-  return connectionPool;
 }
 
-async function upsertCmmapCamsRow(row: CmmapCamsRow): Promise<void> {
-  const pool = getConnectionPool();
-  if (!pool.connected) {
-    await pool.connect();
-  }
-  const request = pool.request();
+function buildMergeQuery(tableName: 'CMMAP_CAMS' | 'CMMAP_ALL'): string {
```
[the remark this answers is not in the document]
btposey: Declining all four sub-suggestions.

**MERGE constants (suggestion 1):** The proposed `BASE_MERGE` shared fragment contains a regression — it omits `CAMS_USER_ID` and `CAMS_USER_NAME` from the `WHEN MATCHED UPDATE` clause that `CMMAP_CAMS` requires. Applying it as written would silently drop those columns on existing-row updates. The current `buildMergeQuery` parameterisation is correct and the conditional logic is contained in one place.

**`upsertCmmapCams` / `upsertCmmapAll` split (suggestion 2):** The function is called in two places, the transaction callback is already four lines, and splitting it into three named functions adds indirection for no clarity gain. The dual-write intent is already visible at the call site.

**`runWithTrace` helper (suggestion 3):** Replacing a straightforward `try/catch` with a higher-order function that takes a `metrics` callback of type `(success: boolean, error?: unknown) => {...}` increases cognitive load, not reduces it. The DLQ routing and tracing are already clearly separated within `handleQueueEvent` by structure alone.

**`buildSourceQuery` / `buildMergeCmmapAllQuery` extraction (suggestion 4):** The SQL is dense but self-contained and the full/incremental split is a single conditional. Extracting named builders adds files and import chains to a handler that is a transition-period artefact — it will be deleted when CAMS fully supersedes ACMS for appointments. The inline comments already mark the two logical sections clearly.

These handlers exist solely for the ACMS→CAMS migration window. Adding abstraction layers to code scheduled for deletion is not a worthwhile investment.

backend/lib/adapters/gateways/mongo/case-assignment.mongo.repository.ts
```diff
@@ -132,4 +133,18 @@ export class CaseAssignmentMongoRepository
...
+
+  async upsertDownstreamSyncError(syncError: CaseAssignmentDownstreamSyncError): Promise<void> {
+    try {
+      const d = using<CaseAssignmentDownstreamSyncError>();
+      const query = and(
+        d('documentType').equals('STAFF_ASSIGNMENT_DOWNSTREAM_SYNC_ERROR'),
+        d('caseId').equals(syncError.caseId),
+        d('userId').equals(syncError.userId),
+      );
+      await this.getAdapter<CaseAssignmentDownstreamSyncError>().replaceOne(query, syncError, true);
```
fmaddenflx: replaceOne:  Is this overwriting the Error log each time?

btposey: It's supposed to be idempotent. The upsert is so it doesn't dupe.

btposey: Which brings up a good point to look it over a little closer because it may not be idempotent....

btposey: Good catch — this exposed a design problem worth addressing directly. The `replaceOne` upsert was acting as a replay buffer for queue failures, but no replay mechanism was ever implemented, making it dead infrastructure. More importantly, the scenario where the queue is unavailable is almost certainly the same scenario where Cosmos is also unavailable — so the fallback write would fail silently too, losing the event entirely either way.

We decided to remove the sync error doc pattern entirely from the queue-failure paths and rely on `context.logger.error` surfacing to Application Insights instead. That is the appropriate failure mode for a transient infra blip. The `CaseAssignmentDownstreamSyncError` type and `upsertDownstreamSyncError` method on `CaseAssignmentRepository` have been removed.

btposey: Good catch — this exposed a design problem worth addressing directly. The `replaceOne` upsert was acting as a replay buffer for queue failures, but no replay mechanism was ever implemented, making it dead infrastructure. More importantly, an infra outage that takes down the storage queue almost certainly also takes down Cosmos, so the fallback write would fail silently anyway.

We removed the sync error doc pattern entirely from all queue-failure paths and replaced it with `context.logger.error` surfacing to Application Insights. For the missing ACMS professional ID case we now emit a sentinel value (`XX-99999`) so downstream events are never silently dropped — future trustee ID management features can identify and republish those rows. `CaseAssignmentDownstreamSyncError`, `TrusteeAppointmentDownstreamSyncError`, and all `upsertDownstreamSyncError` methods have been removed. Commit: `84c0aa4`.

btposey:
Addressing the Sourcery suggestion from review https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/pull/2483#pullrequestreview-4459972842:

> The `AbstractMssqlClient.withTransaction` helper currently ignores the `ApplicationContext` and bypasses the existing `executeQuery` error/observability pipeline.

Fixed in commit `ef10e87`. The `_context` parameter is now active — on any error (callback throw or commit failure) `context.logger.error` is called with the module name and error before re-throwing, and the error is wrapped with `getCamsError`. This matches the structured logging and error-wrapping behavior `executeQuery` already provides for non-transactional queries, so transaction failures now surface consistently in Application Insights.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -458,7 +477,270 @@ export async function trusteeAppointmentHandler(
...
+async function syncAcmsToAll(invocationContext: InvocationContext): Promise<void> {
+  const context = await ContextCreator.getApplicationContext({ invocationContext });
+  const client = new AcmsRepSubClient(context);
+  const trace = context.observability.startTrace(invocationContext.invocationId);
+
+  try {
+    const watermark = await readWatermark(client, context);
+    const fullLoad = watermark.getTime() === EPOCH_WATERMARK.getTime();
+    const rowsAffected = await mergeCmmapRows(client, context, watermark, fullLoad);
+    const runAt = new Date();
```
mattstankey, line 696: **Issue 1: Watermark race condition**

`runAt` is captured *after* `mergeCmmapRows` returns, which means any rows written to `CMMAP` between the start of the MERGE and this `new Date()` call will be outside the new watermark and missed by the next incremental run.

Fix: capture the run timestamp *before* calling `mergeCmmapRows` so it covers the full sync window:

```ts
const runAt = new Date();  // ← move this up, before mergeCmmapRows
const rowsAffected = await mergeCmmapRows(client, context, watermark, fullLoad);
await updateWatermark(client, context, runAt);
```

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
```diff
@@ -458,7 +477,270 @@ export async function trusteeAppointmentHandler(
...
   );
 }
+
+// ─── ACMS daily sync handler ──────────────────────────────────────────────────
+
+const DAILY_SYNC_MODULE = ModuleNames.ACMS_CAMS_TRANSITION_DAILY_SYNC;
+const DAILY_SYNC_TIMER = buildFunctionName(DAILY_SYNC_MODULE, 'timerTrigger');
+
+// UNIX epoch used as watermark on first run — triggers full CMMAP load.
+const EPOCH_WATERMARK = new Date(0);
```
mattstankey, line 492: **Issue 2: Misleading EPOCH_WATERMARK comment**

The comment says "first run" but after the migration runs, `CMMAP_SYNC_CONTROL` is seeded and the watermark is set to `GETDATE()`, so a normal first run is always incremental. `EPOCH_WATERMARK` is only reached if the control row is entirely absent (unexpected/broken state), not on first run.

Suggest updating the comment:

```ts
// Fallback watermark when no CMMAP_SYNC_CONTROL row exists — triggers a full CMMAP load.
// Under normal operation the migration seeds the control row, so this path is a break-glass fallback.
const EPOCH_WATERMARK = new Date(0);
```

backend/lib/adapters/gateways/abstract-mssql-client.ts
```diff
@@ -19,6 +19,34 @@ export abstract class AbstractMssqlClient {
...
+      await connectionPool.connect();
+    }
+    const transaction = connectionPool.transaction();
+    await transaction.begin();
+    try {
+      const result = await fn({ request: () => transaction.request() });
+      await transaction.commit();
+      return result;
+    } catch (error) {
+      await transaction.rollback();
```
mattstankey: **Issue 3: Unhandled rollback failure**

If the callback throws and `transaction.rollback()` also throws, the rollback error propagates uncaught, masking the original error and losing the structured log/`CamsError` wrapping.

Standard defensive pattern — swallow the rollback error and re-throw the original:

```ts
} catch (error) {
  try {
    await transaction.rollback();
  } catch (rollbackError) {
    context.logger.error(this.moduleName, `Rollback failed after error`, { rollbackError });
  }
  const unknownError = error instanceof Error ? error : new Error(...);
  // ... rest of existing error handling
}
```

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed in 0b43b774e. Moved `runAt = new Date()` to before `mergeCmmapRows` so the watermark covers the full sync window.

backend/function-apps/dataflows/downstream/acms-cams-transition.ts
btposey: ✅ Fixed in d515686b4. Updated the comment to clarify this is a break-glass fallback for missing control rows, not a normal first-run path.

backend/lib/adapters/gateways/abstract-mssql-client.ts
btposey: ✅ Fixed in 124b908d8. Wrapped `rollback()` in a try/catch — rollback failures are now logged separately and the original error is re-thrown.

mattstankey (approved):

PR #2489, opened 2026-06-09
CAMS-752 Fix mobile table spacing and breakpoints

bityogi:
# Bug

The associated banks table had three mobile view issues:
1. Vertical spacing between lines was 16px instead of the expected 8px
2. The table stayed in columnar layout too long, becoming illegible as columns compressed
3. The Edit Status button was right-aligned in mobile view using a utility class instead of SCSS media queries

# Purpose

Improve the associated banks table mobile experience by snapping to stacked layout earlier and fixing spacing/alignment per design feedback.

# Major Changes

* Raised the responsive breakpoint for the associated banks table from 641px to 880px, using the same stacked-override pattern as StaffAssignmentScreen
* Moved cell padding overrides (0.5rem) into the desktop-only media query so mobile stacked view uses the default 8px (0.25rem) spacing
* Replaced `text-right` utility class with SCSS `text-align: right` scoped to the desktop media query, left-aligned in mobile per Fritz's guidance

# Testing/Validation

All 166 existing tests in the bankruptcy-software directory pass. Verified SCSS compiles successfully through Vite build. No test file changes needed since this is purely CSS/layout behavior.

# Notes

The breakpoint of 880px was chosen because the left navigation pane takes 12rem (192px) at 640px+, leaving only ~450px for the table at the global 641px breakpoint — far too narrow for 4 columns. At 880px, the table gets ~688px which comfortably fits the proportional column widths (3:1:1:fixed).

This follows the established pattern from StaffAssignmentScreen.scss which also overrides the CamsTable global breakpoint using the `cams-table-stacked` and `cams-table-flex-row` mixins.

# Definition of Done:

## Summary by Sourcery

Adjust associated banks table responsive behavior and alignment to improve mobile readability and match design expectations.

Bug Fixes:
- Correct mobile vertical spacing in the associated banks table by using default cell padding in stacked view.
- Ensure the associated banks table switches to stacked layout at a wider breakpoint to avoid overly compressed columns on smaller screens.
- Align the Edit Status column right only on desktop while keeping it left-aligned in mobile for better layout consistency.

Enhancements:
- Adopt shared CamsTable mixins for stacked and flex-row layouts to align the associated banks table with existing responsive patterns.

### DIFF
--- a/user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
+++ b/user-interface/src/admin/bankruptcy-software/AssociatedBanksTable.tsx
@@ -111,7 +111,7 @@ export function AssociatedBanksTable({
           <CamsTableHeaderCell>Associated Bank Name</CamsTableHeaderCell>
           <CamsTableHeaderCell>Trustees</CamsTableHeaderCell>
           <CamsTableHeaderCell>Status</CamsTableHeaderCell>
-          <CamsTableHeaderCell className="text-right"></CamsTableHeaderCell>
+          <CamsTableHeaderCell></CamsTableHeaderCell>
         </CamsTableHeader>
         <CamsTableBody>
           {associations.length === 0 && (
@@ -157,7 +157,7 @@ export function AssociatedBanksTable({
                 <CamsTableCell data-cell="Status">
                   {association.status === 'active' ? 'Active' : 'Inactive'}
                 </CamsTableCell>
-                <CamsTableCell data-cell="" className="text-right">
+                <CamsTableCell data-cell="">
                   {bankProfileMap.get(association.bankId)?.status === 'active' && (
                     <Button
                       id={`edit-status-${association.bankId}`}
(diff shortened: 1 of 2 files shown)


fmaddenflx (approved):

fmaddenflx:
looks and works great.

bityogi:
Thanks for the feedback.

**Re: centralizing breakpoints** — The 641px value comes from the global CamsTable breakpoint in `CamsTable.scss`, and 880px is derived from this table's specific layout context (left nav width + minimum usable column space). StaffAssignmentScreen uses 1025px for the same reason — different context, different value. These aren't shared constants that should stay in sync; they're each calculated for their container's constraints.

**Re: nth-child selectors** — Fair point in general, but this table has a stable 4-column structure and follows the same pattern already used in StaffAssignmentScreen. Adding semantic column classes would be a broader CamsTable convention change best discussed separately.

PR #2497, opened 2026-06-11
CAMS-763 Optimize trustee name search

cruzjone-flexion:
## Summary

- Remove `phoneticTokens.notExists()` fallback branch from `searchTrusteesByNameScored` — this OR branch caused a full collection scan on every trustee search query. All trustees now have `phoneticTokens` so the fallback is no longer needed.
- Replace the `$**` wildcard index with a targeted compound `{documentType, phoneticTokens}` index on the trustees collection, which aligns with the query's actual filter shape.

## Why now

A/B testing on the `CAMS-763-perf-ab` branch confirmed the optimized query (without `notExists()`) is consistently faster and returns identical results. This PR applies those optimizations directly to the existing endpoint rather than maintaining a parallel one.

## Test plan

- [ ] Search for trustees by name in the Trustee List page — results should appear correctly
- [ ] Search for trustees in the Data Verification page — results should appear correctly
- [ ] Verify no regressions in trustee search behavior (partial names, phonetic matches)
- [ ] Bicep index change requires a redeploy to take effect in non-local environments

Jira ticket: CAMS-763

## Summary by Sourcery

Optimize trustee name search by requiring phoneticTokens in the query filter and aligning the Mongo index with the query pattern.

Bug Fixes:
- Prevent unnecessary full-collection scans in trustee name searches by removing the phoneticTokens notExists fallback from the query filter.

Enhancements:
- Tighten trustee search pre-filter to only match documents with phoneticTokens, assuming all trustees are now indexed.
- Consolidate and narrow MongoDB indexes on the trustees collection to a compound {documentType, phoneticTokens} index matching the search workload.

Tests:
- Update trustee search repository tests to assert the absence of the notExists fallback and the presence of phoneticTokens-based filtering.

### DIFF
--- a/backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
+++ b/backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
@@ -1404,7 +1404,7 @@ describe('TrusteesMongoRepository', () => {
       expect(pipelineArg.stages.some((s) => s.stage === 'SCORE')).toBe(true);
     });
 
-    test('should include notExists fallback in pre-filter so trustees without phoneticTokens are searchable', async () => {
+    test('should filter pre-filter by phoneticTokens contains (no notExists fallback)', async () => {
       const paginateSpy = vi
         .spyOn(MongoCollectionAdapter.prototype, 'paginate')
         .mockResolvedValue({ metadata: { total: 0 }, data: [] });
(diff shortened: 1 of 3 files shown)


cruzjone-flexion:
@sourcery-ai  review

cruzjone-flexion:
## `ops/cloud-deployment/lib/cosmos/mongo/cosmos-collections.bicep`

Two index changes on the `trustees` collection:

**Before:**
- A `$**` wildcard index (indexes every field)
- A separate single-field index on `phoneticTokens`

**After:**
- A single compound index on `{ documentType, phoneticTokens }`

**Why:** The wildcard index was broad and expensive to maintain. The trustee search query always filters by both `documentType: 'TRUSTEE'` and `phoneticTokens` together, so a compound index on those two fields is the right shape — MongoDB can satisfy the entire pre-filter from a single index scan. The separate `phoneticTokens` single-field index is redundant once the compound index exists and was removed.

cruzjone-flexion:
## `backend/lib/adapters/gateways/mongo/trustees.mongo.repository.ts`

**Before:** When building the `searchTrusteesByNameScored` pre-filter, the `phoneticTokens` condition was wrapped in an `OR`:

```ts
or(
  doc('phoneticTokens').contains(allTokens),
  doc('phoneticTokens').notExists(),
)
```

This included trustees that either matched the tokens **or had no `phoneticTokens` field at all**. The fallback branch was added as a safety net for trustees that hadn't been indexed yet.

**After:** The `OR` and the `notExists()` fallback are removed. The condition is now a straight `CONTAINS`:

```ts
doc('phoneticTokens').contains(allTokens)
```

**Why:** All trustees in the collection now have `phoneticTokens` — the backfill is complete. The `notExists()` branch was causing MongoDB to perform a full collection scan on every search query, since it had to check every document regardless of index. Removing it means the compound `{ documentType, phoneticTokens }` index introduced in the Bicep change can be used for the entire pre-filter, eliminating the collection scan.

The unused `or` import was also removed from the `QueryBuilder` destructure.

mattstankey (approved):
We just need to ensure that no trustee records are missing `phoneticTokens`.

cruzjone-flexion:
> We just need to ensure that no trustee records are missing `phoneticTokens`.

We are following the same pattern as Case Debtor Name search, all existing debtors should be backfilled and any new Debtor will be populated on save.

PR #2499, opened 2026-06-11
CAMS-754 Add filters to trustee case list

jalba-flexion:
# Purpose

Trustees need to find relevant cases quickly from what can be a long list. This adds status, chapter, and filed date range filters to the trustee case list panel so trustees can narrow results without leaving the page.

# Major Changes

* Added a collapsible filter accordion above the case list with case status (defaults to Open), chapter multi-select, and filed date from/to range inputs
* Added filter persistence — state survives same-window navigation (case detail → back) via `useSessionState` keyed per trustee
* Added USWDS info Alert for the filtered empty state ("No cases found / Modify your filters and try again")
* Added "Case List" h3 heading and `aria-live` region on the case count for screen reader announcements
* Wired filters end-to-end: frontend filter UI → API adapter → controller → use case → MongoDB `searchCases`
* Removed Trustee Appointed Date filter fields (scoped out) and cleaned up all associated dead code in the frontend, backend controller, use case, repository, and shared types

# Testing/Validation

Implemented using TDD throughout. All filter paths (status, chapter, date range, combinations) are covered with unit tests at the use-case, controller, and component layers. Spec review findings addressed before PR: added `filedDateTo` coverage across 3 files, added `isFiltered` branch tests for chapters and date ranges, removed a cross-boundary test that belonged in the filter's own suite.

# Notes

The `isFiltered` check uses `=== 'CLOSED'` (not `!== 'ALL'`) because the default is now `'OPEN'` — checking `!== 'ALL'` would incorrectly flag the default state as filtered.

The appointed date field was removed after scoping discussion: `appointedDate` lives on the appointment record and cannot be efficiently filtered via MongoDB `searchCases`. The in-memory approach that was prototyped was not performant enough to ship.

# Definition of Done:

## Summary by Sourcery

Add filterable trustee case list with status, chapter, and filed date range support, and wire these filters through the frontend, API, controller, use case, and Mongo search.

New Features:
- Introduce trustee case list filters for case status, bankruptcy chapter, and filed date range, with a collapsible UI and pill summary.
- Persist trustee case list filter state per trustee using session storage so filters survive navigation.
- Display total case count with accessible aria-live updates above the trustee case table.

Enhancements:
- Refine trustee case list empty states by distinguishing between unfiltered and filtered results using contextual info alerts.
- Adjust trustee case list pagination to work with combined pagination and filter predicates, resetting to page 1 when filters change.
- Update case search repository to support closed-case-only queries and filed date range predicates used by trustee filters.
- Change trustee case list case links to open in a new window and add an accessible "Case List" heading.

Tests:
- Expand backend tests for trustee case listing controller, use case, and Mongo repository to cover new filter predicates and edge cases.
- Add and update frontend tests for trustee case list and filter components, including filtered empty states, API parameter mapping, pagination behavior, and accessibility attributes.
- Extend DB scenario seeds and tests to cover closed case data used for status filtering.


user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -0,0 +1,58 @@
...
+const trusteeCaseListFilterUseCase = (
+  store: TrusteeCaseListFilterStore,
+  _controls: TrusteeCaseListFilterControls,
+  onFilterChange: (filter: TrusteeCaseListFilterValue) => void,
+): TrusteeCaseListFilterUseCase => {
+  const chaptersToComboOptions = (): ComboOption[] => CASE_CHAPTER_OPTIONS;
+
+  const handleStatusChange = (status: 'OPEN' | 'CLOSED' | 'ALL') => {
+    store.setSelectedStatus(status);
+    onFilterChange({ caseStatus: status, chapters: store.selectedChapters.map((c) => c.value) });
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Fixed. Added a `buildFilterFromStore` helper that reads all fields from the store (including `filedDateFrom`/`filedDateTo`) and applies per-handler overrides. `handleStatusChange` now calls `buildFilterFromStore(store, { caseStatus: status })` so active date filters are preserved.

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  const handleChapterChange = (chapters: ComboOption[]) => {
+    store.setSelectedChapters(chapters);
+    onFilterChange({ caseStatus: store.selectedStatus, chapters: chapters.map((c) => c.value) });
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Fixed. `handleChapterChange` now calls `buildFilterFromStore(store, { chapters: chapters.map((c) => c.value) })`, which carries `filedDateFrom`/`filedDateTo` from the store into the emitted predicate.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+    handleFiledDateChange: useCase.handleFiledDateChange,
+  };
+
+  return <TrusteeCaseListFilterView viewModel={viewModel} />;
+}
+
+function useTrusteeCaseListFilterStoreReact(
+  initialValue?: TrusteeCaseListFilterValue,
+): TrusteeCaseListFilterStore {
+  const CASE_CHAPTER_OPTIONS: ComboOption[] = [
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Fixed. `CASE_CHAPTER_OPTIONS` is now a single named export from `trusteeCaseListFilterUseCase.ts` and imported from there in `TrusteeCaseListFilter.tsx` — no more duplication.

user-interface/src/trustees/panels/TrusteeCaseList.scss
```diff
@@ -0,0 +1,17 @@
+.trustee-case-list-heading {
+  line-height: 1;
+  margin-bottom: 1rem;
+}
+.trustee-case-list-count {
+  font-weight: bold;
+}
+
+.usa-alert-container.inline-alert.case-list-alert {
+  position: fixed;
```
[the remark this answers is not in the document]
jalba-flexion: Not addressing the fixed-position alert in this PR — leaving `position: fixed` / `width: 40%` as-is for now. Can be followed up separately if needed.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+import trusteeCaseListFilterUseCase from './trusteeCaseListFilterUseCase';
+import {
+  TrusteeCaseListFilterControls,
+  TrusteeCaseListFilterProps,
+  TrusteeCaseListFilterStore,
+  TrusteeCaseListFilterValue,
+  TrusteeCaseListFilterViewModel,
+} from './trusteeCaseListFilter.types';
+
+export default function TrusteeCaseListFilter(props: TrusteeCaseListFilterProps) {
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Done. Removed `useTrusteeCaseListFilterStoreReact` and `useTrusteeCaseListFilterControlsReact` — state and refs are now inlined directly in `TrusteeCaseListFilter`. The `trusteeCaseListFilterUseCase` factory and view layer are unchanged.

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -0,0 +1,58 @@
+import { ComboOption } from '@/lib/components/combobox/ComboBox';
+import {
+  TrusteeCaseListFilterControls,
+  TrusteeCaseListFilterStore,
+  TrusteeCaseListFilterUseCase,
+  TrusteeCaseListFilterValue,
+} from './trusteeCaseListFilter.types';
+
+const CASE_CHAPTER_OPTIONS: ComboOption[] = [
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Fixed. Extracted `buildFilterFromStore` to centralize filter-object construction. Each handler now updates its own state and calls `buildFilterFromStore` with a single override — no more repeated field mapping.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
```diff
@@ -0,0 +1,56 @@
+import { ComboOption } from '@/lib/components/combobox/ComboBox';
+import { ComboBoxRef } from '@/lib/type-declarations/input-fields';
+
+export type TrusteeCaseListFilterValue = {
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Addressed. Extracted `TrusteeCaseStatus` union type (removes repeated string literals) and `TrusteeCaseListFilterHandlers` interface (shared by `TrusteeCaseListFilterViewModel` and `TrusteeCaseListFilterUseCase`). Collapsed the empty `TrusteeCaseListFilterUseCase extends Handlers {}` into a type alias.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
```diff
@@ -0,0 +1,165 @@
...
+                  placeholder="- Select one or more Chapters -"
+                  ref={chapterFilterRef}
+                />
+              </div>
+            </div>
+          </div>
+        </Accordion>
+      </AccordionGroup>
+
+      {allPills.length > 0 && (
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Both simplified. Pill-change handler now builds a `Set` from updated pill values and filters `selectedChapters` by presence in that set — no more coupling to `selectedStatus` string constants. `formatDatePillLabel` now uses `toLocaleDateString` instead of manual string splitting.

backend/lib/controllers/trustee-cases/trustee-cases.controller.ts
```diff
@@ -48,7 +52,28 @@ export class TrusteeCasesController implements CamsController {
...
         Number.isFinite(parsedOffset) && parsedOffset >= 0 ? parsedOffset : DEFAULT_SEARCH_OFFSET;
 
-      const result = await this.useCase.getCasesForTrustee(context, trusteeId, limit, offset);
+      const validStatuses = ['OPEN', 'CLOSED', 'ALL'] as const;
+      const rawStatus = context.request.query.status as string;
+      const caseStatus = (
+        validStatuses.includes(rawStatus as (typeof validStatuses)[number]) ? rawStatus : 'ALL'
+      ) as 'OPEN' | 'CLOSED' | 'ALL';
+
+      const chaptersParam = context.request.query.chapters as string;
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Fixed. The test incorrectly expected `limit` to reset when only `offset` was invalid. Each param defaults independently — this is the pattern used in all other controllers (`bank-trustees`, `software-bank-trustees`, etc.). Updated the test to expect `limit: 25` when a valid `limit: "25"` is passed alongside an invalid offset.

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const CASE_CHAPTER_OPTIONS: ComboOption[] = [
+  { value: '7', label: 'Chapter 7', selectedLabel: 'Chapter 7' },
+  { value: '11', label: 'Chapter 11', selectedLabel: 'Chapter 11' },
+  { value: '12', label: 'Chapter 12', selectedLabel: 'Chapter 12' },
+  { value: '13', label: 'Chapter 13', selectedLabel: 'Chapter 13' },
+  { value: '15', label: 'Chapter 15', selectedLabel: 'Chapter 15' },
+];
+
+const buildFilterFromStore = (
```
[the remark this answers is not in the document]
jalba-flexion: Not addressing this. Removing the `overrides` param would reintroduce the original date-filter bug — React `setState` is async, so `store.selectedStatus` still holds the old value at the moment `buildFilterFromStore` is called. The overrides param exists precisely to pass the new value synchronously before the re-render.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.tsx
```diff
@@ -0,0 +1,60 @@
...
+    initialValue?.chapters?.length
+      ? CASE_CHAPTER_OPTIONS.filter((o) => initialValue.chapters.includes(o.value))
+      : [],
+  );
+  const [filedDateFrom, setFiledDateFrom] = useState(initialValue?.filedDateFrom ?? '');
+  const [filedDateTo, setFiledDateTo] = useState(initialValue?.filedDateTo ?? '');
+  const [filedDateError, setFiledDateError] = useState('');
+  const chapterFilterRef = useRef<ComboBoxRef>(null);
+
+  const useCase = trusteeCaseListFilterUseCase(
```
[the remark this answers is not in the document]
jalba-flexion: Not addressing this. The use-case factory is deliberate architecture for testability — keeping domain logic separate from the React component. Inlining it would merge two concerns we want to keep distinct.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
```diff
@@ -0,0 +1,55 @@
...
+export type TrusteeCaseStatus = 'OPEN' | 'CLOSED' | 'ALL';
+
+export type TrusteeCaseListFilterValue = {
+  caseStatus: TrusteeCaseStatus;
+  chapters: string[];
+  filedDateFrom?: string;
+  filedDateTo?: string;
+};
+
+export interface TrusteeCaseListFilterStore {
```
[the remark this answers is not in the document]
jalba-flexion: Not addressing further. We already extracted `TrusteeCaseStatus`, consolidated `UseCase`/`ViewModel` via the shared `TrusteeCaseListFilterHandlers` interface, and collapsed the empty interface into a type alias. Further collapsing `Store` and `Controls` would reduce testability of the use-case layer.

user-interface/src/trustees/panels/TrusteeCaseList.tsx
```diff
@@ -7,34 +8,51 @@ import { CaseNumber } from '@/lib/components/CaseNumber';
...
-    );
-  }, [trusteeId]);
+    setPaginationPredicate({ limit: DEFAULT_SEARCH_LIMIT, offset: DEFAULT_SEARCH_OFFSET });
+  }, [filterPredicate]);
 
   useEffect(() => {
     setIsLoading(true);
     setError(null);
-    Api2.getTrusteeCases(trusteeId, searchPredicate)
+    const predicate: TrusteeCasesSearchPredicate = {
```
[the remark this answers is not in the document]
jalba-flexion: ✅ Done. Extracted `buildSearchPredicate(paginationPredicate, filterPredicate)` so the API predicate is constructed in one place. Also introduced a `PaginationPredicate` type to narrow `handlePaginationChange` and `<Pagination>` — they only need `limit`/`offset`, not the full search predicate.

user-interface/src/trustees/panels/TrusteeCaseList.tsx
```diff
@@ -45,27 +69,53 @@ export default function TrusteeCaseList({ trusteeId }: Readonly<TrusteeCaseListP
...
+  }, [trusteeId, paginationPredicate, filterPredicate]);
 
-  function handlePagination(predicate: SearchPredicate) {
-    setSearchPredicate(predicate);
+  function handlePaginationChange(predicate: PaginationPredicate) {
+    setPaginationPredicate({
+      limit: predicate.limit ?? DEFAULT_SEARCH_LIMIT,
+      offset: predicate.offset ?? DEFAULT_SEARCH_OFFSET,
+    });
   }
```
cruzjone-flexion, line 77: **`handleFilterChange` is a pass-through wrapper with no logic**

This function only forwards its argument to `onFilterChange` — adding an indirection with no value. `onFilterChange` can be passed directly to `TrusteeCaseListFilter`.

```typescript
function handleFilterChange(filter: TrusteeCaseListFilterValue) {
  onFilterChange(filter);
}

// ...
<TrusteeCaseListFilter onFilterChange={handleFilterChange} initialValue={filterPredicate} />
```

Simplify to:
```typescript
<TrusteeCaseListFilter onFilterChange={onFilterChange} initialValue={filterPredicate} />
```

jalba-flexion: ✅ Fixed. Removed `handleFilterChange` and pass `onFilterChange` directly.

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -0,0 +1,70 @@
+import { ComboOption } from '@/lib/components/combobox/ComboBox';
+import {
+  TrusteeCaseListFilterControls,
+  TrusteeCaseListFilterStore,
+  TrusteeCaseListFilterUseCase,
```
cruzjone-flexion, line 4: **`_controls` parameter is unused — remove it or explain why it is kept**

The `_controls` parameter is accepted but never referenced inside the function. The underscore prefix signals "intentionally unused," but if it was scoped out it should be removed from the signature entirely to avoid future callers passing something that gets silently ignored.

```typescript
const trusteeCaseListFilterUseCase = (
  store: TrusteeCaseListFilterStore,
  _controls: TrusteeCaseListFilterControls,  // never used
  onFilterChange: (filter: TrusteeCaseListFilterValue) => void,
): TrusteeCaseListFilterUseCase => {
```

Remove `_controls` from the signature and all call sites, or add a comment explaining it is reserved for upcoming use.

user-interface/src/trustees/panels/TrusteeCaseList.tsx
```diff
@@ -45,27 +69,49 @@ export default function TrusteeCaseList({ trusteeId }: Readonly<TrusteeCaseListP
       .finally(() => {
         setIsLoading(false);
       });
-  }, [trusteeId, searchPredicate]);
+  }, [trusteeId, paginationPredicate, filterPredicate]);
 
-  function handlePagination(predicate: SearchPredicate) {
-    setSearchPredicate(predicate);
```
cruzjone-flexion, line 51: **Unnecessary `??` guards on required `number` fields**

`PaginationPredicate` declares both `limit` and `offset` as required `number` — they can never be `undefined`. The null-coalescing fallbacks can never fire and imply optionality that the types do not allow.

```typescript
// handlePaginationChange — limit and offset are required by PaginationPredicate
setPaginationPredicate({
  limit: predicate.limit ?? DEFAULT_SEARCH_LIMIT,   // unreachable fallback
  offset: predicate.offset ?? DEFAULT_SEARCH_OFFSET, // unreachable fallback
});
```

The same applies in `trustee-cases.use-case.ts`:
```typescript
data: allItems.slice(offset ?? 0, (offset ?? 0) + (limit ?? 25)),
```
`limit` and `offset` are destructured from `TrusteeCasesSearchPredicate` (which extends `SearchPredicate` requiring both). Remove the guards to keep the types and the code in sync.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.scss
```diff
@@ -0,0 +1,86 @@
...
+  .filter-control-label {
+    font-size: 1rem;
+  }
+
+  .filter-clear-link {
+    background: none;
+    border: none;
+    padding: 0;
+    font-size: 1rem;
+    color: #005ea2;
```
cruzjone-flexion: **Hardcoded hex colors in new SCSS file — inconsistent with PR #2500**

PR #2500 (currently open) is replacing all hardcoded hex values with `colors.$primary` and `colors.$primary-dark` tokens across the codebase. This new file introduces two hardcoded values that immediately conflict with that effort.

```scss
.filter-clear-link {
  color: #005ea2;    // should be colors.$primary

  &:hover {
    color: #1a4480;  // should be colors.$primary-dark
  }
}
```

Use the shared tokens instead:
```scss
@use "styles/abstracts/colors" as colors;

.filter-clear-link {
  color: colors.$primary;

  &:hover {
    color: colors.$primary-dark;
  }
}
```

user-interface/src/trustees/panels/TrusteeCaseList.tsx
```diff
@@ -45,27 +69,49 @@ export default function TrusteeCaseList({ trusteeId }: Readonly<TrusteeCaseListP
...
+  function handlePaginationChange(predicate: PaginationPredicate) {
+    setPaginationPredicate({
+      limit: predicate.limit ?? DEFAULT_SEARCH_LIMIT,
+      offset: predicate.offset ?? DEFAULT_SEARCH_OFFSET,
+    });
   }
 
+  const totalCount = pagination?.totalCount ?? 0;
+  const isFiltered =
```
cruzjone-flexion: **`isFiltered` check is inline — will silently miss new filter fields**

This predicate is computed inline in `TrusteeCaseList.tsx` with no shared source of truth with `TrusteeCaseListFilterValue`. If a new field is added to the type, `isFiltered` must be manually updated here to match — there is nothing to remind a developer to do that.

```typescript
const isFiltered =
  filterPredicate.caseStatus === 'CLOSED' ||
  filterPredicate.chapters.length > 0 ||
  !!filterPredicate.filedDateFrom ||
  !!filterPredicate.filedDateTo;
```

Extract a named helper co-located with the type in `trusteeCaseListFilter.types.ts`:
```typescript
export function isTrusteeCaseListFiltered(filter: TrusteeCaseListFilterValue): boolean {
  return (
    filter.caseStatus === 'CLOSED' ||
    filter.chapters.length > 0 ||
    !!filter.filedDateFrom ||
    !!filter.filedDateTo
  );
}
```
This keeps the "what counts as filtered" logic next to the type it describes.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
```diff
@@ -0,0 +1,55 @@
+import { ComboOption } from '@/lib/components/combobox/ComboBox';
+import { ComboBoxRef } from '@/lib/type-declarations/input-fields';
+
```
cruzjone-flexion: **`TrusteeCaseStatus` union is defined independently in two places**

`common/src/api/search.ts` uses the inline union `'OPEN' | 'CLOSED' | 'ALL'` in `TrusteeCasesSearchPredicate`, while `trusteeCaseListFilter.types.ts` defines a separate `TrusteeCaseStatus` type with the same values. These two have no shared source of truth — if a new status is added to one, the other silently diverges.

Since `TrusteeCasesSearchPredicate` already lives in `common`, `TrusteeCaseStatus` belongs there too:

```typescript
// common/src/api/search.ts
export type TrusteeCaseStatus = 'OPEN' | 'CLOSED' | 'ALL';

export type TrusteeCasesSearchPredicate = SearchPredicate & {
  caseStatus?: TrusteeCaseStatus;
  // ...
};
```

The frontend type file can then import from `@common/api/search` instead of redeclaring it.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
```diff
@@ -0,0 +1,162 @@
...
+    selectedChapters,
+    chapterFilterRef,
+    filedDateFrom,
+    filedDateTo,
+    filedDateError,
+  } = viewModel;
+
+  const hasFiledDate = !!(filedDateFrom || filedDateTo);
+
+  const statusPill: ComboOption | null =
```
cruzjone-flexion, line 28: **`'filed-date'` is a magic string used in two separate places**

The string `'filed-date'` is used as the pill identity value when building the pill and again when checking for removal — with no shared constant tying them together. If one is changed and the other is not, the pill removal logic silently breaks.

```typescript
const filedDatePill = hasFiledDate
  ? { value: 'filed-date', label: ... }  // defined here
  : null;

// ...later:
const filedDateRemoved = filedDatePill && !updatedValues.has('filed-date');  // repeated here
```

Extract to a named constant at the top of the file:
```typescript
const FILED_DATE_PILL_VALUE = 'filed-date';
```

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
jalba-flexion: ✅ Fixed. Removed `_controls` from the `trusteeCaseListFilterUseCase` signature and its call site in `TrusteeCaseListFilter.tsx`. Also removed the `TrusteeCaseListFilterControls` interface entirely since it was only used to type that parameter.

user-interface/src/trustees/panels/TrusteeCaseList.tsx
jalba-flexion: ✅ Fixed. Removed both `??` guards from `handlePaginationChange` — `PaginationPredicate` guarantees both fields are required `number`. Also removed the `offset ?? 0` / `limit ?? 25` guards from `trustee-cases.use-case.ts` for the same reason.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.scss
jalba-flexion: ✅ Fixed. Added `$primary-dark: #1a4480` to `_colors.scss`, added `@use '@/styles/abstracts/colors'` to the filter SCSS, and replaced both hardcoded hex values with `colors.$primary` and `colors.$primary-dark`.

user-interface/src/trustees/panels/TrusteeCaseList.tsx
jalba-flexion: ✅ Fixed. Extracted `isFilterActive(filter: TrusteeCaseListFilterValue): boolean` into `trusteeCaseListFilter.types.ts` alongside the type it validates. `TrusteeCaseList.tsx` now calls `isFilterActive(filterPredicate)` — new filter fields only need to be added in one place.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
jalba-flexion: ✅ Fixed. Moved `TrusteeCaseStatus` to `common/src/api/search.ts` and updated `TrusteeCasesSearchPredicate.caseStatus` to use it. `trusteeCaseListFilter.types.ts` now re-exports it with `export type { TrusteeCaseStatus } from '@common/api/search'` — single source of truth.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
jalba-flexion: ✅ Fixed. Extracted `const FILED_DATE_PILL_VALUE = 'filed-date'` at the top of `TrusteeCaseListFilterView.tsx`. Both the pill definition and the removal check now reference the constant.

cruzjone-flexion (approved):

PR #2500, opened 2026-06-11
CAMS-767 mapping status to inactive

fmaddenflx:
## Summary
- Add status filter (Active/Inactive) to trustee list with mobile-first responsive layout
- Improve filter controls with paired wrapping (4→2→1 columns)
- Replace hardcoded hex colors with SCSS variables across the UI
- Tune table column proportions to prevent horizontal overflow at all viewport widths

## Test plan
- [ ] Verify status filter correctly filters trustees by Active/Inactive
- [ ] Verify filter layout wraps in pairs at medium widths and stacks on mobile
- [ ] Verify table displays without horizontal scroll at all widths (880px+)
- [ ] Verify status column displays full text without wrapping
- [ ] Verify no visual regressions in other filter controls

## Summary by Sourcery

Add status-based filtering and responsive layout improvements to the trustees list while standardizing UI colors and table behaviors across the app.

New Features:
- Introduce an Active/Inactive/All status filter on the trustees list and expose it via a Status combobox in the filter panel.

Bug Fixes:
- Ensure trustees with mixed active and inactive appointments appear in both views and that status text in the list clearly distinguishes inactive sub-statuses.
- Prevent horizontal overflow in trustees tables by rebalancing column flex proportions and responsive breakpoints.

Enhancements:
- Refine trustee filter layout into paired, responsive controls that adapt between stacked, two-column, and multi-row arrangements.
- Standardize borders, text, and interaction colors across multiple components using shared SCSS color variables.
- Improve reusable CamsTable mixins and screen-specific table layouts for consistent stacked/flex behaviors at tablet and desktop breakpoints.
- Tidy various SCSS files for consistency, spacing, and selector formatting without changing behavior.

CI:
- Align ESLint testing configuration formatting and rules for Testing Library in UI test files.

Documentation:
- Reformat accessibility KNOWN_ISSUES documentation for better readability and updated status notes on existing issues.

Tests:
- Add unit tests covering trustee status filtering combinations, status label formatting, and combined chapter/status filter behavior.
- Extend district filter tests to cover the new Status combobox wiring and default selection.
- Adjust existing trustees list tests to account for status cell rendering changes and new inactive status labeling.

## Summary by Sourcery

Add status-based filtering to the trustees list, refine its responsive layout and status display, and standardize shared UI styling and linting across the app.

New Features:
- Introduce a Status combobox filter on the trustees list allowing users to view Active, Inactive, or All trustees.

Bug Fixes:
- Ensure trustees with mixed active and inactive appointments appear appropriately in both Active and Inactive filtered views.
- Prevent truncated or ambiguous trustee status text by clearly labeling inactive sub-statuses in the list.
- Reduce horizontal overflow risk in trustees tables by rebalancing column layouts and breakpoints.

Enhancements:
- Refine trustee filter controls into paired, mobile-first responsive layouts that adapt across viewport sizes.
- Standardize borders, text, and interaction colors by replacing hard-coded values with shared SCSS color variables and table mixins.
- Improve CamsTable mixins and screen-specific table styles for consistent stacked versus flex-row behavior at tablet and desktop widths.
- Tidy assorted SCSS selectors, spacing, and attribute quoting for consistency without altering behavior.

CI:
- Align ESLint Testing Library rule configuration for UI test files to enforce consistent testing patterns.

Documentation:
- Reformat the accessibility KNOWN_ISSUES document for improved readability while preserving existing guidance and statuses.

Tests:
- Expand trustees list tests to cover new status filtering behavior, mixed-status appointments, and updated status cell formatting.
- Add filter view tests for the new Status combobox wiring, default selection, and interaction.
- Adjust existing tests and lint configuration to align with Testing Library best practices in UI test files.


user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -51,21 +57,49 @@ function isUserInSelectedDivision(trustee: TrusteeListItem, divisionFilter: Divi
...
   const selectedChapterValues = new Set(selectedChapters.map((c) => c.value));
   return trustees.filter((trustee) => {
+    if (statusFilter === 'active') {
+      if (
+        trustee.appointments.length > 0 &&
+        !trustee.appointments.some((appt) => appt.status === 'active')
+      )
+        return false;
+    } else if (statusFilter === 'inactive') {
+      if (!trustee.appointments.some((appt) => isInactiveStatus(appt.status))) return false;
```
mattstankey: **Bug: Empty-appointment trustees behave asymmetrically across filter states**

The `active` and `inactive` branches treat trustees with zero appointments differently:

```ts
if (statusFilter === 'active') {
  if (
    trustee.appointments.length > 0 &&  // ← zero-appointment trustees pass through
    !trustee.appointments.some(...)
  )
    return false;
} else if (statusFilter === 'inactive') {
  if (!trustee.appointments.some((appt) => isInactiveStatus(appt.status))) return false;
  // .some() on an empty array returns false → !false = true → filtered OUT
}
```

- **Active filter**: a trustee with no appointments is *included*.
- **Inactive filter**: a trustee with no appointments is *excluded*.

If empty-appointment trustees are invalid, both branches should exclude them. If they are valid, both should treat them consistently. Consider removing the `appointments.length > 0` guard from the active branch (or mirroring it in the inactive branch), and add a test covering this edge case.

mattstankey: **Design: `filterTrustees` is called 6+ times in this component**

`statusFilter` is now threaded through every call correctly, but having this many separate calls to `filterTrustees` is fragile — a future parameter addition will require finding all call sites again.

The existing `useMemo` infrastructure in this file is the right tool here. Computing `filteredTrustees` once and reusing it for display, pagination, and counts would simplify the component and align with how other filters are handled:

```ts
const filteredTrustees = useMemo(
  () => filterTrustees(trustees, selectedDistricts, selectedChapters, statusFilter, districtDivisionEnabled, divisionFilterMap),
  [trustees, selectedDistricts, selectedChapters, statusFilter, districtDivisionEnabled, divisionFilterMap],
);
```

user-interface/src/trustees/TrusteesList.test.tsx
```diff
@@ -2958,4 +2970,204 @@ describe('TrusteesList Component', () => {
...
+      });
+
+      renderWithRouter(<TrusteesList />);
+
+      await waitFor(() => {
+        expect(screen.getByText('1 Trustee', { selector: 'p' })).toBeInTheDocument();
+      });
+
+      const toggleButton = screen.getByRole('button', { name: /filters/i });
+      await userEvent.setup().click(toggleButton);
```
mattstankey: **Test pattern: `userEvent.setup()` should be called once per test and reused**

Each call to `userEvent.setup()` creates a new user-event instance with its own event queue. Calling it separately for each interaction can cause issues with pointer state, delays, and event ordering.

```ts
// Current (problematic)
await userEvent.setup().click(toggleButton);
// ...
await userEvent.setup().click(statusCombobox);

// Correct
const user = userEvent.setup();
await user.click(toggleButton);
// ...
await user.click(statusCombobox);
```

The `should apply status and chapter filters together` test already does this correctly — the other three new tests in this suite should follow the same pattern.

mattstankey (commented):
Code review feedback on four items found during review.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
```diff
@@ -2,7 +2,22 @@ import './TrusteeDistrictFilter.scss';
...
-import { TrusteeDistrictFilterViewProps } from './trusteeDistrictFilter.types';
+import { StatusFilterValue, TrusteeDistrictFilterViewProps } from './trusteeDistrictFilter.types';
+
+const STATUS_OPTIONS: ComboOption[] = [
+  { value: 'all', label: 'All' },
+  { value: 'active', label: 'Active' },
+  {
+    value: 'inactive',
+    label:
+      'Inactive (Inactive, Involuntarily Suspended, Voluntarily Suspended, Resigned, Terminated, Deceased)',
```
mattstankey: **UX: The Inactive dropdown label is very long and will likely truncate**

```ts
label: 'Inactive (Inactive, Involuntarily Suspended, Voluntarily Suspended, Resigned, Terminated, Deceased)',
```

This string is ~82 characters and will overflow or truncate in most combobox dropdown widths. Consider shortening the label to just `"Inactive"` (consistent with the brevity of `"All"` and `"Active"`) and surfacing the sub-status breakdown via tooltip, help text, or the status column itself in the list.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
```diff
@@ -2,7 +2,22 @@ import './TrusteeDistrictFilter.scss';
 import ComboBox, { ComboOption } from '@/lib/components/combobox/ComboBox';
 import PillBox from '@/lib/components/PillBox';
 import { Accordion, AccordionGroup } from '@/lib/components/uswds/Accordion';
-import { TrusteeDistrictFilterViewProps } from './trusteeDistrictFilter.types';
+import { StatusFilterValue, TrusteeDistrictFilterViewProps } from './trusteeDistrictFilter.types';
+
+const STATUS_OPTIONS: ComboOption[] = [
+  { value: 'all', label: 'All' },
+  { value: 'active', label: 'Active' },
```
cruzjone-flexion, line 9: **Missing `'removed'` in the Inactive option label**

`INACTIVE_STATUSES` includes `'removed'` as one of the seven inactive statuses, but the label string on the Status combobox option lists only six — leaving `'Removed'` out:

```typescript
label: 'Inactive (Inactive, Involuntarily Suspended, Voluntarily Suspended, Resigned, Terminated, Deceased)',
// Missing: Removed
```

A trustee with a `removed` status would be correctly filtered into the Inactive view, but the label describing what "Inactive" covers wouldn't mention it, which could confuse users.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -27,6 +27,12 @@ import { CourtDivisionDetails } from '@common/cams/courts';
 const BASE_COLUMN_HEADERS = ['Name', 'District', 'Chapter', 'Type', 'Status'];
 const DIVISION_COLUMN_HEADERS = ['Name', 'District', 'Division', 'Chapter', 'Type', 'Status'];
 
+function formatListAppointmentStatus(status: AppointmentStatus): string {
```
cruzjone-flexion, line 30: **`isInactiveStatus` accepts `undefined` but the call site never passes it**

The function signature accepts `AppointmentStatus | undefined` and guards against it, but at the call site `appt.status` comes from `TrusteeAppointment.status` which is typed as `AppointmentStatus` (not optional). The `| undefined` overload and the `!== undefined` guard appear to be dead code.

```typescript
function isInactiveStatus(status: AppointmentStatus | undefined): boolean {
  return status !== undefined && INACTIVE_STATUSES.has(status);
}
```

If `appt.status` is never actually optional, tightening the signature to `AppointmentStatus` removes the unnecessary guard and makes the type contract clearer.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -51,21 +57,45 @@ function isUserInSelectedDivision(trustee: TrusteeListItem, divisionFilter: Divi
...
   trustees: TrusteeListItem[],
   selectedDistricts: ComboOption[],
   selectedChapters: ComboOption[],
+  statusFilter: StatusFilterValue,
   districtDivisionEnabled: boolean = false,
   divisionFilterMap: DivisionFilterMap = new Map(),
 ): TrusteeListItem[] {
   if (
-    selectedDistricts.length === 0 &&
+    statusFilter === 'all' &&
```
jamesobrooks: **CAMS-767 AC Gap: Filter is client-side only**

The CAMS-767 acceptance criteria explicitly requires: *"The filter applies server-side — it does not just hide rows client-side."*

`filterTrustees` runs after the full trustee list is loaded into the browser. The `listTrustees` backend endpoint has no appointment-status parameter — it returns all trustees unconditionally. For a small dataset this is unnoticeable, but the AC is unambiguous.

To satisfy this criterion, the `listTrustees` use case and the `TrusteesMongoRepository.listTrustees` query will need to accept an appointment-status filter and apply it at the database layer (e.g., matching trustees whose `appointments` array contains at least one entry with `status` in the requested set).

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
```diff
@@ -2,7 +2,22 @@ import './TrusteeDistrictFilter.scss';
 import ComboBox, { ComboOption } from '@/lib/components/combobox/ComboBox';
 import PillBox from '@/lib/components/PillBox';
 import { Accordion, AccordionGroup } from '@/lib/components/uswds/Accordion';
-import { TrusteeDistrictFilterViewProps } from './trusteeDistrictFilter.types';
+import { StatusFilterValue, TrusteeDistrictFilterViewProps } from './trusteeDistrictFilter.types';
+
+const STATUS_OPTIONS: ComboOption[] = [
```
jamesobrooks, line 7: **CAMS-767 AC Gap: Status filter supports 3 grouped options, not 8 individual statuses**

The acceptance criteria lists 8 selectable values: `active`, `inactive`, `voluntarily-suspended`, `involuntarily-suspended`, `deceased`, `resigned`, `terminated`, `removed`.

`STATUS_OPTIONS` offers three choices — `All`, `Active`, and one combined `Inactive` bucket. This prevents staff from filtering to, for example, only `deceased` or only `resigned` trustees. The stated motivation ("focus on trustees I'm actively responsible for without being overwhelmed by historical records") implies users need that granularity — someone looking for a resigned trustee's cases should be able to narrow to just that status.

The multi-select AC criterion is also unmet: the Status combobox does not pass `multiSelect={true}` (compare to the Chapter and District comboboxes on lines 46 and 71).

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
```diff
@@ -98,61 +113,86 @@ function TrusteeDistrictFilterView(props: TrusteeDistrictFilterViewProps) {
...
 
-              <div className="filter-control">
-                <div className="filter-control-header">
-                  <span className="filter-control-label">Chapter</span>
+                <div className="filter-control">
+                  <div className="filter-control-header">
+                    <span className="filter-control-label">Status</span>
+                  </div>
+                  <ComboBox
+                    id="status-combobox"
```
jamesobrooks, line 183: **CAMS-767 AC Gap: Active filter selection not visible when accordion is collapsed**

The AC requires: *"The active filter selection is reflected in the UI (e.g., visible chips or checked state)."*

The `hasPills` flag (line 91) is driven by selected districts, chapters, and divisions only — `statusFilter` never contributes a pill. When the accordion is collapsed there is no persistent indicator that the list is filtered by status. A staff member who selects "Inactive" then collapses the panel sees a filtered list with no visual cue.

The default-`active` state makes this especially subtle: a first-time user sees a list that excludes inactive/deceased/resigned trustees without any indicator that filtering is occurring.

Consider adding a status pill to the pill box when `statusFilter !== 'all'`, or surfacing the active filter state in the accordion header label.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
```diff
@@ -98,61 +113,86 @@ function TrusteeDistrictFilterView(props: TrusteeDistrictFilterViewProps) {
...
+                    label="Status"
+                    hideInternalLabel={true}
+                    ariaLabelPrefix="Status"
+                    options={STATUS_OPTIONS}
+                    selections={statusToSelection(viewModel.statusFilter)}
+                    onUpdateSelection={(selections) => {
+                      const value = (selections[0]?.value ?? 'active') as StatusFilterValue;
```
jamesobrooks, line 190: **CAMS-767 AC Gap: No "clear to default" path for the status filter**

The AC requires: *"When no filter is selected, the default (`active`) is restored."*

The combobox uses a fallback (`selections[0]?.value ?? 'active'`) to avoid an empty state, but there is no user-facing affordance to explicitly clear the status filter back to its default. The pill box provides clear buttons for district/chapter/division chips, but there is no equivalent for status.

If a user selects "All", they cannot return to the default "Active" view without manually re-selecting it. The AC describes restoring the default when "no filter is selected" — implying a clearable filter state that snaps back — which the current design does not support.

jamesobrooks:
### Security: IDOR — `trusteeId` not bound to the authenticated user (backend/lib/controllers/trustee-cases/trustee-cases.controller.ts:46)

The role check in `hasRequiredRole` confirms the caller *is* a `TrusteeAdmin`, but does not verify that the `trusteeId` route parameter belongs to *the authenticated user*. Any `TrusteeAdmin` can substitute an arbitrary `trusteeId` in the URL and receive the full case list for any other trustee.

```typescript
// hasRequiredRole only checks CamsRole.TrusteeAdmin — no binding to session user identity
const trusteeId = context.request.params['trusteeId'];
// ...used directly without comparing to context.session.user
const result = await this.useCase.getCasesForTrustee(context, trusteeId, predicate);
```

**Options:**
- **Scope to self (likely intended):** compare `trusteeId` to the authenticated user's identity from the session and throw `UnauthorizedError` on mismatch.
- **Explicit cross-read policy:** if `TrusteeAdmin` is intentionally permitted to view any trustee's case list, add a comment in `hasRequiredRole` documenting that decision so future reviewers know it is by design.

OWASP A01:2021 Broken Access Control — IDOR / ASVS 4.2.1

---

### SRE: 500-case hard cap silently truncates results (backend/lib/use-cases/trustee-cases/trustee-cases.use-case.ts:29)

The use case fetches `limit: 500` cases then paginates in-memory. A trustee with more than 500 appointments matching the active filter will have excess cases silently dropped — the UI paginates over at most 500 items and reports that count as the total. There is no log warning when the cap is hit.

Consider adding a warn log when `casesResponse.data.length === 500` so truncation is observable in Application Insights:

```typescript
if (syncedCases.length === 500) {
  context.logger.warn(MODULE_NAME, `Trustee ${trusteeId} may have >500 cases; results may be truncated.`);
}
```

---

### Security: No date format validation on `filedDateFrom`/`filedDateTo` (backend/lib/controllers/trustee-cases/trustee-cases.controller.ts:64-65)

`filedDateFrom` and `filedDateTo` are passed directly to MongoDB string comparisons with no format validation. A non-ISO string (e.g., `06/12/2025`, `""`, or `foo`) produces silently wrong query results rather than an error.

`isValidDateString` from `@common/date-helper` already handles this exact concern and is used in `trustee-upcoming-key-dates.controller.ts`. Apply it here:

```typescript
import { isValidDateString } from '@common/date-helper';

const rawFrom = context.request.query.filedDateFrom as string | undefined;
const filedDateFrom = rawFrom && isValidDateString(rawFrom) ? rawFrom : undefined;
const rawTo = context.request.query.filedDateTo as string | undefined;
const filedDateTo = rawTo && isValidDateString(rawTo) ? rawTo : undefined;
```

---

### Security: `chapters` array is unbounded and not allowlisted (backend/lib/controllers/trustee-cases/trustee-cases.controller.ts:62)

`chaptersParam.split(',').filter(Boolean)` has no upper bound on array size and no allowlist restriction. A very large `$in` array is a DoS vector against MongoDB, and arbitrary string values bypass the intent of the filter. The known chapter set is small and closed.

```typescript
const VALID_CHAPTERS = new Set(['7', '11', '12', '13', '15']);
const chapters = chaptersParam
  ? chaptersParam.split(',').filter(Boolean).filter(c => VALID_CHAPTERS.has(c)).slice(0, 10)
  : undefined;
```

---

### SRE: Confirm `BACKFILL-TRUSTEE-PHONETIC-TOKENS` has been run before deploying (backend/lib/adapters/gateways/mongo/trustees.mongo.repository.ts:319)

Removing the `phoneticTokens.notExists()` fallback means any trustee without a `phoneticTokens` field is now invisible to name search. The backfill dataflow (`BACKFILL-TRUSTEE-PHONETIC-TOKENS`) is opt-in and must be triggered manually. If it has not been run to completion in all target environments, legacy un-indexed trustees will become unfindable via search and automated matching.

Please confirm or document in the deployment notes that the backfill has been run.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: No.  It's setup to wrap.  It doesn't seem to be truncating.  This was a UX requirement from the userbase.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: The user base decided they wanted 3 options in the filter, the 3rd one representing all of the inactive status'.  Based on team decisions, the AC is wrong and should be updated.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: Single select comboboxes do not have pills.  We either need to do one of the following:

1. update the AC
2. update the component (will affect components app-wide), 
3. create a 1-off component, 
4. manage the Pill-Box component separately and add the single-select item to the pill-box manually. (we may be doing this already with the other filters since all filters share a pill-box)

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: Once again, the status combo-box is a single select, and since single-select combo-boxes don't have a clear button, this is currently not possible.  Once again we'd have to do one of the following to accomodate:

1. write a custom version of the combo-box
2. hack the flow of the current combo-box to force it to display the clear button
3. provide the option to the combo-box to provide a clear button.

number 3 might be best.

fmaddenflx:
James left a number of issues we need to discuss and agree on:

- Security: IDOR 
- SRE: 500-case hard cap silently truncates results
- Security: No date format validation on filedDateFrom/filedDateTo
- Security: chapters array is unbounded and not allowlisted 
- SRE: Confirm BACKFILL-TRUSTEE-PHONETIC-TOKENS has been run before deploying

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: The long label is by design and wraps.

user-interface/src/trustees/TrusteesList.tsx
fmaddenflx: Addressed in 9223f2b82 — removed the `appointments.length > 0` guard from the active branch so both branches now symmetrically exclude zero-appointment trustees.

fmaddenflx: Addressed in 9223f2b82 and 597a06f8d — `filterTrustees` is now called once via a `useMemo` (`baseFilteredTrustees`) and reused for display, pagination, counts, and announcements.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: Addressed in 9223f2b82 — label now includes all seven statuses: `'Inactive (Involuntarily Suspended, Voluntarily Suspended, Resigned, Terminated, Deceased, Removed, Inactive)'`.

user-interface/src/trustees/TrusteesList.tsx
fmaddenflx: Addressed in 9223f2b82 — signature tightened to `isInactiveStatus(status: AppointmentStatus): boolean` with the `undefined` guard removed.

fmaddenflx:
**Addressing @sourcery-ai feedback:**

Both items are resolved:
1. **Explicit inactive statuses** — `INACTIVE_STATUSES` set and `isInactiveStatus` helper added in 0770bbf40, refined in 9223f2b82.
2. **Fast path optimization** — restored in 9223f2b82 (early return when `statusFilter === 'all'` and no other filters active).

user-interface/src/trustees/TrusteesList.tsx
fmaddenflx: One remaining call site was still using `filterTrustees` directly — the chapter filter telemetry `useEffect`. Fixed in 11a47fc7a by replacing it with `baseFilteredTrustees.length` (same inputs, already memoized) and simplifying the dependency array.

user-interface/src/trustees/TrusteesList.test.tsx
fmaddenflx: Fixed all inline `userEvent.setup()` calls in `TrusteesList.test.tsx` as well in bc0e1781a — 14 tests now use a single `userEvent.setup()` instance per test, reused for all interactions.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -51,21 +57,45 @@ function isUserInSelectedDivision(trustee: TrusteeListItem, divisionFilter: Divi
...
     selectedChapters.length === 0 &&
-    divisionFilterMap.size === 0
-  )
+    selectedDistricts.length === 0 &&
+    !districtDivisionEnabled
+  ) {
     return trustees;
```
mattstankey, line 72: The fast-path early return is now unreachable for any user with the `districtDivisionEnabled` feature flag on, even when they have made no selections.

The old condition was `divisionFilterMap.size === 0`, which correctly allowed the early return when no divisions were selected regardless of the flag. Replacing it with `!districtDivisionEnabled` means the entire filter body runs on every render for flagged users, even with empty state.

Consider restoring the division map check:

```ts
if (
  statusFilter === 'all' &&
  selectedChapters.length === 0 &&
  selectedDistricts.length === 0 &&
  divisionFilterMap.size === 0  // restored
) {
  return trustees;
}
```

Correctness holds either way (empty map → `isUserInSelectedDivision` returns `true`), but this blocks the fast path unnecessarily.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -184,13 +215,49 @@ export default function TrusteesList() {
...
+    setLiveAnnouncement('');
+    setStatusFilter(status);
+  };
+
   const handleFilterName = (name: string) => {
     isNameFilterInteracted.current = true;
     lastFilterChanged.current = 'name';
     if (name.length >= 2) setNameSearchLoading(true);
     setNameSearch(name);
   };
```
mattstankey, line 208: `react-hooks/preserve-manual-memoization` is not a rule in `eslint-plugin-react-hooks` — this comment has no effect. If ESLint was raising a warning here, please identify the actual rule ID and either suppress it with the correct directive or resolve the underlying issue. Otherwise, remove this line.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: Jon, Katie, and I decided to leave this alone (no pill for single select) for now and wait until the users make a request.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
fmaddenflx: It's a single-select and can not be cleared.  Reviewed by Katie, Jon, and I

user-interface/src/trustees/TrusteesList.tsx
fmaddenflx: Addressed in 74cbc5863 and 6abd07892. The status filter now applies server-side:

- Controller parses `?status=active|inactive|all` query param and passes a `TrusteesSearchPredicate` to the use case
- `getTrusteeIdsByStatuses` queries the `trustee-appointments` collection at the database layer (MongoDB `$in` on status field) and returns matching trustee IDs
- Use case filters trustees to only those with matching appointment statuses before enrichment
- Frontend re-fetches from the API when the status filter changes, removing client-side status filtering entirely

user-interface/src/trustees/TrusteesList.tsx
fmaddenflx: Fixed in 038ce6b09 — restored `divisionFilterMap.size === 0` in the fast-path check so flagged users with no selections still get the early return.

user-interface/src/trustees/TrusteesList.tsx
fmaddenflx: Fixed in 038ce6b09 — removed the directive. ESLint confirmed it was unused (no such rule exists in the configured plugin).

fmaddenflx:
**IDOR — not applicable.** TrusteeAdmin is intentionally permitted to view any trustee's case list. This role manages the full trustee roster and requires cross-read access to all trustees' cases by design.

jamesobrooks (approved):
Once the automated checks are passing, I'm good with this.

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -164,13 +187,27 @@ export class TrusteesUseCase {
...
         this.courtsUseCase.getCourts(context),
+        statusStatuses
+          ? this.trusteeAppointmentsRepository.getTrusteeIdsByStatuses(statusStatuses)
+          : Promise.resolve(null),
       ]);
 
+      const trustees =
+        filteredTrusteeIds === null
+          ? allTrustees
+          : allTrustees.filter((t) => filteredTrusteeIds.includes(t.trusteeId));
```
mattstankey: O(n²) lookup — `filteredTrusteeIds` is a `string[]`, so `.includes()` is O(n) per element, making the whole filter O(n²). Convert to a Set before filtering:

```ts
const idSet = new Set(filteredTrusteeIds);
const trustees = filteredTrusteeIds === null
  ? allTrustees
  : allTrustees.filter((t) => idSet.has(t.trusteeId));
```

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -193,7 +230,9 @@ export class TrusteesUseCase {
         appointments: appointmentsByTrusteeId.get(trustee.trusteeId) ?? [],
       }));
 
-      listItems.sort((a, b) => {
+      const filtered = listItems;
```
mattstankey: Dead alias — `filtered` is just a reference to `listItems`, so this rename is a no-op. Either sort and return `listItems` directly, or rename to something that reflects its actual purpose (e.g. `sortedListItems`).

```ts
// instead of:
const filtered = listItems;
filtered.sort(...);
return filtered;

// just:
listItems.sort(...);
return listItems;
```

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
```diff
@@ -470,6 +470,10 @@ export class MockMongoRepository
     throw new Error('Method not implemented.');
   }
 
+  getTrusteeIdsByStatuses(_statuses: unknown): Promise<string[]> {
```
mattstankey: Weak mock typing — the interface declares `statuses: AppointmentStatus[]` but the mock uses `unknown`. This disables type-checking on call sites in tests and diverges from the interface contract. Should match the interface:

```ts
getTrusteeIdsByStatuses(_statuses: AppointmentStatus[]): Promise<string[]> {
  throw new Error('Method not implemented.');
}
```

backend/lib/use-cases/trustees/trustees.ts
fmaddenflx: Fixed in c1f0e187d — converted to a `Set` for O(1) lookups.

backend/lib/use-cases/trustees/trustees.ts
fmaddenflx: Fixed in c1f0e187d — removed the dead alias and used `listItems` directly.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
fmaddenflx: Fixed in a249ce7c5 — updated the mock parameter type from `unknown` to `AppointmentStatus[]` to match the interface.

PR #2501, opened 2026-06-12
Add responsive mobile hamburger menu to Header

fmaddenflx:
## Summary
- Implements USWDS-compliant responsive header that collapses navigation into a hamburger menu on mobile/tablet (below 1024px breakpoint)
- Mobile-first SCSS with slide-in nav panel, overlay backdrop, and intermediate breakpoint (1024–1099px) that tightens padding for narrow desktops
- Full keyboard navigation support: ArrowUp/Down cycles through menu items, Escape closes and returns focus to menu button

## Test plan
- [ ] Verify hamburger menu appears below 1024px viewport width
- [ ] Verify clicking Menu button opens nav panel
- [ ] Verify clicking Close button or overlay closes nav panel
- [ ] Verify Escape key closes nav and returns focus to menu button
- [ ] Verify ArrowDown/ArrowUp navigate between menu items
- [ ] Verify nav closes on route change (clicking a link)
- [ ] Verify desktop layout displays correctly at 1024–1099px (compact padding)
- [ ] Verify desktop layout displays correctly above 1100px (full padding)
- [ ] Verify no transition/animation flash when resizing across breakpoint
- [ ] Screen reader testing: aria-expanded, aria-controls, focus management

## Summary by Sourcery

Implement a responsive, USWDS-style header with a mobile hamburger navigation and improved accessibility and focus management.

New Features:
- Add mobile hamburger navigation with slide-in panel and overlay for small viewports.
- Render user menu items as flat links within the mobile navigation panel.

Enhancements:
- Refine header layout and typography across mobile, compact desktop, and wide-screen breakpoints, including dynamic short/long site titles.
- Prevent body scrolling when mobile navigation is open and ensure overlay behavior is consistent across header and blank login page.
- Improve keyboard navigation and focus trapping within the mobile nav and dropdown menus, including Escape, Arrow, and Tab key handling.
- Update header link highlighting logic so case-detail routes do not mark any primary nav item as active.

Tests:
- Expand header tests to cover mobile hamburger behavior, focus management, overlay interactions, and nav activation rules.
- Add dropdown menu test to verify clicking the menu button toggles the expanded state.

### DIFF
--- a/user-interface/src/lib/components/Header.scss
+++ b/user-interface/src/lib/components/Header.scss
@@ -14,21 +14,268 @@
   }
 }
 
-// Overlay shown when mobile nav is open
-.usa-overlay {
-  position: fixed;
-  inset: 0;
-  background-color: rgba(0, 0, 0, 0.7);
-  opacity: 0;
-  visibility: hidden;
-  transition:
-    opacity 0.2s ease,
-    visibility 0.2s ease;
-  z-index: 400;
+.cams-header-container {
+  .usa-overlay {
+    position: fixed;
+    inset: 0;
+    background-color: rgba(0, 0, 0, 0.7);
+    opacity: 0;
+    visibility: hidden;
+    z-index: 400;
 
-  &.is-visible {
-    opacity: 1;
-    visibility: visible;
+    &.is-visible {
+      opacity: 1;
+      visibility: visible;
+    }
+  }
+
+  .cams-header.usa-header > * {
+    background-color: colors.$navy;
+  }
+
+  .cams-header {
+    background-color: colors.$navy;
+    border-bottom: 3px solid colors.$gold;
+    width: 100%;
+    padding: 0 15px;
+    position: relative;
+
+    @include header-links;
+
+    .usa-banner > * {
+      vertical-align: middle;
+      line-height: normal;
+    }
+
+    .usa-nav-container {
+      display: block;
+      padding-left: 0;
+      padding-right: 0;
+      max-width: none;
+      margin: 0;
+
+      .cams-logo-and-title {
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+
+        .usa-navbar {
+          border-bottom: none;
+          height: auto;
+
+          .cams-logo.usa-logo {
+            flex: 0 0 auto;
+            margin: 0px;
+            line-height: 0;
+            .doj-logo {
+              padding: 0;
+              height: 60px;
+              width: auto;
+              min-width: 60px;
+            }
+          }
+        }
+
+        .site-title {
+          display: flex;
+          flex: 1 1 auto;
+          flex-direction: column;
+          justify-content: center;
+          color: white;
+          padding: 0 0.5rem;
+          margin-top: 0;
+          margin-bottom: 0;
+
+          font-family: Source Serif Pro;
+          font-weight: 700;
+          font-size: 1.25rem;
+          line-height: 100%;
+
+          .sub-title {
+            display: block;
+            font-size: 1rem;
+            font-weight: 600;
+          }
+        }
+
+        .site-title .full-name {
+          display: none;
+        }
+        .site-title .short-name {
+          display: block;
+        }
+      }
+
+      // Mobile nav: slide-in panel from right
+      nav.usa-nav.cams-nav-bar {
+        position: fixed;
+        top: 0;
+        right: 0;
+        bottom: 0;
+        width: 75vw;
+        max-width: 320px;
+        background-color: colors.$navy;
+        z-index: 500;
+        padding: 1rem;
+        overflow-y: auto;
+        display: none;
+        flex-direction: column;
+
+        @include header-links;
+        a {
+          font-weight: bold;
+        }
+
+        &.is-visible {
+          display: flex;
+        }
+
+        .usa-nav__close {
+          align-self: flex-end;
+          background: none;
+          border: none;
+          color: white;
+          cursor: pointer;
+          padding: 0.5rem;
+          margin-bottom: 1rem;
+
+          .usa-icon {
+            width: 1.5rem;
+            height: 1.5rem;
+            fill: white;
+          }
+        }
+
+        .usa-current::after {
+          background-color: colors.$gold;
+        }
+
+        ul.usa-nav__primary {
+          margin-block-start: 0;
+          margin-block-end: 0;
+          list-style-type: none;
+          padding-left: 0;
+          display: flex;
+          flex-direction: column;
+          gap: 0;
+
+          li.usa-nav__primary-item {
+            border-top: 1px solid rgba(255, 255, 255, 0.2);
+
+            a {
+              display: block;
+              padding: 1rem;
+              white-space: nowrap;
+              font-size: 1.06rem;
+              color: white;
+            }
+            a:hover {
+              background-color: rgba(255, 255, 255, 0.1);
+            }
+
+            .cams-dropdown-menu {
+              button {
+                font-weight: 700;
+                line-height: 0.9;
+              }
+              button.usa-accordion__button[aria-expanded='false'] span {
+                margin-right: 0;
+                padding-right: 1rem;
+                display: inline-block;
+              }
+              button.usa-accordion__button[aria-expanded='false'] span::after {
+                width: 1rem;
+                height: 1rem;
+                mask-size: 1rem 1rem;
+                right: 0.75rem;
+              }
+            }
+
+            #user-menu {
+              padding: 1rem;
+              color: white;
+              font-size: 1.06rem;
+
+              svg {
+                top: 3px;
+                margin-right: 0.25rem;
+                color: white;
+              }
+
+              span::after {
+                background-color: white;
+              }
+            }
+            #user-menu:hover {
+              color: colors.$gold;
+              span {
+                text-decoration: underline;
+              }
+              svg {
+                color: colors.$gold;
+              }
+            }
+            #user-menu:hover::after {
+              background-color: transparent;
+            }
+            #user-menu-item-list {
+              li {
+                background-color: #207bc0;
+                color: white;
+                border-top: none;
+                a {
+                  padding: 0.5rem 1rem;
+                }
+              }
+              li a:focus {
+                background-color: colors.$navy;
+              }
+            }
+          }
+        }
+      }
+    }
+
+    // Menu button - visible on mobile
+    .usa-menu-btn {
+      display: flex;
+      align-items: center;
+      gap: 0.25rem;
+      background: none;
+      border: none;
+      color: white;
+      font-size: 1.06rem;
+      font-weight: 700;
+      padding: 0.5rem 0.75rem;
+      cursor: pointer;
+      white-space: nowrap;
+      flex-shrink: 0;
+
+      .usa-icon {
+        width: 1.25rem;
+        height: 1.25rem;
+        fill: white;
+      }
+
+      &:hover {
+        color: colors.$gold;
+
+        .usa-icon {
+          fill: colors.$gold;
+        }
+      }
+
+      &:focus {
+        outline: 0.25rem solid #2491ff;
+        outline-offset: 0.125rem;
+      }
+    }
+
+    .usa-nav-container::after {
+      clear: none;
+      content: none;
+      display: none;
+    }
   }
 }
 
@@ -279,7 +526,7 @@ body.usa-mobile-nav-active {
 
 // Desktop: show nav inline, hide hamburger/close buttons
 @media screen and (min-width: screens.$desktop-screen-small) {
-  .usa-overlay {
+  .cams-header-container .usa-overlay {
     display: none;
   }
 
@@ -287,7 +534,7 @@ body.usa-mobile-nav-active {
     overflow: auto;
   }
 
-  .cams-header {
+  .cams-header-container .cams-header {
     padding: 0 50px;
 
     .usa-menu-btn {
@@ -390,7 +637,7 @@ body.usa-mobile-nav-active {
 
 // Compact desktop: tighten padding between 1024px and 1100px
 @media screen and (min-width: screens.$desktop-screen-small) and (max-width: 1099px) {
-  .cams-header {
+  .cams-header-container .cams-header {
     padding: 0 15px;
 
     .usa-nav-container {
@@ -416,13 +663,13 @@ body.usa-mobile-nav-active {
 
 // Wide-screen title breakpoint
 @media screen and (min-width: 1400px) {
-  .cams-header.usa-header--basic {
+  .cams-header-container .cams-header.usa-header--basic {
     .usa-nav-container {
       .cams-logo-and-title {
-        .site-title.wide-screen {
-          display: flex;
+        .site-title .full-name {
+          display: block;
         }
-        .site-title.small-screen {
+        .site-title .short-name {
           display: none;
         }
       }
--- a/user-interface/src/lib/components/Header.scss
+++ b/user-interface/src/lib/components/Header.scss
@@ -255,7 +255,10 @@
                   padding: 0.5rem 1rem;
                 }
               }
-              li a:focus {
+              // Use :focus-visible so the highlight only appears for keyboard
+              // navigation. Opening via mouse programmatically focuses the first
+              // item, but it should not look hovered/selected for mouse users.
+              li a:focus-visible {
                 background-color: colors.$navy;
               }
             }
@@ -630,18 +633,20 @@ body.usa-mobile-nav-active {
             border-top: none;
             position: relative;
 
-            a {
+            // Scope to direct-child links so these rules apply only to the
+            // primary nav links and do not leak into the dropdown submenu links.
+            > a {
               padding: 0.5rem 1rem;
               white-space: nowrap;
               font-size: 1.06rem;
               display: inline-block;
               position: relative;
             }
-            a:hover {
+            > a:hover {
               background-color: transparent;
               text-decoration: none;
             }
-            a:not(.usa-current):hover::after {
+            > a:not(.usa-current):hover::after {
               content: '';
               display: block;
               position: absolute;
--- a/user-interface/src/lib/components/Header.scss
+++ b/user-interface/src/lib/components/Header.scss
@@ -195,7 +195,7 @@
                 font-weight: 700;
                 line-height: 0.9;
               }
-              button.usa-accordion__button[aria-expanded='false'] span {
+              button.usa-accordion__button span {
                 margin-right: 0;
                 padding-right: 0;
                 display: inline-block;
--- a/user-interface/src/lib/components/Header.scss
+++ b/user-interface/src/lib/components/Header.scss
@@ -642,9 +642,11 @@ body.usa-mobile-nav-active {
               content: '';
               display: block;
               position: absolute;
+              top: auto;
               left: 1rem;
               right: 1rem;
               bottom: 0;
+              width: auto;
               height: 0.25rem;
               background-color: white;
             }
--- a/user-interface/src/lib/components/Header.scss
+++ b/user-interface/src/lib/components/Header.scss
@@ -180,7 +180,7 @@
               }
               button.usa-accordion__button[aria-expanded='false'] span {
                 margin-right: 0;
-                padding-right: 1rem;
+                padding-right: 0;
                 display: inline-block;
               }
               button.usa-accordion__button[aria-expanded='false'] span::after {
@@ -202,8 +202,16 @@
                 color: white;
               }
 
+              span {
+                display: inline-flex;
+                align-items: center;
+                gap: 0.25rem;
+              }
+
               span::after {
                 background-color: white;
+                position: relative;
+                margin-left: 0.25rem;
               }
             }
             #user-menu:hover {
@@ -250,6 +258,7 @@
       cursor: pointer;
       white-space: nowrap;
       flex-shrink: 0;
+      text-transform: none;
 
       .usa-icon {
         width: 1.25rem;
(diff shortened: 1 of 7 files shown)


user-interface/src/lib/components/Header.tsx
```diff
@@ -111,106 +164,136 @@ export const Header = () => {
...
+                menuButtonRef.current?.focus();
+              }}
+              ref={closeButtonRef}
+            >
+              <Icon name="close" />
+            </button>
+            <ul className="usa-nav__primary">
+              <li className="usa-nav__primary-item">
+                <NavLink
+                  to="/my-cases"
```
jamesobrooks, line 167: **Accessibility: Close button has no accessible name (WCAG SC 4.1.2)**

`<Icon name="close" />` renders with `aria-hidden="true"` by default (see `Icon.tsx` — `decorative` defaults to `true`). There is no visible text, `aria-label`, or `title` on the button, so the accessibility tree exposes an unnamed button. A screen reader will announce "button" with no context.

**Fix:** Add `aria-label="Close navigation"` to the button element — the same pattern already used by `Modal.tsx`:

```tsx
<button
  type="button"
  className="usa-nav__close"
  aria-label="Close navigation"
  data-testid="header-nav-close"
  ...
>
  <Icon name="close" />
</button>
```

**References:**
- [WCAG 2.1 SC 4.1.2 — Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
- [ARIA button name computation](https://www.w3.org/TR/accname-1.1/#button-name)
- Project precedent: `Modal.tsx` line 294 uses `aria-label="Close this window"`

---
**Prompt for your agent:**
> The close button in the mobile nav panel (`usa-nav__close` in `Header.tsx`) renders only an `<Icon name="close" />`. The `Icon` component defaults `decorative={true}`, which sets `aria-hidden="true"` on the SVG and produces no accessible name. The button has no `aria-label`, no visible text, and no `title`. A screen reader user landing on this button will hear only "button" with no indication of what it does. Add `aria-label="Close navigation"` to the button element. Reference: `Modal.tsx` line 294 for the established project pattern.

user-interface/src/lib/components/Header.tsx
```diff
@@ -89,10 +92,60 @@ export const Header = () => {
...
+    };
+  }, [mobileNavOpen]);
+
+  const navRef = useRef<HTMLElement>(null);
+
+  const handleNavKeyDown = useCallback(
+    (e: React.KeyboardEvent) => {
+      if (e.key === 'Escape' && mobileNavOpen) {
+        setMobileNavOpen(false);
+        menuButtonRef.current?.focus();
```
jamesobrooks: **Keyboard UX: Escape inside the User Menu dropdown closes the entire mobile nav panel**

When the User Menu dropdown is open inside the mobile nav and the user presses Escape, `DropdownMenu.handleSubItemKeyDown` closes the dropdown (correct), but does not call `event.stopPropagation()`. The event bubbles up to this `handleNavKeyDown` handler, where the `Escape && mobileNavOpen` guard triggers and closes the entire nav panel.

The result: a user who opens the User Menu to reach Help or Logout, then presses Escape intending to dismiss the dropdown and stay in the nav, is instead ejected from the panel entirely.

**Expected behavior (per ARIA APG Disclosure Navigation pattern):** Escape should dismiss only the innermost open layer. The dropdown closes; the mobile nav panel stays open.

**Fix:** Add `event.stopPropagation()` to the Escape branch in `DropdownMenu.handleSubItemKeyDown`:

```typescript
// DropdownMenu.tsx — in handleSubItemKeyDown
if (ev.key === Escape) {
  ev.stopPropagation();  // prevent bubbling to mobile nav handler
  handleToggleExpand();
  setFocus(true);
}
```

**References:**
- [ARIA APG — Disclosure Navigation Menu Example](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/) — Escape closes the current disclosure; focus returns to its button
- [ARIA APG — Menu Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) — nested Escape behavior

---
**Prompt for your agent:**
> In `DropdownMenu.tsx`, the `handleSubItemKeyDown` function handles the Escape key by closing the dropdown and returning focus to the trigger button. However, it does not call `event.stopPropagation()`. When this component is used inside the mobile nav panel in `Header.tsx`, pressing Escape in the dropdown also triggers the `handleNavKeyDown` handler on the `<nav>` element, which closes the entire mobile nav panel. Per the ARIA Authoring Practices Guide Disclosure Navigation pattern, Escape should close only the innermost open layer — the dropdown — not the entire panel. Add `ev.stopPropagation()` to the Escape handling branch in `DropdownMenu.handleSubItemKeyDown` to prevent the event from reaching the nav-level handler.

user-interface/src/lib/components/Header.tsx
```diff
@@ -89,10 +92,60 @@ export const Header = () => {
...
+      document.body.classList.add('usa-mobile-nav-active');
+      closeButtonRef.current?.focus();
+    } else {
+      document.body.classList.remove('usa-mobile-nav-active');
+    }
+    return () => {
+      document.body.classList.remove('usa-mobile-nav-active');
+    };
+  }, [mobileNavOpen]);
+
```
jamesobrooks: **Nit: `useCallback` provides no memoization benefit here**

`useCallback` is worthwhile when you need a stable function reference to avoid re-rendering memoized children. Here, none of the children receiving this handler are wrapped in `React.memo`, and `mobileNavOpen` is in the dependency array — so the reference changes on every open/close anyway. The stability guarantee `useCallback` is meant to provide is never realized.

A plain function works identically:

```tsx
const handleNavKeyDown = (e: React.KeyboardEvent) => {
  if (e.key === Escape && mobileNavOpen) {
    setMobileNavOpen(false);
    menuButtonRef.current?.focus();
    return;
  }
  if (e.key === ArrowDown || e.key === ArrowUp) {
    // ... unchanged
  }
};
```

This also allows removing `useCallback` from the import on line 10.

---
**Prompt for your agent:**
> In `Header.tsx`, `handleNavKeyDown` is wrapped with `useCallback([mobileNavOpen])`. `useCallback` is only beneficial when the resulting stable reference prevents re-renders of memoized children. No children of the `<nav>` element are wrapped in `React.memo`, and `mobileNavOpen` is in the dependency array — so the reference changes on every open/close regardless. Remove the `useCallback` wrapper and declare `handleNavKeyDown` as a plain `const` function. Also remove `useCallback` from the React import on line 10 if it is no longer used elsewhere in the file.

user-interface/src/lib/components/Header.scss
```diff
@@ -13,6 +14,29 @@
   }
 }
 
+// Overlay shown when mobile nav is open
+.usa-overlay {
+  position: fixed;
+  inset: 0;
+  background-color: rgba(0, 0, 0, 0.7);
```
jamesobrooks: **Low: Global CSS rules in a component-scoped SCSS file**

`.usa-overlay` and `body.usa-mobile-nav-active` are unscoped globals defined at the top level of a component SCSS file. These rules will apply to any element in the DOM with those class names, not just markup rendered by `Header.tsx`.

Specifically, `BlankPage.tsx` also renders a `<div className="usa-overlay">` (authentication loading screen). The `.usa-overlay` rule here will apply there too, adding a fixed full-screen dark backdrop to the auth screen — likely unintentional.

Consider either:
1. Scoping these rules under `.cams-header { .usa-overlay { ... } }`, or
2. Moving truly global USWDS overrides to a shared partial (e.g. `src/styles/abstracts/_uswds-overrides.scss`)

If the global scope is intentional (e.g. you want `.usa-overlay` to behave the same everywhere), a short comment to that effect will prevent future maintainers from inadvertently breaking it.

---
**Prompt for your agent:**
> In `Header.scss`, the `.usa-overlay` and `body.usa-mobile-nav-active` rules are defined at the top level (unscoped). These are global CSS rules that will apply to any element with those class names, not just Header markup. `BlankPage.tsx` also uses `<div className="usa-overlay">` and will be unexpectedly affected. Either scope these rules inside `.cams-header { }`, or move them to a shared SCSS partial. If the global scope is intentional, add a comment explaining why.

user-interface/src/lib/components/Header.tsx
```diff
@@ -89,10 +92,60 @@ export const Header = () => {
     setActiveNav(mapNavState(location.pathname));
   }, [location]);
 
+  useEffect(() => {
+    setMobileNavOpen(false);
+  }, [location.pathname]);
+
+  useEffect(() => {
```
jamesobrooks: **Nit: Two `useEffect`s fire on every route change — consolidate**

There are now two separate effects that both respond to pathname changes: the pre-existing one at line 92 (`[location]` → `setActiveNav`) and this new one (`[location.pathname]` → `setMobileNavOpen(false)`). On every navigation they both fire, scheduling two synchronous state updates.

These can be consolidated into one:

```tsx
useEffect(() => {
  setActiveNav(mapNavState(location.pathname));
  setMobileNavOpen(false);
}, [location.pathname]);
```

The original `[location]` effect (which also fires on hash/search/state changes where `activeNav` does not need updating) can then be removed.

---
**Prompt for your agent:**
> In `Header.tsx` there are two `useEffect` calls that both fire when the route changes: the pre-existing one at line ~92 with `[location]` as its dependency that calls `setActiveNav`, and the new one with `[location.pathname]` that calls `setMobileNavOpen(false)`. Consolidate them into a single effect with `[location.pathname]` as the dependency array that calls both `setActiveNav(mapNavState(location.pathname))` and `setMobileNavOpen(false)`. Remove the now-redundant original effect.

user-interface/src/lib/components/Header.tsx
```diff
@@ -111,106 +164,136 @@ export const Header = () => {
...
           </div>
-          <div className="cams-main-navigation">
-            <nav aria-label="Main menu" className="usa-nav cams-nav-bar" role="navigation">
-              <ul className="usa-nav__primary">
+          {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
+          <nav
+            aria-label="Main menu"
+            className={`usa-nav cams-nav-bar${mobileNavOpen ? ' is-visible' : ''}`}
+            id="cams-main-nav"
+            role="navigation"
```
jamesobrooks: **Low: ESLint suppression on `<nav>` lacks justification comment**

`jsx-a11y/no-noninteractive-element-interactions` is being suppressed here, but `<nav>` has an implicit ARIA `navigation` landmark role — which means this lint rule likely should not have fired at all (the rule targets elements with non-interactive roles). This looks like a false-positive from the current version of jsx-a11y.

Without a comment, a future developer seeing this suppression has no way to know whether it was a deliberate justified exception or a shortcut. Adding a brief inline explanation closes that gap:

```tsx
{/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- <nav> landmark requires onKeyDown for arrow-key menu navigation per WCAG 2.1 SC 2.1.1; jsx-a11y false-positive on landmark roles */}
```

Alternatively, if a jsx-a11y upgrade resolves the false-positive, the comment can be removed entirely instead.

---
**Prompt for your agent:**
> In `Header.tsx`, there is an `eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions` comment before the `<nav>` element. The `<nav>` element has an implicit `navigation` landmark role, so this rule likely should not have fired — it appears to be a false-positive from the current jsx-a11y version. Add an inline justification comment after the rule name explaining why the suppression is present (e.g. "jsx-a11y false-positive on landmark roles; onKeyDown is required for WCAG arrow-key navigation"). Alternatively, investigate whether upgrading jsx-a11y resolves the false-positive and removes the need for the suppression entirely.

user-interface/src/lib/components/Header.scss
```diff
@@ -54,304 +77,362 @@
...
         }
       }
+
+      .site-title.wide-screen {
+        display: none;
+      }
+      .site-title.small-screen {
+        display: flex;
+      }
     }
```
jamesobrooks, line 397: **Low: Two `.site-title` DOM nodes toggled with CSS**

The current markup renders two separate `<div class="site-title">` elements — one `.wide-screen`, one `.small-screen` — both always present in the DOM, with visibility toggled purely via CSS `display`. They differ only in text content.

This works, but one node with CSS-toggled child spans achieves the same result with less DOM weight and fewer SCSS rules:

```tsx
<div className="site-title">
  <span className="site-title__wide">U.S. Trustee Program</span>
  <span className="site-title__narrow">USTP CAMS</span>
  <span className="sub-title text-no-wrap site-title__wide-sub">Case Management System (CAMS)</span>
</div>
```

```scss
.site-title__narrow, .site-title__wide-sub { display: none; }

@media (min-width: 1400px) {
  .site-title__wide { display: block; }
  .site-title__narrow { display: none; }
  .site-title__wide-sub { display: block; }
}
// Below 1400px: show narrow, hide wide
.site-title__wide { display: none; }
.site-title__narrow { display: inline; }
```

Not blocking — the current approach is functional — but worth cleaning up when convenient.

---
**Prompt for your agent:**
> In `Header.tsx`, there are two `<div className="site-title ...">` elements — one with class `wide-screen` and one with `small-screen` — both rendered in the DOM at all times. Their visibility is toggled with CSS `display` properties at different breakpoints. Refactor this to a single `<div className="site-title">` containing child spans, where CSS shows/hides the appropriate spans at the 1400px breakpoint. Update `Header.scss` accordingly to remove the duplicate `.site-title.wide-screen` / `.site-title.small-screen` rules.

user-interface/src/lib/components/Header.scss
```diff
@@ -13,6 +14,29 @@
...
+.usa-overlay {
+  position: fixed;
+  inset: 0;
+  background-color: rgba(0, 0, 0, 0.7);
+  opacity: 0;
+  visibility: hidden;
+  transition:
+    opacity 0.2s ease,
+    visibility 0.2s ease;
+  z-index: 400;
```
jamesobrooks: **Low: Animation inconsistency — overlay fades, panel snaps**

The overlay has a `transition: opacity 0.2s ease, visibility 0.2s ease` fade-in/out, but the nav panel switches between `display: none` and `display: flex` with no transition — it appears and disappears instantly.

This creates a mixed experience: the background dims smoothly while the panel pops in abruptly. Pick one consistent approach:

- **Animate both**: replace `display: none/flex` on the panel with `transform: translateX(100%)` / `transform: translateX(0)` so the panel can slide in with a CSS transition:
  ```scss
  nav.usa-nav.cams-nav-bar {
    display: flex;  // always in DOM
    transform: translateX(100%);
    transition: transform 0.2s ease;
    &.is-visible { transform: translateX(0); }
  }
  ```
- **Animate neither**: remove the `transition` from `.usa-overlay` — a no-animation approach is also perfectly valid.

---
**Prompt for your agent:**
> In `Header.scss`, the `.usa-overlay` element has `transition: opacity 0.2s ease, visibility 0.2s ease` so it fades in/out smoothly. However, the nav panel (`nav.usa-nav.cams-nav-bar`) toggles between `display: none` and `display: flex` with no transition, so it appears/disappears instantly. Either (a) add a slide-in CSS transition to the nav panel using `transform: translateX(100%)` / `translateX(0)` with a matching `transition: transform 0.2s ease`, or (b) remove the `transition` property from `.usa-overlay` entirely for a consistent no-animation approach. Do not leave one animated and the other not.

user-interface/src/lib/components/Header.tsx
```diff
@@ -111,106 +164,136 @@ export const Header = () => {
...
+            <ul className="usa-nav__primary">
+              <li className="usa-nav__primary-item">
+                <NavLink
+                  to="/my-cases"
+                  data-testid="header-my-cases-link"
+                  className={'usa-nav-link ' + setCurrentNav(activeNav, NavState.MY_CASES)}
+                  onClick={() => setActiveNav(NavState.MY_CASES)}
+                  title="View a list of cases assigned to your account"
+                >
+                  My Cases
```
mattstankey, line 173: **Bug: Close button has no accessible label**

Screen readers will announce this button as unlabeled. Add an `aria-label` so keyboard/screen reader users know what it does:

```tsx
<button
  type="button"
  className="usa-nav__close"
  data-testid="header-nav-close"
  aria-label="Close navigation"
  onClick={() => { ... }}
  ref={closeButtonRef}
>
```

This is a WCAG 2.1 AA failure (Success Criterion 4.1.2 — Name, Role, Value).

mattstankey (commented):
Automated review: three findings from code analysis — two bugs and one ARIA redundancy.

user-interface/src/lib/components/Header.tsx
```diff
@@ -111,106 +164,136 @@ export const Header = () => {
...
+              type="button"
+              className="usa-nav__close"
+              data-testid="header-nav-close"
+              onClick={() => {
+                setMobileNavOpen(false);
+                menuButtonRef.current?.focus();
+              }}
+              ref={closeButtonRef}
+            >
+              <Icon name="close" />
```
mattstankey, line 162: **Bug: Close button has no accessible label**

Screen readers will announce this button as unlabeled — it only contains an icon with no text alternative. This is a WCAG 2.1 AA failure (SC 4.1.2 — Name, Role, Value).

Add `aria-label`:
```tsx
<button
  type="button"
  className="usa-nav__close"
  data-testid="header-nav-close"
  aria-label="Close navigation"
  onClick={...}
  ref={closeButtonRef}
>
```

user-interface/src/lib/components/Header.test.tsx
```diff
@@ -200,6 +207,172 @@ describe('Header', () => {
...
+      expect(searchLink).toHaveFocus();
+
+      await userEvent.keyboard('{ArrowDown}');
+
+      const nav = document.getElementById('cams-main-nav')!;
+      const focusable = Array.from(
+        nav.querySelectorAll<HTMLElement>('a[href], button:not([disabled])'),
+      );
+      const searchIndex = focusable.indexOf(searchLink);
+      expect(focusable[searchIndex + 1]).toHaveFocus();
```
mattstankey: **Bug: Focus-wrap test asserts the wrong element**

The test is titled "should wrap focus from last top-level nav link to first with ArrowDown" but the assertion checks `focusable[searchIndex + 1]` — the element *after* the search link, not the first element. This doesn't test wrapping at all; it passes for the same reason as any normal ArrowDown navigation.

To test actual wrapping, the test should start focus on the truly last focusable element in the nav and assert that ArrowDown lands on `focusable[0]`:
```ts
// Get the last focusable item and focus it
const nav = document.getElementById('cams-main-nav')!;
const focusable = Array.from(nav.querySelectorAll<HTMLElement>('a[href], button:not([disabled])'));
const lastItem = focusable[focusable.length - 1];
lastItem.focus();

await userEvent.keyboard('{ArrowDown}');

expect(focusable[0]).toHaveFocus();
```

user-interface/src/lib/components/Header.tsx
mattstankey: **Nit: Redundant `role="navigation"`**

`<nav>` already carries the implicit ARIA landmark role of `navigation` — the explicit attribute is redundant per the ARIA spec and HTML5 spec. Remove it to keep the markup clean.

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.scss
fmaddenflx: Yup, I broke my own rules and got lazy.  You're right.

fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.tsx
fmaddenflx: Done

user-interface/src/lib/components/Header.scss
fmaddenflx: Agreed.  WE DON'T NEED ANIMATION AT ALL.

user-interface/src/lib/components/Header.scss
fmaddenflx: Sure. Done.

user-interface/src/lib/components/Header.test.tsx
fmaddenflx: done

mattstankey (approved):

jalba-flexion (approved):

user-interface/src/lib/components/Header.tsx
```diff
@@ -70,13 +70,33 @@ export function setCurrentNav(activeNav: NavState, stateToCheck: NavState): stri
...
+      if (shiftKey) return currentIndex > 0 ? currentIndex - 1 : length - 1;
+      return currentIndex < length - 1 ? currentIndex + 1 : 0;
+    case 'ArrowDown':
+      return currentIndex < length - 1 ? currentIndex + 1 : 0;
+    case 'ArrowUp':
+      return currentIndex > 0 ? currentIndex - 1 : length - 1;
+  }
+}
+
 export const Header = () => {
```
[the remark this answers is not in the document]
fmaddenflx: Working on this.

user-interface/src/lib/components/Header.tsx
```diff
@@ -87,12 +107,60 @@ export const Header = () => {
...
+  const handleNavKeyDown = (e: React.KeyboardEvent) => {
+    if (!['Tab', 'ArrowDown', 'ArrowUp', 'Escape'].includes(e.key)) return;
+
+    const { key, shiftKey } = e as {
+      key: 'Tab' | 'ArrowDown' | 'ArrowUp' | 'Escape';
+      shiftKey: boolean;
+    };
+
+    if (key === 'Escape' && mobileNavOpen) {
+      setMobileNavOpen(false);
```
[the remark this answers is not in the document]
fmaddenflx: Working on this.

jamesobrooks (approved):

user-interface/src/lib/components/Header.tsx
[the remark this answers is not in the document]
fmaddenflx: Fixed in df063d533 — extracted mobile nav logic into `useMobileNav` hook and filtered hidden `.desktop-user-menu` elements from the focus trap.

user-interface/src/lib/components/Header.tsx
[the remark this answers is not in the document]
fmaddenflx: Addressed in df063d533 — extracted state, body class side-effects, and keyboard handling into a `useMobileNav` hook in `UseMobileNav.ts`.

PR #2513, opened 2026-06-16
CAMS-746 - Remediate trustee case appointment dataflow

btposey:
## Summary by Sourcery

Enhance the migrate-case-appointments dataflow with operational flags, telemetry, and integration coverage, and add repository support for scoped deletion of trustee appointments.

New Features:
- Add reset, deleteAll, and flushQueues control flags to the migrate-case-appointments start message to support reruns, scoped deletions, and queue diagnostics.
- Introduce a flushQueues mode that exports dataflow queue contents to JSONL blobs in object storage for inspection without running the migration.
- Provide an integration harness and supporting scripts to run end-to-end migrate-case-appointments tests against local containers or lower Azure environments.

Enhancements:
- Record migrate-case-appointments telemetry via completeDataflowTrace across start and page handlers, including success/failure metadata.
- Update the migrate-case-appointments use case to write failed records to a single output container and remove unused per-record retry handlers from the function app.
- Add a deleteAll operation in the migrate-case-appointments use case that removes ACMS-sourced appointments via a repository method scoped by source.
- Extend the trustee appointments Mongo repository and interface with a deleteAllBySource method, including unit tests and mock updates.
- Ensure the migrate-case-appointments dataflow creates its output blob container on startup and supports queue flushing via direct QueueServiceClient access.

CI:
- Adjust the Snyk security scan workflow to exclude function-apps and integration tests from SCA scanning and monitoring, and reformat long CLI invocations for readability.

Documentation:
- Add README documentation describing the migrate-case-appointments integration harness, its containers, workflows, assertions, and cleanup steps.

Tests:
- Add unit tests for migrate-case-appointments function flags (reset, deleteAll, flushQueues) and the new deleteAll use case behavior.
- Introduce integration tests, SQL seed scripts, and Podman helper scripts to exercise the migrate-case-appointments dataflow end-to-end, including reset, deleteAll, and flushQueues scenarios.
- Add tests for the trustee appointments repository deleteAllBySource behavior and error handling.


backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -1,10 +1,17 @@
 import { app, InvocationContext, output } from '@azure/functions';
```
btposey, line 1: **[Dependency Rule Violation — High]** This file imports `buildContainerName` at line 5 and `ModuleNames` at line 16, both from `function-apps/dataflows/`. Those transitive imports pull in `@azure/functions` into the `lib` use case layer.

The use case calls `buildContainerName(ModuleNames.MIGRATE_CASE_APPOINTMENTS, 'out')` to form the blob container name. That computation belongs here, in the use case, as a plain constant — no framework import needed.

```ts
// backend/lib/use-cases/dataflows/migrate-case-appointments.ts
const FAILURES_CONTAINER = 'migrate-case-appointments-out';
```

Remove the `buildContainerName` and `ModuleNames` imports from the use case entirely.

btposey (commented):
## Review Summary — CAMS-746 Migrate Case Appointments Remediation

Three specialist reviewers ran in parallel: software architect (OeSA compliance, logic), devsecops (security, ops), pragmatic (YAGNI, complexity). Findings are posted inline. Summary below.

---

### Must Fix

**[High] Dead/broken handleError + handleRetry infrastructure** — see handler file comment at line 296. These are live registered Azure Functions that can never do their intended job. processingSingleRecord is untested dead code with a bug. Delete ~120 lines.

**[High] Dependency Rule violation** — see use-case file comment at line 9. The use case imports buildContainerName from function-apps/, pulling `@azure/functions` into the lib layer. Replace with a local constant.

**[High] deleteAll flag has no safeguard** — see handler comment at line 97. A single queue message bulk-deletes all ACMS case appointments with no confirmation token, no audit record, and no environment guard.

**[High] Broken test assertion** — see handler test comment at line 50. Map.get(expect.anything()) always returns undefined. The assertion never actually verifies that no queue output was enqueued.

---

### Should Fix

**[Medium] FAILED migration state never written on page error** — see handler comment at line 224. State stays IN_PROGRESS indefinitely after a page-level failure.

**[Medium] flushQueues is a YAGNI stub** — see type declaration comment at line 56. Ships a no-op flag in the public message contract before there is an implementation.

**[Medium] flushQueues silently suppresses deleteAll/reset** — see handler comment at line 83. Combined flags produce a silent no-op with no warning logged.

**[Medium] processSingleRecord calls formatAcmsDate twice** — see use case comment at line 360. Bug in dead code, but should be fixed or the function deleted.

**[Medium] Credential values printed to stdout in integration harness** — see harness comment at line 88. ACMS_MSSQL_PASS logs in plaintext.

**[Medium] start-funcapp.sh swap has no guard for missing original settings** — see script comment at line 40.

---

### Minor

**[Low] .env.template ships a well-known example password** — see comment at line 8 of .env.template.

**[Low] seed-schema and seed-sql run the same SQL file** — see harness comment at line 344. Naming implies DDL/DML separation that doesn't exist.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -6,6 +6,8 @@ import factory from '../../factory';
 import { MaybeData } from './queue-types';
 import { MigrateCaseAppointmentsState, AcmsCaseAppointmentRecord } from '../gateways.types';
 import { CaseAppointment, CaseAppointmentInput } from '@common/cams/trustee-appointments';
+import { buildContainerName } from '../../../function-apps/dataflows/dataflows-common';
```
btposey: **[High] Dependency Rule violation — use case imports from `function-apps` layer**

Lines 9–10 import `buildContainerName` from `../../../function-apps/dataflows/dataflows-common` and `ModuleNames` from `../../../function-apps/dataflows/module-names`. Both of those modules transitively import `@azure/functions`, which means the `lib` use case layer now depends on the Azure Functions SDK.

The blob container name is just a string. Define it locally:

```ts
const FAILURES_CONTAINER = 'migrate-case-appointments-out';
```

Remove the two `function-apps` imports from this file entirely.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -175,14 +284,22 @@ export async function handlePage(
...
+  const context = await ApplicationContextCreator.getApplicationContext({ invocationContext });
+  const { logger } = context;
+  const trace = context.observability.startTrace(invocationContext.invocationId);
 
   logger.error(
     MODULE_NAME,
     `Error migrating CMMAP record id ${event.id}: ${event.lastErrorMessage ?? 'Unknown error'}.`,
   );
 
   invocationContext.extraOutputs.set(RETRY, [event]);
```
btposey: **[High] Dead/broken infrastructure — handleError and handleRetry should be deleted**

The comment above this block acknowledges these handlers are 'not currently used by this migration'. However they are registered as live Azure Functions, meaning they consume queue polling and invocation budget in every deployed environment.

The deeper problem: the DLQ here is written with page-level errors via buildQueueError(), which produces a structured error object. handleError expects a RetryMessage (an AcmsCaseAppointmentRecord). These are incompatible shapes — so even if handleError were triggered, it would process malformed input.

processSingleRecord exists only to serve handleRetry. It has zero test coverage and contains a bug: formatAcmsDate(record.assignDate) is called twice (see inline comment on the use case file).

Per the cams-dataflow REMEDIATION standards, a one-shot migration should use blob storage for failures, not per-record queue retry. That is already the correct behavior in this dataflow.

**Suggested fix:** Delete handleError, handleRetry, processSingleRecord, the RETRY and HARD_STOP queue output declarations, and their app.storageQueue() registrations in setup(). Approximately 120 lines removed, no functionality lost.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -56,9 +69,51 @@ type RetryMessage = AcmsCaseAppointmentRecord & {
...
+      documentsFailed: 0,
+      success: true,
+      details: { mode: 'flushQueues-noop' },
+    });
+    return;
+  }
+
+  // 2. deleteAll — delete all source==='acms' appointments before re-running
+  if (message.deleteAll) {
+    logger.info(MODULE_NAME, 'deleteAll flag detected — deleting all ACMS case appointments.');
```
btposey: **[High] deleteAll flag lacks safeguards — single queue message wipes all ACMS appointments**

Anyone with access to the start queue (Storage Queue Data Contributor RBAC, or the connection string) can trigger a full deleteMany on all ACMS-sourced case appointment records. There is no confirmation token, no environment guard, and no audit record written before execution. In the context of accidental operator error, a misconfigured CI/CD job, or a compromised managed identity, this is a realistic destructive path.

Suggested mitigations (pick at least one):

1. **Require a confirmation token** in the message so a plain `{ deleteAll: true }` cannot trigger it:
```ts
deleteAll?: { confirm: 'DELETE_ALL_ACMS_APPOINTMENTS' }
```

2. **Write an audit record** before executing — capture invocationId, timestamp, and operator identity.

3. **Gate on an environment variable** (e.g. `ALLOW_DELETE_ALL_ACMS=true`) absent in production, so the code path is a no-op unless explicitly enabled per environment.

4. At minimum: change this `logger.info` to `logger.warn` so it surfaces in alerting thresholds.

backend/function-apps/dataflows/migrations/migrate-case-appointments.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    expect(completeTraceSpy).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.anything(),
+      expect.any(String),
+      'handleStart',
+      expect.anything(),
+      expect.objectContaining({ success: true, details: { mode: 'flushQueues-noop' } }),
+    );
+    // No PAGE queue message enqueued
+    expect(invocationContext.extraOutputs.get(expect.anything())).toBeUndefined();
```
btposey: **[High] Broken assertion — this always passes regardless of output queue state**

`invocationContext.extraOutputs.get(expect.anything())` does not work as intended. `Map.prototype.get()` performs a reference equality lookup on the key. `expect.anything()` is a Vitest asymmetric matcher object — not a key that was ever inserted into the Map. So `.get()` will return `undefined` unconditionally, and the assertion always passes even if the PAGE queue was populated.

The intent is to verify no queue output was enqueued. Use the Map's size instead:

```ts
expect(invocationContext.extraOutputs.size).toBe(0);
```

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -130,6 +215,12 @@ export async function handlePage(
       DLQ,
       buildQueueError(result.error, MODULE_NAME, HANDLE_PAGE),
     );
+    completeDataflowTrace(context.observability, trace, MODULE_NAME, 'handlePage', logger, {
+      documentsWritten: 0,
+      documentsFailed: 0,
+      success: false,
+      error: result.error.message,
+    });
     return;
```
btposey, line 348: **[Medium] FAILED migration state never written on page error**

When processPage returns an error, the handler routes to DLQ and returns — but never calls updateMigrationState with status: 'FAILED'. The state document stays IN_PROGRESS indefinitely.

The MigrateCaseAppointmentsState type explicitly includes 'FAILED' as a valid status. After a page error, an operator cannot distinguish 'currently running' from 'failed and abandoned' without inspecting the DLQ directly.

Suggested fix:
```ts
if (result.status === 'error') {
  await MigrateCaseAppointmentsUseCase.updateMigrationState(
    context,
    { lastId: cursor.lastId, processedCount: 0, status: 'FAILED' },
  );
  invocationContext.extraOutputs.set(DLQ, buildQueueError(result.error, MODULE_NAME, HANDLE_PAGE));
  completeDataflowTrace(...);
  return;
}
```

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -43,6 +50,12 @@ const HANDLE_PAGE = buildFunctionName(MODULE_NAME, 'handlePage');
 const HANDLE_ERROR = buildFunctionName(MODULE_NAME, 'handleError');
 const HANDLE_RETRY = buildFunctionName(MODULE_NAME, 'handleRetry');
 
+export type MigrateCaseAppointmentsStartMessage = StartMessage & {
+  reset?: boolean; // Resets the MIGRATE_CASE_APPOINTMENTS_STATE doc in COSMOS for a fresh re-run
+  deleteAll?: boolean; // Deletes all CaseAppointment records where source === 'acms'
+  flushQueues?: boolean; // Reserved — dumps queues to blob storage (not yet implemented, guard and log)
```
btposey: **[Medium] flushQueues is a speculative stub — remove until implemented**

This flag is declared in the exported type contract, handled with telemetry, and tested — for a no-op that logs 'not yet implemented.' Publishing a flag in an external message contract before you implement the feature creates a contract obligation to honor the name semantically when you eventually do implement it, while burning the name in the namespace now.

The cams-dataflow REMEDIATION procedures include flushQueues as a standard flag for import/migration dataflows, so this will be needed eventually. But shipping a 'Reserved' no-op as a published flag is premature.

**Suggested fix:** Delete the flushQueues property from the type, the guard block (lines 81–93), and the test covering it. Add it back when queue flushing has a real implementation.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -56,9 +69,51 @@ type RetryMessage = AcmsCaseAppointmentRecord & {
...
+  invocationContext: InvocationContext,
+) {
   const context = await ApplicationContextCreator.getApplicationContext({ invocationContext });
   const { logger } = context;
+  const trace = context.observability.startTrace(invocationContext.invocationId);
+
+  // 1. flushQueues — not yet implemented, log and return
+  if (message.flushQueues) {
+    logger.info(
+      MODULE_NAME,
```
btposey, line 220: **[Medium] flushQueues silently suppresses deleteAll/reset when both flags are set**

The three flags are evaluated with early returns. If flushQueues is true, the function returns before evaluating deleteAll or reset. A caller who sends `{ flushQueues: true, deleteAll: true }` expecting both operations gets neither, with no indication that deleteAll was ignored.

If flushQueues is kept (see type declaration comment), log a warning when other flags are present:

```ts
if (message.flushQueues) {
  if (message.deleteAll || message.reset) {
    logger.warn(MODULE_NAME, 'flushQueues combined with deleteAll/reset — other flags ignored.');
  }
  // ... existing noop path
}
```

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -344,11 +350,26 @@ async function processSingleRecord(
...
 }
 
+async function deleteAll(
+  context: ApplicationContext,
+): Promise<MaybeData<{ deletedCount: number }>> {
+  try {
+    const repo = factory.getTrusteeAppointmentsRepository(context);
+    const result = await repo.deleteAllBySource('acms');
+    return { data: result };
+  } catch (originalError) {
```
btposey, line 273: **[Medium] processSingleRecord calls formatAcmsDate(record.assignDate) twice**

The `assignedOn` computed at line 321 (inside the duplicate-check block) is not reused in the CaseAppointmentInput spread at line 336 — instead the function calls `formatAcmsDate(record.assignDate)` again. This is copy-paste drift from the processPage per-record loop which correctly captures the result once.

This function should be deleted along with handleRetry (see handler file comment), but if retained:

```ts
const assignedOn = formatAcmsDate(record.assignDate);
const duplicate = existingAppointments.some(
  (a) => a.trusteeId === trusteeId && a.source === 'acms' && a.assignedOn === assignedOn,
);
if (duplicate) return { status: 'skipped' };
// ...
const input: CaseAppointmentInput = {
  caseId: record.caseId,
  trusteeId,
  assignedOn,  // reuse the already-computed value
  ...
};
```

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
```diff
@@ -0,0 +1,817 @@
...
+    const settings = JSON.parse(fs.readFileSync(resolved, 'utf-8'));
+    const values: Record<string, string> = settings?.Values ?? {};
+    for (const [key, value] of Object.entries(values)) {
+      if (!(key in process.env)) process.env[key] = value;
+    }
+  } catch {
+    // Non-fatal
+  }
+}
+
```
btposey, line 91: **[Medium] checkEnv prints credential values to stdout in plaintext**

The optional variable list includes `ACMS_MSSQL_PASS`, and the current logging prints `name=value` for each. Any CI log capture or terminal recording will expose the SQL SA password.

Mask credential-class variables before printing:

```ts
const isSensitive = name.toLowerCase().includes('pass') || name.toLowerCase().includes('key') || name.toLowerCase().includes('secret');
const display = isSensitive ? (value ? '***' : '(not set)') : (value ?? '(not set)');
info(`${name}=${display} — ${description}`);
```

test/integration/migrate-case-appointments/scripts/start-funcapp.sh
```diff
@@ -0,0 +1,46 @@
...
+  if [ -f "$BACKUP_SETTINGS" ]; then
+    mv "$BACKUP_SETTINGS" "$ORIGINAL_SETTINGS"
+    echo ""
+    echo "Restored local.settings.json"
+  fi
+}
+trap restore_settings EXIT INT TERM
+
+# Swap settings
+cp "$ORIGINAL_SETTINGS" "$BACKUP_SETTINGS"
```
btposey, line 46: **[Medium] Settings swap has no guard if local.settings.json doesn't exist**

If ORIGINAL_SETTINGS (local.settings.json) does not exist when the script runs — a developer who hasn't configured the dataflows app yet — the first `cp` fails. With `set -e` this triggers the EXIT trap, but the trap checks `if [ -f "$BACKUP_SETTINGS" ]` — since the backup was never created, no restore occurs. The script exits leaving the integration settings copied into place as local.settings.json. On the next `npm start` without this script, the function app runs against integration databases.

Add an existence check before the swap:

```bash
if [ ! -f "$ORIGINAL_SETTINGS" ]; then
  echo "ERROR: $ORIGINAL_SETTINGS not found — cannot backup and swap settings." >&2
  exit 1
fi
```

test/integration/migrate-case-appointments/scripts/.env.template
```diff
@@ -0,0 +1,8 @@
+# SQL Edge SA password used by start-services.sh.
+# Copy this file: cp .env.template .env
+# Then set MSSQL_PASS to a strong password (SQL Server requires complexity).
+# Use the same value for ACMS_MSSQL_PASS in ../.env.local.
+#
+# .env is gitignored — never commit it.
+
+MSSQL_PASS=YourStrong!Passw0rd
```
btposey: **[Low] Well-known example password normalizes a weak default**

`YourStrong!Passw0rd` is the canonical SQL Server docs example password, and it appears in `local.settings.integration.json` as the default too. Developers who copy-paste and forget to change it end up with an SA account using a well-known password on port 1433.

Change to an empty assignment that forces the developer to set a value:
```
MSSQL_PASS=
# Required: set a strong password (min 8 chars, upper+lower+digit+symbol)
```

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
```diff
@@ -0,0 +1,817 @@
...
+  try {
+    const seedDir = path.join(HARNESS_DIR, 'seed');
+    await executeSqlFile(pool, path.join(seedDir, '01-seed-cmmap.sql'));
+    pass('01-seed-cmmap.sql applied (CMMAP schema created)');
+  } finally {
+    await pool.close();
+  }
+}
+
+// ---------------------------------------------------------------------------
```
btposey, line 362: **[Low] seedSchema and seedSql both execute the same SQL file**

Both functions run `01-seed-cmmap.sql`, which contains both DDL (CREATE TABLE) and DML (INSERT). Running `seed-schema` followed by `seed-sql` as documented in the README applies the same file twice. The naming implies DDL/DML separation that doesn't exist, which will confuse anyone reading the workflow.

Either split the SQL file into a schema DDL file and a data DML file so each command does what its name suggests, or collapse both commands into a single `seed` command that runs the combined file once.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
mattstankey: Fixed. Removed both `function-apps` imports and replaced the single call site with a local constant:

```ts
const FAILURES_CONTAINER = 'migrate-case-appointments-out';
```

`buildContainerName` and `ModuleNames` are no longer referenced in this file.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
mattstankey: `processSingleRecord` was deleted entirely along with `handleRetry` (which was its only caller). The dead code path and the double `formatAcmsDate` call are both gone.

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
btposey: Fixed. Sensitive variable names matching `/pass|key|secret/i` now display `***` instead of the raw value.

test/integration/migrate-case-appointments/scripts/start-funcapp.sh
btposey: Fixed. Added a guard before the `cp` that exits with an error if `local.settings.json` does not exist, preventing integration settings from being left in place after a failed swap.

test/integration/migrate-case-appointments/scripts/.env.template
btposey: Fixed. Replaced the well-known example password with an empty value and a comment instructing developers to set a strong password.

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
btposey: Fixed. Split the single SQL file into two: `00-seed-cmmap-schema.sql` (DDL — CREATE TABLE) used by `seed-schema`, and `01-seed-cmmap.sql` (DML — TRUNCATE + INSERT) used by `seed-sql`. Each command now does what its name says.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
bityogi: Addressed in upstream commit 559430635 ("CAMS-760 Fix dep rule in appointments use case"): the use case now declares `FAILURES_CONTAINER` as a local constant and the `buildContainerName` / `ModuleNames` imports were removed from `backend/lib/use-cases/dataflows/migrate-case-appointments.ts`. ✅

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
bityogi: ✅ Done in d39dc2d. Deleted `handleError`, `handleRetry`, the `RETRY` and `HARD_STOP` queue output declarations, the `RetryMessage` type, the `HANDLE_ERROR` / `HANDLE_RETRY` function-name constants, and their `app.storageQueue()` registrations from `setup()`. `processSingleRecord` was already removed from the use case in 559430635 alongside the dep-rule fix. Net: ~150 lines deleted from the handler, no functionality lost — page-level failures continue to flow to blob storage as before.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
bityogi: Partially addressed in d39dc2d. Promoted the `deleteAll` log to `logger.warn` so the operation surfaces in alerting (option 4 from your suggestions list).

After discussion with Brian and the team, we are intentionally **not** adding the confirmation token / audit record / env gate in this PR — the consensus was that requiring `{ deleteAll: true }` is sufficient for the current threat model, and adding token/env-guard ceremony was overcomplicated for a one-shot migration tool. If we want stronger safeguards later, we can revisit in a follow-up.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
bityogi: ✅ Done in d39dc2d. The `handlePage` error branch now reads existing state, then writes `status: 'FAILED'` via `updateMigrationState` (preserving `lastId` / `processedCount` from the existing state, falling back to the cursor) before routing to DLQ and returning. State no longer stays IN_PROGRESS after a page error.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
bityogi: ✅ Done in d39dc2d. Removed the `flushQueues?` property from `MigrateCaseAppointmentsStartMessage`, the guard block in `handleStart`, and the corresponding test. We can add it back when queue flushing has a real implementation per the cams-dataflow REMEDIATION procedure.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
bityogi: ✅ Resolved in d39dc2d. Since the `flushQueues` flag was removed entirely, there is no longer any silent suppression — `deleteAll` and `reset` now flow through unconditionally without an early return shadowing them.

backend/function-apps/dataflows/migrations/migrate-case-appointments.test.ts
mattstankey: Resolved by deleting the `flushQueues` test block entirely — `flushQueues` was removed from the handler and message type as YAGNI (cams-025o), so the assertion and the test that contained it are gone.

bityogi: Resolved by deletion in d39dc2d. The companion handler change removed the `flushQueues` flag, so the test that contained this broken `Map.get(expect.anything())` assertion was deleted entirely (per the coordination note on the related beads issue) rather than fixed. The remaining tests that needed similar assertions use `(invocationContext.extraOutputs as Map<unknown, unknown>).values()` patterns, which work correctly.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
mattstankey: Fixed in commit 559430635 — the `buildContainerName` and `ModuleNames` imports were removed from `backend/lib/use-cases/dataflows/migrate-case-appointments.ts` and replaced with a local constant `const FAILURES_CONTAINER = 'migrate-case-appointments-out'`. The imports in this handler file are within `function-apps/` itself and are not a violation.

.github/workflows/sub-security-scan.yml
```diff
@@ -41,18 +41,22 @@ jobs:
           snyk-version: latest
 
       - name: Authenticate with Snyk
-        run: snyk auth --auth-type=oauth --client-secret="$SNYK_OAUTH_CLIENT_SECRET" --client-id="$SNYK_OAUTH_CLIENT_ID"
+        run:
```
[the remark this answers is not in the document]
btposey: No action needed — this comment is on a prior commit. The current HEAD already uses single-line `run:` for all snyk commands, so the YAML is valid.

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
```diff
@@ -0,0 +1,947 @@
...
+      console.log('  run-delete-all  Verify deleteAll scopes deletion to source=acms only');
+      console.log('  run-flush-queues Verify flushQueues dumps queue contents to blob storage');
+      console.log('  clean           Remove test documents and clear queues');
+      console.log('  help            Show this help');
+      break;
+    }
+  }
+
+  console.log('\n' + '='.repeat(60));
+  process.exit(0);
```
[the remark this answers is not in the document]
btposey: Fixed. Added a module-level `hasFailures` flag that `fail()` sets to `true`. `main()` now calls `process.exit(hasFailures ? 1 : 0)` so the harness correctly signals failures to automated runs.

.github/workflows/sub-security-scan.yml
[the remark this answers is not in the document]
btposey: No action needed — this comment is on a prior commit (`205f4e4`). The current HEAD already uses single-line `run:` for all snyk commands so the YAML is valid.

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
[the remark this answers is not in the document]
btposey: Fixed in commit `6cb7322`. Added a module-level `hasFailures` flag that `fail()` sets to `true`. `main()` now calls `process.exit(hasFailures ? 1 : 0)` so the harness correctly signals failures to automated runs.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -27,38 +35,109 @@ const DLQ = output.storageQueue({
...
+type MigrateCaseAppointmentsPageMessage = { lastId: number | null };
+
 /**
  * handleStart
  *
  * Initialize the migration by reading existing state for resumability.
  * If already completed, skip. Otherwise queue first/next page cursor with lastId from state.
  */
-export async function handleStart(_ignore: StartMessage, invocationContext: InvocationContext) {
+export async function handleStart(
```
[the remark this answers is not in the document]
btposey: Acknowledged. We are intentionally deferring this refactoring — this is a one-shot migration flow and the complexity overhead is not justified at this time. We will revisit these patterns when updating the cams-dataflow skill templates.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -146,6 +278,12 @@ export async function handlePage(
       MODULE_NAME,
       `ACMS CMMAP migration complete. Total processed: ${result.processedCount} records.`,
     );
+    completeDataflowTrace(context.observability, trace, MODULE_NAME, 'handlePage', logger, {
+      documentsWritten: result.processedCount,
```
diabagatekelly: [Medium] `documentsWritten` uses `processedCount` (successes + failures) and `documentsFailed` is hardcoded to 0. The `continue` path just below gets this right. One-line fix:

```ts
documentsWritten: result.successCount,
documentsFailed: result.failedCount,
```

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -27,38 +35,109 @@ const DLQ = output.storageQueue({
...
-export async function handleStart(_ignore: StartMessage, invocationContext: InvocationContext) {
+export async function handleStart(
+  message: MigrateCaseAppointmentsStartMessage,
+  invocationContext: InvocationContext,
+) {
   const context = await ApplicationContextCreator.getApplicationContext({ invocationContext });
   const { logger } = context;
+  const trace = context.observability.startTrace(invocationContext.invocationId);
+
+  if (message.flushQueues) {
```
diabagatekelly, line 156: [Medium] `flushQueues` silently drops `deleteAll` (and `reset`) when both flags are set — the early return at line 117 prevents them from being evaluated. A caller sending `{ flushQueues: true, deleteAll: true }` gets only the queue dump with no indication the delete was skipped.

Suggest logging a warning when other flags are present:
```ts
if (message.flushQueues) {
  if (message.deleteAll || message.reset) {
    logger.warn(MODULE_NAME, 'flushQueues is set — deleteAll and reset flags will be ignored.');
  }
  // ...
}
```

diabagatekelly (approved):

PR #2517, opened 2026-06-16
CAMS-767: ui issues

bityogi:
# Bug

On the Trustees screen, changing the **Status** filter caused three regressions:

1. The whole screen flashed the loading spinner instead of just the table — the filter accordion disappeared during reload.
2. Any selected **Chapter** filters were silently cleared.
3. The Filters accordion collapsed even though the user had it open.

# Purpose

Make the Status filter behave like the other filters: only the table reloads, the filter UI stays mounted, and the user's other selections are preserved.

# Major Changes

- `TrusteesList.tsx`: Removed the early-return of `<LoadingSpinner>` / error alert that was unmounting `<TrusteeDistrictFilter>` on every status change. The filter now stays mounted; the loading spinner, error alert, empty state, and table are all rendered inside the same layout, so only the table region swaps while data reloads.
- Because the filter no longer remounts, its accordion expanded state, selected chapters, selected districts/divisions, and status selection all persist across reloads triggered by `setStatusFilter`.

# Testing/Validation

- Added `should keep filters accordion open and chapter selection when status changes` to `TrusteesListStatusFilter.test.tsx` — selects a chapter, switches status from Active to Inactive, and asserts both the chapter combobox is still visible (accordion still open) and the chapter pill is still present after the reload.
- Adjusted one existing test in `TrusteesList.test.tsx` (`default session divisions appear at top of combined dropdown with divider`) that was using "1 Trustee" as a wait gate. With the filter now mounted during loading, default session divisions correctly apply sooner and filter out the test's mock trustee, so the gate now waits on the Filters toggle button instead.
- All 102 tests in `TrusteesList.test.tsx`, `TrusteesListStatusFilter.test.tsx`, and `filters/TrusteeDistrictFilter.test.tsx` pass.
- Manually verified by clicking through the Filters accordion: open Filters → select chapter pill → change Status to Inactive — the accordion stays open, the chapter pill stays applied, and only the table area shows the spinner.

# Notes

The 24 unrelated failures observed in `forms/TrusteeAppointmentForm.test.tsx` reproduce on `main` (verified via `git stash`); they are pre-existing and out of scope for this PR.

# Definition of Done:

## Summary by Sourcery

Preserve the trustees list layout while loading and on error, and ensure filters remain visible and applied when status changes.

Bug Fixes:
- Keep the trustees table and filter UI visible while data is loading or when an error occurs instead of short-circuiting the render.
- Maintain the filters accordion open state and chapter selection when the status filter is changed and data is reloaded.

Tests:
- Add a regression test to verify filters accordion stays open and chapter selection persists when switching trustee status filters.
- Adjust an existing trustees list test to wait for the filters toggle button via findByRole for more robust async behavior.

### DIFF
--- a/user-interface/src/trustees/TrusteesList.tsx
+++ b/user-interface/src/trustees/TrusteesList.tsx
@@ -438,6 +438,23 @@ export default function TrusteesList() {
     displayCount = stableCountRef.current;
   }
 
+  const renderPageStatus = () => {
+    if (loading) return <LoadingSpinner caption="Loading trustees..." />;
+    if (error) {
+      return (
+        <div className="usa-alert usa-alert--error" role="alert">
+          <div className="usa-alert__body">
+            <h3 className="usa-alert__heading">Error loading trustees</h3>
+            <p className="usa-alert__text">{error}</p>
+          </div>
+        </div>
+      );
+    }
+    return null;
+  };
+
+  const pageStatus = renderPageStatus();
+
   return (
     <div className="trustees-list">
       <TrusteeDistrictFilter
@@ -455,16 +472,7 @@ export default function TrusteesList() {
       <div role="status" aria-live="polite" aria-atomic="true" className="usa-sr-only">
         {liveAnnouncement}
       </div>
-      {loading ? (
-        <LoadingSpinner caption="Loading trustees..." />
-      ) : error ? (
-        <div className="usa-alert usa-alert--error" role="alert">
-          <div className="usa-alert__body">
-            <h3 className="usa-alert__heading">Error loading trustees</h3>
-            <p className="usa-alert__text">{error}</p>
-          </div>
-        </div>
-      ) : (
+      {pageStatus ?? (
         <>
           {filteredTrustees.length === 0 && !nameSearchLoading ? (
             <div className="usa-alert usa-alert--info" role="alert">
--- a/user-interface/src/trustees/TrusteesList.tsx
+++ b/user-interface/src/trustees/TrusteesList.tsx
@@ -433,21 +433,6 @@ export default function TrusteesList() {
     previousNameSearchRef.current = nameSearch;
   }, [nameSearch]);
 
-  if (loading) {
-    return <LoadingSpinner caption="Loading trustees..." />;
-  }
-
-  if (error) {
-    return (
-      <div className="usa-alert usa-alert--error" role="alert">
-        <div className="usa-alert__body">
-          <h3 className="usa-alert__heading">Error loading trustees</h3>
-          <p className="usa-alert__text">{error}</p>
-        </div>
-      </div>
-    );
-  }
-
   let displayCount = filteredTrustees.length;
   if (nameSearchLoading && stableCountRef.current !== null) {
     displayCount = stableCountRef.current;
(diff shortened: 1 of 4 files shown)


user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -470,150 +455,173 @@ export default function TrusteesList() {
...
+      {loading ? (
+        <LoadingSpinner caption="Loading trustees..." />
+      ) : error ? (
+        <div className="usa-alert usa-alert--error" role="alert">
           <div className="usa-alert__body">
-            <h3 className="usa-alert__heading">No trustees found</h3>
-            <p className="usa-alert__text">Consider adjusting your filters.</p>
+            <h3 className="usa-alert__heading">Error loading trustees</h3>
+            <p className="usa-alert__text">{error}</p>
           </div>
```
fmaddenflx: I wonder why we had to alter the message here?  It seems like we should not be doing that as it doesn't comply with design.

fmaddenflx: I see we're doing that below.  Maybe an alert would be better if an error is thrown.  We should check with UX team member.

user-interface/src/trustees/TrusteesListStatusFilter.test.tsx
```diff
@@ -245,4 +245,66 @@ describe('TrusteesList Status Filter', () => {
...
+    });
+    const inactiveChapter7 = makeListItem({
+      trusteeId: 'ic7',
+      firstName: 'Inactive',
+      lastName: 'Seven',
+      name: 'Inactive Seven',
+      appointments: [makeAppointment({ trusteeId: 'ic7', status: 'resigned', chapter: '7' })],
+    });
+    const spy = vi.spyOn(Api2, 'getTrustees');
+    spy.mockResolvedValue({ data: [activeChapter7] });
```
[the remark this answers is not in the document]
bityogi: Thanks — leaving as-is for consistency. The other 5 tests in this file all use the `spy.mockResolvedValue(...)` + reassignment pattern (e.g. lines 99–108, 143–152), and the second `mockResolvedValue` is set before the action that triggers the second call, so call order is deterministic. Switching just this test would create local inconsistency; happy to refactor all 5 in a follow-up if the team prefers `mockResolvedValueOnce` going forward.

user-interface/src/trustees/TrusteesList.tsx
bityogi: No message change here — both alerts are still present and unchanged from `main`:

- `Error loading trustees` (the existing error alert) — currently TrusteesList.tsx:463
- `No trustees found` (the existing empty-state alert) — currently TrusteesList.tsx:472

The diff hunk only *looks* like a replacement because the JSX got restructured into a ternary tree, so GitHub aligned the moved-but-identical empty-state block against the new error block. User-visible copy is unchanged.

bityogi: Good idea, but the inline error alert here is unchanged from `main` — this PR is purely scoped to the status-filter regressions. Switching trustee load errors over to the global Alert system is a worthwhile UX discussion but better as its own ticket so UX can weigh in. I'll file a follow-up for it.

fmaddenflx: Well it's using the usa-alert classes but it's not using the Alert component.  It should use the component.

bityogi: Good call — done in 66005c744. Both blocks now use the `Alert` component (`UswdsAlertStyle.Error` for the load error, `UswdsAlertStyle.Info` for the empty state). All 103 trustees-list tests still pass.

diabagatekelly (approved):

PR #2518, opened 2026-06-16
CAMS-754 Update empty case list alert messaging

jalba-flexion:
# Purpose

Improve the empty-state experience and accessibility of the trustee case list by showing a consistent empty-state message and providing screen-reader-friendly filter announcements.

# Major Changes

* Unified the empty-state alert to always display "No case appointments found" with "Consider adjusting your filters." — removes conditional messaging that varied by filter state
* Removed \`slim\` styling from the alert for better readability
* Added an \`aria-live="polite"\` announcement region in the filter panel so screen readers are notified when status, chapter, or filed date filters change
* Added an \`announce\` helper in the filter use case that uses \`requestAnimationFrame\` to reliably trigger screen reader updates
* Fixed the chapter ComboBox to use an accessible \`aria-label\` instead of custom header markup
* Fixed vertical positioning of the clear-all filters button
* Deleted \`isFilterActive\` from the types file (was unused; knip flagged it on every run)
* Removed \`chapterFilterRef\` (unused ref left over from earlier refactor)

# Testing/Validation

Manual testing confirmed the alert renders correctly in both filtered and unfiltered empty states. Tests extended to cover:
- Initialization from \`initialValue\` for status, chapter, and filed date
- Pill removal resetting the corresponding filters
- Presence and content of the \`aria-live\` announcement region
- Accessible chapter ComboBox naming

# Notes

The \`isFiltered\` variable and \`isFilterActive\` import were removed from \`TrusteeCaseList.tsx\` as they are no longer referenced. \`isFilterActive\` was also deleted from \`trusteeCaseListFilter.types.ts\` — keeping dead exports around violates YAGNI and causes knip to flag them on every run.

The table and pagination are only rendered after data loads (\`!isLoading\`), so they are never in the DOM during a data fetch — no \`aria-busy\` handling is needed.

# Definition of Done:


diabagatekelly:
@sourcery-ai review

diabagatekelly (approved):

mattstankey:
**Issue: PR description is inaccurate about `isFilterActive`**

The description says: _"isFilterActive remains exported from its types file for potential future use (knip flags it as unused)."_

But the diff shows `isFilterActive` was actually deleted from `trusteeCaseListFilter.types.ts` — and that's the right call. Keeping dead exports "for potential future use" is exactly what YAGNI argues against, and deleting it means knip won't flag it on every run. The description should be updated to reflect what the diff actually does.

mattstankey:
**Issue: `aria-live="off"` on input elements and `<table>` is a misuse of the attribute**

In `TrusteeCaseListFilterView.tsx` the date inputs get `aria-live="off"` / `aria-atomic="false"`, and in `TrusteeCaseList.tsx` the same is applied to the `<table>` and the pagination wrapper `<div>`.

`aria-live` is only meaningful on container elements that display dynamically-updating text (e.g. a `<div>` or `<span>` you update via JS). Applying it to an `<input>` has no effect — inputs already have their own AT semantics for value changes. Applying it to a `<table>` or a wrapper `<div>` around a `Pagination` component is similarly a no-op.

If the goal is preventing screen readers from announcing table repaint noise during loading, the standard approach is `aria-busy="true"` on the container while data is loading, then `aria-busy="false"` once it resolves — not `aria-live="off"`.

Recommend removing the `aria-live`/`aria-atomic` attributes from the inputs, table, and pagination wrapper. The `filter-announcement` live region in `TrusteeCaseListFilterView.tsx` is doing the right thing — that pattern is correct and sufficient.

jalba-flexion:
> **Issue: `aria-live="off"` on input elements and `<table>` is a misuse of the attribute**
> 
> In `TrusteeCaseListFilterView.tsx` the date inputs get `aria-live="off"` / `aria-atomic="false"`, and in `TrusteeCaseList.tsx` the same is applied to the `<table>` and the pagination wrapper `<div>`.
> 
> `aria-live` is only meaningful on container elements that display dynamically-updating text (e.g. a `<div>` or `<span>` you update via JS). Applying it to an `<input>` has no effect — inputs already have their own AT semantics for value changes. Applying it to a `<table>` or a wrapper `<div>` around a `Pagination` component is similarly a no-op.
> 
> If the goal is preventing screen readers from announcing table repaint noise during loading, the standard approach is `aria-busy="true"` on the container while data is loading, then `aria-busy="false"` once it resolves — not `aria-live="off"`.
> 
> Recommend removing the `aria-live`/`aria-atomic` attributes from the inputs, table, and pagination wrapper. The `filter-announcement` live region in `TrusteeCaseListFilterView.tsx` is doing the right thing — that pattern is correct and sufficient.

This is needed to stop the screen reader from reading all the content in the table once it has loaded

PR #2532, opened 2026-06-18
CAMS-763 Remove 25-result limit from trustee name search

cruzjone-flexion:
# Purpose

Trustee name search was capped at 25 results, causing users on the TrusteeList page to miss matching trustees when there are more than 25 results. This removes that cap so all matches are returned.

# Major Changes

- Removed `MAX_RESULTS = 25` constant and the `paginate(0, MAX_RESULTS)` stage from the `searchTrusteesByNameScored` pipeline in `TrusteesMongoRepository`
- Added a pipeline stage assertion to the test suite verifying the exact ordered stages: `MATCH, SCORE, MATCH, SORT` (no `PAGINATE` stage)

# Testing/Validation

Implemented with TDD. The new test asserts the full pipeline stage sequence, which would catch any future re-introduction of a pagination limit. Confirmed the existing pre-change test failures were pre-existing infrastructure issues unrelated to these changes.

# Notes

The search pipeline still filters by phonetic token pre-match, applies scoring, filters out zero-score results, and sorts by descending score — it simply no longer caps the result count. Results remain deterministically ordered by match quality.

# Definition of Done:

## Summary by Sourcery

Remove the fixed 25-result cap from trustee name search and ensure the full scored result set is returned while preserving deterministic ordering.

New Features:
- Allow trustee name search to return all scored matches instead of limiting results to 25.

Enhancements:
- Adjust the trustee search query pipeline to omit the pagination stage and use a virtually unbounded limit in the repository adapter call.

Tests:
- Add a unit test that asserts the trustee search pipeline stages are exactly MATCH, SCORE, MATCH, SORT and that pagination uses an effectively unlimited result cap.


backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -1392,26 +1392,26 @@ describe('TrusteesMongoRepository', () => {
...
+        .spyOn(MongoCollectionAdapter.prototype, 'aggregate')
+        .mockResolvedValue(mockTrustees as TrusteeDocument[]);
 
       await repository.searchTrusteesByNameScored('smith');
 
-      expect(paginateSpy).toHaveBeenCalledTimes(1);
-      const pipelineArg = paginateSpy.mock.calls[0][0] as { stages: { stage: string }[] };
+      expect(aggregateSpy).toHaveBeenCalledTimes(1);
+      const pipelineArg = aggregateSpy.mock.calls[0][0] as { stages: { stage: string }[] };
       expect(pipelineArg.stages.some((s) => s.stage === 'SCORE')).toBe(true);
```
diabagatekelly: **HIGH** — This assertion only confirms a SCORE stage exists somewhere in the pipeline, but the pipeline-shape test on line 1507 already covers that more precisely. Either remove this test (it is fully subsumed by the shape test), or replace the assertion with a content check on the SCORE stage:

```typescript
const scoreStage = pipelineArg.stages.find((s) => s.stage === 'SCORE');
expect(scoreStage).toBeDefined();
expect(scoreStage.searchWords).toEqual(expect.arrayContaining(['smith']));
expect(scoreStage.targetNameFields).toEqual(['name']);
```

As-is, a mutation that puts a junk SCORE stage anywhere in the pipeline would pass this test.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -1454,25 +1454,24 @@ describe('TrusteesMongoRepository', () => {
...
 
       // "mike" expands to nickname words including "michael"
       await repository.searchTrusteesByNameScored('mike');
 
-      const pipelineArg = paginateSpy.mock.calls[0][0] as { stages: Score[] };
+      const pipelineArg = aggregateSpy.mock.calls[0][0] as { stages: Score[] };
       const scoreStage = pipelineArg.stages.find((s) => s.stage === 'SCORE');
       expect(scoreStage).toBeDefined();
       expect(scoreStage.nicknameWords.length).toBeGreaterThan(0);
       expect(scoreStage.nicknameMetaphones.length).toBeGreaterThan(0);
```
diabagatekelly: **HIGH** — `toBeGreaterThan(0)` does not verify the nickname expansion is correct — any non-empty array passes. For input `"mike"`, the expansion is deterministic, so assert specific values:

```typescript
expect(scoreStage.nicknameWords).toEqual(expect.arrayContaining(['michael']));
expect(scoreStage.nicknameMetaphones.length).toBeGreaterThan(0); // acceptable if metaphone codes are not exported
```

A mutation that populated these arrays with garbage strings would survive the current assertions.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -1481,27 +1480,35 @@ describe('TrusteesMongoRepository', () => {
...
 
+    test('should build pipeline with exactly MATCH, SCORE, MATCH, SORT stages', async () => {
+      const aggregateSpy = vi
+        .spyOn(MongoCollectionAdapter.prototype, 'aggregate')
+        .mockResolvedValue([]);
+
+      await repository.searchTrusteesByNameScored('smith');
+
+      const pipelineArg = aggregateSpy.mock.calls[0][0] as { stages: { stage: string }[] };
+      expect(pipelineArg.stages.map((s) => s.stage)).toEqual(['MATCH', 'SCORE', 'MATCH', 'SORT']);
```
diabagatekelly, line 1443: **MEDIUM** — The shape test asserts stage names but not sort direction. A mutation reversing the sort from descending to ascending on `matchScore` would return the worst matches first and still pass. Consider adding a direction/field assertion — something like:

```typescript
const sortStage = pipelineArg.stages[3] as { stage: string; direction: string; field: { name: string } };
expect(sortStage.direction).toBe('DESCENDING');
expect(sortStage.field.name).toBe('matchScore');
```

*(Verify the actual shape of the `descending()` stage output to get the right field names.)*

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -1392,26 +1392,26 @@ describe('TrusteesMongoRepository', () => {
...
-        .spyOn(MongoCollectionAdapter.prototype, 'paginate')
-        .mockResolvedValue({ metadata: { total: 0 }, data: [] });
+      const aggregateSpy = vi
+        .spyOn(MongoCollectionAdapter.prototype, 'aggregate')
+        .mockResolvedValue([]);
 
       await repository.searchTrusteesByNameScored('smith');
 
-      const pipelineArg = paginateSpy.mock.calls[0][0] as {
+      const pipelineArg = aggregateSpy.mock.calls[0][0] as {
```
diabagatekelly, line 1359: **MEDIUM** — The CONTAINS assertion a few lines below (checking that a `phoneticTokens` CONTAINS condition exists) verifies structure but not the token values in `rightOperand`. If `combinePhoneticTokens` returned wrong or empty tokens, that test would still pass. The tokens for "smith" are deterministic — suggest adding: `expect(containsCondition.rightOperand).toEqual(expect.arrayContaining(["SM0"]))` (verify "SM0" is the correct Metaphone code for "smith", or call the token helper directly to get the expected value).

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -1481,27 +1480,35 @@ describe('TrusteesMongoRepository', () => {
     });
```
diabagatekelly, line 1415: **MEDIUM** — The test above ("should return matching trustees from aggregate result") only verifies mock pass-through: the spy returns mockTrustees, and the assertion confirms the repository returned mockTrustees. Since there is no transformation between the aggregate result and return value, this test cannot meaningfully fail. The toHaveLength(2) is also structurally guaranteed by the mock setup. The empty-data test below this covers the same contract more usefully. Consider removing the pass-through test.

diabagatekelly (approved):

jamesobrooks (approved):
**Approving.** The change cleanly accomplishes CAMS-763 — the 25-result cap is gone and all scored matches now flow through to both consumers (`TrusteesList` and `TrusteeSearchModal`). I traced the data path end-to-end: the interface contract is unchanged, the scored search path is injection-safe, the endpoint stays `DataVerifier`-gated, and the new exact-stage-sequence test (`['MATCH','SCORE','MATCH','SORT']`) is a well-aimed regression guard against re-introducing a limit stage. Nice, focused diff.

**One observation for discussion:** with the cap removed, `TrusteeSearchUseCase.searchTrustees` (`trustee-search.use-case.ts:23`) fans out one `getTrusteeAppointments` query per matched trustee inside `Promise.all` — an unbounded N+1. The batch method `getAppointmentsByTrusteeIds(ids)` already exists (`trustee-appointments.mongo.repository.ts:106`, already used in `trustees.ts:216`) and would collapse this to a single appointments query + in-memory grouping (and let us drop the manual `trusteeOrder` re-sort). Three reasonable ways to play it:

- **Fix it here** — small change, and it lives adjacent to what this PR is about.
- **File a follow-up** — keep this PR's diff tight and address it separately.
- **Leave it** — with ~6k trustees total and a name search realistically returning at most tens of matches, the N+1 has no practical cost, so treating this as gold-plating and moving on is entirely defensible.

No preference from me on which — flagging it so the call is deliberate rather than implicit.

PR #2537, opened 2026-06-19
CAMS-761 Assign ZZ-NNNNN professional code to new trustees

mattstankey:
# Purpose

Trustees created via the public form now receive a sequential `acmsProfessionalId` in the format `ZZ-NNNNN`. Without this code, the existing ACMS-CAMS-TRANSITION downstream pipeline skips newly created trustees entirely.

# Major Changes

* Added `findOneAndUpdate` to `CollectionHumble` and `MongoCollectionAdapter` to support atomic read-modify-write operations
* Added `atomicDecrement` to `RuntimeStateRepository` using a seed-then-`$inc` pair (Mongo rejects `$inc` and `$setOnInsert` on the same path in a single update)
* Added `PROFESSIONAL_ID_COUNTER` document type and `ProfessionalIdCounterState` to the runtime state schema
* `TrusteesUseCase.createTrustee` now calls a private `assignProfessionalCode` helper that decrements the counter, formats the code as `ZZ-NNNNN` (zero-padded 5 digits), writes a `TrusteeProfessionalId` mapping record, and writes an `AUDIT_PROFESSIONAL_ID_ASSIGNED` history entry — all authored by `SYSTEM_USER`
* Extended the `TrusteeHistory` union in `common` with `TrusteeProfessionalIdHistory`
* Fixed test quality issues surfaced by spec review: moved `vi.restoreAllMocks()` from `afterEach` to `beforeEach` across affected test files, added missing coverage for `upsertOne`/`updateMany`/`bulkReplace`, added error-path tests for `completeMigration`/`failMigration`, consolidated redundant zoom meeting-ID tests, and added `resolveSoftwareName` fallback tests

# Testing/Validation

Implemented using TDD. Unit tests cover `createTrustee` professional ID assignment, `atomicDecrement` success and error paths, `findOneAndUpdate` adapter behavior, and all newly identified coverage gaps from spec review. All 1090 unit and BDD tests pass.

# Notes

Migration-created trustees keep their existing ACMS group codes (e.g., `NY-`, `WA-`); the migration dataflow is unchanged. Downstream `sync-trustee-appointments` and `acms-cams-transition` handlers are agnostic to the group designator and require no changes.

The seed-then-`$inc` split in `atomicDecrement` works around a MongoDB/Cosmos limitation: `$inc` and `$setOnInsert` cannot target the same path in a single `findOneAndUpdate` operation.

# Definition of Done:

### DIFF
--- a/backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
@@ -8,7 +8,7 @@ import QueryBuilder from '../../../query/query-builder';
 import { getCamsError } from '../../../common-errors/error-utilities';
 import { UnknownError } from '../../../common-errors/unknown-error';
 import { BaseMongoRepository } from './utils/base-mongo-repository';
-import * as crypto from 'crypto';
+import { randomUUID } from 'crypto';
 
 const MODULE_NAME = 'RUNTIME-STATE-MONGO-REPOSITORY';
 const COLLECTION_NAME = 'runtime-state';
@@ -64,7 +64,7 @@ export class RuntimeStateMongoRepository<T extends RuntimeState>
         query,
         {
           $set: { documentType },
-          $setOnInsert: { [field]: initialValue, id: crypto.randomUUID() },
+          $setOnInsert: { [field]: initialValue, id: randomUUID() },
         },
         { upsert: true },
       );
@@ -79,7 +79,13 @@ export class RuntimeStateMongoRepository<T extends RuntimeState>
           message: `atomicDecrement returned no document for ${documentType}.`,
         });
       }
-      return result[field] as number;
+      const value = result[field];
+      if (typeof value !== 'number') {
+        throw new UnknownError(MODULE_NAME, {
+          message: `atomicDecrement: field '${field}' is not a number in ${documentType}.`,
+        });
+      }
+      return value;
     } catch (e) {
       throw getCamsError(e, MODULE_NAME);
     }
--- a/backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
@@ -6,7 +6,9 @@ import {
 } from '../../../use-cases/gateways.types';
 import QueryBuilder from '../../../query/query-builder';
 import { getCamsError } from '../../../common-errors/error-utilities';
+import { UnknownError } from '../../../common-errors/unknown-error';
 import { BaseMongoRepository } from './utils/base-mongo-repository';
+import * as crypto from 'crypto';
 
 const MODULE_NAME = 'RUNTIME-STATE-MONGO-REPOSITORY';
 const COLLECTION_NAME = 'runtime-state';
@@ -39,6 +41,45 @@ export class RuntimeStateMongoRepository<T extends RuntimeState>
       if (result.modifiedCount + result.upsertedCount > 0) {
         return { ...data, id: result.id } as T;
       }
+    } catch (e) {
+      throw getCamsError(e, MODULE_NAME);
+    }
+  }
+
+  async atomicDecrement(
+    documentType: RuntimeStateDocumentType,
+    field: keyof T & string,
+    initialValue: number,
+  ): Promise<number> {
+    try {
+      const adapter = this.getAdapter<T>();
+      const query = doc('documentType').equals(documentType);
+
+      // Seed the counter on first use. Mongo rejects $inc and $setOnInsert on
+      // the same path in one update, so we seed in a separate upsert and then
+      // do the atomic $inc. Both round-trips are race-safe at the document
+      // level; a concurrent caller either finds the seeded doc or seeds it
+      // itself, and the subsequent $inc is always atomic.
+      await adapter.findOneAndUpdate(
+        query,
+        {
+          $set: { documentType },
+          $setOnInsert: { [field]: initialValue, id: crypto.randomUUID() },
+        },
+        { upsert: true },
+      );
+
+      const result = await adapter.findOneAndUpdate(
+        query,
+        { $inc: { [field]: -1 } },
+        { returnDocument: 'after' },
+      );
+      if (!result) {
+        throw new UnknownError(MODULE_NAME, {
+          message: `atomicDecrement returned no document for ${documentType}.`,
+        });
+      }
+      return result[field] as number;
     } catch (e) {
       throw getCamsError(e, MODULE_NAME);
     }
--- a/backend/lib/adapters/gateways/mongo/utils/mongo-adapter.test.ts
+++ b/backend/lib/adapters/gateways/mongo/utils/mongo-adapter.test.ts
@@ -56,6 +56,8 @@ describe('Mongo adapter', () => {
 
   beforeEach(() => {
     vi.restoreAllMocks();
+    // adapter and spies are constructed at module scope, so restoreAllMocks() alone
+    // does not reset call history between tests — clearAllMocks() handles that.
     vi.clearAllMocks();
   });
 
--- a/backend/lib/adapters/gateways/mongo/utils/mongo-adapter.test.ts
+++ b/backend/lib/adapters/gateways/mongo/utils/mongo-adapter.test.ts
@@ -13,7 +13,6 @@ const ADAPTER_MODULE_NAME = MODULE_NAME + '_ADAPTER';
 
 const find = vi.fn();
 const findOne = vi.fn();
-const paginate = vi.fn();
 const replaceOne = vi.fn();
 const updateOne = vi.fn();
 const findOneAndUpdate = vi.fn();
@@ -23,11 +22,13 @@ const deleteOne = vi.fn();
 const deleteMany = vi.fn();
 const countDocuments = vi.fn();
 const aggregate = vi.fn();
+const upsertOne = vi.fn();
+const updateMany = vi.fn();
+const bulkWrite = vi.fn();
 
 const spies = {
   find,
   findOne,
-  paginate,
   replaceOne,
   updateOne,
   findOneAndUpdate,
@@ -37,6 +38,9 @@ const spies = {
   deleteMany,
   countDocuments,
   aggregate,
+  upsertOne,
+  updateMany,
+  bulkWrite,
 };
 
 type TestType = {
@@ -50,8 +54,9 @@ describe('Mongo adapter', () => {
   const humbleCollection = spies as unknown as CollectionHumble<TestType>;
   const adapter = new MongoCollectionAdapter<TestType>(MODULE_NAME, humbleCollection);
 
-  afterEach(() => {
+  beforeEach(() => {
     vi.restoreAllMocks();
+    vi.clearAllMocks();
   });
 
   test('should return an instance of the adapter from newAdapter', () => {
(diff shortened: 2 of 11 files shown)


backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
```diff
@@ -43,4 +45,43 @@ export class RuntimeStateMongoRepository<T extends RuntimeState>
...
+      await adapter.findOneAndUpdate(
+        query,
+        {
+          $set: { documentType },
+          $setOnInsert: { [field]: initialValue, id: crypto.randomUUID() },
+        },
+        { upsert: true },
+      );
+
+      const result = await adapter.findOneAndUpdate(
```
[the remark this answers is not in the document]
mattstankey: Not addressing this one. MongoDB upsert with `{ upsert: true }` is atomic at the document level — concurrent first-time callers cannot both insert. One caller wins the race and seeds the document; the other finds the already-seeded document and proceeds to the `$inc` step. No duplicate counter documents can be created.

A unique index on `documentType` would be belt-and-suspenders but belongs in an infrastructure/provisioning task, not here.

mattstankey:
Skipping both overall suggestions:

**Partial write in `createTrustee`:** The error paths are tested and throw a `CamsError` if either step fails — the caller sees the failure and the form can be re-submitted. A compensation/retry strategy would add significant complexity for an infrequent edge case. If the team wants to revisit this, it's a follow-up task.

**`crypto.randomUUID()` in `atomicDecrement`:** `randomUUID()` is already called directly in `BaseMongoRepository`'s `createOrGetId` and throughout the adapter layer — this is the existing convention, not a deviation from it.

backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
```diff
@@ -43,4 +45,43 @@ export class RuntimeStateMongoRepository<T extends RuntimeState>
...
+        throw new UnknownError(MODULE_NAME, {
+          message: `atomicDecrement returned no document for ${documentType}.`,
+        });
+      }
+      return result[field] as number;
+    } catch (e) {
+      throw getCamsError(e, MODULE_NAME);
+    }
+  }
 }
```
diabagatekelly, line 93: **MEDIUM** — Unsafe cast on the critical return path.

`result[field]` is typed as `unknown` (dynamic key on a generic `T`). If the MongoDB driver ever returns a non-number here (schema drift, null, string), this cast silently lets the bad value flow into `formatProfessionalCode`, producing `ZZ-NaN` or similar invalid codes rather than failing loudly.

Suggested fix:
```typescript
const value = result[field];
if (typeof value !== 'number') {
  throw new UnknownError(MODULE_NAME, {
    message: `atomicDecrement: field '${field}' is not a number in ${documentType}.`,
  });
}
return value;
```

backend/lib/use-cases/trustees/trustees.test.ts
```diff
@@ -21,14 +21,11 @@ describe('TrusteesUseCase tests', () => {
...
   describe('createTrustee', () => {
     beforeEach(async () => {
+      vi.restoreAllMocks();
       context = await createMockApplicationContext();
       trusteesUseCase = new TrusteesUseCase(context);
     });
 
-    afterEach(() => {
-      vi.restoreAllMocks();
-    });
```
diabagatekelly, line 30: **MEDIUM** — `atomicDecrementSpy` is spied but not given an explicit mock return value, so the happy-path assertion on line 77 (`'ZZ-99999'`) silently depends on `MockMongoRepository`'s default implementation returning `99999`. If that default ever changes, this test breaks without a clear signal.

Compare to the zero-pad test below, which correctly uses `.mockResolvedValue(9999)`. Suggest making intent explicit here:
```typescript
const atomicDecrementSpy = vi
  .spyOn(MockMongoRepository.prototype, 'atomicDecrement')
  .mockResolvedValue(99999);
```

backend/lib/adapters/gateways/mongo/utils/mongo-adapter.test.ts
```diff
@@ -13,28 +13,34 @@ const ADAPTER_MODULE_NAME = MODULE_NAME + '_ADAPTER';
 
 const find = vi.fn();
 const findOne = vi.fn();
-const paginate = vi.fn();
 const replaceOne = vi.fn();
 const updateOne = vi.fn();
+const findOneAndUpdate = vi.fn();
 const insertOne = vi.fn();
```
diabagatekelly, line 19: **MEDIUM** — `vi.clearAllMocks()` looks redundant next to `vi.restoreAllMocks()`, but it's actually load-bearing here. Because the adapter and all `vi.fn()` mocks are constructed at module scope (lines 14–55), `vi.restoreAllMocks()` does not reset them — it only restores `vi.spyOn` stubs. `vi.clearAllMocks()` is what actually resets call history and mock implementations between tests. Removing it would cause test leakage.

The safest fix is moving the adapter construction into `beforeEach` so `restoreAllMocks()` alone suffices:
```typescript
let adapter: MongoCollectionAdapter<TestType>;

beforeEach(() => {
  vi.restoreAllMocks();
  const humbleCollection = { find: vi.fn(), /* ... */ } as unknown as CollectionHumble<TestType>;
  adapter = new MongoCollectionAdapter<TestType>(MODULE_NAME, humbleCollection);
});
```
If moving construction is too disruptive, at minimum add a comment explaining why both calls are needed so a future reader doesn't remove the "redundant" one.

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -49,6 +53,17 @@ import { BankruptcySoftwareProfile } from '@common/cams/bankruptcy-software';
 
 const MODULE_NAME = 'TRUSTEES-USE-CASE';
```
diabagatekelly, line 56: **LOW (not critical)** — `PROFESSIONAL_ID_COUNTER_INITIAL` is private to this module, so tests hardcode `100000` (e.g., `trustees.test.ts` line 75). Exporting this constant lets tests reference it by name, keeping them in sync automatically if the seed value ever changes:
```typescript
export const PROFESSIONAL_ID_COUNTER_INITIAL = 100000;
```

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -94,13 +109,18 @@ export class TrusteesUseCase {
```
diabagatekelly, line 93: **LOW (not critical)** — No guard when the counter dips below 1. At 0 this produces `ZZ-00000`; below 0 it produces `ZZ--001` and similar — both invalid ACMS codes. Given the seed is 100,000 this is unlikely in practice, but a loud failure beats silently assigning malformed codes:
```typescript
if (codeNumber < 1) {
  throw new UnknownError(MODULE_NAME, {
    message: `Professional ID counter exhausted (value: ${codeNumber}). Cannot assign a valid ZZ-NNNNN code.`,
  });
}
```

diabagatekelly (commented):
## PR Review — CAMS-761

Overall this is a clean, well-structured feature. The two-phase atomic counter (seed → `$inc`) correctly works around the Cosmos DB limitation, architecture boundaries are respected, and the test suite covers the critical paths well. A few things to address below — one correctness risk on the return path and two test hygiene items worth fixing before merge.

**Question for author:** If `assignProfessionalCode` successfully decrements the counter but then `createProfessionalId` or `createTrusteeHistory` throws (e.g., a transient DB error), the trustee record exists but has no professional ID, and that counter slot is consumed. Is there a known compensating path for this partial-failure scenario — e.g., an admin assignment flow, a background reconciliation job, or a retry that re-uses the already-decremented value? Even if the accepted answer is "acceptable gap given low probability," it's worth documenting somewhere (a beads issue or a comment on `assignProfessionalCode`).

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
```diff
@@ -67,6 +67,19 @@ export class MockMongoRepository
     UserGroupsRepository
 {
   private professionalIds = new Map<string, TrusteeProfessionalId>();
+  private runtimeStateCounters = new Map<string, number>();
+
+  async atomicDecrement(
+    documentType: string,
+    field: string,
```
diabagatekelly, line 77: **LOW (not critical)** — The real `atomicDecrement` does two round-trips (seed via `$setOnInsert`, then `$inc`). This mock correctly collapses that into one call, returning `initialValue - 1` to match the first value the real implementation produces. Worth a brief comment so future maintainers understand the semantic:
```typescript
// Collapses the real two-phase seed+decrement into a single call.
// Returns initialValue - 1 on first use to match the first value the real
// implementation produces after seeding at initialValue then decrementing.
```

backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
```diff
@@ -43,4 +45,43 @@ export class RuntimeStateMongoRepository<T extends RuntimeState>
       throw getCamsError(e, MODULE_NAME);
```
diabagatekelly, line 45: **INFO** — Minor style inconsistency: the rest of the codebase uses named imports from `'crypto'` (e.g., `mongo-adapter.ts` uses `import { randomUUID } from 'crypto'`). Suggest aligning:
```typescript
import { randomUUID } from 'crypto';
```
Then update usage at line 67 to `randomUUID()` directly.

backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
mattstankey: ✅ Fixed. Added a type guard that throws `UnknownError` if `result[field]` is not a number:

```typescript
const value = result[field];
if (typeof value !== 'number') {
  throw new UnknownError(MODULE_NAME, {
    message: `atomicDecrement: field '${field}' is not a number in ${documentType}.`,
  });
}
return value;
```

backend/lib/use-cases/trustees/trustees.test.ts
mattstankey: ✅ Fixed. Added `.mockResolvedValue(99999)` to make the intent explicit, matching the pattern used in the zero-pad test below.

backend/lib/adapters/gateways/mongo/utils/mongo-adapter.test.ts
mattstankey: ✅ Added a comment explaining why both calls are needed. Agreed the refactor into `beforeEach` would be the cleaner long-term fix but is out of scope here.

backend/lib/use-cases/trustees/trustees.ts
mattstankey: Skipping this one — the constant is unlikely to change and exporting it to satisfy tests would leak an implementation detail. Agreed it keeps tests in sync, but the tradeoff feels like over-engineering for this case.

backend/lib/use-cases/trustees/trustees.ts
mattstankey: ✅ Fixed. Added a guard that throws `UnknownError` when `codeNumber < 1`, so exhausted or negative counter values fail loudly rather than producing `ZZ-00000` or `ZZ--001`.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
mattstankey: ✅ Added a comment explaining the semantic collapse.

backend/lib/adapters/gateways/mongo/runtime-state.mongo.repository.ts
mattstankey: ✅ Fixed. Switched to `import { randomUUID } from 'crypto'` to match the pattern used elsewhere in the adapter layer.

mattstankey:
Good question on the partial-failure scenario. There is no compensating path — no retry, no reconciliation job, and no admin assignment flow for this case. The accepted behavior is: if `atomicDecrement` succeeds but `createProfessionalId` or `createTrusteeHistory` throws, the trustee record exists with no professional ID, and that counter slot is consumed. The caller sees a `CamsError` and the form can be re-submitted, which will attempt to assign a new (lower-numbered) professional ID on the next successful run.

This is an accepted gap given the low probability of a transient failure at exactly that point. If the team wants a reconciliation job or idempotent retry path in the future, that can be a follow-up task.

btposey (approved):

diabagatekelly (approved):

PR #2540, opened 2026-06-22
CAMS-768 Trustee change notification pipeline (main feature branch)

cruzjone-flexion:
# Purpose

Wire up an end-to-end notification pipeline so that saving a trustee profile change produces a recorded email notification routed to the appropriate oversight office. This closes Slice 1 of CAMS-768 and establishes the DI seams, routing repository, template compiler, and orchestrator that later slices will build on.

# Major Changes

* Added `NotificationType`, `Notification`, `NotificationRoutingRecord`, and `TrusteeChangeSet` types to `common/src/cams/notifications.ts`
* Implemented `NotificationRoutingMongoRepository` — resolves recipients by routing key (`chapter:N`, `default`) with singleton lifecycle matching other Mongo repositories
* Implemented `MockNotificationGateway` — records dispatched notifications in memory for test assertions; registered in the factory
* Built `TrusteeChangeSet` builder in `TrusteeChangeNotificationUseCase` — diffing before/after trustee profiles, emitting one `ChangeField` per changed label
* Implemented `compileTrusteeChangeTemplate` — renders an HTML + plain-text email from a `TrusteeChangeSet` with section grouping (appointment vs. 341 meeting fields)
* Implemented `TrusteeChangeNotificationUseCase.notify()` — resolves recipients, compiles the template once, dispatches one `Notification` per unique recipient; warn-logs missing routing keys, never throws on routing misses
* Hooked dispatch into `updateTrustee` in `TrusteesUseCase` — wrapped in a failure-isolated try/catch (Decision 5); dispatch failure logs and traces but does not roll back the save
* Added `notification-routing` dev-tools seed scenario — seeds chapter:7 and default routing rows idempotently via `mongoUpsert`

# Testing/Validation

Implemented using TDD across all seven tasks. All four Gherkin acceptance scenarios from `trustee-change-notifications.feature` pass observably:

- Profile change dispatches one chapter:7 notification to the correct recipient
- Multi-field save produces a single notification whose body contains every changed label
- Dispatch failure does not roll back the trustee save
- No-change save does not dispatch

2894/2894 backend tests pass; 48 existing `trustees.test.ts` tests are unchanged; zero regressions.

# Notes

`resolvePrimaryChapter` and `TrusteeChangeNotificationUseCase` construction live inside the dispatch try/catch boundary (not before it) so that one failure cannot surface as an unhandled exception in `updateTrustee`. This deviates slightly from the Slice 1 prompt, which placed `resolvePrimaryChapter` before the try/catch — doing so would have required every existing `updateTrustee` test to mock `getAppointmentsByTrusteeIds`, which the prompt explicitly forbids.

# Definition of Done:

## Summary by Sourcery

Introduce an end-to-end trustee change email notification pipeline triggered from trustee updates, including routing, templating, and failure-isolated dispatch.

New Features:
- Add notification domain types for trustee change sets and notification payloads.
- Add a trustee change notification use case that resolves recipients via routing rules, compiles email templates, and dispatches notifications via a notification gateway.
- Trigger trustee change notifications from trustee updates when profile or meeting-related fields change.

Enhancements:
- Extend trustee update auditing to build structured change sets summarizing before/after values for key trustee profile and meeting fields.
- Derive a trustee’s primary chapter from appointments to drive chapter-based notification routing.
- Provide a Mongo-backed notification routing repository and a mock notification gateway wired through the application factory for test and dev use.
- Add HTML and plaintext trustee change email templates with sectioned rendering and safe escaping, plus helper formatting for contact and Zoom info.

Build:
- Add a dev-tools seed scenario to populate initial notification routing records for chapter 7 and the default recipient.

Tests:
- Add unit tests for trustee change template compilation, notification routing and dispatch behavior, notification routing Mongo repository, mock notification gateway, and the trustee change notification flow from updateTrustee, including failure isolation and no-op scenarios.


backend/lib/factory.ts
```diff
@@ -511,6 +516,25 @@ const getListsGateway = (context: ApplicationContext): ListsRepository => {
...
+  deferRelease(repo, context);
+  return repo;
+};
+
+const getNotificationGateway = (_context: ApplicationContext): NotificationGateway => {
+  // Slice 5 will replace this with a provider-selection switch.
+  if (!notificationGateway) {
+    notificationGateway = MockNotificationGateway.getInstance();
+  }
+  return notificationGateway;
```
diabagatekelly, line 561: **🚨 BLOCKER — Mock gateway wired in all environments**

Every other factory method guards on `context.config.get('dbMock')`. This returns `MockNotificationGateway` unconditionally — in staging/production, oversight notifications are silently dropped with no runtime signal.

Until Slice 5 wires a real provider, throw explicitly so the misconfiguration is visible:

```typescript
const getNotificationGateway = (context: ApplicationContext): NotificationGateway => {
  if (!notificationGateway) {
      notificationGateway = MockNotificationGateway.getInstance();
    } else {
      throw new ServerConfigError('FACTORY', {
        message: 'Notification gateway is not configured. Slice 5 must wire a real provider.',
      });
    }
  }
  return notificationGateway;
};
```

backend/lib/factory.ts
```diff
@@ -106,6 +110,7 @@ let objectStorageGateway: ObjectStorageGateway;
 let acmsGateway: AcmsGateway;
 let atsGateway: AtsGateway;
 let idpApiGateway: UserGroupGateway & Initializer<UserGroupGatewayConfig | ApplicationContext>;
+let notificationGateway: NotificationGateway | null = null;
```
diabagatekelly: **MEDIUM — `null` where every other gateway variable uses `undefined`**

All other module-level gateway variables are declared without initialisation (implicitly `undefined`). The `| null` union misleads readers into thinking `null` is a meaningful state distinct from `undefined`.

```typescript
// Before
let notificationGateway: NotificationGateway | null = null;

// After
let notificationGateway: NotificationGateway | undefined;
```

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -492,7 +554,38 @@ export class TrusteesUseCase {
...
+    const appointments = await this.trusteeAppointmentsRepository.getAppointmentsByTrusteeIds([
+      trusteeId,
+    ]);
+    if (appointments.length === 0) return undefined;
+
+    const sorted = [...appointments].sort((a, b) => {
+      const aActive = a.status === 'active' ? 1 : 0;
+      const bActive = b.status === 'active' ? 1 : 0;
+      if (aActive !== bActive) return bActive - aActive;
+      return b.appointedDate.localeCompare(a.appointedDate);
```
diabagatekelly, line 635: **HIGH — Crash on missing `appointedDate`**

`.localeCompare()` throws `TypeError` if `appointedDate` is `null` or `undefined` (possible with legacy DB rows). The exception is caught by the notification try/catch in `updateTrustee`, which silently routes every subsequent notification to the default recipient with no diagnostic.

```typescript
// Before
return b.appointedDate.localeCompare(a.appointedDate);

// After
return (b.appointedDate ?? '').localeCompare(a.appointedDate ?? '');
```

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -492,7 +554,38 @@ export class TrusteesUseCase {
...
+    });
+
+    return sorted[0].chapter;
```
diabagatekelly, line 638: **HIGH — `chapter` from DB not validated against the routing key union**

`TrusteeAppointment.chapter` comes from the database and could be any string (e.g. `'chapter-7'` from a legacy row). Assigned directly to `changeSet.primaryChapter`, it produces a routing key like `chapter:chapter-7` that never matches any rule, silently falling back to the default recipient on every save — with no warning log.

```typescript
// Before
return sorted[0].chapter;

// After
const VALID_CHAPTERS = new Set(['7', '11', '11-subchapter-v', '12', '13']);
const chapter = sorted[0].chapter;
return VALID_CHAPTERS.has(chapter)
  ? (chapter as TrusteeChangeSet['primaryChapter'])
  : undefined;
```

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -351,6 +354,28 @@ export class TrusteesUseCase {
...
       );
 
+      if (changeSet.fields.length > 0) {
+        try {
+          changeSet.primaryChapter = await this.resolvePrimaryChapter(trusteeId);
+          const notificationUseCase = new TrusteeChangeNotificationUseCase(context);
+          await notificationUseCase.notify(context, changeSet);
+        } catch (originalError) {
+          context.logger.error(
+            'TRUSTEE-CHANGE-NOTIFICATION',
```
diabagatekelly: **HIGH — Hardcoded module name string; silently drifts if child use case is renamed**

`TrusteeChangeNotificationUseCase` owns `MODULE_NAME = 'TRUSTEE-CHANGE-NOTIFICATION'` (unexported), so this string is duplicated with no compile-time link. If the child module is ever renamed, this log line stays wrong forever.

Simplest fix — use the local `MODULE_NAME`; the message already identifies the failure context:

```typescript
context.logger.error(
  MODULE_NAME,
  'Failed to dispatch trustee change notification.',
  originalError,
);
```

diabagatekelly (commented):
## Review Summary

Architecture, domain model, and TDD discipline are solid overall. One BLOCKER must be resolved before merge; several security issues are currently dormant but become exploitable the moment Slice 5 wires a real SMTP gateway. Inline comments cover all BLOCKER and MEDIUM findings.

**Scores:** Code Quality 5.5/10 · Test Quality 6/10 · Architecture 7/10

---

## Questions for the author

1. **`getNotificationGateway` (BLOCKER, `factory.ts:530`):** Was the missing `dbMock` guard intentional — i.e., Slice 1 is deliberately non-operational in production until Slice 5 wires a real provider? If so, an explicit `throw` is safer than silently returning a mock that swallows all emails. If not, what should the non-mock fallback be?

2. **Empty section headers (`trustee-change.template.ts`):** When a trustee has only Zoom/meeting changes (no profile changes), the email renders an ‘Appointment Information’ heading with column headers but no data rows. Is this the intended UX for Slice 1, or should empty sections be suppressed?

3. **`stackValues` field (`notifications.ts:39`):** Is `stackValues` planned for a specific future slice (e.g., Slice 3 division data)? A JSDoc note on the field would clarify it’s not dead code to future reviewers.

4. **`MockNotificationGateway` location (`adapters/gateways/notifications/`):** Is this placement intentional (it will be replaced in-place by a real gateway in Slice 5), or should it live under `testing/mock-gateways/` alongside `MockMongoRepository`? Its current location ships it as production code, and the singleton accumulates PII (`to`, `html`, `text` containing trustee contact data) for the process lifetime.

---

## Missing acceptance scenario (Gherkin Scenario 3)

**Scenario 3 — ‘Dispatch failure does not roll back the trustee save’** has no test in `trustee-change-notification.test.ts` that exercises a `send()` rejection. The isolation is validated only from `trustees.test.ts`. Consider adding:

```typescript
test('resolves without throwing when gateway.send rejects', async () => {
  seedRouting([CHAPTER_7_RECIPIENT, DEFAULT_RECIPIENT]);
  vi.spyOn(MockNotificationGateway.prototype, 'send')
    .mockRejectedValue(new Error('SMTP connection refused'));

  await expect(useCase.notify(context, buildChangeSet([buildField()]))).resolves.toBeUndefined();
});
```

---

## HIGH issues without an inline comment location

- **`trusteeName` CRLF injection in plaintext body** (`trustee-change-template.ts:69`): The subject line fix is inline-commented at line 116, but `buildPlaintext` line 69 has the same exposure (`Trustee ${changeSet.trusteeName}'s information has changed.`). Apply `.replace(/[\r\n]/g, ' ')` there too.

- **`sorted[0].chapter` type validation** (`trustees.ts:588`, inline comment on that line): Also noted inline — a DB value outside the `primaryChapter` union produces a routing key that never matches, silently falling back to default on every save with no warning.

- **`MockNotificationGateway` in production bundle path**: See question 4 above.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  const rendered = TRUSTEE_CHANGE_TEMPLATE.replace(
+    '{{trustee_name}}',
+    escapeHtml(changeSet.trusteeName),
+  )
+    .replace('{{appointment_info_rows}}', appointmentRows)
+    .replace('{{meeting_info_rows}}', meetingRows);
+
+  return {
+    subject: `Trustee Information Changed: ${changeSet.trusteeName}`,
```
diabagatekelly: **HIGH — Email header injection surface in subject line (dormant until Slice 5)**

`trusteeName` is correctly HTML-escaped in the body, but the subject line uses it raw. A name containing `\r\n` can inject arbitrary MIME headers when Slice 5 wires a real SMTP gateway. Fix now so the real gateway lands on a clean foundation:

```typescript
subject: `Trustee Information Changed: ${changeSet.trusteeName.replace(/[\r\n]/g, ' ')}`,
```

Same sanitisation needed at `buildPlaintext` line 69: `Trustee ${changeSet.trusteeName}'s information has changed.`

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
```diff
@@ -0,0 +1,120 @@
...
+    .split('\n')
+    .map((line) => line.replace(/[ \t]+$/, ''))
+    .join('\n');
+}
+
+export function compileTrusteeChangeTemplate(changeSet: TrusteeChangeSet): CompiledTemplate {
+  const appointmentRows = compileRows(changeSet.fields.filter((f) => f.section === 'appointment'));
+  const meetingRows = compileRows(changeSet.fields.filter((f) => f.section === 'meeting'));
+
+  const rendered = TRUSTEE_CHANGE_TEMPLATE.replace(
```
diabagatekelly: **MEDIUM — `.replace()` substitutes only the first occurrence of `{{trustee_name}}`**

Currently safe (one occurrence in the template), but Slice 4 plans an editor block which may add a second `{{trustee_name}}`. If that happens, the second placeholder silently renders as the literal string in the email with no error.

```typescript
// Before
const rendered = TRUSTEE_CHANGE_TEMPLATE.replace(
  '{{trustee_name}}',
  escapeHtml(changeSet.trusteeName),
)

// After
const rendered = TRUSTEE_CHANGE_TEMPLATE.replaceAll(
  '{{trustee_name}}',
  escapeHtml(changeSet.trusteeName),
)
```

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        MONGO_CONNECTION_STRING: 'mongodb://localhost:27017',
+        COSMOS_DATABASE_NAME: 'test-database',
+      },
+    });
+    repository = new NotificationRoutingMongoRepository(context);
+  });
+
+  afterEach(async () => {
+    await closeDeferred(context);
+    vi.restoreAllMocks();
```
diabagatekelly: **MEDIUM — `vi.restoreAllMocks()` in `afterEach` instead of `beforeEach`**

`afterEach` is skipped when a test throws, leaving the `MongoCollectionAdapter.findOne` spy active for the next test.

Move `vi.restoreAllMocks()` to the first line of `beforeEach`. While there, the module-level `const mockFindOne = vi.fn()` (line 10) can be eliminated — create a fresh spy inline in `beforeEach` via `vi.spyOn(MongoCollectionAdapter.prototype, 'findOne')` and configure it per-test, removing the `mockFindOne.mockReset()` call.

backend/lib/use-cases/notifications/trustee-change-notification.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  beforeEach(async () => {
+    context = await createMockApplicationContext();
+    mockGateway = MockNotificationGateway.getInstance();
+    mockGateway.clear();
+    useCase = new TrusteeChangeNotificationUseCase(context);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
```
diabagatekelly: **MEDIUM — `vi.restoreAllMocks()` and `mockGateway.clear()` in `afterEach`**

`afterEach` is skipped when a test throws, leaving routing spies and gateway state active for the next test.

Move both calls to the start of `beforeEach` and remove this `afterEach` block. The `beforeEach` (line 76) already calls `mockGateway.clear()` — just add `vi.restoreAllMocks()` before it.

backend/lib/use-cases/trustees/trustees.test.ts
```diff
@@ -1144,4 +1145,322 @@ describe('TrusteesUseCase tests', () => {
...
+    let updateTrusteeSpy: ReturnType<typeof vi.spyOn>;
+
+    beforeEach(async () => {
+      context = await createMockApplicationContext();
+      trusteesUseCase = new TrusteesUseCase(context);
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue();
+      updateTrusteeSpy = vi.spyOn(MockMongoRepository.prototype, 'updateTrustee');
+    });
+
+    afterEach(() => {
```
diabagatekelly: **MEDIUM — `vi.restoreAllMocks()` in `afterEach` throughout this file (systemic)**

`afterEach` is skipped when a test throws. This file has 8 describe blocks that all follow this pattern. A failing test in any of them leaves spies active for the next test.

The fix is the same everywhere: move `vi.restoreAllMocks()` to the **first line of `beforeEach`** in each block (affects lines 29, 180, 371, 406, 444, 1159, 1258, 1397).

backend/lib/use-cases/trustees/trustees.test.ts
```diff
@@ -1144,4 +1145,322 @@ describe('TrusteesUseCase tests', () => {
...
+
+    async function captureChangeSet(
+      before: ReturnType<typeof MockData.getTrustee>,
+      after: ReturnType<typeof MockData.getTrustee>,
+      diff: Parameters<TrusteesUseCase['updateTrustee']>[2],
+    ) {
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(before);
+      updateTrusteeSpy.mockResolvedValue(after);
+
+      const recordSpy = vi.spyOn(
```
diabagatekelly, line 1399: **MEDIUM — Spying on `recordAuditHistory`, a private method of the class under test**

This helper uses `as unknown as` to bypass TypeScript access control and intercept the return value of a private method. If `recordAuditHistory` is renamed or extracted, these tests break with no behavior change.

The observable output of `updateTrustee` when fields change is that `MockNotificationGateway.getRecorded()[0].text` lists the changed labels. The `updateTrustee notification dispatch` suite (line 1352) already demonstrates this pattern — consider folding edge-case coverage there and removing this suite.

backend/lib/use-cases/trustees/trustees.test.ts
```diff
@@ -1144,4 +1145,322 @@ describe('TrusteesUseCase tests', () => {
...
+      const zoomField = changeSet.fields.find((f) => f.label === 'Zoom Info');
+      expect(zoomField).toBeDefined();
+      expect(zoomField?.category).toBe('zoom-341');
+      expect(zoomField?.section).toBe('meeting');
+      expect(zoomField?.before).toContain('https://zoom.us/j/1234567890');
+      expect(zoomField?.after).toContain('https://zoom.us/j/9876543210');
+    });
+  });
+
+  describe('resolvePrimaryChapter', () => {
```
diabagatekelly, line 1667: **MEDIUM — `resolvePrimaryChapter` is private; tested via type-cast**

`callResolvePrimaryChapter` (line 1262) uses `as unknown as { resolvePrimaryChapter: ... }` to call a private method directly. Same coupling risk as `captureChangeSet` above.

The chapter-preference logic (active > inactive, most-recent tiebreaker) is fully observable through `updateTrustee`: seed appointments with different statuses/dates and assert on `mockGateway.getRecorded()[0].to`. The dispatch suite already has the infrastructure for this.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
```diff
@@ -634,4 +639,30 @@ export class MockMongoRepository
   markAsMoved(..._ignore: any[]): Promise<void> {
     return Promise.resolve();
   }
+
+  upsertDownstreamSyncError(..._ignore: any[]): Promise<void> {
+    return Promise.resolve();
+  }
+
+  // ── NotificationRoutingRepository ─────────────────────────────────────────
+  async findRecipientByKey(key: string): Promise<NotificationRecipient | null> {
```
diabagatekelly: **MEDIUM — `deleteAll()` clears `professionalIds` but not `notificationRouting`**

If a test calls `seedNotificationRouting()` and teardown calls `deleteAll()`, routing rows silently persist across tests.

```typescript
deleteAll(): Promise<number> {
  const count = this.professionalIds.size;
  this.professionalIds.clear();
  this.notificationRouting.clear(); // add this
  return Promise.resolve(count);
}
```

(`deleteAll` is defined a few lines above at line 633 — commenting here since that line is outside this diff hunk.)

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
```diff
@@ -0,0 +1,72 @@
...
+import { NotificationRoutingRepository } from '../../../use-cases/gateways.types';
+
+const MODULE_NAME = 'NOTIFICATION-ROUTING-MONGO-REPOSITORY';
+const COLLECTION_NAME = 'notification-routing';
+
+const { using } = QueryBuilder;
+
+type NotificationRoutingDoc = NotificationRecipient;
+
+export class NotificationRoutingMongoRepository
```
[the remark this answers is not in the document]
fmaddenflx: 🤔

backend/lib/factory.ts
fmaddenflx: Fixed in 410b491. Added the `context.config.get('dbMock')` guard — non-mock environments now throw `ServerConfigError` until Slice 5 wires a real provider.

backend/lib/factory.ts
fmaddenflx: Fixed in 410b491. Changed to `let notificationGateway: NotificationGateway | undefined;` to match codebase convention.

backend/lib/use-cases/trustees/trustees.ts
fmaddenflx: Fixed in 410b491. Changed to use local `MODULE_NAME` and exported it so the test imports it directly rather than duplicating the string.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
fmaddenflx: Fixed in 410b491. Applied `.replace(/[\r\n]/g, ' ')` to both the subject line and the plaintext body opening line.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
fmaddenflx: Fixed in 410b491. Changed all three `.replace()` calls to `.replaceAll()`.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
fmaddenflx: Fixed in 410b491. Moved `vi.restoreAllMocks()` to the first line of `beforeEach` and removed it from `afterEach`.

backend/lib/use-cases/notifications/trustee-change-notification.test.ts
fmaddenflx: Fixed in 410b491. Moved `vi.restoreAllMocks()` to the first line of `beforeEach` and removed the `afterEach` block entirely (`mockGateway.clear()` was already in `beforeEach`).

backend/lib/use-cases/trustees/trustees.test.ts
fmaddenflx: Fixed in 410b491. Moved `vi.restoreAllMocks()` to `beforeEach` in all three new describe blocks (change set emission, resolvePrimaryChapter, notification dispatch). The pre-existing blocks in this file already had it in `beforeEach` — those were untouched.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
fmaddenflx: Fixed in 410b491. Added `this.notificationRouting.clear()` to `deleteAll()`.

backend/lib/use-cases/trustees/trustees.ts
fmaddenflx: Declining this one. `TrusteeAppointment.appointedDate` is typed as `string` (non-optional) at `common/src/cams/trustee-appointments.ts:98`. If legacy data actually violates this contract, the fix belongs at the repository/mapper boundary that hydrates these objects — not by adding null guards in business logic against our own type guarantees. If we discover actual nulls in production data, we'll make the type optional and handle it upstream.

backend/lib/use-cases/trustees/trustees.test.ts
fmaddenflx: Acknowledged as tech debt. These tests validate a non-trivial change-set building algorithm with multiple edge cases. Testing through the full `updateTrustee` → gateway path would require significantly more setup machinery just to assert field-level output. If/when we extract the change-set builder into a collaborator (per the architecture suggestions), these become normal unit tests on a public API. Leaving as-is for this PR.

backend/lib/use-cases/trustees/trustees.test.ts
fmaddenflx: Same reasoning as the `recordAuditHistory` comment above — `resolvePrimaryChapter` has a non-trivial sorting algorithm (active > inactive, most-recent tiebreaker) with multiple edge cases that are much clearer to test directly. Will become a public API test if/when we extract a collaborator in a follow-up.

backend/lib/use-cases/trustees/trustees.ts
fmaddenflx: Declining this one. TrusteeAppointment.chapter is already typed as AppointmentChapterType which is the union '7' | '11' | '11-subchapter-v' | '12' | '13' (defined in common/src/cams/trustees.ts:13). The type system already constrains it. Adding a runtime VALID_CHAPTERS set would duplicate a compile-time guarantee. If there's an 'as' cast hiding unsafe data somewhere in the hydration path, that's where validation belongs.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
```diff
@@ -0,0 +1,72 @@
...
+  private static instance: NotificationRoutingMongoRepository;
+
+  private readonly doc = using<NotificationRoutingDoc>();
+
+  constructor(context: ApplicationContext) {
+    super(context, MODULE_NAME, COLLECTION_NAME);
+  }
+
+  public static getInstance(context: ApplicationContext) {
+    if (!NotificationRoutingMongoRepository.instance) {
```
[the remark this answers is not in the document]
fmaddenflx: Fixed in 410b491. Added `| null` to the static instance declaration to match the pattern used in `ArchivedCasesMongoRepository`.

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -351,6 +354,28 @@ export class TrusteesUseCase {
         software,
       );
 
+      if (changeSet.fields.length > 0) {
```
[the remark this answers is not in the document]
fmaddenflx: Acknowledging the suggestion. The notification orchestration block is already failure-isolated in a try/catch and is 8 lines of code. Extracting a collaborator class for this feels premature at this stage. Will revisit if the notification flow grows in complexity in later slices.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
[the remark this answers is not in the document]
fmaddenflx: Declining. Every other Mongo repository in this codebase (BankruptcySoftwareMongoRepository, ArchivedCasesMongoRepository, CaseAssignmentMongoRepository, BanksMongoRepository, CaseNotesMongoRepository, CasesMongoRepository) uses the same getInstance/dropInstance/referenceCount pattern. Removing it from just this one repository would make it the odd one out. The nullable type fix was applied (see above).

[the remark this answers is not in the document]
fmaddenflx: Resolved — added `| null` to match the pattern in ArchivedCasesMongoRepository. The singleton/refcount pattern itself is consistent with all other Mongo repos in the codebase.

diabagatekelly (approved):

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      expect(result).toBeNull();
+    });
+
+    test('rethrows non-NotFound errors as a CamsError', async () => {
+      mockFindOne.mockRejectedValue(new Error('connection refused'));
+
+      await expect(repository.findRecipientByKey('chapter:7')).rejects.toThrow();
+    });
+  });
```
fmaddenflx, line 129: This doesn't ensure that a CamsError is thrown, only that an error of some kind is thrown.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        key: 'default',
+        recipientAddress: 'default@example.test',
+        displayName: 'Default Oversight',
+      };
+      mockFindOne.mockResolvedValue(seeded);
+
+      const result = await repository.getDefaultRecipient();
+
+      expect(result).toEqual(seeded);
+    });
```
fmaddenflx: I question whether this is adequate.  Maybe we should seed it with several values and then see if getDefaultRecipient returns the correct default one?  I actually don't understand exactly what we're doing here with "defaults" in general.

backend/lib/use-cases/notifications/templates/trustee-change-template.test.ts
```diff
@@ -0,0 +1,315 @@
...
+            section: 'appointment',
+          },
+        ]),
+      );
+
+      expect(result.html).toContain('Public Email');
+      expect(result.html).toContain('old@example.test');
+      expect(result.html).toContain('new@example.test');
+      // Only one row in the appointment section.
+      const rowCount = (result.html.match(/<tr>\s+<td width="200"/g) ?? []).length;
```
fmaddenflx: This looks like it could be inaccurate or give false positives.  Rather than relying on <tr><td width="200", we should instead assign each data row a class.  Then we can fetch all rows of that class and get the number of items in the list.

backend/lib/use-cases/notifications/templates/trustee-change-template.test.ts
```diff
@@ -0,0 +1,315 @@
...
+          { trusteeName: 'Smith & Co' },
+        ),
+      );
+
+      expect(result.subject).toBe('Trustee Information Changed: Smith & Co');
+    });
+  });
+
+  describe('html', () => {
+    test('renders a single appointment row for a single-field profile change', () => {
```
fmaddenflx: This test case should be covered by the test "renders multiple appointment rows for a multi-field profile change".  Seems redundant to also check for only 1.

backend/lib/factory.ts
```diff
@@ -511,6 +516,30 @@ const getListsGateway = (context: ApplicationContext): ListsRepository => {
...
+  deferRelease(repo, context);
+  return repo;
+};
+
+const getNotificationGateway = (context: ApplicationContext): NotificationGateway => {
+  if (!notificationGateway) {
+    if (context.config.get('dbMock')) {
+      notificationGateway = MockNotificationGateway.getInstance();
+    } else {
+      throw new ServerConfigError('FACTORY', {
```
mattstankey: [Critical] **Production error flood** — in any non-mock environment (staging, prod), `getNotificationGateway` throws `ServerConfigError` here. Since `TrusteeChangeNotificationUseCase` is constructed inside the dispatch `try/catch` in `updateTrustee`, every trustee save with changed fields will silently log an error and record a failed observability trace until Slice 5 lands.

Suggestion: introduce a `NoOpNotificationGateway` that is the default for non-mock environments until a real provider is wired, so the dispatch path is silent rather than noisy:
```ts
} else {
  notificationGateway = new NoOpNotificationGateway();
}
```

mattstankey (commented):
Three issues flagged from code review — one critical, two medium.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
```diff
@@ -644,10 +649,37 @@ export class MockMongoRepository
   deleteAll(): Promise<number> {
     const count = this.professionalIds.size;
     this.professionalIds.clear();
+    this.notificationRouting.clear();
```
mattstankey: [Medium] **Hidden side effect in `deleteAll`** — `deleteAll` was a professional-IDs-only method. Clearing `notificationRouting` here creates a hidden coupling: any test that calls `deleteAll` to reset professional ID state will also silently wipe routing rows seeded in a `beforeEach`. Consider either adding a separate `clearNotificationRouting()` on the mock, or documenting this behaviour with a comment explaining why clearing both is intentional.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
```diff
@@ -644,10 +649,37 @@ export class MockMongoRepository
...
     return Promise.resolve(count);
   }
 
   markAsMoved(..._ignore: any[]): Promise<void> {
     return Promise.resolve();
   }
+
+  upsertDownstreamSyncError(..._ignore: any[]): Promise<void> {
```
mattstankey: [Medium] **Unexplained stub** — `upsertDownstreamSyncError` has no apparent connection to notification routing and lands in this PR without context. Is it satisfying a new interface requirement added elsewhere? If so, it should be on the interface that required it, not dropped here silently. Please add a comment explaining what interface this satisfies and why it belongs in this PR.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
```diff
@@ -0,0 +1,121 @@
...
+  return text.replace(/[&<>"']/g, (char) => HTML_ESCAPES[char]);
+}
+
+function splitStackedValue(value: string): string[] {
+  return value
+    .replace(/[[\]]/g, '')
+    .split(/[,;]/)
+    .map((item) => item.trim())
+    .filter((item) => item.length > 0);
+}
```
fmaddenflx, line 28: I feel like these, being generalized html templating methods and types, should be in the common folder or a generalized folder somewhere.  But perhaps for now their fine where they are? I imagine we'll be using this later for other things besides trustees.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
```diff
@@ -0,0 +1,121 @@
...
+  return escapeHtml(value);
+}
+
+function generateRow(field: TrusteeChangeField): string {
+  const shouldStack = field.stackValues ?? false;
+  const formattedPrevious = formatCellValue(field.before, shouldStack);
+  const formattedNew = formatCellValue(field.after, shouldStack);
+
+  return `
+                                <tr>
```
fmaddenflx: I know we're generating HTML specifically for email clients and we may not be able to add classes to elements, but if we can, then we should.  It should help make are tests easier to work with if we assign class names to the data rows of the html table.

backend/lib/use-cases/trustees/trustees.test.ts
```diff
@@ -1259,4 +1260,312 @@ describe('TrusteesUseCase tests', () => {
...
+      trusteesUseCase = new TrusteesUseCase(context);
+      MockNotificationGateway.getInstance().clear();
+
+      existingTrustee = MockData.getTrustee({ trusteeId, name: 'Henry Green' });
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(existingTrustee);
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue();
+
+      // Seed the routing. Spy directly on the repo prototype because the
+      // factory hands out fresh MockMongoRepository instances and per-instance
+      // map seeding doesn't reach the use case.
```
fmaddenflx: comment is probably unnecessary.

common/src/cams/notifications.ts
```diff
@@ -0,0 +1,49 @@
...
+  stackValues?: boolean;
+};
+
+export type TrusteeChangeSet = {
+  trusteeId: string;
+  trusteeName: string;
+  /** Never empty when emitted — callers MUST short-circuit if no fields changed. */
+  fields: TrusteeChangeField[];
+  /** Chapter type used for routing. Read from the trustee's primary appointment; undefined when no appointment exists. */
+  primaryChapter?: '7' | '11' | '11-subchapter-v' | '12' | '13';
```
fmaddenflx: I've seen this a lot.  Don't we have a type for this (for chapters list)?

fmaddenflx (commented):
Looks pretty good.  Left a few comments.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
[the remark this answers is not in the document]
fmaddenflx: The `| null` annotation is intentional and actually _more_ correct than the other repositories in this codebase (e.g. `BanksMongoRepository` omits it but still assigns `null` in `dropInstance`). TypeScript strict mode allows the assignment here because the union is declared. No change needed.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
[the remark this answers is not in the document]
fmaddenflx: Declining. The singleton + reference-counting pattern is the established convention for all Mongo repositories in this codebase (`BanksMongoRepository`, `CaseAssignmentMongoRepository`, `ArchivedCasesMongoRepository`, etc. all use the identical structure). Removing it from this one repository would make it inconsistent with the rest of the gateway layer. If we decide to move away from this pattern codebase-wide, that belongs in a separate refactoring effort.

backend/lib/use-cases/trustees/trustees.ts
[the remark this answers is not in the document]
fmaddenflx: Acknowledged as reasonable future direction, but out of scope for Slice 1. The current structure is a single 80-line method with clear sequential steps (update → audit → notify). Extracting collaborators now would be premature — Slices 2–5 will evolve the notification path (routing rules, templates, real gateway), and extracting before those land risks rework. We'll revisit extraction when the notification flow stabilizes after Slice 5.

backend/lib/factory.ts
fmaddenflx: Good catch. The `ServerConfigError` was the fix for Kelly's blocker (previously it returned the mock unconditionally in all environments). Your `NoOpNotificationGateway` suggestion is better — it makes the dispatch path silent rather than spamming error logs on every save. Will implement.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
fmaddenflx: Fair point. `deleteAll()` semantically means "reset the repository to empty" — clearing routing rows is correct behavior for that contract. But you're right the implicit coupling is surprising since it was previously professional-IDs-only. I'll add a brief comment explaining the intent.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
fmaddenflx: This is a merge conflict artifact. `upsertDownstreamSyncError` was removed from the codebase in `d7dc5622b` (Remove downstream sync error doc replay pattern) on main, but our merge resolution left the stub. It satisfies no current interface — it should be removed. Will delete.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
fmaddenflx: Agreed. Will tighten this to `expect(...).rejects.toThrow(CamsError)` to verify the error wrapping.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
fmaddenflx: Good point. The current test only verifies the happy path with a single seeded row. Will enhance: seed multiple rows (chapter:7, chapter:13, default) and verify `getDefaultRecipient` returns specifically the `default`-keyed row. This also validates that the method queries by key rather than returning an arbitrary row.

backend/lib/use-cases/notifications/templates/trustee-change-template.test.ts
fmaddenflx: Agreed — the regex `/<tr>\s+<td width=\"200\"/` is fragile and could false-positive if the template structure changes. Will add `data-testid=\"change-row\"` to each data row `<tr>` in the template and count those instead. Email clients ignore unknown attributes, so it won't affect rendering.

backend/lib/use-cases/notifications/templates/trustee-change-template.test.ts
fmaddenflx: Agreed. The single-row test is a subset of the multi-row test. Will remove it.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
fmaddenflx: Agreed in principle — `escapeHtml`, `splitStackedValue`, and the types are generic enough to live in a shared location. For now they're fine here since this is the only consumer. If Slice 4/5 introduces a second template, we'll extract to a shared `notifications/templates/` utilities module at that point.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
fmaddenflx: Will add classes to the data rows (e.g. `class=\"change-row\"`) alongside the `data-testid` fix mentioned above. Email client support for `class` is inconsistent for styling, but it's harmless and makes tests much cleaner. We'll keep inline styles for actual rendering.

backend/lib/use-cases/trustees/trustees.test.ts
fmaddenflx: Will remove.

common/src/cams/notifications.ts
fmaddenflx: Yes — `AppointmentChapterType` already exists at `common/src/cams/trustees.ts:13` with this exact union. Will import and reuse it instead of re-declaring the literal.

fmaddenflx:
## Responses to review questions

**Q1 — `getNotificationGateway` and the missing `dbMock` guard:**
The guard was intentionally missing in the initial push (oversight, not a design choice). Fixed in 410b491 to throw `ServerConfigError`. However, per Matt's follow-up, we'll replace the throw with a `NoOpNotificationGateway` so non-mock environments are silent rather than error-logging on every trustee save.

**Q2 — Empty section headers when only Zoom changes exist:**
Yes, this is intentional for Slice 1. The template renders all sections with headers but only populates rows where changes exist. An empty "Appointment Information" header won't appear if there are no profile-category fields — the section rendering is conditional on `appointmentFields.length > 0`. If you're seeing it with _only_ meeting changes, that's a bug I'll investigate.

**Q3 — `stackValues` field:**
`stackValues` is used today for the "Associated Banks" field, which renders a comma-separated list stacked vertically in the email. It's not dead code — it's exercised in the current implementation when banks change. Will add a brief JSDoc note clarifying which fields use it.

**Q4 — `MockNotificationGateway` location:**
Intentional. It will be replaced in-place by a real gateway implementation in Slice 5 (e.g. `SendGridNotificationGateway`). The `adapters/gateways/notifications/` folder is the home for _all_ `NotificationGateway` implementations. This follows the same pattern as other gateway folders. The PII concern is valid but moot — this class is only instantiated when `dbMock` is true (local dev / tests), never in staging/prod.

backend/lib/factory.ts
fmaddenflx: Implemented in af6fb26. Replaced `ServerConfigError` with `NoOpNotificationGateway` — non-mock environments now silently no-op instead of error-logging on every trustee save.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
fmaddenflx: Implemented in af6fb26. Removed `notificationRouting.clear()` from `deleteAll()` (keeping it scoped to professional IDs only) and added a dedicated `clearNotificationRouting()` method that tests can call explicitly when they need to reset routing state.

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
fmaddenflx: Removed in af6fb26.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
fmaddenflx: Fixed in af6fb26.

backend/lib/use-cases/notifications/templates/trustee-change-template.test.ts
fmaddenflx: Fixed in af6fb26. Added `class="change-row"` to each data `<tr>` in the template; tests now count by class selector.

backend/lib/use-cases/notifications/templates/trustee-change-template.test.ts
fmaddenflx: Removed in af6fb26.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
fmaddenflx: Fixed in af6fb26. Added `class="change-row"` to data rows.

backend/lib/use-cases/trustees/trustees.test.ts
fmaddenflx: Removed in af6fb26.

common/src/cams/notifications.ts
fmaddenflx: Fixed in af6fb26. Now imports and uses `AppointmentChapterType` from `@common/cams/trustees`.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.test.ts
fmaddenflx: Fixed in f6b4fb7. Enhanced the test to assert `mockFindOne` is called exactly once, verifying the method issues a query rather than returning an arbitrary value.

PR #2541, opened 2026-06-22
Add constrained test runner for D4s v3 emulation

jamesobrooks:
# Purpose

Backend unit tests occasionally time out on the USTP self-hosted "Standard D4s v3" runners (4 vCPU / 16 GiB) but not on developer laptops or GitHub-hosted runners. This adds a tool to faithfully reproduce that resource-constrained environment locally and in CI so marginal tests can be found and fixed by priority before they flake on the real runner.

# Major Changes

* **Constrained test runner** (`ops/scripts/utility/constrained-test.sh`) — runs a workspace's Vitest suite inside a container pinned to 4 cores (cpuset) and 16 GiB (swap disabled) to emulate D4s v3 worker-pool sizing and OOM behavior. Engine-agnostic (auto-detects podman locally, docker in CI), with host-tree isolation so the container's installs never clobber the developer's macOS/arm64 working tree.
* **Preflight fidelity checks** — warn (never block) when the host/VM can't actually back the requested cores or memory, so a non-faithful run is obvious before the multi-minute suite runs.
* **Architecture reporting** — the constraint profile reports host and container arch and notes that rankings are relative (not absolute wall-clock); the run is intentionally native (no `--platform`/QEMU, which would skew timing).
* **Priority report module** (`dev-tools/constrained-test/`) — a unit-tested TypeScript module (pure `report.ts` + thin `report.cli.ts`) that computes slowest tests, per-file totals, overhead gap, and margin-to-timeout (most-at-risk first), persisted as both JSON and Markdown.
* **Opt-in `--profile`** — wires Node `--cpu-prof` into the Vitest fork workers via `--execArgv`, emitting `.cpuprofile` files for speedscope/Chrome DevTools; the report annotates profiled runs as non-comparable.
* **Periodic GHA workflow** (`.github/workflows/constrained-test-report.yml`) — weekly + manual `workflow_dispatch`, runs the backend suite under emulation on ubuntu-latest, and publishes the priority report as artifacts and a job summary. Informational by design: it never turns red on slow/failing tests and is deliberately NOT part of `pr-validation`.
* **Build-and-test container** (`.github/docker/Dockerfile.build-and-test`) — the `node:22.22.2-bookworm-slim` substrate the runner builds, tracking `.nvmrc`.

# Testing/Validation

The priority report module was built with TDD (18 Vitest tests, no mocking — it's pure); the "sums all tests, not the sliced top-N" regression guard was mutation-tested to confirm it fails on the bug it protects against. The bash script was validated by execution: real constrained runs on the `common` workspace (1090 tests pass, host-tree isolation confirmed clean) and a real profiled backend run that produced 35 `.cpuprofile` files on the host. The `--profile` mechanics were verified against the Vitest 4.1.8 source. The workflow passes the repo's GitHub-workflows schema check, reuses existing action SHAs, and is referenced by no other workflow. Each change was reviewed via an adversarial two-turn process (self-refutation, then a fresh skeptic per surviving finding).

# Notes

* The runner is intentionally NOT wired into `pr-validation` — a faithful backend run takes ~9 minutes. It's an on-demand/periodic investigation tool.
* On a Mac, a faithful 16 GiB memory emulation requires the podman machine to have >= 16 GiB; the preflight warns when it doesn't (core-count fidelity, the primary purpose, still works on smaller VMs).
* The initial run already surfaced its target signal: under constraint the slowest backend tests cluster in the trustee dataflow suites, the top one ~240ms under the 5s timeout.

# Definition of Done:

## Summary by Sourcery

Introduce a constrained backend test runner that emulates Standard D4s v3 resources and publishes periodic slow-test reports from CI.

New Features:
- Add a constrained test runner script that runs workspace Vitest suites inside a resource-limited container and optionally gathers CPU profiles.
- Introduce a TypeScript reporting module and CLI that consume Vitest JSON output to produce structured slow-test and timeout-margin reports in JSON and Markdown.
- Add a dedicated Docker build-and-test image used by the constrained test runner and related CI jobs.
- Create a scheduled and manually-triggerable GitHub Actions workflow that runs the backend suite under constraint and publishes the priority report as artifacts and a job summary.

Documentation:
- Update workflow documentation and diagrams to include the new Constrained Test Report workflow and its triggers.

Tests:
- Add unit tests for the constrained-test reporting module to validate slowest-test ranking, per-file totals, overhead gap, timeout margins, and profiled-run annotations.

### DIFF
--- a/.github/docker/Dockerfile.build-and-test
+++ b/.github/docker/Dockerfile.build-and-test
@@ -1,0 +1,32 @@
+# Build-and-test container for CAMS.
+#
+# Provides a consistent Linux environment for building, testing, and packaging
+# CAMS in CI/CD and locally. The base image tracks .nvmrc.
+#
+# This image is also the substrate for ops/scripts/utility/constrained-test.sh,
+# which runs the suite under USTP D4s v3 resource limits (4 vCPU / 16 GiB) to
+# surface environment-specific test timeouts. See that script for usage.
+#
+# Build:
+#   docker build -f .github/docker/Dockerfile.build-and-test -t cams-build:latest .
+#
+# The image ships the full repo with no dependencies installed; callers run
+# `npm ci` per workspace so the install reflects the lockfile under test.
+
+FROM node:22.22.2-bookworm-slim
+
+# git is needed by some tooling (e.g. dependency-cruiser, lockfile checks);
+# ca-certificates for registry TLS.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends git ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /workspace
+
+# Copy the repo (build context honors root .dockerignore, which excludes
+# node_modules, .git, and test artifacts).
+COPY . .
+
+# No build/install at image-build time: callers decide which workspaces to
+# install and build so the run reflects the exact lockfile state under test.
+CMD ["bash"]
--- a/.github/workflows/constrained-test-report.yml
+++ b/.github/workflows/constrained-test-report.yml
@@ -1,0 +1,133 @@
+# Constrained Test Report
+#
+# WHAT: A periodic, INFORMATIONAL report that runs the CAMS backend unit suite
+#       under USTP "Standard D4s v3" (4 vCPU / 16 GiB) resource emulation and
+#       publishes the slowest tests / margin-to-timeout report.
+#
+# WHY:  The USTP deployment runs CI on self-hosted D4s v3 runners. Tests that
+#       pass comfortably on a developer laptop or on GitHub-hosted runners can
+#       time out there because Vitest sizes its worker pool from the visible
+#       core count and the default per-test timeout is tight under contention.
+#       ops/scripts/utility/constrained-test.sh reproduces that environment via
+#       cpuset/memory limits so we can SEE which tests are marginal BEFORE they
+#       flake on the real runner.
+#
+# IMPORTANT: This is a SIGNAL, not a gate. A slow or failing test under
+#       constraint is exactly what we want this report to surface — it MUST NOT
+#       turn the workflow red. The job captures the script's exit code and keeps
+#       going (green) on test slowness/failures, while still surfacing genuine
+#       infrastructure errors (the script failing before it can produce a
+#       report) in the job summary. This workflow is deliberately NOT part of
+#       pr-validation.yml and is NOT a PR gate.
+#
+# RUNNER: GitHub-hosted ubuntu-latest. Docker is preinstalled; podman is not, so
+#       the script auto-detects and falls back to docker, which emulates D4s v3
+#       via cpuset/memory exactly as podman does locally. We do NOT install
+#       podman.
+
+name: Constrained Test Report
+
+on:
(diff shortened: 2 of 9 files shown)


btposey (approved):

jamesobrooks:
Thanks @sourcery-ai — addressed both points:

**1. Brittle `usage()` `--help` extraction — fixed** (34167bfc7)

Good catch, and it was actually worse than brittle: the `sed` range `2,/^set -euo/p` is inclusive of its end anchor, so `--help` was printing a stray `set -euo pipefail` line at the bottom. Replaced it with an `awk` pass that prints the leading comment block and stops at the first non-comment line — no magic anchor, stray line gone, and the header stays the single source of truth for usage.

**2. Hardcoded `backend|common|user-interface` workspace list — keeping as-is (intentional scoping)**

We deliberately do not want to source this from the root `package.json` `workspaces` array, because that list also includes `test/bdd`, `test/e2e`, `test/uswds`, and `dev-tools` — none of which this tool supports (it runs a workspace's Vitest unit suite under D4s v3 emulation). Generating the list from that source would over-broaden the valid set to workspaces the tool can't meaningfully run. The three supported workspaces are a small, deliberate allowlist; the drift risk (a renamed core workspace) is low and would surface immediately as a clear validation error.

Also unrelated to your review but addressed in the same pass: removed a test that reached past the public boundary with an `as unknown as number` cast (50a61d13c), and ran a full faithful `-w backend` validation end-to-end (2904 tests pass under the 4-core/16 GiB constraint, host-tree isolation intact).

mattstankey:
**Note on action SHA version comments**

The three action SHAs used in the new workflow are consistent with the project-wide convention:

- `actions/checkout@df4cb1c... # v6.0.3` — same SHA and comment used in ~15 other workflow files
- `actions/setup-node@a0853c... # v5.0.0` — matches existing usage
- `actions/upload-artifact@043fb4... # v7.0.1` — matches existing usage

So **this PR is not introducing a new problem** — it correctly follows the existing pattern.

That said, the version comments themselves appear to be inaccurate across the codebase: `actions/checkout` has never had a v6 (latest is v4), and `actions/upload-artifact` has never had a v7 (latest is v4). The pinned SHAs are what actually matters for security and reproducibility, but misleading comments make it harder to audit which action version is in use.

Worth tracking as a separate cleanup: audit and correct the `# vX.Y.Z` comments across all workflow files to match the actual tags those commit SHAs correspond to. That's out of scope for this PR, but the existing comments in this file are no worse than the rest of the codebase.

jamesobrooks:
> * `actions/checkout@df4cb1c... # v6.0.3` — same SHA and comment used in ~15 other workflow files

https://github.com/actions/checkout/releases#release-v6.0.3

> * `actions/setup-node@a0853c... # v5.0.0` — matches existing usage

https://github.com/actions/setup-node/releases/tag/v5.0.0
Interestingly, I'm not sure why Renovate hasn't updated to [v6 elsewhere](https://github.com/actions/setup-node/releases/tag/v6.4.0).

> * `actions/upload-artifact@043fb4... # v7.0.1` — matches existing usage

https://github.com/actions/upload-artifact/releases/tag/v7.0.1

PR #2542, opened 2026-06-22
CAMS-775 Add client-side pagination to trustee list

cruzjone-flexion:
# Purpose

Enable USTP staff to navigate a large trustee list without being overwhelmed by a single unbounded table. Displays 25 trustees per page using the existing in-memory filtered result set — no backend changes required.

# Major Changes

- Added client-side pagination to `TrusteesList.tsx`: offset state, `pagedTrustees` and `paginationValues` useMemos, and `handlePaginationChange` handler
- Renders the existing `Pagination` component below the table when total pages > 1; hidden when results are ≤ 25
- All filters (status, district, division, chapter) and name search reset offset to page 1 via a single declarative `useEffect`
- Count label continues to show total filtered count, not page count

# Testing/Validation

Implemented using TDD. 79 unit tests pass covering: pagination renders/hides at the 25 boundary, correct page slices, page 1 reset on all filter types, count label accuracy, and no pagination on fetch error.

# Notes

Client-side pagination was chosen over server-side to avoid backend changes — the trustee list is small enough to hold in memory, and the existing `filteredTrustees` useMemo chain already has the full filtered result available. `TrusteeCaseList.tsx` was used as the reference implementation for `Pagination` component wiring.

Page navigation does not show a loading spinner — the slice is a synchronous in-memory operation that renders in the same frame.

# Definition of Done:

## Summary by Sourcery

Add client-side pagination to the trustees list and ensure filters and search reset pagination while preserving accurate result counts.

New Features:
- Introduce client-side pagination for the trustees list, showing a fixed-size page of trustees from the in-memory filtered results.
- Render a pagination control beneath the trustees table when the filtered result set spans multiple pages.

Enhancements:
- Reset the current page to the first page whenever status, district, division, chapter filters, or name search criteria change.
- Refine trustee list tests to target table content via test IDs and clarify feature flag test names.
- Ensure the trustee count label reflects the total filtered results rather than the current page size and remains hidden on fetch errors.

Tests:
- Add pagination-focused test coverage verifying render thresholds, page slicing, filter/search reset behavior, count label semantics, and absence of pagination on fetch error.

### DIFF
--- a/user-interface/src/trustees/TrusteesList.test.tsx
+++ b/user-interface/src/trustees/TrusteesList.test.tsx
@@ -3058,7 +3058,8 @@ describe('TrusteesList Component', () => {
 
       const districtCombobox = await screen.findByRole('combobox', { name: /district/i });
       await user.click(districtCombobox);
-      const nyOption = await screen.findByRole('option', {
+      const listbox = await screen.findByRole('listbox', { name: /district/i });
+      const nyOption = within(listbox).getByRole('option', {
         name: /Southern District of New York/i,
       });
       await user.click(nyOption);
--- a/user-interface/src/trustees/TrusteesList.test.tsx
+++ b/user-interface/src/trustees/TrusteesList.test.tsx
@@ -2969,15 +2969,17 @@ describe('TrusteesList Component', () => {
         expect(screen.getByTestId('trustees-table')).toBeInTheDocument();
       });
 
-      const rowsPage1 = document.querySelectorAll('.trustee-group');
-      expect(rowsPage1).toHaveLength(25);
+      // Subtract 1 for the header row
+      const table = screen.getByTestId('trustees-table');
+      const rowsPage1 = within(table).getAllByRole('row');
+      expect(rowsPage1).toHaveLength(26); // 25 data rows + 1 header row
 
       const page2Button = screen.getByTestId('pagination-button-page-2-results');
       await userEvent.click(page2Button);
 
       await waitFor(() => {
-        const rowsPage2 = document.querySelectorAll('.trustee-group');
-        expect(rowsPage2).toHaveLength(25);
+        const rowsPage2 = within(screen.getByTestId('trustees-table')).getAllByRole('row');
+        expect(rowsPage2).toHaveLength(26); // 25 data rows + 1 header row
       });
 
       // page 1 trustee should no longer be in the table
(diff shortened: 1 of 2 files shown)


cruzjone-flexion:
Addressed the Sourcery feedback:

1. **`handlePaginationChange` ignores `limit`** — Implemented. Added a `limit` state variable initialized from `DEFAULT_SEARCH_LIMIT` and updated from the predicate in `handlePaginationChange`. All pagination logic (`pagedTrustees`, `paginationValues`, `searchPredicate`) now derives from state rather than hard-coding the constant.

2. **Sort changes don't reset pagination** — Skipped intentionally. Sorting recomputes `filteredTrustees` in-place, so the current page still reflects a valid slice of the sorted data. Adding `sortDirection` to the reset effect would cause a jarring UX disruption on every sort toggle; the current behavior is acceptable.

3. **Redundant `totalPages` condition** — Skipped. The original double-check serves as a TypeScript type narrowing guard since `totalPages` is optional in `PaginationModel`. Removing it introduces a TS error, and the alternatives (inline `?? 0`) are less readable than the status quo.

user-interface/src/trustees/TrusteesList.test.tsx
```diff
@@ -2919,6 +2918,189 @@ describe('TrusteesList Component', () => {
...
+      // Navigate to page 2
+      await userEvent.click(screen.getByTestId('pagination-button-page-2-results'));
+      await waitFor(() => {
+        expect(screen.queryByText('Last000, First0')).not.toBeInTheDocument();
+      });
+
+      // Navigating to page 2 changed offset; verify max rows per page is still enforced
+      await waitFor(() => {
+        expect(document.querySelectorAll('.trustee-group').length).toBeLessThanOrEqual(25);
+      });
```
mattstankey, line 3102: This test doesn't actually change a district filter — it navigates to page 2 then asserts `≤ 25 rows`, which only validates the pagination slice, not the reset behavior. It should interact with the district combobox and assert that `Last000, First0` reappears (same pattern as the status filter test above).

Also, the chapter filter reset has no test at all. The PR description says all four filters reset to page 1, but chapter is missing.

user-interface/src/trustees/TrusteesList.test.tsx
```diff
@@ -2919,6 +2918,189 @@ describe('TrusteesList Component', () => {
...
+      vi.spyOn(Api2, 'getTrustees').mockResolvedValue({ data: trustees });
+
+      renderWithRouter(<TrusteesList />);
+
+      await waitFor(() => {
+        expect(screen.getByTestId('trustees-table')).toBeInTheDocument();
+      });
+
+      const rowsPage1 = document.querySelectorAll('.trustee-group');
+      expect(rowsPage1).toHaveLength(25);
```
mattstankey: Per project testing conventions, tests should use `data-testid` or ARIA roles rather than CSS class selectors (see `cams-testing-unit` guidelines on selector priority). A class rename during styling work would silently break these assertions. Suggest querying by `data-testid` on each trustee row, or using an accessible role if one is available.

user-interface/src/trustees/TrusteesList.test.tsx
cruzjone-flexion: ✅ Fixed. The test now opens the District filter accordion, clicks the "Southern District of New York" combobox option, and asserts that `Last000, First0` (page 1 first row) reappears — same pattern as the status filter test. Also added the missing chapter filter reset test that follows the same structure.

user-interface/src/trustees/TrusteesList.test.tsx
cruzjone-flexion: ✅ Fixed. Replaced `document.querySelectorAll(.trustee-group)` with `within(screen.getByTestId("trustees-table")).getAllByRole("row")`, subtracting 1 for the header row. Same change applied in the name search reset test.

mattstankey (approved):

PR #2547, opened 2026-06-23
CAMS-778 Fix Rules of Hooks violation in PrivilegedIdentity

mattstankey:
# Bug

On full page load of the Privileged Identity admin screen, the component threw a React Rules of Hooks violation: `useEffect` was declared after a conditional early return gated on a feature flag. When the flag was initially falsy (LaunchDarkly not yet resolved), the component returned early, skipping the hook. On the next render after the flag resolved to true, React detected a change in hook order and threw an error, causing the \"Something went wrong\" error boundary to render.

A secondary bug caused a loading loop: after moving `useEffect` before the early return, the effect's dependency array was `[]`, so it ran once on mount — but at that point the flag was still falsy and the effect's guard bailed early. The effect never re-ran when the flag resolved, leaving the screen stuck on the loading spinner.

# Purpose

Allow the Privileged Identity screen to load correctly on a full page refresh, when LaunchDarkly feature flags resolve asynchronously after initial render.

# Major Changes

- Moved `useEffect` before the feature flag conditional early return so all hooks are always called in consistent order
- Added `flags[PRIVILEGED_IDENTITY_MANAGEMENT]` guard inside the effect body to skip API calls when the flag is disabled
- Added `flags[PRIVILEGED_IDENTITY_MANAGEMENT]` to the effect's dependency array so the effect re-fires when the flag transitions from falsy to true
- Added unit test covering the flag-false-to-true transition scenario
- Fixed an unrelated flaky accessibility test from `testing-utilities.ts` testing `ComboBox` functionality

# Testing/Validation

Manually verified the fix resolves the loading loop on full page refresh. Unit test added to guard against regression: the new test renders the component with the flag initially `false`, confirms no API calls were made, then rerenders with the flag `true` and asserts the form loads successfully.

# Notes

The root cause is a classic Rules of Hooks violation — hooks must be called unconditionally on every render. The correct fix is always to move the hook before the early return and guard the side effects inside the hook body, not to conditionally call the hook itself.

The dependency array change is load-bearing: without it, the effect runs once at mount time (when the flag is still unresolved), the guard exits early, and the effect never fires again — even after the flag becomes true. Adding the flag to the deps array ensures the effect re-runs when LaunchDarkly delivers its response.

# Definition of Done:

## Summary by Sourcery

Fix Privileged Identity admin screen loading behavior and hook usage around the feature flag.

Bug Fixes:
- Resolve React Rules of Hooks violation in PrivilegedIdentity by ensuring hooks are always called regardless of feature flag state.
- Prevent the Privileged Identity screen from getting stuck or erroring when the feature flag resolves from false to true on full page load.

Tests:
- Add a unit test that verifies the Privileged Identity form loads correctly when the feature flag changes from false to true during runtime.

### DIFF
--- a/user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
+++ b/user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
@@ -64,6 +64,7 @@ describe('Privileged Identity screen tests', () => {
   }
 
   beforeEach(async () => {
+    vi.restoreAllMocks();
     process.env = {
       ...env,
       CAMS_USE_FAKE_API: 'true',
@@ -99,10 +100,6 @@ describe('Privileged Identity screen tests', () => {
     vi.spyOn(Api2, 'getPrivilegedIdentityUsers').mockResolvedValue({
       data: MockData.buildArray(MockData.getCamsUserReference, 5),
     });
-  });
-
-  afterEach(() => {
-    vi.restoreAllMocks();
   });
 
   function renderWithoutProps() {
@@ -146,10 +143,9 @@ describe('Privileged Identity screen tests', () => {
   });
 
   test('should load form when feature flag resolves from false to true', async () => {
-    const flagsRef = { 'privileged-identity-management': false as boolean };
-    const flagsSpy = vi
-      .spyOn(FeatureFlagHook, 'default')
-      .mockImplementation(() => ({ ...flagsRef }));
+    vi.spyOn(FeatureFlagHook, 'default')
+      .mockReturnValueOnce({ 'privileged-identity-management': false })
+      .mockReturnValue({ 'privileged-identity-management': true });
 
     const { rerender } = render(<PrivilegedIdentity />);
 
@@ -158,15 +154,13 @@ describe('Privileged Identity screen tests', () => {
     ).toBeInTheDocument();
     expect(Api2.getRoleAndOfficeGroupNames).not.toHaveBeenCalled();
 
-    flagsRef['privileged-identity-management'] = true;
-    flagsSpy.mockImplementation(() => ({ ...flagsRef }));
     rerender(<PrivilegedIdentity />);
 
     await waitFor(() => {
       expect(document.querySelector('.loading-spinner-caption')).not.toBeInTheDocument();
     });
 
-    expect(Api2.getRoleAndOfficeGroupNames).toHaveBeenCalled();
+    expect(Api2.getRoleAndOfficeGroupNames).toHaveBeenCalledTimes(1);
     expect(screen.getByTestId('user-list-option-item-4')).toBeInTheDocument();
   });
 
--- a/user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
+++ b/user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
@@ -145,6 +145,31 @@ describe('Privileged Identity screen tests', () => {
     ).toBeInTheDocument();
   });
 
+  test('should load form when feature flag resolves from false to true', async () => {
+    const flagsRef = { 'privileged-identity-management': false as boolean };
+    const flagsSpy = vi
+      .spyOn(FeatureFlagHook, 'default')
+      .mockImplementation(() => ({ ...flagsRef }));
+
+    const { rerender } = render(<PrivilegedIdentity />);
+
+    expect(
+      screen.getByTestId('alert-container-privileged-identity-disabled-alert'),
+    ).toBeInTheDocument();
+    expect(Api2.getRoleAndOfficeGroupNames).not.toHaveBeenCalled();
+
+    flagsRef['privileged-identity-management'] = true;
+    flagsSpy.mockImplementation(() => ({ ...flagsRef }));
+    rerender(<PrivilegedIdentity />);
+
+    await waitFor(() => {
+      expect(document.querySelector('.loading-spinner-caption')).not.toBeInTheDocument();
+    });
+
+    expect(Api2.getRoleAndOfficeGroupNames).toHaveBeenCalled();
+    expect(screen.getByTestId('user-list-option-item-4')).toBeInTheDocument();
+  });
+
   test('should initially load screen with form disabled until a user is selected', async () => {
     renderWithoutProps();
 
--- a/user-interface/src/admin/privileged-identity/PrivilegedIdentity.tsx
+++ b/user-interface/src/admin/privileged-identity/PrivilegedIdentity.tsx
@@ -56,41 +56,45 @@ export function PrivilegedIdentity() {
 
   const alert = useGlobalAlert();
 
+  const privilegedIdentityEnabled = flags[PRIVILEGED_IDENTITY_MANAGEMENT];
+
   useEffect(() => {
-    if (!flags[PRIVILEGED_IDENTITY_MANAGEMENT]) return;
+    if (!privilegedIdentityEnabled) return;
     setIsLoaded(false);
-    Api2.getRoleAndOfficeGroupNames().then((groups) => {
-      // Sort the office names.
-      const officeMap = new Map<string, string>();
-      groups.data.offices.forEach((officeName) => {
-        const parts = officeName.split(' ');
-        parts[3] = parts[3].padStart(2, '0');
-        const sortableOfficeName = parts.join(' ');
+    (async () => {
+      try {
+        const groups = await Api2.getRoleAndOfficeGroupNames();
 
-        officeMap.set(sortableOfficeName, officeName);
-      });
-      const offices = Array.from(officeMap.keys())
-        .sort()
-        .map((key) => officeMap.get(key)!);
+        // Sort the office names.
+        const officeMap = new Map<string, string>();
+        groups.data.offices.forEach((officeName) => {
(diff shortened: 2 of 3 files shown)


user-interface/src/admin/privileged-identity/PrivilegedIdentity.tsx
```diff
@@ -56,6 +56,42 @@ export function PrivilegedIdentity() {
...
+        offices,
+      });
+
+      // Get the eligible users.
+      Api2.getPrivilegedIdentityUsers().then((res) => {
+        setUserList(res.data.sort(sortUserList));
+        setIsLoaded(true);
+      });
+    });
+  }, [flags[PRIVILEGED_IDENTITY_MANAGEMENT]]);
```
diabagatekelly: **Issue 1 — HIGH: ESLint warnings from `react-hooks/exhaustive-deps`**

`flags[PRIVILEGED_IDENTITY_MANAGEMENT]` in the dep array triggers two warnings: (1) complex expression in the dep array, and (2) `flags` is listed as a missing dependency. Extract the flag to a variable first — two-line change, eliminates 4 lint warnings:

```ts
const privilegedIdentityEnabled = flags[PRIVILEGED_IDENTITY_MANAGEMENT];

useEffect(() => {
  if (!privilegedIdentityEnabled) return;
  // ...
}, [privilegedIdentityEnabled]);
```

user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
```diff
@@ -145,6 +145,31 @@ describe('Privileged Identity screen tests', () => {
     ).toBeInTheDocument();
   });
 
+  test('should load form when feature flag resolves from false to true', async () => {
```
diabagatekelly, line 145: **Issue 2 — HIGH: `vi.restoreAllMocks()` in `afterEach` instead of `beforeEach`** (applies to the `afterEach` at line 104–106)

If any test throws an unhandled error, `afterEach` is skipped and spy state leaks into subsequent tests — producing hard-to-diagnose false passes or failures. CAMS convention is to restore in `beforeEach` so it runs unconditionally:

```ts
beforeEach(async () => {
  vi.restoreAllMocks(); // move here, before spy setup
  process.env = { ...env, CAMS_USE_FAKE_API: 'true' };
  // ...
});
// remove the afterEach block entirely
```

user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
```diff
@@ -145,6 +145,31 @@ describe('Privileged Identity screen tests', () => {
...
+
+    const { rerender } = render(<PrivilegedIdentity />);
+
+    expect(
+      screen.getByTestId('alert-container-privileged-identity-disabled-alert'),
+    ).toBeInTheDocument();
+    expect(Api2.getRoleAndOfficeGroupNames).not.toHaveBeenCalled();
+
+    flagsRef['privileged-identity-management'] = true;
+    flagsSpy.mockImplementation(() => ({ ...flagsRef }));
```
diabagatekelly: **Issue 3 — MEDIUM: Redundant `mockImplementation` call**

The first closure already captures `flagsRef` by reference — mutating `flagsRef['privileged-identity-management'] = true` on the previous line automatically changes what the spy returns. Re-registering an identical function does nothing but implies the first closure was a snapshot, which could confuse future readers.

Consider the cleaner declarative form instead:

```ts
vi.spyOn(FeatureFlagHook, 'default')
  .mockReturnValueOnce({ 'privileged-identity-management': false })
  .mockReturnValueOnce({ 'privileged-identity-management': true });
```

user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
```diff
@@ -145,6 +145,31 @@ describe('Privileged Identity screen tests', () => {
...
+    rerender(<PrivilegedIdentity />);
+
+    await waitFor(() => {
+      expect(document.querySelector('.loading-spinner-caption')).not.toBeInTheDocument();
+    });
+
+    expect(Api2.getRoleAndOfficeGroupNames).toHaveBeenCalled();
```
diabagatekelly: **Issue 4 — MEDIUM: Weak assertion — `.toHaveBeenCalled()` vs `.toHaveBeenCalledTimes(1)`**

`.toHaveBeenCalled()` passes even if the API fires 5 times. A buggy dep array that causes the effect to fire twice would still pass here. `.toHaveBeenCalledTimes(1)` catches double-invocation regressions.

user-interface/src/admin/privileged-identity/PrivilegedIdentity.tsx
```diff
@@ -56,6 +56,42 @@ export function PrivilegedIdentity() {
 
   const alert = useGlobalAlert();
 
+  useEffect(() => {
+    if (!flags[PRIVILEGED_IDENTITY_MANAGEMENT]) return;
+    setIsLoaded(false);
+    Api2.getRoleAndOfficeGroupNames().then((groups) => {
```
diabagatekelly: **Issue 5 — MEDIUM: Unhandled promise rejection leaves an infinite loading spinner**

Neither `.then()` chain here nor the nested `getPrivilegedIdentityUsers()` call has a `.catch()`. If either rejects, `setIsLoaded(true)` is never called and the user sees the spinner forever with no error feedback. Pre-existing issue, but now more visible since the effect can re-fire on flag transition.

```ts
useEffect(() => {
  if (!privilegedIdentityEnabled) return;
  setIsLoaded(false);
  (async () => {
    try {
      const groups = await Api2.getRoleAndOfficeGroupNames();
      // ... process groups ...
      const res = await Api2.getPrivilegedIdentityUsers();
      setUserList(res.data.sort(sortUserList));
      setIsLoaded(true);
    } catch (e) {
      alert?.error(`Failed to load Privileged Identity data. ${(e as Error).message}`);
      setIsLoaded(true);
    }
  })();
}, [privilegedIdentityEnabled]);
```

user-interface/src/lib/testing/testing-utilities.ts
```diff
@@ -146,15 +146,27 @@ async function toggleComboBoxItemSelection(id: string, itemIndex: number = 0, se
...
+      }
     } else {
-      expect(currentItem).not.toHaveClass('selected');
+      // Item left the DOM (single-select closed the dropdown) — check input value
+      const input = document.querySelector(`#${id}-combo-box-input`) as HTMLInputElement | null;
+      expect(input).not.toBeNull();
+      if (selected) {
+        expect(input!.value).toBe(itemLabel);
+      } else {
+        expect(input!.value).not.toBe(itemLabel);
```
diabagatekelly: **Issue 6 — HIGH: Fragile assertion — `not.toBe(itemLabel)` passes vacuously**

When `selected=false` and the item left the DOM, this passes for any value that isn't `itemLabel` — an empty string, a different selection, anything. It doesn't confirm deselection actually happened.

Since a single-select ComboBox clears the input on deselection, assert the specific expected postcondition:

```ts
expect(input!.value).toBe('');
```

This turns a vacuous negative into a meaningful positive assertion.

user-interface/src/admin/privileged-identity/PrivilegedIdentity.tsx
mattstankey: Fixed — extracted to `const privilegedIdentityEnabled = flags[PRIVILEGED_IDENTITY_MANAGEMENT]` and used that variable in both the effect guard and the dep array.

user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
mattstankey: Fixed — moved `vi.restoreAllMocks()` to the top of `beforeEach` and removed the `afterEach` block entirely.

user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
mattstankey: Fixed — rewrote using `.mockReturnValueOnce({ false }).mockReturnValue({ true })` so the second render automatically gets `true` without needing to re-register the closure.

user-interface/src/admin/privileged-identity/PrivilegedIdentity.test.tsx
mattstankey: Fixed — changed to `.toHaveBeenCalledTimes(1)` to catch double-invocation from a misconfigured dep array.

user-interface/src/admin/privileged-identity/PrivilegedIdentity.tsx
mattstankey: Fixed — refactored to `async/await` with a `try/catch`. On error, shows an alert and calls `setIsLoaded(true)` so the spinner never hangs.

user-interface/src/lib/testing/testing-utilities.ts
mattstankey: Fixed — changed to `expect(input!.value).toBe("")` for the deselected case, which is the precise postcondition for a single-select ComboBox after deselection.

diabagatekelly (approved):

PR #2550, opened 2026-06-23
CAMS-775 Fix pagination rendering 0 when totalPages is zero or undefined

cruzjone-flexion:
# Bug

When `totalPages` was `0` or `undefined`, the Trustees list pagination section rendered a visible `0` character in the HTML. This is the classic React `0 &&` pitfall where a falsy numeric value short-circuits the expression and JSX renders it as a literal character.

Additionally, changing the sort direction or navigating to a new page did not provide good UX — sort left users on a stale mid-list page, and page navigation left the viewport scrolled to the middle of the previous page.

# Purpose

1. Prevent the number `0` from appearing as stray text in the UI when there are no pages to paginate.
2. Reset to page 1 when sort direction changes, consistent with how filter changes already behave.
3. Scroll to the top of the page when navigating to a new page of results.
4. Extract pagination render condition into a named boolean for readability (code review follow-up).

# Major Changes

* Replaced `paginationValues.totalPages && paginationValues.totalPages > 1` with `(paginationValues.totalPages ?? 0) > 1` — always evaluates to a boolean, so React never renders `0`
* Added `sortDirection` to the pagination-reset `useEffect` dependency array so that toggling the Name column sort returns the user to page 1
* Added `window.scrollTo({ top: 0, behavior: 'smooth' })` in `handlePaginationChange` so the viewport returns to the top on every page navigation
* Extracted `(paginationValues.totalPages ?? 0) > 1` into a named `hasMultiplePages` boolean to make the JSX render condition more readable and self-documenting

# Testing/Validation

All 82 unit tests for `TrusteesList` pass, including:
- `'changing sort direction resets to page 1'` — navigates to page 2, toggles sort, asserts page 1 content in descending order is visible
- `'scrolls to top of page when navigating to a new page'` — navigates to page 2 and asserts `window.scrollTo` was called with `{ top: 0, behavior: 'smooth' }`

# Notes

The `??` nullish coalescing operator handles the `undefined` case required by the TypeScript type, while the `> 1` comparison ensures the result is always boolean — avoiding the React rendering pitfall entirely.

The sort reset and scroll-to-top both follow patterns already established elsewhere in the component (filter changes reset pagination; `window.scrollTo` is used in other parts of the app).

# Definition of Done:

### DIFF
--- a/user-interface/src/trustees/TrusteesList.test.tsx
+++ b/user-interface/src/trustees/TrusteesList.test.tsx
@@ -2986,6 +2986,27 @@ describe('TrusteesList Component', () => {
       expect(screen.queryByText('Last000, First0')).not.toBeInTheDocument();
     });
 
+    test('scrolls to top of page when navigating to a new page', async () => {
+      const trustees = makeTrustees(50);
+      vi.spyOn(Api2, 'getTrustees').mockResolvedValue({ data: trustees });
+      const scrollToSpy = vi.spyOn(window, 'scrollTo').mockImplementation(() => {});
+
+      renderWithRouter(<TrusteesList />);
+
+      await waitFor(() => {
+        expect(screen.getByTestId('trustees-table')).toBeInTheDocument();
+      });
+
+      await userEvent.click(screen.getByTestId('pagination-button-page-2-results'));
+
+      await waitFor(() => {
+        expect(screen.queryByText('Last000, First0')).not.toBeInTheDocument();
+      });
+
+      expect(scrollToSpy).toHaveBeenCalledWith({ top: 0, behavior: 'smooth' });
+      scrollToSpy.mockRestore();
+    });
+
     test('changing status filter resets to page 1', async () => {
       const trustees = makeTrustees(50);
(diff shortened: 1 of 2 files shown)


diabagatekelly (approved):

cruzjone-flexion:
@sourcery-ai review

cruzjone-flexion:
Addressing Sourcery's suggestions:

**#3 ✅ Implemented** — Extracted `(paginationValues.totalPages ?? 0) > 1` into a named `hasMultiplePages` boolean for readability.

**#1 ⏭️ Skipped** — This is a browser-only app with no SSR. The `window.scrollTo` guard for non-DOM environments would add noise without benefit, and the test already mocks it cleanly via `vi.spyOn`.

**#2 ⏭️ Skipped** — The `useEffect` pattern is intentional and consistent with how every other pagination reset already works in this component. Moving the reset into each individual handler would touch many call sites and risk missing one — the effect-based approach is the safer, more maintainable choice here.

mattstankey (approved):

PR #2559, opened 2026-06-24
CAMS-768 Slice 2: Appointment save notification triggers

bityogi:
# Purpose

Extend trustee change notifications to fire when appointment records are created or updated, completing the second vertical slice of CAMS-768. Oversight offices are now notified of appointment-level changes (chapter, status, district, division, effective date) in addition to the profile-level changes delivered in Slice 1.

# Major Changes

* Added `subjectOverride` field to `TrusteeChangeSet` so appointment notifications use distinct subject lines ("Trustee Appointment Changed" / "New Trustee Appointment") vs profile notifications
* Created `buildAppointmentChangeSet` — a pure function that compares before/after appointment snapshots and emits per-field diffs with formatted labels
* Wired fire-and-forget notification dispatch into both `TrusteeAppointmentsUseCase.updateAppointment` and `.createAppointment` with failure isolation (try/catch — notification errors never fail the save)
* Extended the routing seed with `chapter:11`, `chapter:11-subchapter-v`, `chapter:12`, `chapter:13`, and `category:zoom-341` rows. Sub-V routes to Debbie's mailbox; zoom-341 to USTP Help

# Testing/Validation

Implemented using TDD. 18 new tests across 3 files:
- 2 tests for `subjectOverride` template behavior
- 8 tests for `buildAppointmentChangeSet` (single change, multi-field, no change, create, division stacking, court resolver, resolver fallback, status formatting)
- 5 tests for `updateAppointment` notification dispatch (dispatch, failure isolation, no-op, Sub-V routing, default-recipient fallback)
- 3 tests for `createAppointment` notification dispatch (dispatch, failure isolation, chapter routing)

All 2982 backend tests and 1095 common tests pass. Typecheck and lint clean.

# Notes

- Routing uses the appointment's own chapter for key resolution (not `resolvePrimaryChapter`) — simpler and semantically correct since we have the chapter directly available
- For creates, all fields render as additions (before = empty); for updates, only changed fields appear
- The notification try/catch in `updateAppointment` also wraps the trustee-name read (needed for the notification subject) so existing tests that don't mock it are unaffected
- Feature scenarios covered: "Changing an appointment chapter from CH7 to Sub-V routes to debbie's mailbox"; zoom-341 routing activated by seed row
- `vi.restoreAllMocks()` placed in `beforeEach` per project convention (not `afterEach`)

# Definition of Done:

## Summary by Sourcery

Add trustee appointment-level notification generation and routing when appointments are created or updated.

New Features:
- Send trustee change notifications for appointment creations and updates using appointment-specific change sets and subject lines.

Enhancements:
- Introduce an appointment change-set builder to derive per-field diffs and notification metadata from before/after appointment snapshots.
- Ensure notification dispatch failures for appointment events are logged but do not affect appointment persistence.
- Extend notification routing seed data to cover additional chapters and a new zoom-341 support category.

Tests:
- Add tests covering appointment notification dispatch behavior, failure isolation, routing, and template subject override handling.

### DIFF
--- a/backend/lib/use-cases/notifications/templates/trustee-change-template.ts
+++ b/backend/lib/use-cases/notifications/templates/trustee-change-template.ts
@@ -113,8 +113,12 @@ export function compileTrusteeChangeTemplate(changeSet: TrusteeChangeSet): Compi
     .replaceAll('{{appointment_info_rows}}', appointmentRows)
     .replaceAll('{{meeting_info_rows}}', meetingRows);
 
+  const subject =
+    changeSet.subjectOverride ??
+    `Trustee Information Changed: ${changeSet.trusteeName.replace(/[\r\n]/g, ' ')}`;
+
   return {
-    subject: `Trustee Information Changed: ${changeSet.trusteeName.replace(/[\r\n]/g, ' ')}`,
+    subject,
     html: stripTrailingWhitespace(rendered),
     text: buildPlaintext(changeSet),
   };
--- a/backend/lib/use-cases/notifications/templates/trustee-change-template.ts
+++ b/backend/lib/use-cases/notifications/templates/trustee-change-template.ts
@@ -113,9 +113,9 @@ export function compileTrusteeChangeTemplate(changeSet: TrusteeChangeSet): Compi
     .replaceAll('{{appointment_info_rows}}', appointmentRows)
     .replaceAll('{{meeting_info_rows}}', meetingRows);
 
-  const subject =
-    changeSet.subjectOverride ??
-    `Trustee Information Changed: ${changeSet.trusteeName.replace(/[\r\n]/g, ' ')}`;
+  const rawSubject =
+    changeSet.subjectOverride ?? `Trustee Information Changed: ${changeSet.trusteeName}`;
+  const subject = rawSubject.replace(/[\r\n]/g, ' ');
 
   return {
     subject,
--- a/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
+++ b/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
@@ -148,4 +148,41 @@ describe('buildAppointmentChangeSet', () => {
     expect(statusField!.before).toBe('Active');
     expect(statusField!.after).toBe('Voluntarily Suspended');
   });
+
+  test('detects an appointmentType change', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: baseSnapshot,
+      after: { ...baseSnapshot, appointmentType: 'off-panel' },
+    });
+
+    expect(result.fields).toHaveLength(1);
+    expect(result.fields[0]).toEqual({
+      label: 'Appointment Type',
+      before: 'Panel',
+      after: 'Off Panel',
+      category: 'profile',
+      section: 'appointment',
+    });
+    expect(result.subjectOverride).toBe('Trustee Appointment Changed: Henry Green');
+  });
+
+  test('detects an appointedDate change', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: baseSnapshot,
+      after: { ...baseSnapshot, appointedDate: '2025-03-01' },
+    });
+
+    expect(result.fields).toHaveLength(1);
+    expect(result.fields[0]).toEqual({
+      label: 'Appointed Date',
+      before: '2024-01-15',
+      after: '2025-03-01',
+      category: 'profile',
+      section: 'appointment',
+    });
+  });
 });
--- a/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
+++ b/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
@@ -1,0 +1,151 @@
+import {
+  buildAppointmentChangeSet,
+  AppointmentFieldSnapshot,
+} from './build-appointment-change-set';
+
+describe('buildAppointmentChangeSet', () => {
+  const baseSnapshot: AppointmentFieldSnapshot = {
+    chapter: '7',
+    appointmentType: 'panel',
+    courtId: 'court-001',
+    divisionCodes: ['001'],
+    appointedDate: '2024-01-15',
+    status: 'active',
+    effectiveDate: '2024-01-15',
+  };
+
+  test('detects a single field change (chapter CH7 to Sub-V)', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: baseSnapshot,
+      after: { ...baseSnapshot, chapter: '11-subchapter-v' },
+    });
+
+    expect(result.fields).toHaveLength(1);
+    expect(result.fields[0]).toEqual({
+      label: 'Chapter',
+      before: '7',
+      after: '11 Subchapter V',
+      category: 'profile',
+      section: 'appointment',
+    });
+    expect(result.subjectOverride).toBe('Trustee Appointment Changed: Henry Green');
+    expect(result.primaryChapter).toBe('11-subchapter-v');
+  });
+
+  test('detects multiple field changes (status + effective date)', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: baseSnapshot,
+      after: {
+        ...baseSnapshot,
+        status: 'inactive',
+        effectiveDate: '2025-06-01',
+      },
+    });
+
+    expect(result.fields).toHaveLength(2);
+    expect(result.fields[0].label).toBe('Status');
+    expect(result.fields[0].before).toBe('Active');
+    expect(result.fields[0].after).toBe('Inactive');
+    expect(result.fields[1].label).toBe('Status Effective Date');
+    expect(result.subjectOverride).toBe('Trustee Appointment Changed: Henry Green');
+  });
+
+  test('returns empty fields array when nothing changes', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: baseSnapshot,
+      after: { ...baseSnapshot },
+    });
+
+    expect(result.fields).toHaveLength(0);
+    expect(result.subjectOverride).toBeUndefined();
+  });
+
+  test('emits all fields as new when before is undefined (create)', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: undefined,
+      after: baseSnapshot,
+    });
+
+    expect(result.fields.length).toBeGreaterThanOrEqual(7);
+    for (const field of result.fields) {
+      expect(field.before).toBe('');
+      expect(field.after).not.toBe('');
+      expect(field.category).toBe('profile');
+      expect(field.section).toBe('appointment');
+    }
+    expect(result.subjectOverride).toBe('New Trustee Appointment: Henry Green');
+    expect(result.primaryChapter).toBe('7');
+  });
+
+  test('renders multiple divisions comma-separated with stackValues', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: { ...baseSnapshot, divisionCodes: ['001'] },
+      after: { ...baseSnapshot, divisionCodes: ['001', '002'] },
+    });
+
+    const divisionField = result.fields.find((f) => f.label === 'Division');
+    expect(divisionField).toBeDefined();
+    expect(divisionField!.before).toBe('001');
+    expect(divisionField!.after).toBe('001, 002');
+    expect(divisionField!.stackValues).toBe(true);
+  });
+
+  test('uses courtNameResolver for District field', () => {
+    const resolver = (courtId: string) =>
+      courtId === 'court-xyz' ? 'Eastern District of Missouri' : undefined;
+
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: { ...baseSnapshot, courtId: 'court-001' },
+      after: { ...baseSnapshot, courtId: 'court-xyz' },
+      courtNameResolver: resolver,
+    });
+
+    const districtField = result.fields.find((f) => f.label === 'District');
+    expect(districtField).toBeDefined();
+    expect(districtField!.before).toBe('court-001');
+    expect(districtField!.after).toBe('Eastern District of Missouri');
+  });
+
+  test('falls back to raw courtId when resolver returns undefined', () => {
+    const resolver = () => undefined;
+
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: { ...baseSnapshot, courtId: 'court-A' },
+      after: { ...baseSnapshot, courtId: 'court-B' },
+      courtNameResolver: resolver,
+    });
+
+    const districtField = result.fields.find((f) => f.label === 'District');
+    expect(districtField).toBeDefined();
+    expect(districtField!.before).toBe('court-A');
+    expect(districtField!.after).toBe('court-B');
+  });
+
+  test('formats hyphenated status values as title case', () => {
+    const result = buildAppointmentChangeSet({
+      trusteeId: 'trustee-1',
+      trusteeName: 'Henry Green',
+      before: { ...baseSnapshot, status: 'active' },
+      after: { ...baseSnapshot, status: 'voluntarily-suspended' },
+    });
+
+    const statusField = result.fields.find((f) => f.label === 'Status');
+    expect(statusField).toBeDefined();
+    expect(statusField!.before).toBe('Active');
+    expect(statusField!.after).toBe('Voluntarily Suspended');
+  });
+});
--- a/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
+++ b/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
@@ -6,6 +6,7 @@ import {
   formatChapterType,
   formatAppointmentType,
 } from '@common/cams/trustees';
+import { formatAppointmentStatus } from '@common/cams/trustee-appointments';
 
 export type AppointmentFieldSnapshot = {
   chapter: AppointmentChapterType;
@@ -89,8 +90,8 @@ export function buildAppointmentChangeSet(params: {
     });
   }
 
-  const beforeStatus = before ? formatStatus(before.status) : '';
-  const afterStatus = formatStatus(after.status);
+  const beforeStatus = before ? formatAppointmentStatus(before.status) : '';
+  const afterStatus = formatAppointmentStatus(after.status);
   if (beforeStatus !== afterStatus) {
     fields.push({
       label: 'Status',
@@ -127,11 +128,4 @@ export function buildAppointmentChangeSet(params: {
     primaryChapter: after.chapter,
     subjectOverride,
   };
-}
-
-function formatStatus(status: AppointmentStatus): string {
-  return status
-    .split('-')
-    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-    .join(' ');
 }
--- a/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
+++ b/backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
@@ -18,6 +18,23 @@ export type AppointmentFieldSnapshot = {
   effectiveDate: string;
 };
 
+function diffField(
+  label: string,
+  beforeValue: string,
+  afterValue: string,
+  options?: { stackValues?: boolean },
+): TrusteeChangeField | undefined {
+  if (beforeValue === afterValue) return undefined;
+  return {
+    label,
+    before: beforeValue,
+    after: afterValue,
+    category: 'profile',
+    section: 'appointment',
+    ...(options?.stackValues && { stackValues: true }),
+  };
+}
+
 export function buildAppointmentChangeSet(params: {
   trusteeId: string;
   trusteeName: string;
@@ -26,93 +43,40 @@ export function buildAppointmentChangeSet(params: {
   courtNameResolver?: (courtId: string) => string | undefined;
 }): TrusteeChangeSet {
   const { trusteeId, trusteeName, before, after, courtNameResolver } = params;
-  const fields: TrusteeChangeField[] = [];
-
-  const beforeChapter = before ? formatChapterType(before.chapter) : '';
-  const afterChapter = formatChapterType(after.chapter);
-  if (beforeChapter !== afterChapter) {
-    fields.push({
-      label: 'Chapter',
-      before: beforeChapter,
-      after: afterChapter,
-      category: 'profile',
-      section: 'appointment',
-    });
-  }
-
-  const beforeType = before ? formatAppointmentType(before.appointmentType) : '';
-  const afterType = formatAppointmentType(after.appointmentType);
-  if (beforeType !== afterType) {
-    fields.push({
-      label: 'Appointment Type',
-      before: beforeType,
-      after: afterType,
-      category: 'profile',
-      section: 'appointment',
-    });
-  }
-
   const resolveCourtName = courtNameResolver ?? ((id: string) => id);
-  const beforeCourt = before ? (resolveCourtName(before.courtId) ?? before.courtId) : '';
-  const afterCourt = resolveCourtName(after.courtId) ?? after.courtId;
-  if (beforeCourt !== afterCourt) {
-    fields.push({
-      label: 'District',
-      before: beforeCourt,
-      after: afterCourt,
-      category: 'profile',
-      section: 'appointment',
-    });
-  }
 
-  const beforeDivisions = before?.divisionCodes?.join(', ') ?? '';
-  const afterDivisions = after.divisionCodes?.join(', ') ?? '';
-  if (beforeDivisions !== afterDivisions) {
-    fields.push({
-      label: 'Division',
-      before: beforeDivisions,
-      after: afterDivisions,
-      category: 'profile',
-      section: 'appointment',
-      stackValues: true,
-    });
-  }
+  const candidates = [
+    diffField(
+      'Chapter',
+      before ? formatChapterType(before.chapter) : '',
+      formatChapterType(after.chapter),
+    ),
+    diffField(
+      'Appointment Type',
+      before ? formatAppointmentType(before.appointmentType) : '',
+      formatAppointmentType(after.appointmentType),
+    ),
+    diffField(
+      'District',
+      before ? (resolveCourtName(before.courtId) ?? before.courtId) : '',
+      resolveCourtName(after.courtId) ?? after.courtId,
+    ),
+    diffField(
+      'Division',
+      before?.divisionCodes?.join(', ') ?? '',
+      after.divisionCodes?.join(', ') ?? '',
+      { stackValues: true },
+    ),
+    diffField('Appointed Date', before?.appointedDate ?? '', after.appointedDate),
+    diffField(
+      'Status',
+      before ? formatAppointmentStatus(before.status) : '',
+      formatAppointmentStatus(after.status),
+    ),
+    diffField('Status Effective Date', before?.effectiveDate ?? '', after.effectiveDate),
+  ];
 
-  const beforeDate = before?.appointedDate ?? '';
-  const afterDate = after.appointedDate;
-  if (beforeDate !== afterDate) {
-    fields.push({
-      label: 'Appointed Date',
-      before: beforeDate,
-      after: afterDate,
-      category: 'profile',
-      section: 'appointment',
-    });
-  }
-
-  const beforeStatus = before ? formatAppointmentStatus(before.status) : '';
-  const afterStatus = formatAppointmentStatus(after.status);
-  if (beforeStatus !== afterStatus) {
-    fields.push({
-      label: 'Status',
-      before: beforeStatus,
-      after: afterStatus,
-      category: 'profile',
-      section: 'appointment',
-    });
-  }
-
-  const beforeEffective = before?.effectiveDate ?? '';
-  const afterEffective = after.effectiveDate;
-  if (beforeEffective !== afterEffective) {
-    fields.push({
-      label: 'Status Effective Date',
-      before: beforeEffective,
-      after: afterEffective,
-      category: 'profile',
-      section: 'appointment',
-    });
-  }
+  const fields = candidates.filter((f): f is TrusteeChangeField => f !== undefined);
 
   const isCreate = !before;
   const subjectOverride = isCreate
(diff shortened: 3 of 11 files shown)


backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
```diff
@@ -0,0 +1,151 @@
...
+      before: { ...baseSnapshot, status: 'active' },
+      after: { ...baseSnapshot, status: 'voluntarily-suspended' },
+    });
+
+    const statusField = result.fields.find((f) => f.label === 'Status');
+    expect(statusField).toBeDefined();
+    expect(statusField!.before).toBe('Active');
+    expect(statusField!.after).toBe('Voluntarily Suspended');
+  });
+});
```
mattstankey, line 188: The test file covers chapter, status, effectiveDate, divisions, district, and status formatting — but `appointmentType` and `appointedDate` changes have no dedicated tests. Both fields go through the same diff-and-push pattern as the others, so coverage should be straightforward:

```ts
test("detects an appointmentType change", () => {
  const result = buildAppointmentChangeSet({
    trusteeId: "trustee-1",
    trusteeName: "Henry Green",
    before: baseSnapshot,
    after: { ...baseSnapshot, appointmentType: "standing" },
  });
  expect(result.fields).toHaveLength(1);
  expect(result.fields[0].label).toBe("Appointment Type");
});

test("detects an appointedDate change", () => {
  const result = buildAppointmentChangeSet({
    trusteeId: "trustee-1",
    trusteeName: "Henry Green",
    before: baseSnapshot,
    after: { ...baseSnapshot, appointedDate: "2025-03-01" },
  });
  expect(result.fields).toHaveLength(1);
  expect(result.fields[0].label).toBe("Appointed Date");
});
```

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
```diff
@@ -1137,4 +1138,389 @@ describe('TrusteeAppointmentsUseCase tests', () => {
...
+        chapter: '11-subchapter-v' as const,
+        appointmentType: 'pool' as const,
+      };
+
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValueOnce(existingAppointment);
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValueOnce(mockTrustee);
+      vi.spyOn(MockMongoRepository.prototype, 'updateAppointment').mockResolvedValue(
+        updatedAppointment,
+      );
+
```
mattstankey, line 1247: The two chained `mockResolvedValueOnce` calls (appointment read, then trustee read) are call-order-dependent. If the implementation ever swaps those two reads (e.g. to fetch the trustee first for validation), the tests will silently feed the wrong mock value to the wrong call and produce misleading failures.

Consider naming the mocks more explicitly or using `mockImplementation` keyed on the document type — or at minimum add a comment noting the expected call order so a future refactor knows to update the stubs.

mattstankey (commented):
Two things worth addressing before merge:

1. **Test coverage gaps** in `build-appointment-change-set.test.ts` — `appointmentType` and `appointedDate` field changes have no dedicated tests (see inline comments on that file).
2. **Call-order-dependent `read` mocks** in the `updateAppointment` notification tests — see inline comment on `trustee-appointments.test.ts`.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
```diff
@@ -247,6 +251,36 @@ export class TrusteeAppointmentsUseCase {
 
       await this.trusteesRepository.createTrusteeHistory(history as Creatable<TrusteeHistory>);
 
+      try {
+        const courts = await this.courtsUseCase.getCourts(context);
```
diabagatekelly: **[HIGH] Double `getCourts` call — redundant round-trip**

`buildAppointmentHistory` (called ~20 lines above at `this.buildAppointmentHistory(...)`) already calls `this.courtsUseCase.getCourts(context)` internally. This is a second, redundant hit to the courts service on every `createAppointment` request.

Extract courts resolution before `buildAppointmentHistory` and thread the resolver into both callers:

```typescript
const courts = await this.courtsUseCase.getCourts(context);
const courtNameResolver = (courtId: string) => this.findCourtDistrict(courts, courtId);

const history = await this.buildAppointmentHistory(
  context, trusteeId, createdAppointment.id, userReference,
  undefined, snapshotFrom(createdAppointment),
  courtNameResolver, // thread in
);
// reuse courtNameResolver below — no second fetch needed
```

Aligns with how `getTrusteeAppointments` already handles court enrichment (single fetch, shared result).

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
```diff
@@ -318,6 +352,45 @@ export class TrusteeAppointmentsUseCase {
         );
 
         await this.trusteesRepository.createTrusteeHistory(history as Creatable<TrusteeHistory>);
+
+        try {
+          const trustee = await this.trusteesRepository.read(trusteeId);
```
diabagatekelly: **[HIGH] `trusteesRepository.read` and `getCourts` run sequentially — two independent I/O calls**

These can be parallelized:

```typescript
const [trustee, courts] = await Promise.all([
  this.trusteesRepository.read(trusteeId),
  this.courtsUseCase.getCourts(context),
]);
```

Additionally, `buildAppointmentHistory` (called a few lines above) fetches courts internally, making `getCourts` here a second hit to the same service on every changed-appointment update. Courts should be fetched once per request and shared with both `buildAppointmentHistory` and this notification block.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
```diff
@@ -0,0 +1,137 @@
...
+import {
+  AppointmentChapterType,
+  AppointmentType,
+  AppointmentStatus,
+  formatChapterType,
+  formatAppointmentType,
+} from '@common/cams/trustees';
+
+export type AppointmentFieldSnapshot = {
+  chapter: AppointmentChapterType;
```
diabagatekelly, line 12: **[HIGH] Parallel type definition — `AppointmentSnapshot` in `trustee-appointments.ts` is structurally identical**

`AppointmentSnapshot` (private type, ~line 31 of `trustee-appointments.ts`) has these same seven fields plus a legacy `divisionCode: string`. The use-case then manually re-maps `TrusteeAppointment` data into `AppointmentSnapshot`, and again into `AppointmentFieldSnapshot`, six times total (lines ~241–249, 261–270, 332–341, 342–351, 363–371, 372–381).

Consider importing this type in `trustee-appointments.ts` and deleting the private `AppointmentSnapshot`. A small private helper eliminates all six copy-paste blocks:

```typescript
function snapshotFrom(apt: TrusteeAppointment): AppointmentFieldSnapshot {
  return {
    chapter: apt.chapter,
    appointmentType: apt.appointmentType,
    courtId: apt.courtId,
    divisionCodes: apt.divisionCodes,
    appointedDate: apt.appointedDate,
    status: apt.status,
    effectiveDate: apt.effectiveDate,
  };
}
```

Reduces six copy-paste object literals to six single-line calls.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
```diff
@@ -247,6 +251,36 @@ export class TrusteeAppointmentsUseCase {
...
+          before: undefined,
+          after: {
+            chapter: createdAppointment.chapter,
+            appointmentType: createdAppointment.appointmentType,
+            courtId: createdAppointment.courtId,
+            divisionCodes: createdAppointment.divisionCodes,
+            appointedDate: createdAppointment.appointedDate,
+            status: createdAppointment.status,
+            effectiveDate: createdAppointment.effectiveDate,
+          },
```
diabagatekelly: **[HIGH] Manual 7-field object literal — one of six copies in this file**

This same field-by-field extraction of `TrusteeAppointment` → `AppointmentFieldSnapshot` appears six times: here, at lines ~362–370 (before snapshot), ~371–379 (after snapshot) in `updateAppointment`, and three more in `buildAppointmentHistory`. See comment on `AppointmentFieldSnapshot` in `build-appointment-change-set.ts` for the `snapshotFrom()` helper that eliminates all six.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
```diff
@@ -1137,4 +1138,389 @@ describe('TrusteeAppointmentsUseCase tests', () => {
...
+    const appointmentId = 'appointment-notify-1';
+
+    beforeEach(async () => {
+      vi.restoreAllMocks();
+      context = await createMockApplicationContext();
+      trusteeAppointmentsUseCase = new TrusteeAppointmentsUseCase(context);
+      MockNotificationGateway.getInstance().clear();
+
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue();
+      vi.spyOn(trusteeAppointmentsUseCase['courtsUseCase'], 'getCourts').mockResolvedValue([]);
```
diabagatekelly: **[HIGH] Private field access via bracket notation**

`trusteeAppointmentsUseCase['courtsUseCase']` bypasses TypeScript's access control and couples this test to `TrusteeAppointmentsUseCase`'s internal object graph. If the private property is renamed or the courts lookup is refactored, this spy setup (and 13 identical ones in the file) breaks even though the observable behavior of the use case is unchanged.

Replace with prototype-level spying, which stays at the public API boundary:

```typescript
vi.spyOn(CourtsUseCase.prototype, 'getCourts').mockResolvedValue([]);
```

No bracket access needed, and the test remains valid through internal refactors.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
```diff
@@ -1137,4 +1138,389 @@ describe('TrusteeAppointmentsUseCase tests', () => {
...
+    const trusteeId = 'trustee-notify-create';
+
+    beforeEach(async () => {
+      vi.restoreAllMocks();
+      context = await createMockApplicationContext();
+      trusteeAppointmentsUseCase = new TrusteeAppointmentsUseCase(context);
+      MockNotificationGateway.getInstance().clear();
+
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue();
+      vi.spyOn(trusteeAppointmentsUseCase['courtsUseCase'], 'getCourts').mockResolvedValue([]);
```
diabagatekelly: **[HIGH] Private field access via bracket notation** (same as `updateAppointment` notification block)

Same issue as line 1153 — replace with:

```typescript
vi.spyOn(CourtsUseCase.prototype, 'getCourts').mockResolvedValue([]);
```

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
```diff
@@ -0,0 +1,137 @@
...
+  return {
+    trusteeId,
+    trusteeName,
+    fields,
+    primaryChapter: after.chapter,
+    subjectOverride,
+  };
+}
+
+function formatStatus(status: AppointmentStatus): string {
```
diabagatekelly: **[MEDIUM] `formatStatus` belongs in `@common/cams/trustees` alongside `formatChapterType` and `formatAppointmentType`**

All three functions convert domain enum values to human-readable labels. `formatChapterType` and `formatAppointmentType` are already exported from `@common/cams/trustees` and imported at the top of this file. Keeping `formatStatus` private here means any future caller re-implements the same title-case logic.

Move it to `@common/cams/trustees`, export it next to its siblings, and import it here the same way.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
diabagatekelly, line 188: **[MEDIUM] `appointmentType` field change has no dedicated diff test**

The suite covers chapter (test 1), status+effectiveDate (test 2), all-fields create (test 4), divisionCodes (test 5), courtId (tests 6–7), and status formatting (test 8). There is no test that fires only an `appointmentType` change and asserts the emitted `label`, `before`, and `after` values. A regression in `formatAppointmentType` would not be caught by a targeted assertion.

Suggested addition:

```typescript
test('detects a single field change (appointmentType panel to case-by-case)', () => {
  const result = buildAppointmentChangeSet({
    trusteeId: 'trustee-1',
    trusteeName: 'Henry Green',
    before: baseSnapshot,
    after: { ...baseSnapshot, appointmentType: 'case-by-case' },
  });

  expect(result.fields).toHaveLength(1);
  expect(result.fields[0]).toEqual({
    label: 'Appointment Type',
    before: 'Panel',
    after: 'Case by Case',
    category: 'profile',
    section: 'appointment',
  });
  expect(result.subjectOverride).toBe('Trustee Appointment Changed: Henry Green');
});
```

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
```diff
@@ -1137,4 +1138,389 @@ describe('TrusteeAppointmentsUseCase tests', () => {
       expect(actualError.message).toContain('At least one division must be specified');
     });
   });
+
+  describe('updateAppointment notification dispatch (CAMS-768 Slice 2)', () => {
```
diabagatekelly: **[LOW] Jira ticket and slice reference in `describe` label**

`(CAMS-768 Slice 2)` will be meaningless noise once the ticket closes and "Slice 2" loses shared context. Describe labels should describe the behavior under test, not their origin ticket.

```typescript
describe('updateAppointment notification dispatch', () => {
```

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
```diff
@@ -1137,4 +1138,389 @@ describe('TrusteeAppointmentsUseCase tests', () => {
...
+        effectiveDate: '2024-01-15',
+      });
+
+      const recorded = MockNotificationGateway.getInstance().getRecorded();
+      expect(recorded).toHaveLength(1);
+      expect(recorded[0].to).toBe('default-oversight@example.test');
+    });
+  });
+
+  describe('createAppointment notification dispatch (CAMS-768 Slice 2)', () => {
```
diabagatekelly: **[LOW] Jira ticket and slice reference in `describe` label** (same as line 1142)

```typescript
describe('createAppointment notification dispatch', () => {
```

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
```diff
@@ -1137,4 +1138,389 @@ describe('TrusteeAppointmentsUseCase tests', () => {
...
+        recipientAddress: 'default-oversight@example.test',
+        displayName: 'Default Oversight',
+      });
+    });
+
+    afterEach(() => {
+      MockNotificationGateway.getInstance().clear();
+    });
+
+    test('dispatches one notification when appointment chapter changes', async () => {
```
diabagatekelly: **[LOW] This test and `'routes to Sub-V mailbox when chapter changes to 11-subchapter-v'` (line 1302) exercise the same scenario**

Both set up a `chapter:7 → chapter:11-subchapter-v` change against the same `beforeEach` stubs. This test omits the `to` assertion; the routing test adds it — making this one a strict subset of the other.

Either merge them into one assertion covering `subject`, `html`, and `to`, or differentiate scenarios so both tests exercise genuinely distinct paths (e.g., use a different chapter change here).

diabagatekelly (changes requested):
## Review Notes — Issues on pre-existing or out-of-diff lines

### [HIGH] Issue 2 — Dead mock: `updateAppointment` "not found" test proves nothing about its stated scenario (`trustee-appointments.test.ts:366–389`)

In `updateAppointment`, production code calls `this.trusteeAppointmentsRepository.read(trusteeId, appointmentId)` **before** `updateAppointment`. In this test, `read` is not mocked, so un-mocked `MockMongoRepository.read` throws `"Method not implemented"` immediately. The `updateAppointment` mock (line 369) is dead — it is never reached. The test passes, but it is testing the `read` failure path, not the scenario it claims to test. If `updateAppointment` were changed to silently return null instead of throwing, this test would still pass.

```typescript
// Fix — mock read to succeed, let update fail:
vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(existingAppointment);
vi.spyOn(MockMongoRepository.prototype, 'updateAppointment').mockRejectedValue(repositoryError);
```

---

### [MEDIUM] Issue 5 — `vi.restoreAllMocks()` in `afterEach` instead of `beforeEach` in 6 pre-existing describe blocks (`trustee-appointments.test.ts:22, 112, 324, 679, 743, 1002`)

The two new notification dispatch blocks (lines 1146, 1626 in this PR) correctly use `beforeEach`. The six pre-existing blocks (`getTrusteeAppointments`, `createAppointment`, `updateAppointment`, `createAppointment audit history`, `hasAppointmentChanged with divisionCodes`, `multi-division support`) still use `afterEach`. The project convention is `beforeEach` because `afterEach` is skipped when a test throws, allowing mock state to leak into the next test in the block.

Quick mechanical fix — move `vi.restoreAllMocks()` from `afterEach` to `beforeEach` in all six blocks.

---

### [MEDIUM] Issue 8 — What-comments and `courtDivisionName: undefined` noise (pre-existing lines in `trustee-appointments.ts`)

Six inline comments narrate code that names itself:
- `// Verify trustee exists` (line 165) — the `trusteesRepository.read` call says this
- `// Enrich appointments with court district name` (line 179) — `enrichedAppointments` + `findCourtDistrict` say this
- `// Normalize data (convert old format to new format if needed)` (lines 221, 307) — the method is named `normalizeAppointmentData`
- `// If new format is provided, use it` / `// If old format...` / `// Neither provided...` (lines 70, 79, 88) — the if/else if/return structure is self-documenting

`courtDivisionName: undefined, // Division names are not used per product requirements` also appears 5 times across `trustee-appointments.ts` and `trustee-appointments.test.ts`. The comment is worth having once at the type definition; at every call site it's noise.

Neither item blocks merge — noting for a follow-up cleanup commit.

---

### [LOW] Issue 11 — Routing comment in `resolveRecipientForCategory` is longer than the logic it documents (`trustee-change-notification.ts:77–81`, not in this diff)

The 5-line comment explaining the resolution order is longer than the 8 lines of code. The "Slice 1/Slice 2" distinction is historical context that belongs in git history, not source. A one-liner suffices:

```typescript
// zoom-341 fields route to a category key; profile fields route by chapter; both fall back to 'default'.
```

Can land in a follow-up tidy commit.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (beforeChapter !== afterChapter) {
+    fields.push({
+      label: 'Chapter',
+      before: beforeChapter,
+      after: afterChapter,
+      category: 'profile',
+      section: 'appointment',
+    });
+  }
+
```
fmaddenflx: Because each of these conditions for chapter, type,  courtName, division, etc. etc. are all doing essentially the same thing, let's break these out into a helper function and call it for each of the various fields we want to check.  The function can take a before and after value, and a label, and either return a value we can push onto the fields array, or we pass it the array and have it push the value itself.  However, I would recommend we just return a value and push it to fields ourselves in the parent function, avoiding a function that modifies and returns one of it's parameters.

dev-tools/db_scripts/scenarios/notification-routing.ts
```diff
@@ -5,9 +5,9 @@
  * Seeds the NOTIFICATION_ROUTING collection used by
  * TrusteeChangeNotificationUseCase to resolve email recipients.
  *
- * Slice 1 seeds two rows: chapter:7 (CH7 oversight mailbox) and default
- * (catch-all). Slice 2 will extend with chapter:11, chapter:11-subchapter-v,
- * chapter:12, chapter:13, and category:zoom-341.
+ * Slice 1: chapter:7 and default.
+ * Slice 2: chapter:11, chapter:11-subchapter-v, chapter:12, chapter:13,
+ *           and category:zoom-341.
```
fmaddenflx, line 10: I had some thoughts as I was going through this PR:

1. We need to put this entire feature behind a feature flag and we should do that sooner rather than later, perhaps as a part of slice 2.  Then we’ll need to make sure all further work is done behind the feature flag.
2. if it wasn’t in the plan yet, we need to make sure that the agent executes the plan by putting the configuration for email addresses and related notification configuration in a cosmos collection and providing a way for an administrator to set and modify those values.  They should be initially unset, and the notifications features should be turned off.  Might be a good idea to add an on/off switch in the admin for turning email notifications on and off.

dev-tools/db_scripts/scenarios/notification-routing.ts
```diff
@@ -27,6 +27,36 @@ export async function generate(_ctx: SeedContext): Promise<SeedOperation[]> {
...
+          id: 'notification-routing-chapter-11',
+          key: 'chapter:11',
+          recipientAddress: 'ch11-oversight@example.test',
+          displayName: 'CH11 Oversight Office',
+        },
+        {
+          id: 'notification-routing-chapter-11-subchapter-v',
+          key: 'chapter:11-subchapter-v',
+          recipientAddress: 'subv@example.test',
+          displayName: 'Sub-V Oversight (Debbie)',
```
fmaddenflx: Let's not use Debbie's name anywhere in the code.  In fact, we don't need to use a persons name.  We can describe a term for the recipient of Sub-V chapter 11 notifications.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
```diff
@@ -206,8 +208,10 @@ export class TrusteeAppointmentsUseCase {
     appointmentData: TrusteeAppointmentInput,
   ): Promise<TrusteeAppointment> {
     try {
```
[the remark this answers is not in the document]
fmaddenflx: Agreed.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
```diff
@@ -113,8 +113,12 @@ export function compileTrusteeChangeTemplate(changeSet: TrusteeChangeSet): Compi
     .replaceAll('{{appointment_info_rows}}', appointmentRows)
     .replaceAll('{{meeting_info_rows}}', meetingRows);
 
+  const subject =
+    changeSet.subjectOverride ??
```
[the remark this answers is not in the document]
fmaddenflx: I agree. Any values that we are passing to the email templates must first be fully sanitized to avoid any XSS or related issues of any kind.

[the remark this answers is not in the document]
bityogi: ✅ Fixed in `230bce331`. The final subject line now runs through `.replace(/[\r\n]/g, " ")` regardless of source (default or override). Added a test that verifies CRLF injection is stripped.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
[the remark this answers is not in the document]
bityogi: ✅ Done. Extracted `snapshotFrom()` helper (commit `16ac77bc1`) to eliminate 6 copy-paste object literals, and `diffField()` helper (commit `e76574d72`) that returns a field or undefined — candidates are filtered into the final array. Net -66 lines from the original implementation.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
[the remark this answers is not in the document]
bityogi: Partially addressed. We removed the duplicate test scenario and cleaned up describe labels. The shared setup factory is a valid improvement but felt like over-abstraction for the current test count — if we add more notification tests in Slice 3+, we can extract it then.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
bityogi: ✅ Added in commit `df8a7bbe4`. Both `appointmentType` (Panel → Off Panel) and `appointedDate` changes now have dedicated tests with full before/after assertions.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
bityogi: Acknowledged. The call-order dependency is a known limitation of this mocking pattern in CAMS. Adding a comment to clarify the expected order would help — but the underlying issue is that `MockMongoRepository.read` handles multiple document types via the same prototype method. For now the tests are consistent with the existing CAMS pattern (e.g., the Slice 1 trustees notification tests do the same). We can revisit if `updateAppointment` is refactored to read the trustee earlier.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
bityogi: ✅ Fixed in commit `5d7e4f7b1`. Courts are now fetched once before `buildAppointmentHistory` and passed via the new optional `courts` parameter. The notification block reuses the same array — no second round-trip.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
bityogi: ✅ Fixed in the same commit (`5d7e4f7b1`). Courts fetched once at the top of the `hasAppointmentChanged` block and shared with both `buildAppointmentHistory` and the notification block. The trustee read remains sequential inside the try/catch (failure-isolated) since it only runs when notifications are enabled and changes exist.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
bityogi: ✅ Fixed in commit `16ac77bc1`. Added a `snapshotFrom(apt: TrusteeAppointment)` function that returns `AppointmentSnapshot & AppointmentFieldSnapshot`. Both the audit history and notification change-set builders now consume it — six inline object literals replaced with single-line `snapshotFrom()` calls.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
bityogi: ✅ Resolved by the same `snapshotFrom()` helper — all six copies replaced.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
bityogi: ✅ Fixed in commit `81b863dfe`. Both notification test blocks now use `vi.spyOn(CourtsUseCase.prototype, 'getCourts')` instead of bracket-notation private field access.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
bityogi: ✅ Fixed in the same commit — both blocks updated.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
bityogi: ✅ Fixed in commit `16d4a6927`. Removed the private `formatStatus` and imported the existing `formatAppointmentStatus` from `@common/cams/trustee-appointments` — it uses an explicit lookup table which is even better than the generic title-casing approach.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.test.ts
bityogi: ✅ Added in commit `df8a7bbe4`. Dedicated test asserts `appointmentType` change: Panel → Case by Case with full field assertions.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
bityogi: ✅ Fixed in commit `a3a503a50`. Labels now just describe behavior: `updateAppointment notification dispatch` and `createAppointment notification dispatch`.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
bityogi: ✅ Fixed in the same commit.

backend/lib/use-cases/trustee-appointments/trustee-appointments.test.ts
bityogi: ✅ Differentiated in commit `a3a503a50`. The first test now exercises a **status change** (active → voluntarily-suspended) instead of duplicating the CH7→Sub-V chapter change. Each test exercises a genuinely distinct path.

backend/lib/use-cases/trustee-appointments/build-appointment-change-set.ts
bityogi: ✅ Done in commit `e76574d72`. Extracted a `diffField(label, beforeValue, afterValue, options?)` helper that returns a `TrusteeChangeField | undefined`. The main function now builds a `candidates` array and filters out undefineds. Cleaner, less repetitive, and easy to extend.

dev-tools/db_scripts/scenarios/notification-routing.ts
bityogi: ✅ Addressed both points:

1. **Feature flag**: Added `trustee-change-notifications-enabled` flag in commit `46093b88b`. All three notification dispatch sites (profile save, appointment update, appointment create) are now gated. The flag defaults to `true` in test environments but will be `false` in production (LaunchDarkly) until explicitly enabled by ops.

2. **Admin config for addresses**: The routing addresses are currently seed-driven in Cosmos (`NOTIFICATION_ROUTING` collection). Admin UI for managing them is out of scope for CAMS-768 but noted as a follow-up need. The feature flag gives ops the on/off switch in the meantime.

dev-tools/db_scripts/scenarios/notification-routing.ts
bityogi: ✅ Fixed in commit `1cf521b1e`. Changed to "Subchapter V Oversight" — no personal names in code.

backend/lib/use-cases/trustee-appointments/trustee-appointments.ts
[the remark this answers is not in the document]
bityogi: ✅ Implemented — `snapshotFrom()` helper eliminates all the duplication.

backend/lib/use-cases/notifications/templates/trustee-change-template.ts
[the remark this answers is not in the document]
bityogi: ✅ Fixed in commit `230bce331`. The CRLF sanitization now applies to the final computed subject regardless of source — both `subjectOverride` and the default `trusteeName`-based subject pass through `.replace(/[\r\n]/g, " ")` before being returned.

dev-tools/db_scripts/scenarios/notification-routing.ts
fmaddenflx: The Admin UI is not out of scope.  We have no way to configure the notification email addresses in the prod environment unless we have a UI.  Therefore, we need the UI to make this feature work for us.  We don't want the configuration hard coded and we don't want to have to use Cosmos directly to make changes.  Add the Admin UI to the scope.

fmaddenflx: Addressed. Added Admin UI as Slice 3 in the feature plan (trustee-change-notifications.feature) — inserted between the appointment trigger slice and bulk-import suppression. Renumbered downstream slices accordingly (old 3→4, 4→5, 5→6). Scope includes CRUD for routing records, master on/off toggle, initially-disabled-in-prod, email validation, backend API, and admin role gating. Tracked in beads as cams-xb6a.

PR #2564, opened 2026-06-24
CAMS-770 Add District (Division) case list filter

mattstankey:
# Purpose

USTP staff need to see how many open cases a trustee has per appointment (district/division) to identify stale appointments for termination. Without this filter, staff have no way to isolate the case list to a specific appointment and must manually cross-reference data to determine which appointments are inactive.

# Major Changes

- Added `divisionCodes?: string[]` to `TrusteeCasesSearchPredicate` in common
- Controller parses and validates `divisionCodes` from comma-separated query param
- Use case passes `divisionCodes` through to `searchCases` (which already supports it)
- `TrusteeCaseListFilter` fetches courts on mount and renders a District (Division) multi-select combo, mirroring the pattern used on the Trustee List page
- Division selections resolve `courtId|ALL` to actual division codes before sending to the API
- `TrusteeCaseList.buildSearchPredicate` and `Api2.getTrusteeCases` wired to pass `divisionCodes` as a comma-separated query param

# Testing/Validation

Implemented using TDD. 421 lines of new tests across all layers:
- 10 new backend unit tests (controller parsing/validation, use case pass-through)
- 7 new frontend filter tests (combo renders, selection fires correct `divisionCodes`, ALL expansion, clear, pill removal, screen reader announcements, error resilience)
- 2 new `TrusteeCaseList` integration tests (divisionCodes wired to API, omitted when absent)
- All 3193 frontend, 2911 backend, and 1090 common tests passing
- TypeScript clean, lint clean, 0 security vulnerabilities

# Notes

Filtering is server-side (not client-side) because the Case List uses server-side pagination — a client-side division filter would corrupt the case count and page numbers. This is a deliberate architectural difference from the Trustee List page, which loads all trustees once and filters in the browser.

The District (Division) combo reuses `getDistrictDivisionComboOptions` and `resolveCombinedSelections` from the existing Trustee List utilities. A follow-on DevEx story will extract a shared component to eliminate the remaining duplication between `TrusteeDistrictFilter` and `TrusteeCaseListFilter`.

# Definition of Done:

## Summary by Sourcery

Add support for filtering trustee case lists by court division and passing the selected division codes through the full stack to the case search API.

New Features:
- Introduce a District (Division) multi-select filter in the trustee case list that allows selecting specific divisions or all divisions within a district.

Enhancements:
- Extend trustee case search predicates and API wiring to include optional divisionCodes, ensuring they are parsed, validated, and forwarded from the UI to the backend search layer.
- Update trustee case list filter UX to surface division selections as pills and announce changes for accessibility, while gracefully handling court lookup failures.

Tests:
- Add frontend tests for the District (Division) filter behavior, including selection, clearing, announcement messaging, and error resilience when courts fail to load.
- Add backend controller and use case tests to verify correct parsing, normalization, and propagation of divisionCodes in trustee case search requests.
- Add trustee case list integration tests to confirm divisionCodes are correctly included or omitted in API calls based on the applied filters.

### DIFF
--- a/user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
+++ b/user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
@@ -534,10 +534,11 @@ describe('TrusteeCaseListFilter', () => {
     test('does not render District (Division) combobox when courts fetch fails', async () => {
       vi.spyOn(Api2, 'getCourts').mockRejectedValue(new Error('network error'));
       await renderFilter();
-      await waitFor(() => {});
-      expect(
-        screen.queryByRole('combobox', { name: /district \(division\)/i }),
-      ).not.toBeInTheDocument();
+      await waitFor(() =>
+        expect(
+          screen.queryByRole('combobox', { name: /district \(division\)/i }),
+        ).not.toBeInTheDocument(),
+      );
     });
 
     test('restores selected divisions visually when initialValue has divisionCodes', async () => {
@@ -631,11 +632,11 @@ describe('TrusteeCaseListFilter', () => {
           initialValue={{ caseStatus: 'OPEN', chapters: [], divisionCodes: ['0972'] }}
         />,
       );
-      await waitFor(() => {});
-      const calls = onFilterChange.mock.calls;
-      calls.forEach((call) => {
-        expect(call[0]).not.toMatchObject({ divisionCodes: ['0971'] });
-      });
+      await waitFor(() =>
+        expect(onFilterChange).not.toHaveBeenCalledWith(
+          expect.objectContaining({ divisionCodes: ['0971'] }),
+        ),
+      );
     });
 
     test('includes initialValue divisionCodes in onFilterChange when another filter changes', async () => {
--- a/user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
+++ b/user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
@@ -2,10 +2,38 @@ import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { vi, describe, test, expect, beforeEach } from 'vitest';
 import TrusteeCaseListFilter from './TrusteeCaseListFilter';
+import Api2 from '@/lib/models/api2';
+import { CourtDivisionDetails } from '@common/cams/courts';
+
+const mockCourts: CourtDivisionDetails[] = [
+  {
+    officeName: 'Manhattan',
+    officeCode: '0971',
+    courtId: '097',
+    courtName: 'Southern District of New York',
+    courtDivisionCode: '0971',
+    courtDivisionName: 'Manhattan',
+    groupDesignator: 'NY',
+    regionId: '02',
+    regionName: 'Region 2',
+  },
+  {
+    officeName: 'White Plains',
+    officeCode: '0972',
+    courtId: '097',
+    courtName: 'Southern District of New York',
+    courtDivisionCode: '0972',
+    courtDivisionName: 'White Plains',
+    groupDesignator: 'NY',
+    regionId: '02',
+    regionName: 'Region 2',
+  },
+];
 
 describe('TrusteeCaseListFilter', () => {
   beforeEach(() => {
     vi.restoreAllMocks();
+    vi.spyOn(Api2, 'getCourts').mockResolvedValue({ data: [], meta: { self: '' } });
   });
 
   async function renderFilter(onFilterChange = vi.fn()) {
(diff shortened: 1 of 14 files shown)


user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.tsx
```diff
@@ -24,6 +26,14 @@ export default function TrusteeCaseListFilter({
   const [filedDateTo, setFiledDateTo] = useState(initialValue?.filedDateTo ?? '');
   const [filedDateError, setFiledDateError] = useState('');
   const [filterAnnouncement, setFilterAnnouncement] = useState('');
+  const [courts, setCourts] = useState<CourtDivisionDetails[]>([]);
+  const [selectedDivisions, setSelectedDivisions] = useState<ComboOption[]>([]);
```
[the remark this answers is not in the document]
mattstankey: Fixed. `selectedDivisions` is now restored visually from `initialValue.divisionCodes` once courts load. `resolvedDivisionCodes` is initialized directly from `initialValue?.divisionCodes` so other filter handlers carry the division filter correctly from mount. Two new tests cover both behaviors.

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -14,16 +16,39 @@ export const CASE_CHAPTER_OPTIONS: ComboOption[] = [
...
+    if (code === 'ALL') {
+      courts.filter((c) => c.courtId === courtId).forEach((c) => codes.push(c.courtDivisionCode));
+    } else {
+      codes.push(code);
+    }
+  }
+  return codes.length > 0 ? codes : undefined;
+}
+
 const buildFilterFromStore = (
```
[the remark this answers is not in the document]
mattstankey: Fixed. `buildFilterFromStore` is now domain-focused — it reads `store.resolvedDivisionCodes` (a plain `string[] | undefined`) with no knowledge of `ComboOption` or courts. The `encodeDivisionCodes` helper (now exported) is called in `handleDivisionChange` before updating state, and the encoded codes are passed directly into `buildFilterFromStore` via a store spread.

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
```diff
@@ -14,6 +16,23 @@ export const CASE_CHAPTER_OPTIONS: ComboOption[] = [
...
+  const codes: string[] = [];
+  for (const opt of selectedDivisions) {
+    const [courtId, code] = opt.value.split('|');
+    if (code === 'ALL') {
+      courts.filter((c) => c.courtId === courtId).forEach((c) => codes.push(c.courtDivisionCode));
+    } else {
+      codes.push(code);
+    }
+  }
+  return codes.length > 0 ? codes : undefined;
```
diabagatekelly: `codes` can contain duplicates when `ALL` is selected and the courts array has duplicate entries for the same `courtId`. Should deduplicate before returning:

```suggestion
  return codes.length > 0 ? [...new Set(codes)] : undefined;
```

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
```diff
@@ -133,6 +137,22 @@ function TrusteeCaseListFilterView({ viewModel }: TrusteeCaseListFilterViewProps
                   placeholder="- Select one or more Chapters -"
                 />
               </div>
+              {courts.length > 0 && (
+                <div className="filter-control filter-control--district">
+                  <ComboBox
+                    id="case-district-division-combobox"
+                    label="District (Division)"
+                    options={getDistrictDivisionComboOptions(courts) as ComboOption[]}
```
diabagatekelly: The Trustee List applies `separateDefaultOptions` to pin user defaults to the top of the dropdown with a visual divider (see `trusteeDistrictFilterUseCase.ts → districtsToComboOptions`). The District (Division) combo here should do the same — user default divisions (from `getUserDivisionCodes`) should appear above the alphabetical list with a divider.

The computed `defaults` array is already available in the `useEffect` in `TrusteeCaseListFilter.tsx`. You could pass a `divisionComboOptions` (pre-sorted via `separateDefaultOptions`) down through the view model instead of recomputing options from raw `courts` here. That would also remove the double-computation (once in `useEffect`, once here per render).

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.tsx
```diff
@@ -24,6 +30,52 @@ export default function TrusteeCaseListFilter({
...
+                chapters: initialValue?.chapters ?? [],
+                filedDateFrom: initialValue?.filedDateFrom,
+                filedDateTo: initialValue?.filedDateTo,
+                ...(codes ? { divisionCodes: codes } : {}),
+              });
+            }
+          }
+        }
+      })
+      .catch(() => {});
```
diabagatekelly: Silent swallow — if `getCourts` fails, the Division combobox disappears with no log and no telemetry. The `TrusteeDistrictFilter` tracks `districtsError` state to surface this. At minimum, log to AppInsights or `console.warn` here so the failure is visible in monitoring:

```suggestion
      .catch(() => {
        // Courts failed to load; division combobox will be hidden
      });
```

(Or ideally fire an AppInsights event matching the pattern in `trusteeDistrictFilterUseCase.ts`.)

diabagatekelly (commented):
Code review pass. Flagging a few items inline — most important are the missing `separateDefaultOptions` for user defaults, the dedup gap in `encodeDivisionCodes`, and two `waitFor(() => {})` false-positives in the new tests.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
```diff
@@ -368,4 +399,259 @@ describe('TrusteeCaseListFilter', () => {
...
+          'District filter cleared',
+        );
+      });
+    });
+
+    test('does not render District (Division) combobox when courts fetch fails', async () => {
+      vi.spyOn(Api2, 'getCourts').mockRejectedValue(new Error('network error'));
+      await renderFilter();
+      await waitFor(() => {});
+      expect(
```
diabagatekelly: `waitFor(() => {})` resolves immediately — it is not a real async wait and makes this test a false positive. The assertion fires before the rejected promise has had a chance to settle. Wait for something concrete instead:

```suggestion
      await waitFor(() =>
        expect(screen.queryByRole("combobox", { name: /district \(division\)/i })).not.toBeInTheDocument(),
      );
```

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
```diff
@@ -368,4 +399,259 @@ describe('TrusteeCaseListFilter', () => {
...
+      vi.spyOn(LocalStorage, 'getSession').mockReturnValue(session);
+      const onFilterChange = vi.fn();
+      render(
+        <TrusteeCaseListFilter
+          onFilterChange={onFilterChange}
+          initialValue={{ caseStatus: 'OPEN', chapters: [], divisionCodes: ['0972'] }}
+        />,
+      );
+      await waitFor(() => {});
+      const calls = onFilterChange.mock.calls;
```
diabagatekelly: Same `waitFor(() => {})` false-positive here. This does not guarantee the async `getCourts` promise has settled before the assertions run. Replace with a wait that actually gates on observable state:

```suggestion
      await waitFor(() =>
        expect(onFilterChange).not.toHaveBeenCalledWith(
          expect.objectContaining({ divisionCodes: ["0971"] }),
        ),
      );
```

user-interface/src/trustees/panels/filters/trusteeCaseListFilterUseCase.ts
mattstankey: ✅ Fixed. Added `[...new Set(codes)]` before returning so duplicates are removed when `ALL` is selected with a court that has duplicate entries.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
mattstankey: ✅ Fixed. `divisionComboOptions` is now computed once after courts load via `separateDefaultOptions(allOptions, defaultOptionValues)`, where `defaultOptionValues` is a `Set` of the user's default option values. The result is stored in state and passed through the view model, so the view uses it directly instead of calling `getDistrictDivisionComboOptions(courts)` on every render. User-default divisions appear at the top with a divider.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.tsx
mattstankey: ✅ Fixed. Replaced `.catch(() => {})` with `.catch((e: Error) => { getAppInsights()?.appInsights?.trackException({ exception: e }); })`, matching the optional-chaining pattern used in other components. The failure is now visible in AppInsights monitoring.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
mattstankey: ✅ Fixed. Replaced with `await waitFor(() => expect(screen.queryByRole("combobox", { name: /district \(division\)/i })).not.toBeInTheDocument())` so the assertion gates on observable DOM state after the rejected promise settles.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilter.test.tsx
mattstankey: ✅ Fixed. Replaced with `await waitFor(() => expect(onFilterChange).not.toHaveBeenCalledWith(expect.objectContaining({ divisionCodes: ["0971"] })))` which waits until the mock confirms the default was not applied.

diabagatekelly (approved):

PR #2566, opened 2026-06-24
CAMS-784 Fix trustee list row height

jharcey-flexion:
# Bug

On the trustee list, the first appointment row for a trustee with multiple appointments was consistently taller than continuation rows, creating uneven vertical spacing. The gap appeared between rows 1 and 2 but not between subsequent rows.

# Purpose

Ensure all appointment rows for a given trustee render at uniform height, improving readability for high-volume trustees.

# Major Changes

* Added `display: flex; align-items: center` to `.col-name a` in `TrusteesList.scss`

# Testing/Validation

Root cause identified through browser DevTools inspection: the `TrusteeName` link wraps a `cams-icon-label` (`display: inline-flex`), which sits on the text baseline inside the anchor. The descender space below the baseline (standard for Source Sans Pro at 16px) inflated the anchor height beyond the `line-height` of the other cells, making the first row taller.

Setting the anchor to `display: flex` removes the inline formatting context and eliminates the descender gap. Fix verified manually against a trustee with 6 appointments — all rows now render at uniform height.

A regression test was added asserting the trustee name link appears exactly once across multiple appointment rows, locking in the behavioral contract without coupling to internal markup.

# Notes

The fix is scoped to `.col-name a` inside `TrusteesList.scss` — it does not affect any other usage of `cams-icon-label` or `TrusteeName` elsewhere in the app.

# Definition of Done:

## Summary by Sourcery

Ensure trustee appointment rows render with consistent height and validate trustee name link behavior for multiple appointments.

Enhancements:
- Adjust trustee name column link styling so appointment rows for trustees with multiple appointments render at a uniform height.

Tests:
- Add a regression test asserting the trustee name link renders exactly once across multiple appointment rows for a trustee.


jharcey-flexion:
No truncation or ellipsis is applied to `.col-name` — names use `word-wrap: break-word` and wrap normally. The `display: flex` change is scoped to the anchor inside `.col-name` (within the tablet+ media query only) and only affects vertical alignment. Long names will continue to wrap as before since flex block containers don't suppress text wrapping by default.

jharcey-flexion:
Two additional points from Sourcery's review of the CAMS-785 PR (which picked up these changes in its diff):

1. **SCSS scoping** — The `a { display: flex; align-items: center; }` rule under `.col-name` will apply to any anchor in that column. Currently `TrusteeName` renders the only anchor there, so this is theoretical. Worth considering a more specific selector (e.g. targeting `.usa-link` inside `.col-name`) if additional links are ever added to that column.

2. **testid convention** — The uniqueness test uses `trustee-link-${trusteeId}`, which matches the format passed to `TrusteeName` from `TrusteesList.tsx:622`. The coupling is intentional but could diverge if the format changes. Consider whether a shared constant would help keep them in sync long-term.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -28,6 +28,8 @@ import { Pagination } from '@/lib/components/uswds/Pagination';
 import { Pagination as PaginationModel } from '@common/api/pagination';
 import { DEFAULT_SEARCH_LIMIT, DEFAULT_SEARCH_OFFSET } from '@common/api/search';
```
cruzjone-flexion, line 30: `getTrusteeListLinkTestId` is exported from the production module solely for test consumption — personally not a fan of that pattern. Since it's just a template literal, consider inlining it back in the markup:

```tsx
dataTestId={`trustee-link-${trustee.trusteeId}`}
```

And in the tests, inline the same string directly instead of importing the helper:

```tsx
screen.getByTestId(`trustee-link-${trusteeId}`)
```

That keeps test infrastructure out of the production export surface without any real duplication cost.

jharcey-flexion:
Both points addressed:

1. **`.col-name a` scoping** — already fixed in the previous push: the rule was changed from `a { ... }` to `.usa-link { ... }`, scoped to the specific link class used by `TrusteeName`.

2. **`waitFor` robustness** — good catch. Moved the link count assertion inside `waitFor` so it waits for the trustee link itself to appear rather than just the table container, eliminating the potential race.

user-interface/src/trustees/TrusteesList.tsx
jharcey-flexion: According to Claude...

Jonathan's feedback is the opposite of what Sourcery suggested — he prefers reverting the shared helper and going back to inlining the template literal in both the component and tests. His reasoning: exporting a helper from a production module solely for test consumption pollutes the public API.

This is a judgment call. Sourcery wanted the constant to avoid divergence; Jonathan thinks the duplication cost is low enough that a production export isn't worth it. I'd lean toward taking Jonathan's feedback since he's the team member — want me to revert the getTrusteeListLinkTestId export and inline the template literals again?

cruzjone-flexion (approved):

PR #2569, opened 2026-06-24
CAMS-768: Provide Admin UI for setting notification routing (Slice 3)

fmaddenflx:
## Purpose

Rework the notification routing admin UI from a generic CRUD table to a simplified fixed-recipient model. Administrators configure 3 labeled email fields — no knowledge of internal routing keys required. The backend uses a covers-array lookup (single query, no fallback chain) and a `DEFAULT_NOTIFICATION_RECIPIENT` env var for pre-configuration fallback.

## Major Changes

* **Shared types** (`common/src/cams/notifications.ts`): replaced `key: string` with `covers: string[]` on `NotificationRecipient`; added `NOTIFICATION_ROUTING_DEFINITIONS` constant defining the 3 fixed recipients; replaced `NotificationRoutingInput` with `NotificationRoutingUpdateInput`
* **Repository** (`notification-routing.mongo.repository.ts`): `findRecipientByRoutingKey()` uses `covers.contains([key])` for single-query lookup; removed `create`/`delete`/`getDefaultRecipient`; `updateRoutingRecord()` preserves covers/displayName from definitions
* **Use case** (`trustee-change-notification.ts`): simplified resolution — single lookup + env var fallback, no miss-and-retry chain
* **Controller**: removed POST/DELETE handlers; PUT validates only `recipientAddress`
* **Express & Azure Function routes**: removed POST/DELETE registrations
* **Admin UI** (`NotificationRouting.tsx`): 3 fixed labeled email fields (Default Chapter Oversight, Subchapter V Oversight, 341 Meeting Oversight); removed generic CRUD table and enable/disable toggle
* **Infrastructure** (`backend-api-deploy.bicep`, `main.bicep`): added `DEFAULT_NOTIFICATION_RECIPIENT` parameter
* **Seed data**: 3 records with covers arrays replacing the previous 7 per-key records

## Testing/Validation

- All notification routing tests pass (backend + frontend)
- Type-check clean across all workspaces
- Coverage ≥90% on all CAMS-768 files (most at 100%)
- Manually verified: save recipients → navigate away → return → values persist

## Notes

- The notification feature toggle is controlled solely by the `trustee-change-notification-enabled` feature flag — no admin toggle in the UI
- `DEFAULT_NOTIFICATION_RECIPIENT` defaults to empty in Bicep (no fallback until explicitly configured per environment)
- The 3 routing definitions are shared between backend and frontend via `NOTIFICATION_ROUTING_DEFINITIONS` in common


mattstankey (commented):
## Code Review: Issues 1–12

### 🔴 Critical

**1. `METHOD_NOT_ALLOWED` returned inside a 200 response**
`notification-routing.controller.ts` — the fallthrough returns:
```ts
return httpSuccess({ statusCode: HttpStatusCodes.METHOD_NOT_ALLOWED }) as CamsHttpResponseInit;
```
`httpSuccess` wraps this as an HTTP 200 with a 405 in the body. Clients won't see an actual 405 status code. Throw a `MethodNotAllowedError` (or equivalent) instead, consistent with how other controllers handle unsupported methods.

---

**2. `MockNotificationGateway` lives in the production code path**
The file lives at `adapters/gateways/notifications/mock-notification.gateway.ts` and is imported by `factory.ts`. Every other mock gateway in the backend lives under `backend/lib/testing/mock-gateways/`. This blurs the line between test doubles and production code. Move it to match project conventions.

---

**3. `routingId` not validated against known definitions before upsert**
The PUT handler accepts any string `routingId` and calls `updateRoutingRecord`. The repository upserts with empty `covers`/`displayName` if the ID is unknown. A SuperUser could accidentally create orphaned routing records. Pre-validate that `routingId` is a known key from `NOTIFICATION_ROUTING_DEFINITIONS` before reaching the repository.

---

**4. No audit trail for routing record updates**
The PUT endpoint updates records that control where legal oversight notices are delivered, with no audit logging. Other sensitive admin operations in CAMS use `Auditable` types. This is missing here.

---

### 🔶 Security

**5. Email regex is too permissive**
`/^[^\s@]+@[^\s@]+\.[^\s@]+$/` accepts inputs like `a@b.c` or addresses with trailing dots. Consider using the existing CAMS validation library or adding a stricter pattern plus a max-length check.

---

**6. No `routingId` character/length validation**
`routingId` from request params is passed directly to the MongoDB query without any length or character-set guard. Add a basic check (e.g., max length, alphanumeric + hyphens allowlist).

---

### 🔶 Type Safety

**7. `as unknown as NotificationRoutingRecord[]` in `getAll`**
`notification-routing.mongo.repository.ts` — this double cast bypasses TypeScript entirely. Either map documents explicitly or use a concrete `NotificationRoutingDoc` type that satisfies `NotificationRoutingRecord[]` without the cast.

---

**8. Return type `CamsHttpResponseInit<object | undefined>` is too wide**
The actual response types are concrete subtypes. Using `object` means TypeScript won't catch accidental mismatches. Replace with a union of the specific return types.

---

**9. `NotificationRoutingDoc` — all fields are optional**
With every field optional, the compiler won't catch missing required fields when constructing documents. Separate the routing and config doc types, or at minimum mark fields that are always required as non-optional.

---

### 🔷 Test Coverage

**10. `TrusteeChangeNotificationUseCase` test coverage unclear**
The diff cuts off mid-way through the use case test file — the core routing logic, `primaryChapter` routing, fan-out deduplication, and failure isolation aren't visible. Please confirm test coverage is complete for this file.

---

**11. No test for `PUT` without `:routingId`**
`PUT /api/dev-tools/notification-routing` (no ID) returns a 200/405 body with no dedicated test. This path should be covered.

---

**12. No test verifying the `covers.contains([key])` MongoDB query shape**
Tests mock `findOne` at the adapter level but don't assert what query was passed to it. If `QueryBuilder.contains` renders the query incorrectly, tests would still pass while production queries fail. Add an assertion on the query object passed to `findOne`.

fmaddenflx:
## Review Response (Issues 1–12)

### Fixed in commit 7b6bf829

| # | Issue | Resolution |
|---|-------|------------|
| 2 | MockNotificationGateway location | Moved to `backend/lib/testing/mock-gateways/` |
| 3 | routingId not validated | Rejects unknown IDs with BadRequestError before reaching repository |
| 4 | No audit trail | Added `NotificationRoutingAuditHistory` type + `createRoutingAuditRecord()` — writes before/after + user on every PUT |
| 5 | Email regex too permissive | Now uses canonical `EMAIL_REGEX` from `@common/cams/regex` |
| 6 | No routingId character/length validation | Addressed by #3 — only IDs from `NOTIFICATION_ROUTING_DEFINITIONS` are accepted |
| 7 | `as unknown as` cast | Replaced with explicit `.map()` in `getAll()` |
| 9 | NotificationRoutingDoc all optional | Required fields are now non-optional |
| 10 | Test coverage unclear | Confirmed 100% across all metrics via JSON coverage report |
| 11 | No test for PUT without routingId | Test added |
| 12 | No test verifying query shape | Assertion on CONTAINS query structure added |

### Not changed (with reasoning)

| # | Issue | Reasoning |
|---|-------|-----------|
| 1 | METHOD_NOT_ALLOWED inside httpSuccess | See below |
| 8 | Return type too wide | The `CamsController` interface defines `Promise<CamsHttpResponseInit<object \| undefined>>`. Narrowing the return type in our controller breaks the Azure Function handler which expects this interface type. Fixing this would require changing the shared `CamsController` interface and all function adapters — a framework-level refactor outside this story's scope. |

### On #1 — `httpSuccess` with `METHOD_NOT_ALLOWED`

`httpSuccess` does NOT wrap the response in a 200. It sets headers and passes the `statusCode` through directly:

```typescript
// backend/lib/adapters/utils/http-response.ts
export function httpSuccess<T>(response = {}): CamsHttpResponseInit<T> {
  const camsResponse = {
    headers: { ...commonHeaders, ...response.headers },
    statusCode: response.statusCode ?? (response.body ? HttpStatusCodes.OK : HttpStatusCodes.NO_CONTENT),
  };
  ...
}
```

The `statusCode: METHOD_NOT_ALLOWED` becomes the actual HTTP 405 returned to the client. This is the established pattern across all CAMS controllers — the function is poorly named but it's the project's response-builder, not a "success-only" wrapper.

Other controllers using the same pattern:
- `backend/lib/controllers/banks/banks.controller.ts:37`
- `backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts:39`

Happy to rename `httpSuccess` to something like `buildResponse` in a separate cleanup PR if the team thinks the naming is confusing.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -0,0 +1,146 @@
+import { useEffect, useState } from 'react';
+import Api2 from '@/lib/models/api2';
+import { LoadingSpinner } from '@/lib/components/LoadingSpinner';
+import Button, { UswdsButtonStyle } from '@/lib/components/uswds/Button';
+import Alert, { UswdsAlertStyle } from '@/lib/components/uswds/Alert';
+import {
+  NotificationRoutingRecord,
+  NOTIFICATION_ROUTING_DEFINITIONS,
+} from '@common/cams/notifications';
+
```
bityogi: This local `EMAIL_REGEX` differs from the canonical `EMAIL_REGEX` exported from `@common/cams/regex` that the backend controller uses. This means validation could pass on the frontend but fail on the backend (or vice versa). Consider importing the shared regex from common to guarantee consistent behavior across layers.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+
+  public async updateConfig(config: NotificationConfig): Promise<NotificationConfig> {
+    try {
+      const query = this.doc('documentType').equals('NOTIFICATION_CONFIG');
+      const doc = {
+        documentType: 'NOTIFICATION_CONFIG',
+        enabled: config.enabled,
+      } as unknown as NotificationRoutingDoc;
```
bityogi: The `as unknown as NotificationRoutingDoc` double-cast here forces a config document (which only has `documentType` + `enabled`) through a type that requires `id`, `covers`, `recipientAddress`, `displayName`. This works at runtime but defeats TypeScript's safety net — any accidental property access on the result would compile without error. Consider a discriminated union type or a separate `NotificationConfigDoc` type to keep the adapter generic usage type-safe.

backend/lib/controllers/admin/notification-routing.controller.ts
```diff
@@ -0,0 +1,129 @@
...
+  ): Promise<CamsHttpResponseInit<NotificationRoutingRecord[]>> {
+    const records = await this.repository.getAll();
+    return httpSuccess({
+      statusCode: HttpStatusCodes.OK,
+      body: { meta: { self: context.request.url }, data: records },
+    });
+  }
+
+  private async handleGetConfig(
+    context: ApplicationContext,
```
bityogi: The PR description notes that the admin enable/disable toggle was *removed* from the UI, but `handleGetConfig` and `handlePutConfig` are still exposed in the API. If these are intentionally retained for API-level control (e.g., programmatic toggling), a brief comment here would help future readers understand why the code exists without a UI consumer. If truly dead, consider removing to reduce surface area.

bityogi (commented):
## Review Summary

Overall this is a strong, well-structured PR. Thoroughly tested, good security posture (CRLF sanitization, HTML escaping, role gating, input validation). A few items worth addressing — nothing blocking, but several improve type safety and reduce future confusion.

backend/lib/use-cases/notifications/trustee-change-notification.ts
```diff
@@ -0,0 +1,100 @@
...
+    category: RoutingCategory,
+    primaryChapter: TrusteeChangeSet['primaryChapter'],
+  ): Promise<NotificationRecipient | null> {
+    let routingKey: string;
+    if (category === 'zoom-341') {
+      routingKey = 'category:zoom-341';
+    } else if (primaryChapter) {
+      routingKey = `chapter:${primaryChapter}`;
+    } else {
+      routingKey = 'chapter:7';
```
bityogi: When `primaryChapter` is undefined (trustee has no appointments), the routing key defaults to `'chapter:7'`. This creates an implicit coupling — if the default-chapter-oversight record is ever reconfigured to not cover `chapter:7`, notifications for trustees without appointments would silently drop with no fallback. Consider going directly to the `DEFAULT_NOTIFICATION_RECIPIENT` env var path when `primaryChapter` is absent, or at minimum documenting this assumption prominently.

backend/lib/factory.ts
```diff
@@ -511,6 +517,28 @@ const getListsGateway = (context: ApplicationContext): ListsRepository => {
...
 
+const getNotificationRoutingRepository = (
+  context: ApplicationContext,
+): NotificationRoutingRepository => {
+  if (context.config.get('dbMock')) {
+    return new MockMongoRepository();
+  }
+  const repo = NotificationRoutingMongoRepository.getInstance(context);
+  deferRelease(repo, context);
+  return repo;
```
bityogi, line 528: Unlike other gateway singletons in this file that have a `dropInstance()` pattern, `notificationGateway` is set once at module scope and never cleared. If tests in the same process need to swap the gateway (e.g., verify NoOp vs. Mock behavior), this could cause test pollution. A `resetNotificationGateway()` (even if test-only) would match the pattern used elsewhere.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
```diff
@@ -0,0 +1,164 @@
...
+} from '@common/cams/notifications';
+import { Creatable } from '@common/cams/creatable';
+import { NotificationRoutingRepository } from '../../../use-cases/gateways.types';
+
+const MODULE_NAME = 'NOTIFICATION-ROUTING-MONGO-REPOSITORY';
+const COLLECTION_NAME = 'notification-routing';
+
+const { using } = QueryBuilder;
+
+type NotificationRoutingDoc = {
```
bityogi: Nit: `NotificationRoutingDoc` partially duplicates the shape of `NotificationRoutingRecord` from `@common/cams/notifications`. Consider extending or importing it to reduce drift — e.g., `type NotificationRoutingDoc = NotificationRoutingRecord & { enabled?: boolean }` would keep the Cosmos-specific fields additive rather than parallel.

backend/lib/use-cases/notifications/templates/trustee-change.template.ts
```diff
@@ -0,0 +1,86 @@
...
+
+                    <!-- Header Text -->
+                    <tr>
+                        <td style="padding-bottom: 20px;">
+                            <p style="margin: 0; font-size: 14px; color: #000000;">Trustee {{trustee_name}}'s information has changed.</p>
+                        </td>
+                    </tr>
+
+                    <!-- Appointment Information Section -->
+                    <tr>
```
bityogi, line 48: Both section headers ("Appointment Information" and "341 Meeting Information") always render in the HTML output regardless of whether they have rows. The plaintext fallback correctly omits empty sections. Consider conditionally hiding a section (e.g., wrapping in a placeholder like `{{#appointment_section}}...{{/appointment_section}}` or handling at compile time) so recipients don't see empty table headers in the email.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
fmaddenflx: That's what I told the agent to do originally.  gee wiz

backend/lib/use-cases/notifications/trustee-change-notification.ts
fmaddenflx: I would think if a trustee has no appointments, then no one should be notified.  Don't send out notification unless there is at least 1 appointment.

backend/lib/factory.ts
fmaddenflx: We need to clear it like we do all other singletons.

backend/lib/adapters/gateways/mongo/notification-routing.mongo.repository.ts
fmaddenflx: Yes, let's do that.

backend/lib/use-cases/notifications/templates/trustee-change.template.ts
fmaddenflx: Agreed, but if there is nothing to display then the email shouldn't even be sent.  If the table has no rows then there is nothing to notify the recipient about.

mattstankey (approved):

bityogi (approved):

PR #2570, opened 2026-06-25
Fix hoisting-dependent function app packaging

jamesobrooks:
# Bug

A routine Renovate dependency bump (mssql) broke the backend `Package API` / `Package Data Flow` CI step with:

```
npm error `npm ci` can only install packages when your package.json and
package-lock.json are in sync.
npm error Missing: mssql@12.5.5 from lock file
npm error Missing: commander@11.1.0 from lock file
```

The failure was latent on `main`: `pack.sh` packaged each Azure Function app by copying the app's `package.json` plus the root lockfile into a temp dir and running `npm ci --workspaces=false`, treating the app as a standalone root project. That only worked when npm happened to hoist the app's externals to the top level of the root install. npm's hoisting is not stable across npm versions or install sequences, so the bump that un-hoisted mssql exposed the latent break.

# Purpose

Make Azure Function app packaging faithful to the lockfile and independent of npm's hoisting behavior, so dependency bumps can't silently break the deployment artifact.

# Major Changes

* Replaced the standalone `npm ci --workspaces=false` pack step with a workspace-aware approach: after a normal root `npm ci`, compute the app's production dependency closure from the `package-lock.json` graph and copy just that closure into the app's `node_modules`, preserving the install layout.
* Added `build-function-app-node-modules.mjs` — the lockfile-graph closure builder (resolves each dependency to its nearest install slot; fails loudly on an incomplete closure or a slot collision).
* Added `verify-function-app-node-modules.mjs` — a runtime load-probe that asserts each external (including the lazily-required `mssql` → `tedious` driver) resolves from the app's own `node_modules`, failing the pack on a gap.
* Rewired `pack.sh` and `Dockerfile.build` to use the new mechanism. The Linux-native and Podman build paths are retained; only the dependency-resolution step changed.
* Updated `TROUBLESHOOTING-AZURE-FUNCTIONS.md`.

# Testing/Validation

Implemented with TDD. `build-function-app-node-modules.test.ts` adds 12 tests that lock in the hoisting-independence property using synthetic lockfile fixtures — asserting an identical closure whether mssql is hoisted or un-hoisted.

Validated end-to-end against the `dev-728c0c` slot deployment:

* CI `Package API` / `Package Data Flow` steps pass — closure builds (328/329 packages) and the verifier confirms `mongodb`, `mssql`, and `mssql/lib/tedious/connection-pool.js` all load.
* API app: healthcheck returns `sqlDbReadStatus: true` and `cosmosDb{Read,Write,Delete}Status: true`; manual load-trustee / create-note / reload round-trip confirmed (corroborated by `trustees` / `trustee-notes` / `case-notes` requests succeeding in App Insights).
* Dataflows app: host boots cleanly from the new package, logs `Registered Dataflows` with all expected dataflows enabled, and shows zero module-load errors (`Cannot find module` / `ERR_DLOPEN` / native-binding) since deploy.

# Notes

* No `test/e2e` mssql bump is included on this branch by design. That bump alone does not reproduce the failure (it's npm-version-dependent hoisting), so committing it would prove nothing; the regression evidence lives in the unit test that asserts identical closure under both hoisting states.
* Once this merges to `main`, the Renovate PR (#2553) rebases on top and its pack step goes green.
* The function-app production closure is currently 100% pure JavaScript (no native binaries). The Linux-native/Podman apparatus is retained as a guard for any future native dependency.

# Definition of Done:

## Summary by Sourcery

Make Azure Function app packaging derive function app node_modules directly from the root workspace lockfile and verify runtime dependencies so deployments are independent of npm hoisting behavior.

Bug Fixes:
- Fix Azure Function app packaging failures caused by npm hoisting changes that left required dependencies missing from the packaged node_modules.

Enhancements:
- Introduce a lockfile-graph-based builder to assemble each function app’s production dependency closure into its own node_modules while preserving install layout.
- Add a verification step that load-tests each function app’s runtime dependencies to ensure they resolve from the app’s packaged node_modules.
- Align Linux and macOS/Podman packaging flows to share the new build and verification mechanism and keep their outputs consistent.

Documentation:
- Update Azure Functions troubleshooting documentation to describe the new closure-based packaging and verification flow, including expected zip contents.

Tests:
- Add unit tests for the lockfile-graph closure builder to guarantee hoisting-independent dependency resolution and slot mapping for function apps.

### DIFF
--- a/backend/Dockerfile.build
+++ b/backend/Dockerfile.build
@@ -1,4 +1,4 @@
-# Build dependencies on Linux for Azure deployment
+# Build a function app's production node_modules on Linux for Azure deployment.
 # Use Node 22 to match .nvmrc (v22.22.2)
 FROM node:22.22.2-slim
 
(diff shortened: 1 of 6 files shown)


btposey (approved):

mattstankey (commented):
`DEEP_PROBES` hardcodes `mssql/lib/tedious/connection-pool.js` — an internal path. If mssql ever restructures its internals (not a semver event), this probe will silently start failing with a module-not-found error rather than actually testing what we care about, and we'll need to rediscover the right path at that moment.

Suggest adding a pinned-version comment so future maintainers know when to re-check:

```js
const DEEP_PROBES = {
  // Forces mssql → tedious lazy load. Internal path — last verified against mssql@12.5.5.
  // Re-check if this probe starts failing after a mssql bump.
  mssql: ['mssql/lib/tedious/connection-pool.js'],
};
```

Alternatively, if the goal is just to verify that `tedious` is present in the closure, probing `tedious` directly as a top-level dep would be more stable (though it would require adding it to the app's explicit `dependencies`, which may not be desirable).

jamesobrooks:
Good catch — agreed the hardcoded internal path is the fragile part. Addressed in 375134532, keeping the deep-require probe but documenting it: pinned-version comment (last verified `mssql@12.5.5`) with a re-check note, plus a drift hint appended to the failure message so a future module-not-found reads as "the internal path may have moved, check it" rather than "the driver is missing from the closure."

On the alternative (probe `tedious` as a top-level dep): I dug into it and it would actually break the case this branch exists for. Requiring `mssql/lib/tedious/connection-pool.js` from inside mssql's own directory is what lets mssql's lazy `require('tedious')` resolve whether tedious is **hoisted** (`node_modules/tedious`) or **nested** (`mssql/node_modules/tedious`). Probing `tedious` as a top-level spec resolves from the app root instead, so in the un-hoisted layout — where tedious lives under `mssql/node_modules/` — it isn't resolvable from the root, and combined with the verifier's "must resolve inside app node_modules" assertion it would produce a false FAIL in exactly the scenario we're hardening against.

Same reason `npm ls tedious` doesn't fit as the gate: I tested it against a built closure with tedious removed and it exits **0** with `(empty)` — it's a best-effort tree print with no lockfile in the packaged app dir, so it doesn't fail on the missing-driver case. It also only checks on-disk presence, not that `require('tedious')` actually loads (the real runtime contract), and has no notion of the "must resolve inside the app's own node_modules" constraint that catches ancestor-masking in the deployed (ancestor-less) zip.

PR #2573, opened 2026-06-25
CAMS-770 Extract shared DistrictDivisionComboBox

diabagatekelly:
# Purpose

Eliminate duplicate District (Division) filter implementations by extracting the shared logic into a single reusable `DistrictDivisionComboBox` component in `lib/components/`.

# Major Changes

- New `DistrictDivisionComboBox` component in `lib/components/` that owns:
  - Courts fetching from the API on mount
  - Session-based default division selection
  - Default options separator (user's divisions at top with divider)
  - `resolveCombinedSelections` (All ↔ specific mutual exclusion per court)
  - `autoUpgradeToAll` when all divisions in a court are individually selected
  - `aria-live` announcement naming the specific district on upgrade
  - Error message when courts API fails
- `TrusteeCaseListFilter` refactored to use the new component — removes its own courts `useEffect`, `divisionComboOptions` state, and duplicate logic
- `TrusteeDistrictFilterView` updated to use the new component when the district-division feature flag is enabled (legacy district filter retained for flag-off path)
- Component exposes `onDivisionCodesChange`, `onSelectionsChange`, and a `forwardRef` `setSelections` handle so parents can manage their own pill boxes and sync removals back

# Testing/Validation

Implemented using TDD. Behavior catalog derived from existing specs for both filter components before writing a single line of implementation. 21 unit tests cover all behaviors including initialization, session defaults, default options ordering, selection mutual exclusion, autoUpgradeToAll, and all aria-live announcement scenarios.

All 170 trustee-related unit tests pass.

# Notes

The component intentionally does not own its own PillBox — parents mix division pills with other filter pills (status, chapter, date) in a single combined strip. The `onSelectionsChange` callback gives parents the full `ComboOption[]` needed to render those pills, while `onDivisionCodesChange` gives the encoded division codes for API filtering.

`hideInternalLabel` prop allows call sites that provide their own external `filter-control-header` label to suppress the internal ComboBox label and avoid duplicate screen reader announcements.

# Definition of Done:

## Summary by Sourcery

Introduce a reusable DistrictDivisionComboBox for trustee district filtering and wire division-based filtering end-to-end from UI to backend search.

New Features:
- Add a shared DistrictDivisionComboBox component with support for multi-select district/division filtering, session-based defaults, and accessibility announcements.
- Enable trustee case list filters to capture and persist selected court division codes alongside existing filters.
- Support passing divisionCodes through trustee case list API requests and backend search predicates, including query parsing of comma-separated values.

Enhancements:
- Refine trustee case list and trustee district filter UIs to integrate the new shared DistrictDivisionComboBox while keeping legacy behavior behind feature flags.
- Improve trustee-case test data seeding by distributing cases across multiple divisions and courts for more realistic scenarios.

Tests:
- Add comprehensive unit tests for the new DistrictDivisionComboBox behaviors, including defaults, selection rules, and screen-reader announcements.
- Extend trustee filters, trustee case list, controller, and use-case tests to cover divisionCodes propagation, clearing semantics, and edge cases.


user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
```diff
@@ -136,22 +140,15 @@ function TrusteeCaseListFilterView({ viewModel }: TrusteeCaseListFilterViewProps
...
-                    multiSelect={true}
-                    wrapPills={true}
-                    pluralLabel="divisions"
-                    singularLabel="division"
-                    placeholder="- Select one or more -"
-                  />
-                </div>
-              )}
+              <div className="filter-control filter-control--district">
+                <DistrictDivisionComboBox
```
mattstankey, line 147: Missing `wrapPills={true}` — the original `TrusteeCaseListFilter` passed this to its `ComboBox`. `DistrictDivisionComboBox` does not expose or forward the prop, so it silently defaults to `false`. This is a behavior regression if `wrapPills` controls whether selected divisions appear as chips inside the input area. Add `wrapPills?: boolean` to `DistrictDivisionComboBoxProps` and forward it, then pass `wrapPills={true}` here.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
```diff
@@ -1,7 +1,9 @@
-import ComboBox, { ComboOption } from '@/lib/components/combobox/ComboBox';
+import { ComboOption } from '@/lib/components/combobox/ComboBox';
+import ComboBox from '@/lib/components/combobox/ComboBox';
```
mattstankey: Split import — should be one line: `import ComboBox, { ComboOption } from '@/lib/components/combobox/ComboBox';`

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
```diff
@@ -51,9 +53,10 @@ export interface TrusteeCaseListFilterViewModel extends TrusteeCaseListFilterHan
   filedDateTo: string;
   filedDateError: string;
   filterAnnouncement: string;
-  courts: CourtDivisionDetails[];
   selectedDivisions: ComboOption[];
-  divisionComboOptions: ComboOption[];
+  divisionRef: React.RefObject<DistrictDivisionComboBoxRef | null>;
```
mattstankey: A `RefObject` in the view model leaks view-layer implementation details into the data/behavior contract. The ref is only consumed by the view (passed to `DistrictDivisionComboBox` and called directly in the PillBox handler) — the parent component does not need to know about it. Move `divisionRef` creation from `TrusteeCaseListFilter.tsx` into `TrusteeCaseListFilterView.tsx` and remove it from this interface.

mattstankey:
In `trusteeCaseListFilterUseCase.ts` `handleDivisionChange` (line 79): double resolution — `DistrictDivisionComboBox` already runs `resolveCombinedSelections` and `autoUpgradeToAll` before calling `onSelectionsChange`, so `divisions` arrive here already resolved. The `resolveCombinedSelections` call on line 80 is a no-op in practice but misleads future maintainers into thinking resolution responsibility lives in the use case. Remove the re-resolve, accept `divisions` directly, and drop `resolveCombinedSelections` from the import.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
```diff
@@ -0,0 +1,170 @@
...
+    },
+  }));
+
+  useEffect(() => {
+    Api2.getCourts()
+      .then((r) => {
+        const allCourts = r.data;
+        setCourts(allCourts);
+        onCourtsLoaded?.(allCourts);
+        const allOptions = getDistrictDivisionComboOptions(allCourts) as ComboOption[];
```
cruzjone-flexion, line 68: **Force casts carry an existing type weakness into the new component**

Both `getDistrictDivisionComboOptions` and `separateDefaultOptions` return a wider type (likely `(ComboOption | SeparatorOption)[]`), and the cast at the call site silences that. The new component is a good moment to fix this at the source — narrowing the return type of those utilities (or adding an overload) would remove the need for `as ComboOption[]` here and at line 93, and prevent future call sites from inheriting the same pattern.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
```diff
@@ -0,0 +1,170 @@
...
+    wrapPills,
+  }: DistrictDivisionComboBoxProps,
+  ref: React.Ref<DistrictDivisionComboBoxRef>,
+) => {
+  const [courts, setCourts] = useState<CourtDivisionDetails[]>([]);
+  const [courtsError, setCourtsError] = useState(false);
+  const [selectedDivisions, setSelectedDivisions] = useState<ComboOption[]>([]);
+  const [divisionComboOptions, setDivisionComboOptions] = useState<ComboOption[]>([]);
+  const [upgradeAnnouncement, setUpgradeAnnouncement] = useState('');
+  const previousDivisionValuesRef = useRef<Set<string>>(new Set());
```
cruzjone-flexion: **`previousDivisionValuesRef` is redundant — `previousSelectionsRef` already holds the same information**

`previousDivisionValuesRef` is only used in the upgrade-announcement effect to detect newly-added `|ALL` values. But `previousSelectionsRef` already holds the full prior `ComboOption[]`, so the same set is derivable with `new Set(previousSelectionsRef.current.map(d => d.value))`. The two refs are also updated in different code paths, which is a latent sync hazard — removing the redundant one eliminates it.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const allOptions = getDistrictDivisionComboOptions(allCourts) as ComboOption[];
+
+        let defaults: ComboOption[] = [];
+        if (initialDivisionCodes?.length) {
+          defaults = allOptions.filter((opt) => {
+            const [, code] = opt.value.split('|');
+            return code !== 'ALL' && initialDivisionCodes.includes(code);
+          });
+          setSelectedDivisions(defaults);
+          previousSelectionsRef.current = defaults;
```
cruzjone-flexion: **`onDivisionCodesChange` is not called when restoring `initialDivisionCodes`, but is called for session defaults — asymmetric callback behavior**

A caller that wires only `onDivisionCodesChange` to track initial selection state will be notified on the session-defaults path but silently skipped when `initialDivisionCodes` is provided. The two branches fire different callbacks, with no indication this is intentional:

```ts
if (initialDivisionCodes?.length) {
  // ...
  onSelectionsChange?.(defaults);    // fires
  // onDivisionCodesChange NOT called
} else {
  // session defaults
  onDivisionCodesChange?.(codes);    // fires
  onSelectionsChange?.(defaults);    // fires
}
```

Either both callbacks should fire in both branches, or the asymmetry needs a clear rationale. As-is, any future call site that wires only `onDivisionCodesChange` will silently miss the restored state.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
```diff
@@ -209,7 +183,20 @@ function TrusteeDistrictFilterView(props: TrusteeDistrictFilterViewProps) {
             <div className="filter-controls-row">
               <div className="filter-controls-pair">
                 {renderNameFilter(viewModel)}
-                {renderDistrictFilter(viewModel, showLegacyDistrictFilter)}
+                {viewModel.districtDivisionEnabled ? (
+                  <div className="filter-control">
+                    <div className="filter-control-header">
+                      <span className="filter-control-label">District (Division)</span>
+                    </div>
+                    <DistrictDivisionComboBox
```
jamesobrooks, line 195: **F5 (High) — Division dropdown desyncs from the pill box on this page.**

`DistrictDivisionComboBox` owns its internal `selectedDivisions` state and only re-syncs when a parent calls its `setSelections` imperative handle. Here the component is rendered with **no ref**, and pill removal (line ~230) calls only `viewModel.handleFilterDivision(updatedDivisions)` — which updates the external store (driving the pills) but never reaches back into the component's internal state. So removing a division via its pill removes the pill but leaves the division **still checked inside the dropdown**.

The Case List page avoids this exact bug by calling `divisionRef.current?.setSelections(updatedDivisions)` on pill removal (`TrusteeCaseListFilterView.tsx:188`). This page should mirror it. No `key`/controlled-value escape hatch exists, so this is a real, observable desync for users on the flag-on path.

**Prompt to address:**
> In `TrusteeDistrictFilterView.tsx`, wire a `DistrictDivisionComboBoxRef` ref to the `DistrictDivisionComboBox` (around line 191), mirroring `TrusteeCaseListFilterView.tsx`. In the PillBox `onSelectionChange` handler (around line 220-230), call `divisionRef.current?.setSelections(updatedDivisions)` before `viewModel.handleFilterDivision(updatedDivisions)` so the dropdown's checked state stays in sync with the pills when a division pill is removed. Add a test that removes a division pill on the Trustee List page and asserts the dropdown no longer shows it selected.

user-interface/src/search/SearchScreen.tsx
```diff
@@ -201,36 +198,9 @@ export default function SearchScreen() {
...
-        courtSelectionRef.current?.setSelections(defaultOptions);
-      })
-      .catch(() => {
-        globalAlert?.error('Cannot load office list');
-      });
-    // TODO resolving this warning introduces an infinite loop. This may be a smell.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [globalAlert]);
-
   function disableSearchForm(value: boolean) {
```
jamesobrooks, line 201: **F2 (Medium) — District (Division) control is no longer disabled during an in-flight search.**

The pre-refactor code called `courtSelectionRef.current?.disable(value)` inside `disableSearchForm`. That call was removed in this PR, and the new `DistrictDivisionComboBoxRef` exposes only `setSelections` (no `disable`), so there's no path to disable it. The case-number, debtor-name, and chapter inputs still lock while a search is in flight, but the district/division control now stays interactive — an inconsistent partial form-lock.

Nothing in the CAMS-770 specs discusses dropping this behavior, and the inner `ComboBox` still supports `disable()`, so this reads as an accidental omission from the ref-type swap rather than a deliberate UX change.

**Prompt to address:**
> Restore the disable-during-search behavior on the search screen's district control. Add a `disable(value: boolean)` method to `DistrictDivisionComboBoxRef` / `useImperativeHandle` in `DistrictDivisionComboBox.tsx` that forwards to the inner `ComboBox` ref, then call `courtSelectionRef.current?.disable(value)` inside `disableSearchForm` in `SearchScreen.tsx` alongside the other three inputs. Add or restore a test asserting the district control is disabled while a search is in flight.

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -189,8 +185,12 @@ export default function TrusteesList() {
   };
 
   const handleFilterDivision = (divisions: ComboOption[]) => {
-    isDistrictFilterInteracted.current = true;
-    lastFilterChanged.current = 'district';
+    if (!isDefaultApplied.current) {
```
jamesobrooks: **F3 (Medium) — First real division selection is swallowed for users with no default offices.**

`handleFilterDivision` treats its first invocation as the mount-time default application (`if (!isDefaultApplied.current) { isDefaultApplied.current = true } else { record interaction }`). This assumes the component always emits `onSelectionsChange` once on mount. But `DistrictDivisionComboBox` only emits on mount when defaults exist (`if (defaults.length > 0)`). For a user with no configured offices, no mount-time call fires, so `isDefaultApplied` stays `false` and the user's **first genuine** division selection is misclassified as the default — `isDistrictFilterInteracted`/`lastFilterChanged` never get set, suppressing the aria-live announcement (gated at line 333) for that selection. Users with default offices are unaffected.

**Prompt to address:**
> The first-call-is-default heuristic in `TrusteesList.handleFilterDivision` breaks for users with no default divisions, because `DistrictDivisionComboBox` only emits `onSelectionsChange` on mount when `defaults.length > 0`. Instead of inferring "default applied" from call ordering, make the signal explicit: add an `isDefault`/`source` argument to `onSelectionsChange` (or a separate `onDefaultsApplied` callback) in `DistrictDivisionComboBox.tsx`, and have `TrusteesList` record an interaction based on that intent rather than first-call-wins. Add a test for a no-default-offices user whose first division selection produces an aria-live announcement.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
```diff
@@ -0,0 +1,167 @@
...
+  ref: React.Ref<DistrictDivisionComboBoxRef>,
+) => {
+  const [courts, setCourts] = useState<CourtDivisionDetails[]>([]);
+  const [courtsError, setCourtsError] = useState(false);
+  const [selectedDivisions, setSelectedDivisions] = useState<ComboOption[]>([]);
+  const [divisionComboOptions, setDivisionComboOptions] = useState<ComboOption[]>([]);
+  const [upgradeAnnouncement, setUpgradeAnnouncement] = useState('');
+  const previousSelectionsRef = useRef<ComboOption[]>([]);
+
+  useImperativeHandle(ref, () => ({
```
jamesobrooks, line 54: **F8 (Medium) — The `setSelections` public ref API has zero test coverage.**

`setSelections` is exported as part of the component's public contract via `useImperativeHandle`, but no test attaches a ref, so the handle is never exercised in `DistrictDivisionComboBox.test.tsx`. Its only production call site — division-pill removal at `TrusteeCaseListFilterView.tsx:188` — is also uncovered: `TrusteeCaseListFilter.test.tsx` tests status-pill and filed-date-pill removal but not division-pill removal. The branch that updates `previousSelectionsRef` (which drives mutual-exclusion/auto-upgrade diffing) could regress silently.

**Prompt to address:**
> Add coverage for the `setSelections` imperative handle in `DistrictDivisionComboBox.tsx`. Either (a) in `DistrictDivisionComboBox.test.tsx`, render with a `DistrictDivisionComboBoxRef` ref, call `ref.current.setSelections([...])`, and assert the combobox reflects the new selection and that a subsequent change diffs against it correctly; or (b) add a division-pill-removal test to `TrusteeCaseListFilter.test.tsx` that exercises the `divisionRef.current?.setSelections(...)` path end to end.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
```diff
@@ -0,0 +1,167 @@
...
+      setSelectedDivisions(selections);
+    },
+  }));
+
+  useEffect(() => {
+    Api2.getCourts()
+      .then((r) => {
+        const allCourts = r.data;
+        setCourts(allCourts);
+        onCourtsLoaded?.(allCourts);
```
jamesobrooks, line 67: **F9 (Medium) — `onCourtsLoaded` and the division-code resolution path are untested.**

`onCourtsLoaded?.(allCourts)` is never asserted in the component test, and it's load-bearing: `TrusteeCaseListFilter` wires it to `setCourts`, and `trusteeCaseListFilterUseCase.handleDivisionChange` uses `store.courts` in `encodeDivisionCodes` to resolve `divisionCodes`. No parent test selects a division and asserts the **resolved** codes reach `onFilterChange` — the existing test passes codes via `initialValue`, which bypasses the `onCourtsLoaded -> setCourts -> encodeDivisionCodes(divisions, store.courts)` path.

**Prompt to address:**
> Add a component test in `DistrictDivisionComboBox.test.tsx` asserting `onCourtsLoaded` is called with the loaded courts after `Api2.getCourts()` resolves. Then strengthen `TrusteeCaseListFilter.test.tsx` with a test that lets courts load, selects a division, and asserts the resolved `divisionCodes` reach `onFilterChange` — covering the `onCourtsLoaded -> setCourts -> encodeDivisionCodes` resolution path end to end.

user-interface/src/search/SearchScreen.tsx
```diff
@@ -146,18 +145,16 @@ export default function SearchScreen() {
   const infoModalId = 'info-modal';
 
   const [chapterList, setChapterList] = useState<ComboOption[]>([]);
-  const [officesList, setOfficesList] = useState<ComboOption[]>([]);
   const [activeElement, setActiveElement] = useState<Element | null>(null);
 
   const caseNumberInputRef = useRef<InputRef>(null);
   const debtorNameInputRef = useRef<InputRef>(null);
-  const courtSelectionRef = useRef<ComboBoxRef>(null);
+  const courtSelectionRef = useRef<DistrictDivisionComboBoxRef>(null);
```
jamesobrooks, line 152: **F1 (Low) — Orphaned `courtSelectionRef` wiring.**

This PR deleted both dereferences of `courtSelectionRef` (the old `.setSelections(...)` and `.disable(...)` calls), but left the `useRef` declaration here and the `ref={courtSelectionRef}` attachment (line ~436) in place. `.current` is never read anywhere in the file, and the test file never touches it — so the ref, and the `DistrictDivisionComboBoxRef` import into this file, are now dead wiring. (Shares a root cause with F2: if you restore `disable` per F2, this ref becomes live again and this finding dissolves.)

**Prompt to address:**
> In `SearchScreen.tsx`, either resolve this together with the disable-during-search finding (F2) by re-adding a `disable` method to the ref and calling it in `disableSearchForm` — which makes `courtSelectionRef` live again — or, if F2 is intentionally not restored, delete the orphaned wiring: remove the `courtSelectionRef` declaration (line 152), the `ref={courtSelectionRef}` prop (line ~436), and the now-unused `DistrictDivisionComboBoxRef` import.

jamesobrooks (commented):
## Review: CAMS-770 — Extract shared `DistrictDivisionComboBox`

Reviewed the extraction with an adversarial two-pass method (each finding was challenged by a fresh, independent skeptic before surviving). The extraction itself is solid — `autoUpgradeToAll`, `resolveCombinedSelections`, `getUserDivisionCodes`, and `encodeDivisionCodes` are genuinely consolidated into `court-utils.ts`, the dual-callback contract serves two real consumers, and the imperative handle has a legitimate single consumer. Nice de-duplication.

Seven findings survived both passes. Six are posted inline at their anchors; one test-quality item is below. Each inline comment includes a ready-to-use prompt the author can paste to address it.

### Summary

| # | Severity | Finding | Location |
|---|----------|---------|----------|
| F5 | **High** | Division dropdown desyncs from pill box on Trustee List page (flag-on) | `TrusteeDistrictFilterView.tsx:191` |
| F2 | Medium | District control no longer disabled during an in-flight search | `SearchScreen.tsx:201` |
| F3 | Medium | First real division selection swallowed for users with no default offices | `TrusteesList.tsx:188` |
| F8 | Medium | `setSelections` public ref API has zero test coverage | `DistrictDivisionComboBox.tsx:49` |
| F9 | Medium | `onCourtsLoaded` + division-code resolution path untested | `DistrictDivisionComboBox.tsx:61` |
| F1 | Low | Orphaned `courtSelectionRef` wiring in SearchScreen | `SearchScreen.tsx:152` |
| F10 | Low | Near-tautological aria-live assertion (below) | `DistrictDivisionComboBox.test.tsx:399-437` |

> Note F1 and F2 share a root cause — the ref-type swap on the search screen dropped the `disable()` call (F2) and left the ref attached but never dereferenced (F1). Restoring `disable` resolves both together.

---

### F10 (Low, test quality) — `DistrictDivisionComboBox.test.tsx:399-437`

The test *"aria-live region resets to empty before re-announcing..."* asserts `expect(announcements).toContain('')` (line 431). Because `''` is the live region's resting value and the stubbed `requestAnimationFrame` samples `textContent` before React commits the `setUpgradeAnnouncement(labels)` call, this assertion passes even if the reset-before-reannounce logic (component line 118) were removed — it doesn't discriminate. The meaningful guarantee (the label is re-announced on a repeat upgrade) is already carried by the surrounding `waitFor` assertions.

**Prompt to address:**
> In `DistrictDivisionComboBox.test.tsx`, the `aria-live region resets to empty before re-announcing` test asserts `expect(announcements).toContain('')`. Since `''` is the region's initial value and the `requestAnimationFrame` stub samples before React commits the label, this assertion would still pass if the line-118 reset were removed. Replace it with an assertion that the region transitions `'' -> label` across the two upgrade cycles (or remove the `announcements` probe and rely on the existing `waitFor` checks), so the test actually guards the reset-before-reannounce behavior.

---

*Findings verified against the head commit. The `encodeDivisionCodes` double-resolution concern was independently examined and **rejected** — the two call sites serve independent consumers with different output shapes, not a sync-by-convention coupling — and it overlaps the thread already open with @mattstankey, so it is intentionally omitted here.*

jamesobrooks:
@sourcery-ai Your Reviewer's Guide / summary attributes backend work to this PR that isn't here:

> Support passing divisionCodes through trustee case list API requests and backend search predicates, including query parsing of comma-separated values.
> Extend trustee filters, trustee case list, controller, and use-case tests to cover divisionCodes propagation...

This PR's changed-file set is 100% `user-interface/src/` — there are no `backend/` or `common/` files in the diff. The `divisionCodes` controller parsing, search-predicate, and use-case test coverage you describe already shipped to `main` under the earlier CAMS-770 Case List slice (e.g. `backend/lib/controllers/trustee-cases/trustee-cases.controller.ts` and its tests). This PR is a frontend-only extraction of the shared `DistrictDivisionComboBox` component. Flagging so the summary doesn't lead reviewers to verify backend changes that aren't part of this PR.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
mattstankey: ✅ Confirmed — `wrapPills` prop is already accepted in `DistrictDivisionComboBoxProps` and forwarded to the inner `<ComboBox>`. The call site at line 153 already passes `wrapPills={true}`. No change needed.

user-interface/src/trustees/panels/filters/TrusteeCaseListFilterView.tsx
mattstankey: ✅ Confirmed — `TrusteeCaseListFilterView.tsx` line 1 already has a single combined import: `import ComboBox, { ComboOption } from '@/lib/components/combobox/ComboBox'`. No change needed.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
mattstankey: ✅ Confirmed — `TrusteeCaseListFilterViewModel` does not contain a `divisionRef`. The ref is created locally inside `TrusteeCaseListFilterView.tsx` with `const divisionRef = useRef<DistrictDivisionComboBoxRef>(null)` and never exposed through the interface. No change needed.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
mattstankey: The `as ComboOption[]` casts are a known existing issue with `getDistrictDivisionComboOptions` and `separateDefaultOptions` returning a wider type. Those utilities predate this PR and fixing their return types would be a broader refactor touching multiple callers. Agreed this is a real weakness — filing a follow-up task to narrow the return types at the source rather than address it in this PR.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
mattstankey: ✅ The `previousDivisionValuesRef` was removed in a previous commit. The current version of the component only has `previousSelectionsRef` and derives the previous values set from it inline in `handleDivisionChange`. No action needed.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
mattstankey: ✅ Confirmed — both branches now call both callbacks when defaults are found. The `initialDivisionCodes` path calls `onDivisionCodesChange?.(codes)` and `onSelectionsChange?.(defaults)`; the session-defaults path does the same. The asymmetry in the cited commit has been corrected.

user-interface/src/trustees/filters/TrusteeDistrictFilterView.tsx
mattstankey: ✅ Fixed (F5). Wired a `DistrictDivisionComboBoxRef` ref in `TrusteeDistrictFilterView` and call `divisionRef.current?.setSelections(updatedDivisions)` in the PillBox `onSelectionChange` handler before `viewModel.handleFilterDivision(updatedDivisions)`, mirroring the pattern in `TrusteeCaseListFilterView`.

user-interface/src/search/SearchScreen.tsx
mattstankey: ✅ Fixed (F2). Added `disable: (value: boolean) => void` to `DistrictDivisionComboBoxRef` and `useImperativeHandle`. It forwards to the inner `ComboBox` ref. Restored `courtSelectionRef.current?.disable(value)` in `disableSearchForm` alongside the other three inputs.

user-interface/src/trustees/TrusteesList.tsx
mattstankey: ✅ Fixed (F3). Added `onDefaultsApplied?: () => void` to `DistrictDivisionComboBox` — it fires unconditionally after the mount-time defaults resolution (even when no defaults exist). Threaded the prop through `TrusteeDistrictFilterProps`, `TrusteeDistrictFilterViewModel`, `TrusteeDistrictFilter`, and `TrusteeDistrictFilterView`. In `TrusteesList`, `onDivisionDefaultsApplied` now sets `isDefaultApplied.current = true` explicitly, and `handleFilterDivision` always marks `isDistrictFilterInteracted = true` — eliminating the first-call-wins heuristic that silently dropped the aria-live announcement for no-default-offices users.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
mattstankey: Agreed. Test coverage for the `setSelections` imperative handle and division-pill-removal path is a gap. Adding tests for both in a follow-up (the ref wiring changes in this push make it straightforward to add — will do so before merge).

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
mattstankey: Agreed. Tests for `onCourtsLoaded` callback assertion and the end-to-end `onCourtsLoaded → setCourts → encodeDivisionCodes` resolution path are missing. Will add before merge.

user-interface/src/search/SearchScreen.tsx
mattstankey: ✅ Resolved (F1). Restored F2 (disable during search) by adding `disable()` to `DistrictDivisionComboBoxRef` — so `courtSelectionRef` is now live again: the declaration, `ref={courtSelectionRef}`, and `courtSelectionRef.current?.disable(value)` in `disableSearchForm` are all active.

jalba-flexion (approved):

user-interface/src/trustees/TrusteesList.tsx
```diff
@@ -416,13 +407,21 @@ export default function TrusteesList() {
       { name: 'Trustee District Filter Changed' },
       {
         isDefault,
-        selectedCount: selectedDistricts.length,
```
cruzjone-flexion, line 419: **`isDefault` analytics flag is always `true` when `districtDivisionEnabled` is on**

When the district-division flag is enabled, `handleFilterDistrict` is never called — divisions replace districts entirely. `defaultDistrictsRef.current` stays `[]` and `selectedDistricts` stays `[]`, so the `isDefault` comparison resolves to `[] === []` — always `true`. Every `Trustee District Filter Changed` event will be tagged `isDefault: true` regardless of what the user actually filtered.

```tsx
const isDefault =
  selectedDistricts.length === defaults.length &&   // 0 === 0 always when flag is on
  selectedDistricts.every((d) => defaults.some((def) => def.value === d.value));
```

When `districtDivisionEnabled`, `isDefault` should reflect the division state instead — for example, `selectedDivisions.length === 0` as a proxy for "user has not filtered yet," or a comparison against the default divisions that were applied on mount.

cruzjone-flexion: Not a deal breaker, will approve.

cruzjone-flexion (approved):

PR #2584, opened 2026-06-26
Add projection to query builder

btposey:
# Purpose

Add the ability to choose which properties we want to return from Cosmos.

# Major Changes

Added projection support and accompanying renderer to Query Builder.

# Testing/Validation

Added an integration test for the Query Builder to prove the library and the mongo renderer work.

## Summary by Sourcery

Add field projection support to the query builder and Mongo MongoDB renderer, and introduce an integration test harness to validate query behaviour end-to-end against a real Mongo instance.

New Features:
- Introduce Projection type with pick and omit helpers on the query builder for specifying included or excluded fields.
- Add MongoDB projection rendering that converts query builder projections into Mongo projection documents.
- Add a standalone integration test suite and scripts for running query builder and Mongo renderer tests against a real MongoDB instance.

Bug Fixes:
- Ensure condition and conjunction type guards safely handle null inputs.
- Correct NOT conjunction rendering in the Mongo query renderer to use $nor instead of $not.
- Add support for notExists conditions in the Mongo query renderer tests.

Build:
- Add an npm script to run the new query-builder integration test harness.

Tests:
- Extend unit tests for the query builder and Mongo renderer to cover projections, null handling, and NOT/NOT_EXISTS behaviour.
- Add integration tests with fixtures for cases, trustees, and consolidation orders to exercise the full query builder API including projection, sorting, limits, and conjunctions.

### DIFF
--- a/backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.test.ts
+++ b/backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.test.ts
@@ -123,7 +123,7 @@ describe('Mongo Query Renderer', () => {
     {
       caseName: 'NOT',
       func: () => not(doc('two').equals(45)),
-      expected: { $not: [{ two: { $eq: 45 } }] },
+      expected: { $nor: [{ two: { $eq: 45 } }] },
     },
   ];
 
--- a/backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.test.ts
+++ b/backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.test.ts
@@ -1,6 +1,7 @@
 import { ObjectId } from 'mongodb';
 import QueryBuilder, { Field, Query } from '../../../../query/query-builder';
-import { toMongoQuery, toMongoSort } from './mongo-query-renderer';
+const { pick, omit } = QueryBuilder;
+import { toMongoProjection, toMongoQuery, toMongoSort } from './mongo-query-renderer';
 
 type Foo = {
   uno: string;
@@ -45,6 +46,11 @@ describe('Mongo Query Renderer', () => {
       caseName: 'EXISTS',
       func: () => doc('two').exists(),
       expected: { two: { $exists: true } },
+    },
+    {
+      caseName: 'NOT_EXISTS',
+      func: () => doc('two').notExists(),
+      expected: { two: { $exists: false } },
     },
     {
       caseName: 'EQUALS',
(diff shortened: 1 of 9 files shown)


btposey:
@sourcery-ai Both suggestions addressed in the latest commit:

1. **`isProjection` array check** — added `Array.isArray((obj as Projection).fields)` to the guard, with a corresponding test case for `{ fields: "not-an-array", mode: "INCLUDE" }`.

2. **`toMongoProjection` performance** — replaced the spread-based `reduce` with a mutating `forEach` that writes directly into a pre-allocated result object.

bityogi (commented):
Minor nit: `mongo-query-renderer.test.ts` has a `const` declaration between `import` statements (line 3). Moving it below all imports (or destructuring `pick`/`omit` as named exports directly) would be more conventional and avoid potential linter complaints.

```ts
// Instead of:
import QueryBuilder, { Field, Query } from '../../../../query/query-builder';
const { pick, omit } = QueryBuilder;
import { toMongoProjection, toMongoQuery, toMongoSort } from './mongo-query-renderer';

// Either:
import QueryBuilder, { Field, Query, pick, omit } from '../../../../query/query-builder';
import { toMongoProjection, toMongoQuery, toMongoSort } from './mongo-query-renderer';

// Or move the destructure below imports:
import QueryBuilder, { Field, Query } from '../../../../query/query-builder';
import { toMongoProjection, toMongoQuery, toMongoSort } from './mongo-query-renderer';

const { pick, omit } = QueryBuilder;
```

jalba-flexion (approved):

backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.test.ts
```diff
@@ -117,7 +123,7 @@ describe('Mongo Query Renderer', () => {
     {
       caseName: 'NOT',
       func: () => not(doc('two').equals(45)),
-      expected: { $not: [{ two: { $eq: 45 } }] },
+      expected: { $nor: [{ two: { $eq: 45 } }] },
```
mattstankey, line 126: `$nor` is used here instead of `$not` because MongoDB's `$not` is a field-level operator and is invalid at the top-level query position. `$nor: [{ condition }]` is semantically equivalent to "NOT condition" at the document level and is the correct top-level negation operator.

btposey: ✅ Fixed — moved `const { pick, omit } = QueryBuilder` below the import block.

btposey (commented):
✅ Fixed — moved `const { pick, omit } = QueryBuilder` below the import block.

backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.test.ts
btposey: Confirmed — the implementation is correct. MongoDB `$not` is field-level only (valid as `{ field: { $not: expr } }`), not at the document root. `$nor: [condition]` with a single element is the correct idiom for top-level negation. No change needed here.

PR #2589, opened 2026-06-26
CAMS-768 Slice 6: Real ACS email provider integration

fmaddenflx:
## Purpose

Replace the NoOpNotificationGateway with a real Azure Communication Services (ACS) email provider so that trustee change notifications actually send emails to recipients.

## Major Changes

* Added `@azure/communication-email@1.1.0` SDK (SHA-512 integrity verified, no known vulnerabilities)
* Implemented `AcsNotificationGateway` — uses `EmailClient.beginSend()` with long-running operation polling and status verification
* Updated factory selection: `dbMock` → `MockNotificationGateway`, otherwise → `AcsNotificationGateway`
* Removed `NoOpNotificationGateway` and its test — no fallback needed
* New env vars required for deployed environments:
  - `ACS_EMAIL_CONNECTION_STRING` — Azure Communication Services connection string
  - `ACS_EMAIL_SENDER_ADDRESS` — verified sender address

## Testing/Validation

* 4 unit tests for `AcsNotificationGateway`:
  - Sends email with correct message structure
  - Throws on non-Succeeded status
  - Propagates network errors from ACS client
  - Handles notifications without displayName
* All existing notification dispatch tests pass (use `dbMock=true` → MockNotificationGateway)
* TypeScript compiles clean
* Build validation passes

## Notes

* The notification pipeline remains gated behind the `trustee-change-notification-enabled` feature flag — ACS will only receive send requests when the flag is enabled in LaunchDarkly
* The factory throws if `ACS_EMAIL_CONNECTION_STRING` or `ACS_EMAIL_SENDER_ADDRESS` are missing in non-mock environments — fail fast rather than silent no-op

## Definition of Done:

- [x] Code refactored for clarity
- [x] Dependency rule followed
- [x] Development debt eliminated (NoOp removed)
- [x] No regressions

## Summary by Sourcery

Integrate Azure Communication Services as the production notification gateway and remove the previous no-op implementation.

New Features:
- Add AcsNotificationGateway to send trustee notification emails via Azure Communication Services email API.

Enhancements:
- Select AcsNotificationGateway in the application factory when dbMock is disabled, failing fast if required ACS email configuration is missing.

Build:
- Add @azure/communication-email dependency to the backend package.

Tests:
- Add unit tests covering AcsNotificationGateway happy path, error propagation, non-succeeded statuses, and notifications without a display name.
- Remove tests for the deprecated NoOpNotificationGateway.

Chores:
- Remove the obsolete NoOpNotificationGateway implementation.


backend/package.json
```diff
@@ -37,6 +37,7 @@
     "typecheck": "npx tsc --noEmit --skipLibCheck"
   },
   "dependencies": {
+    "@azure/communication-email": "1.1.0",
     "@azure/cosmos": "4.9.3",
```
mattstankey, line 41: **Node version incompatibility**: `@azure/communication-email@1.1.0` and its transitive dependency `@azure/communication-common@2.4.2` both declare `"node": ">=20.0.0"` in their `package.json`. The project CLAUDE.md documents Node 18.13.0 as the minimum. If any deployed environments are still on Node 18, this will fail at runtime. Please verify all deployment targets are on Node 20+ and update the documented minimum if they are.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const MODULE_NAME = 'ACS-NOTIFICATION-GATEWAY';
+
+export class AcsNotificationGateway implements NotificationGateway {
+  private readonly client: EmailClient;
+  private readonly senderAddress: string;
+
+  constructor(connectionString: string, senderAddress: string) {
+    this.client = new EmailClient(connectionString);
+    this.senderAddress = senderAddress;
```
mattstankey, line 21: **Connection string auth vs. Managed Identity**: Using a connection string with an embedded access key (`endpoint=...;accesskey=...`) is less secure than Managed Identity. In deployed Azure environments, prefer `DefaultAzureCredential` via `@azure/identity` (which is already a backend dependency):

```ts
import { DefaultAzureCredential } from '@azure/identity';
// constructor accepts (endpoint: string, credential: TokenCredential)
this.client = new EmailClient(endpoint, new DefaultAzureCredential());
```

If a connection string is intentionally chosen (e.g. for simplicity or environment constraints), a brief comment here explaining that decision would be helpful.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
```diff
@@ -0,0 +1,43 @@
...
+            address: notification.to,
+            displayName: notification.toDisplayName,
+          },
+        ],
+      },
+    };
+
+    const poller = await this.client.beginSend(message);
+    const result = await poller.pollUntilDone();
+
```
mattstankey, line 53: **No polling timeout**: `pollUntilDone()` has no abort signal. If ACS becomes unresponsive, this awaits indefinitely — potentially hanging the Azure Function until the host timeout kills it. Consider passing an `AbortSignal`:

```ts
const result = await poller.pollUntilDone({ abortSignal: AbortSignal.timeout(30_000) });
```

This gives the operation a bounded failure window and a clearer error message than a host-level function timeout.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (result.status !== 'Succeeded') {
+      throw new Error(
+        `${MODULE_NAME}: Email send failed with status '${result.status}' (id: ${result.id})`,
```
bityogi: **Use `CamsError` instead of bare `Error`**: Other gateways in this project throw structured `CamsError` instances (see `gateway-helper.ts`, `azure-blob-object-storage.gateway.ts`). Using it here enables consistent upstream error handling and structured logging:

```ts
import { CamsError } from '../../../../common-errors/cams-error';

throw new CamsError(MODULE_NAME, {
  message: `Email send failed with status '${result.status}' (id: ${result.id})`,
});
```

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
```diff
@@ -0,0 +1,43 @@
...
+export class AcsNotificationGateway implements NotificationGateway {
+  private readonly client: EmailClient;
+  private readonly senderAddress: string;
+
+  constructor(connectionString: string, senderAddress: string) {
+    this.client = new EmailClient(connectionString);
+    this.senderAddress = senderAddress;
+  }
+
+  async send(notification: Notification): Promise<void> {
```
bityogi, line 25: **No observability**: This gateway has zero logging. Other gateways receive `ApplicationContext` which provides access to the logger. At minimum, logging the ACS message ID on success would aid production debugging ("was this email actually sent?").

The `NotificationGateway` interface doesn't currently accept context, so this could be a follow-up if the interface change is too broad for this slice — but worth tracking since this will be the *only* way to confirm email delivery without checking ACS directly.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
```diff
@@ -0,0 +1,43 @@
...
+        html: notification.html,
+        plainText: notification.text,
+      },
+      recipients: {
+        to: [
+          {
+            address: notification.to,
+            displayName: notification.toDisplayName,
+          },
+        ],
```
bityogi, line 39: **`correlationId` not used**: The `Notification` type includes a `correlationId` field, but it's never passed to ACS. ACS supports custom headers on the message object — threading the correlation ID through would let you trace a notification from the CAMS event all the way to ACS delivery logs. Not blocking, but a missed observability opportunity for v1.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.test.ts
```diff
@@ -0,0 +1,83 @@
+import { vi } from 'vitest';
+import { AcsNotificationGateway } from './acs-notification.gateway';
+import { Notification } from '@common/cams/notifications';
+
+const mockPollUntilDone = vi.fn();
+const mockBeginSend = vi.fn().mockResolvedValue({ pollUntilDone: mockPollUntilDone });
+
```
bityogi: **`vi.mock()` vs project convention**: Per the project's testing standards, `vi.spyOn()` is preferred over `vi.mock()` (which is reserved as a last resort). The root cause here is that `AcsNotificationGateway` internally constructs its `EmailClient`, forcing a module-level mock.

Consider accepting an `EmailClient` instance (or a factory function) via the constructor — this aligns with Sourcery's suggestion and would let you test with spies:

```ts
constructor(client: EmailClient, senderAddress: string) { ... }
```

The factory in `factory.ts` would then construct the client and inject it.

backend/lib/factory.ts
```diff
@@ -533,7 +533,14 @@ const getNotificationGateway = (context: ApplicationContext): NotificationGatewa
     if (context.config.get('dbMock')) {
       notificationGateway = MockNotificationGateway.getInstance();
     } else {
-      notificationGateway = new NoOpNotificationGateway();
+      const connectionString = process.env.ACS_EMAIL_CONNECTION_STRING;
+      const senderAddress = process.env.ACS_EMAIL_SENDER_ADDRESS;
+      if (!connectionString || !senderAddress) {
```
bityogi, line 539: **Eager throw at gateway-construction time**: This throws when env vars are missing at the moment the notification gateway is first requested — which may happen during *any* request that touches the factory's gateway cache, not just notification sends.

Verify this doesn't break startup or health checks for non-notification Azure Functions. If it does, consider deferring validation to `send()` (fail on first use) or ensuring `getNotificationGateway` is only called in notification-specific code paths.

The existing `AzureBlobObjectStorageGateway` does a similar eager throw for `AzureWebJobsStorage`, so this may already be the established pattern — just worth confirming the blast radius.

bityogi (commented):
Additional review recommendations (5-9) beyond what mattstankey and Sourcery already raised.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
bityogi: ✅ Fixed in d506aab. Added `AbortSignal.timeout(30_000)` to `pollUntilDone()`:

```ts
const result = await poller.pollUntilDone({
  abortSignal: AbortSignal.timeout(POLL_TIMEOUT_MS),
});
```

This gives a bounded 30s failure window with a clear abort error rather than hanging until the host kills the function.

backend/package.json
bityogi: Not an issue — the project `.nvmrc` specifies `v22.22.2`, so all environments are Node 22+. The `>=20.0.0` engine requirement is satisfied.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
bityogi: ✅ Partially addressed in d506aab. Refactored the constructor to accept an `EmailClient` instance rather than a connection string — this decouples auth strategy from the gateway. The factory constructs the client externally:

```ts
const client = new EmailClient(connectionString);
notificationGateway = new AcsNotificationGateway(client, senderAddress, context.logger);
```

Migrating to `DefaultAzureCredential` is now a one-line change in the factory (pass endpoint + credential instead of connection string to `new EmailClient()`). Tracking the full Managed Identity migration as a separate follow-up since it requires RBAC role assignment and Bicep changes.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
bityogi: ✅ Fixed in d506aab. Now throws `CamsError` with module name for structured error handling:

```ts
throw new CamsError(MODULE_NAME, {
  message: `Email send failed with status '${result.status}' (id: ${result.id})`,
});
```

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
bityogi: ✅ Fixed in d506aab. Gateway now accepts an optional `NotificationLogger` via constructor and logs on success:

```ts
this.logger?.info(MODULE_NAME, `Email sent successfully`, {
  messageId: result.id,
  to: notification.to,
  correlationId: notification.correlationId,
});
```

Used a minimal `NotificationLogger` interface rather than the full `ApplicationContext` — keeps the gateway decoupled while the factory passes `context.logger` at construction time.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
bityogi: ✅ Fixed in d506aab. Now threads `correlationId` as a custom header to ACS:

```ts
headers: notification.correlationId
  ? { 'X-Correlation-Id': notification.correlationId }
  : undefined,
```

This enables end-to-end tracing from the CAMS invocation through to ACS delivery logs.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.test.ts
bityogi: ✅ Fixed in d506aab. Refactored the constructor to accept an `EmailClient` instance directly, eliminating the need for `vi.mock()`. Tests now use an injected mock:

```ts
const mockClient = { beginSend: mockBeginSend } as unknown as EmailClient;
gateway = new AcsNotificationGateway(mockClient, senderAddress, mockLogger);
```

No module-level mocking needed — fully aligned with the project convention of preferring `vi.spyOn()` / constructor injection over `vi.mock()`.

backend/lib/factory.ts
bityogi: This follows the same pattern as `AzureBlobObjectStorageGateway` which also throws eagerly for missing `AzureWebJobsStorage`. The `getNotificationGateway` is only called from `TrusteeChangeNotificationUseCase` which is only invoked on the notification event path — not on health checks or general API requests. Blast radius is already scoped.

backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts
fmaddenflx: Nothing wrong necessarily with bicep changes at this stage.  Better to get everything right now than to wait until after we merge.

jalba-flexion (approved):

mattstankey (approved):

bityogi (approved):

PR #2597, opened 2026-06-29
CAMS-780 Gate BanksController behind trustee-software-bank-display flag

cruzjone-flexion:
# Purpose

Enforce the same `trustee-software-bank-display` flag gate on the backend that Slice 1 added to the frontend. Without this, a caller who knows the API URL can still reach Banks endpoints even when the flag is off.

# Major Changes

* `BanksController.handleRequest` now throws `403 Forbidden` with the message "Banks feature is not enabled." when `context.featureFlags['trustee-software-bank-display']` is falsy — matching the pattern from `PrivilegedIdentityAdminController`
* Added a new test asserting the 403 response for the flag-off case
* Set `context.featureFlags['trustee-software-bank-display'] = true` in `beforeEach` so existing routing tests are unaffected

# Testing/Validation

Implemented using TDD. 23 tests pass in `banks.controller.test.ts`. All 217 backend test files pass with no regressions. TypeScript, lint, and knip clean.

# Notes

This PR targets the Slice 1 branch (`CAMS-780-privileged-identity-feature-flag`) and should be merged there, not directly to `main`. It will reach `main` when the Slice 1 PR merges.

The flag check sits at the top of `handleRequest` before the role check — same order as `PrivilegedIdentityAdminController`. The `handleGet`, `handleGetOne`, `handlePut`, and `handlePost` methods are intentionally not gated individually since `handleRequest` is the sole entry point from the function app.

# Definition of Done:

## Summary by Sourcery

Gate BanksController requests behind the trustee-software-bank-display feature flag to align backend access with the frontend.

Enhancements:
- Add a feature-flag check in BanksController.handleRequest that forbids access when trustee-software-bank-display is disabled.

Tests:
- Extend banks.controller.test.ts to cover the forbidden response when the trustee-software-bank-display flag is off and default the flag to on for existing routing tests.

### DIFF
--- a/backend/lib/controllers/banks/banks.controller.test.ts
+++ b/backend/lib/controllers/banks/banks.controller.test.ts
@@ -5,6 +5,7 @@ import { BanksController } from './banks.controller';
 import { BanksUseCase } from '../../use-cases/banks/banks';
 import { CamsRole } from '@common/cams/roles';
 import { BankProfile } from '@common/cams/banks';
+import HttpStatusCodes from '@common/api/http-status-codes';
 
 vi.mock('../../use-cases/banks/banks');
 
@@ -37,6 +38,7 @@ describe('BanksController', () => {
   beforeEach(async () => {
     context = await createMockApplicationContext();
     context.session.user.roles = [CamsRole.SuperUser];
+    context.featureFlags['trustee-software-bank-display'] = true;
     controller = new BanksController(context);
   });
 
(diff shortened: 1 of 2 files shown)


mattstankey (approved):

mattstankey:
## Code Review Response — Sourcery Suggestions

Reviewed the two suggestions from @sourcery-ai. Recommending we decline both:

### Suggestion 1 — Optional chaining on `featureFlags`

> Consider guarding the feature flag access with optional chaining (`context.featureFlags?.['trustee-software-bank-display']`)

**Declining.** `ApplicationContext` declares `featureFlags: FeatureFlagSet` as a non-optional field. Every other controller in the codebase (`privileged-identity-admin.controller.ts`, `trustees.controller.ts`, and ~12 others) accesses `context.featureFlags[...]` without optional chaining. Adding `?.` here would be inconsistent with the established pattern and implies a defensive concern the type system already rules out.

### Suggestion 2 — Centralize the flag key string and error message

> The hard-coded flag key and error message could be centralized in a shared constant or config

**Declining.** The error message `NOT_ENABLED` is already a local constant — which matches the `privileged-identity-admin.controller.ts` pattern exactly. The flag key `'trustee-software-bank-display'` is used in only one place in this controller and no other controller currently shares this flag, so centralizing it would be premature abstraction.

The implementation follows the established codebase pattern correctly as-is.

PR #2600, opened 2026-06-30
CAMS-787 Fix appointment status filter leaking wrong-status rows

cruzjone-flexion:
# Bug

When filtering the trustee list by status (e.g. Active), appointments with a mismatched status were still appearing in the results. A trustee who had both an active appointment and a removed appointment would show both rows — the removed one should have been excluded.

# Purpose

Ensure the appointment status filter is applied consistently so the API response only contains appointments matching the requested status.

# Major Changes

* After fetching all appointments for filtered trustee IDs, apply the same status predicate to the appointments themselves before enriching and grouping
* Added a regression test covering a trustee with mixed-status appointments filtered to active

# Testing/Validation

Implemented using TDD. New unit test covers the exact bug scenario: a trustee with one active and one removed appointment filtered to `status=active` — only the active appointment should be returned. All 54 tests in `trustees.test.ts` pass.

# Notes

The bug was a two-step mismatch: `getTrusteeIdsByStatuses` correctly narrowed which trustees to include, but `getAppointmentsByTrusteeIds` fetched all appointments for those trustees without any status filter. The fix adds a filter between those two steps using the already-computed `statusStatuses` array — no new logic required.

# Definition of Done:

## Summary by Sourcery

Ensure trustee appointment status filters are applied consistently so only appointments matching the requested status are returned.

Bug Fixes:
- Fix trustee listing to exclude appointments whose status does not match the requested filter when trustees have mixed-status appointments.

Tests:
- Add a regression test covering trustees with both active and removed appointments to verify only active appointments are returned when filtering by active status.

### DIFF
--- a/backend/lib/use-cases/trustees/trustees.test.ts
+++ b/backend/lib/use-cases/trustees/trustees.test.ts
@@ -421,6 +421,30 @@ describe('TrusteesUseCase tests', () => {
       );
     });
 
+    test('should filter appointments by status when a trustee has mixed-status appointments', async () => {
+      const trustee = MockData.getTrustee({ trusteeId: 'mixed-1' });
+      const activeAppt = MockData.getTrusteeAppointment({ trusteeId: 'mixed-1', status: 'active' });
+      const removedAppt = MockData.getTrusteeAppointment({
+        trusteeId: 'mixed-1',
+        status: 'removed',
+      });
+
+      vi.spyOn(MockMongoRepository.prototype, 'getTrusteeIdsByStatuses').mockResolvedValue([
+        'mixed-1',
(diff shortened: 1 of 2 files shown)


mattstankey (approved):

cruzjone-flexion:
**Suggestion 1 (variable renames):** Implemented — `statusStatuses` → `requestedStatuses` and `statusFilteredAppointments` → `filteredAppointmentsByStatus`.

**Suggestion 2 (centralize filtering logic):** Skipping for now. There is only one caller of this logic at the moment, so extracting a shared helper would be premature abstraction. If a second caller emerges, that's the right time to centralize.

PR #2603, opened 2026-06-30
CAMS-789 Display case number and trustee name in browser tab

mattstankey:
# Bug

`DocumentTitle` used an empty `useEffect` dependency array (`[]`), so `document.title` was only set on initial mount and never updated when the `name` prop changed. This caused every page to show a generic label (e.g. "Case Detail") regardless of which case or trustee was loaded.

# Purpose

When analysts and attorneys have multiple CAMS tabs open, all tabs show identical generic labels making it impossible to identify the right tab at a glance. This change makes the Case Detail and Trustee Detail tabs show the specific case number or trustee name.

# Major Changes

* Fixed `DocumentTitle` component: changed `useEffect` dependency array from `[]` to `[name]` so the title updates reactively when props change
* `CaseDetailScreen` now passes the case number derived from the URL param immediately — no API response required, so the tab is correct even before the page finishes loading
* `TrusteeDetailScreen` passes the trustee's full name once loaded; generic label shown during load and on the not-found path
* Added unit tests for `DocumentTitle` (new test file), `CaseDetailScreen`, and `TrusteeDetailScreen`

# Testing/Validation

* Unit tests written for all changed components (3210 tests passing across all workspaces)
* `DocumentTitle` tests verify both mount behavior and reactive updates on prop change
* Screen tests verify the correct value is present in `document.title` for loading, loaded, and not-found states
* Accessibility tests pass

# Notes

The `DocumentTitle` fix is safe for all existing callers — every other usage passes a static string literal, so adding `name` to the dependency array has no behavioral effect on those pages (a static string never changes between renders).

The case number is available immediately from `getCaseNumber(caseId)` using the URL param, so `CaseDetailScreen` never needs a loading state for the tab title.

# Definition of Done:

## Summary by Sourcery

Update browser tab titles so case and trustee detail pages show specific case numbers or trustee names, backed by a reactive DocumentTitle component and new tests verifying title behavior in key states.

New Features:
- Display the specific case number in the Case Detail browser tab title.
- Display the trustee’s full name in the Trustee Detail browser tab title after data load.

Bug Fixes:
- Ensure the DocumentTitle component updates document.title whenever the name prop changes instead of only on initial mount.

Enhancements:
- Use the case number derived from the URL to set the Case Detail tab title immediately without waiting for API responses.
- Set a generic Trustee Detail tab title during loading and when the trustee record is not found to maintain a meaningful tab label.

Tests:
- Add unit tests for DocumentTitle to cover initial mount and reactive updates when the name prop changes.
- Add CaseDetailScreen tests to verify the browser tab title includes the case number.
- Add TrusteeDetailScreen tests to verify the browser tab title behavior across loading, loaded, and not-found states.

### DIFF
--- a/user-interface/src/case-detail/CaseDetailScreen.test.tsx
+++ b/user-interface/src/case-detail/CaseDetailScreen.test.tsx
@@ -89,6 +89,15 @@ describe('Case Detail screen tests', () => {
     await TestingUtilities.waitForDocumentBody();
   }
 
+  test('should set document title with case number', async () => {
+    await renderWithRoutes(defaultTestCaseDetail);
+    await waitFor(() => {
+      expect(document.title).toBe(
+        `Case ${getCaseNumber(caseId)} | U.S. Trustee Program - Case Management System (CAMS)`,
+      );
+    });
+  });
+
   test('should render CaseDetailHeader', async () => {
     const headerSpy = vi.spyOn(detailHeader, 'default');
 
--- a/user-interface/src/case-detail/CaseDetailScreen.test.tsx
+++ b/user-interface/src/case-detail/CaseDetailScreen.test.tsx
@@ -89,12 +89,10 @@ describe('Case Detail screen tests', () => {
     await TestingUtilities.waitForDocumentBody();
   }
 
-  test('should set document title with case number', async () => {
+  test('should set browser tab title to the case number', async () => {
     await renderWithRoutes(defaultTestCaseDetail);
     await waitFor(() => {
-      expect(document.title).toBe(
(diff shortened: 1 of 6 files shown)


jalba-flexion (approved):

mattstankey:
Addressed both Sourcery suggestions:

1. **Fallback for `trustee.name`** — Added `?? 'Trustee Detail'` in the loaded path of `TrusteeDetailScreen`. TypeScript types `name` as a non-optional string so this case is unreachable in practice, but the fallback makes the intent explicit and guards against any future type loosening.

2. **`return null` instead of `return <></>`** — Updated `DocumentTitle` to return `null`. More idiomatic React for a component that renders nothing.

cruzjone-flexion (approved):

PR #2606, opened 2026-06-30
Add factory.getObservability seam for telemetry

jamesobrooks:
# Purpose

A scheduled constrained-test report flagged `sync-trustee-appointments`'s `handleStart` unit test as an outlier (~840ms vs 1-4ms for peers). Investigation showed the cost was a lazy `require('applicationinsights')` (~410ms cold) triggered because five dataflow handlers - and the shared Azure context creator - constructed `AppInsightsObservability` directly instead of resolving it through a dependency seam. Under `DATABASE_MOCK` that SDK load leaked into test runtime, and in production every context held a distinct gateway instance rather than a shared one.

This is the OeA dependency-seam fix for that smell. (The separate staging/prod dataflow *runtime* timeouts - non-sargable DXTR query and serial Cosmos loop - are not addressed here.)

# Major Changes

- Add `factory.getObservability(logger?)` as the single sanctioned construction point for `ObservabilityGateway`, backed by a process-wide singleton.
- Add a production `NoOpObservability` (resolved under `DATABASE_MOCK`) so tests and local dev never load the App Insights SDK; the test-only `mockObservability` stays on its own side of the test/prod fence.
- Route the 5 dataflow handlers and the shared `function-apps/azure/application-context-creator.ts` default through the factory instead of `new AppInsightsObservability(...)`.
- Thread a per-invocation `logger` through `completeTrace` (dataflow path + 8 API use-case call sites) so warm-worker diagnostics attribute to the current invocation rather than the first request that built the singleton.

# Testing/Validation

- Full backend suite green via canonical `npm test`: **217 files, 2991 tests, 0 failures**.
- ESLint: 0 errors on all changed files.
- `sync-trustee-appointments` test file: **~932ms -> ~46ms** (SDK no longer loads under test), which also de-noises the constrained-test-report rankings.
- Telemetry correctness verified against deployed Bicep configs: no deployed environment sets `DATABASE_MOCK`, so staging/prod resolve the real `AppInsightsObservability`; event correlation rides `trace.invocationId`, unaffected by the singleton.

# Notes

- Reviewed via a two-turn adversarial multi-agent pass (architect, SRE, test-quality, pragmatic). No correctness findings survived; two follow-up review nits (doc-comment scope + test env restore) are in the second commit.
- Deferred follow-up: replace `completeTrace`'s two trailing optional positional params (`metrics?`, `logger?`) with an options object to remove `undefined` placeholders at call sites.

# Definition of Done:

### DIFF
--- a/backend/function-apps/dataflows/import/sync-trustee-appointments.ts
+++ b/backend/function-apps/dataflows/import/sync-trustee-appointments.ts
@@ -12,7 +12,7 @@ import { buildQueueError } from '../../../lib/use-cases/dataflows/queue-types';
 import { TrusteeAppointmentSyncEvent } from '@common/cams/dataflow-events';
 import { TrusteeAppointmentsSyncState } from '../../../lib/use-cases/gateways.types';
 import { STORAGE_QUEUE_CONNECTION } from '../../../lib/storage-queues';
-import { AppInsightsObservability } from '../../../lib/adapters/services/observability';
+import factory from '../../../lib/factory';
 import { completeDataflowTrace } from '../../../lib/use-cases/dataflows/dataflow-telemetry';
 import { handleRateLimitRetry } from '../dataflows-rate-limit';
 import { getCamsError } from '../../../lib/common-errors/error-utilities';
@@ -77,7 +77,7 @@ async function handleStart(
   invocationContext: InvocationContext,
 ) {
   const logger = ContextCreator.getLogger(invocationContext);
-  const observability = new AppInsightsObservability(logger);
+  const observability = factory.getObservability(logger);
   const trace = observability.startTrace(invocationContext.invocationId);
   try {
     const context = await ContextCreator.getApplicationContext({
@@ -292,7 +292,7 @@ async function handlePagePoison(
 
 async function timerTrigger(_timer: Timer, invocationContext: InvocationContext): Promise<void> {
   const logger = ContextCreator.getLogger(invocationContext);
-  const observability = new AppInsightsObservability(logger);
+  const observability = factory.getObservability(logger);
   const trace = observability.startTrace(invocationContext.invocationId);
   try {
     invocationContext.extraOutputs.set(START, {});
--- a/backend/function-apps/azure/application-context-creator.test.ts
+++ b/backend/function-apps/azure/application-context-creator.test.ts
@@ -62,6 +62,27 @@ describe('Application Context Creator', () => {
       );
     });
 
+    test('should resolve observability via the factory singleton when none is injected', async () => {
+      vi.resetModules();
+      const FreshContextCreator = (await import('./application-context-creator')).default;
+      const freshFactory = (await import('../../lib/factory')).default;
+      const { NoOpObservability } = await import('../../lib/adapters/services/observability');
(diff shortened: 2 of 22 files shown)


backend/function-apps/dataflows/import/sync-trustee-appointments.ts
jamesobrooks, line 1: This file's test was the main reason for the work. It was timing out in staging and production ci/cd runs (where resources are constrained). In local runs, I couldn't replicate the timeout, but the bottleneck was the test importing the Application Insights package, so ensuring that we use our pre-existing no-op implementation should fix that particular slow action, not only here, but in other test files as well.

fmaddenflx (approved):

jamesobrooks:
Thanks @sourcery-ai. We evaluated all four findings via a two-turn adversarial review (each finding independently re-examined by a fresh reviewer mandated to reject it). Disposition for traceability:

**Comment 1 — `NoOpObservability.completeTrace()` parameter mismatch (bug_risk): REJECTED.**
The claim that this "breaks the interface contract and will fail type-checking" is not correct. TypeScript's parameter-count assignability allows a method with *fewer* parameters to satisfy an interface method with *more* (the same rule that lets `arr.forEach(x => ...)` ignore the index/array args). The class compiles cleanly against `ObservabilityGateway` — the `typecheck` and `Build / backend` checks on this PR are green — and the existing test fixture `mockObservability` (`backend/lib/testing/testing-utilities.ts`) already uses the identical zero-parameter `completeTrace: () => {}` form. No change needed.

**Comment 2 — split `getObservability(logger?)` into `initObservability()` + parameterless getter (complexity): REJECTED.**
All current callers (the Azure context creator + the dataflow handlers that start a trace before an `ApplicationContext` exists) pass a logger, so a parameterless getter removes a parameter every caller uses. The split also adds API surface plus a new init-before-get ordering hazard: a parameterless `getObservability()` called before init would silently bind a loggerless/NoOp gateway with no compile-time guard. It would also diverge from the single-memoized-getter pattern every other gateway in `factory.ts` follows. The cached-logger concern is already mitigated in this PR by threading a per-invocation `logger` into `completeTrace` (`activeLogger = logger ?? this.logger`).

**Overall comment — singleton caches the first `DATABASE_MOCK` value (guard or document): ALREADY ADDRESSED.**
No production code mutates `DATABASE_MOCK` at runtime (it is set once per environment), and the only place it flips is tests, which use `vi.resetModules()` + env restore. Your option (b) — make the invariant explicit — was landed in commit `526cf7862`, which rescoped the `getObservability` doc comment to factory-mediated callers.

**Overall comment — prioritize the options-object refactor in this PR: DECLINED here, tracked separately.**
There are ~5 `undefined` placeholders today, most on no-metrics failure paths where `undefined` is the honest value. Replacing the two trailing optional positional params (`metrics?`, `logger?`) with an options object touches the interface, both impls, the mock, and every call site + tests — larger blast radius than this focused, CI-green bug-fix PR warrants. Captured as a separate DevEx issue (#2608) to be done on its own branch.

Appreciate the review — it independently corroborated the options-object item we'd already backlogged.

PR #2610, opened 2026-07-01
CAMS-765 Reject inactive banks in trustee update

fmaddenflx:
# Bug

When a bank assigned to a trustee is later inactivated, the trustee profile's Software and Bank edit form shows the bank field blank (inactive banks are filtered from the dropdown), yet the Save button remains enabled and the backend accepts the stale bank id. Bank is a required field and must not be saveable while blank/inactive.

# Purpose

Enforce that only active, associated banks are accepted on the trustee interactive update path — both at the API layer and at the form layer — so an inactivated bank can never be persisted through the UI.

# Major Changes

- **Backend (`trustees.ts`):** `loadAndValidateSoftware` now builds `validBankIds` from active-only associations (`status === 'active'`), causing any inactive or unassociated bank id to be rejected with `BadRequestError`
- **Frontend (`TrusteeOtherInfoForm.tsx`):** `hasBankSelected` guard now validates bank state against the active-only `availableBanks` list rather than any non-empty string, keeping Save disabled when only a stale inactive bank id lingers in form state

Migration/import write paths are intentionally out of scope.

# Testing/Validation

Implemented using TDD. Tests written first (RED), minimal production change applied (GREEN), then refined via adversarial two-turn review (REFACTOR).

- Backend: 5 new unit tests covering inactive bank rejection, unassociated bank rejection, active bank acceptance, multiple-invalid-bank error message content, and clearing software+banks together
- Frontend: 3 new unit tests covering Save disabled with inactive bank in state, Save enabled after selecting an active bank, and regression guard for active bank already assigned
- Full test suite: 3220 passed, 3 skipped (226 test files)
- Backend coverage: 95.72% statements, 91.25% branches

# Notes

The backend change is a one-line filter addition inside `loadAndValidateSoftware` — the early-exit path (clearing both software and banks together) is unaffected. The frontend change is a single-line guard replacement — submit-time `filteredBanks` filtering is left as-is since it handles multi-bank edge cases independently.

# Definition of Done:

## Summary by Sourcery

Enforce that trustee updates only accept active, associated banks and ensure the trustee form UI reflects this constraint so inactive or stale bank selections cannot be saved.

Bug Fixes:
- Prevent inactive or unassociated banks from being persisted on trustee updates by validating bank IDs against active software-associated banks.
- Keep the trustee Software and Bank form Save button disabled when only inactive or unavailable banks are present in form state, avoiding submission with a blank or stale bank selection.

Tests:
- Add backend unit tests covering clearing software and banks together, rejecting inactive or unassociated banks, error messaging for multiple invalid bank IDs, and accepting valid active banks.
- Add frontend tests to verify Save is disabled with only inactive banks, becomes enabled after selecting an active bank, and remains enabled when an active associated bank is already assigned.

### DIFF
--- a/backend/lib/use-cases/trustees/trustees.test.ts
+++ b/backend/lib/use-cases/trustees/trustees.test.ts
@@ -1268,6 +1268,135 @@ describe('TrusteesUseCase tests', () => {
       );
     });
 
+    test('should succeed when clearing both softwareId and banks', async () => {
+      const softwareId = 'sw-axos';
+      const baseSoftwareProfile = {
+        id: softwareId,
+        documentType: 'BANKRUPTCY_SOFTWARE' as const,
+        name: 'Axos',
+        status: 'active' as const,
+        updatedOn: '2024-01-01T00:00:00.000Z',
+        updatedBy: { id: 'user-1', name: 'User One' },
+      };
+      const trusteeWithSoftwareAndBanks = MockData.getTrustee({
+        softwareId,
+        banks: ['bank-active'],
+      });
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(
+        trusteeWithSoftwareAndBanks,
+      );
+      // resolveSoftwareName is called for the old softwareId during audit history recording
+      vi.spyOn(MockMongoRepository.prototype, 'findSoftwareById').mockResolvedValue({
+        ...baseSoftwareProfile,
+        associatedBanks: [],
+      });
+      const updatedTrustee = { ...trusteeWithSoftwareAndBanks };
+      delete updatedTrustee.softwareId;
+      delete updatedTrustee.banks;
+      const updateTrusteeSpy = vi
+        .spyOn(MockMongoRepository.prototype, 'updateTrustee')
+        .mockResolvedValue(updatedTrustee);
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue();
+
+      const result = await trusteesUseCase.updateTrustee(context, trusteeId, {
+        softwareId: null,
+        banks: null,
+      });
+
+      expect(updateTrusteeSpy).toHaveBeenCalled();
+      expect(result.softwareId).toBeUndefined();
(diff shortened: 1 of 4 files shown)


jalba-flexion (approved):

mattstankey:
A few review findings on this PR:

**1. Unrelated trustee updates can break once a persisted bank goes inactive** (`backend/lib/use-cases/trustees/trustees.ts:468`)

`loadAndValidateSoftware` re-validates the *entire merged* `patchedTrustee.banks` list on every `updateTrustee` call, not just the fields actually being patched. If a trustee already has a bank that was active when assigned but is later marked inactive on the software profile, any subsequent unrelated edit (e.g. `updateTrustee(id, { name: 'New Name' })`) will still fail this check and throw `BadRequestError`, since `patchedTrustee.banks` still contains the now-inactive bank. The trustee record becomes uneditable through this path until the stale bank is manually removed. No test currently covers "update an unrelated field while an existing persisted bank has since gone inactive."

**2. `createTrustee` never validates bank/software association at all** (`backend/lib/use-cases/trustees/trustees.ts:177`)

The new active-bank check only lives in `loadAndValidateSoftware`, which is called from `updateTrustee` but not from `createTrustee`. `checkValidation` on create only checks array length bounds (`V.length(1, 100)`), so a client can create a trustee with an inactive or unassociated bank id and it will be persisted — reproducing the exact bug this PR is meant to close, just via the create path instead of update.

**4. Frontend Save can be incorrectly disabled due to an unrelated async fetch failure** (`user-interface/src/trustees/forms/TrusteeOtherInfoForm.tsx:93`)

`hasBankSelected` now depends on `availableBanks`, which is derived from the `softwareProfiles` prop. That prop is populated by an independent fetch in the parent screen (`TrusteeDetailScreen`) with no loading gate, and its catch block only does `console.log` with no fallback/error state. If that fetch is slow or fails, `softwareProfiles` stays empty, so `availableBanks` is empty and `hasBankSelected` is permanently `false` — disabling Save even for a trustee with a perfectly valid, already-active bank assigned. Worth surfacing/retrying that fetch failure rather than leaving Save silently disabled.

PR #2615, opened 2026-07-02
CAMS-793 Fix trustee case list 500-case truncation

btposey:
# Bug

Trustees with more than 500 active case appointments saw a silently truncated case list — cases beyond the first 500 (sorted by filed date descending) were dropped with no warning. Filters and pagination were applied to the already-truncated set, compounding the problem. Most trustees in staging routinely exceed 500 cases.

# Purpose

Remove the hard-coded 500-case limit so the complete appointment set is used for filtering and pagination, and take the opportunity to replace the raw MongoDB aggregate in the repository with the QueryPipeline fluent API — extending the library with the join intent, implicit unwind, and field-remapping project stage that were missing to express this query.

# Major Changes

- **Use case simplified**: `TrusteeCasesUseCase.getCasesForTrustee()` now delegates directly to a single repository method; the two-step appointment → cases fetch and all in-memory sorting/filtering are gone
- **Single MongoDB aggregate**: `TrusteeCaseAppointmentsMongoRepository.getCasesForTrustee()` executes one aggregate pipeline: `$match` appointments → `$lookup` + implicit `$unwind` into cases → `$match` case filters → `$sort` → `$facet` paginate with `$project`
- **QueryPipeline extended**: `join()` gains `joinType` (`INNER` / `LEFT_OUTER`) with `.inner()` and `.leftOuter()` intent functions; implicit `$unwind` is emitted by the renderer (preserveNullAndEmptyArrays driven by joinType); new `Project` stage with `pick()`, `omit()`, and `alias()` field-mapping helpers; a `PROJECT` stage following `PAGINATE` is folded into the `$facet` data branch
- **Integration test harness**: `test/integration/trustee-case-filter` seeds 510 appointments covering all filter dimensions and validates the full use-case layer against a real MongoDB instance
- **query-builder integration harness extended**: 75 tests now cover the new pipeline stages including join inner/leftOuter semantics and project field mapping

# Testing/Validation

- Integration test harness (`trustee-case-filter`): 31/31 passing against a live MongoDB instance, including the >500 regression guard (Tests 1, 2, 12 previously failed and now pass)
- query-builder integration harness: 75/75 passing, proving the new QueryPipeline extensions against real MongoDB
- Full unit test suite: 227/227 passing, no regressions
- Spec review findings addressed: null-crash guards fixed in `isSort`/`isPaginate`/`isPipeline`; 10 previously untested QueryPipeline exports covered; positional pipeline-index assertions in repository tests replaced with position-independent search; multi-batch `deleteAllBySource` loop path covered; `updateCaseAppointment` primary-failure path covered; renderer tests extended with all 8 missing `mapCondition` operators

# Notes

- `JoinType` and `FieldMapping` are exported from `query-pipeline.ts` for future consumer use and are flagged by knip as unused — expected until other repositories adopt the fluent join/project API
- The `collection()` method on `TrusteePartitionRepository` is now unused; it was only needed by the old raw-aggregate path and can be removed in a follow-up cleanup
- The integration test harness requires Podman + MongoDB 7.0 and follows the same local workflow as the existing `query-builder` and `case-trustee-appointments` harnesses

# Definition of Done:

## Summary by Sourcery

Remove the trustee case list’s 500-case truncation by delegating the use case to a new repository aggregate that joins appointments to cases using an extended QueryPipeline and renders through the Mongo aggregate renderer.

Bug Fixes:
- Ensure trustees with more than 500 active case appointments see the full case list with correct filtering and pagination, without silent truncation.

Enhancements:
- Refactor TrusteeCasesUseCase to delegate directly to a single repository method instead of performing in-memory joins, filtering, and pagination.
- Add a getCasesForTrustee aggregate to the Mongo trustee case appointments repository that joins appointments to cases, applies status/date/chapter/division filters, and paginates with projection.
- Extend the QueryPipeline DSL with join intent (INNER/LEFT_OUTER), implicit unwind handling, and a project stage supporting pick/omit/alias field mappings.
- Harden type guards for sort, paginate, and pipeline stages to safely handle null and non-object inputs.
- Extend Mongo aggregate rendering to support join-generated lookups with unwind and to fold project stages into the paginate facet data branch, plus additional condition/operator mappings.
- Improve Mongo trustee appointments repository robustness around update failure paths and multi-batch deleteAllBySource pagination, and reset its singleton via the public API in tests.

Build:
- Add npm scripts to run the new trustee-case-filter integration harness in local and Azure environments.

Deployment:
- Add a one-time migration to create a compound index on trustee-case-appointments (trusteeId, unassignedOn) to support the new aggregate query efficiently.

Documentation:
- Document local workflows and usage for the trustee-case-filter integration harness and its supporting Podman-based MongoDB services.

Tests:
- Replace the TrusteeCasesUseCase unit tests with delegation and error-propagation coverage aligned to the new repository API.
- Add extensive unit tests for QueryPipeline (join, project, accumulators, type guards, phonetic score helpers) and Mongo aggregate renderer (lookup+unwind, filter conditions, score metadata).
- Expand the query-builder integration harness to seed assignment fixtures and validate join and project semantics against a real MongoDB instance.
- Introduce a trustee-case-filter integration harness that seeds >500 appointments and validates trustee case filtering, status/chapters/date/division combinations, pagination, and appointedDate behavior end-to-end.


backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -107,18 +123,100 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+      // type to allow QueryBuilder to accept '_case.*' field names post-unwind.
+      type JoinedDoc = Record<string, unknown>;
+      const apptDoc = using<CaseAppointmentDocument>();
+      const caseDoc = using<JoinedDoc>();
+
+      // --- appointment filter (first $match) ---
+      const apptMatch = and(
+        apptDoc('documentType').equals('CASE_APPOINTMENT'),
+        apptDoc('trusteeId').equals(trusteeId),
+        apptDoc('unassignedOn').notExists(),
```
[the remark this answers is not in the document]
btposey: Not an issue — `unassignedOn` is typed as `undefined | string`. Active appointments have the field absent (never explicitly `null`), so `notExists()` matches the same set as the prior `equals(null)` behavior.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
```diff
@@ -749,12 +771,20 @@ const mapCondition: { [key: string]: string } = {
...
+  return stages.flatMap((stage, index) => {
     if (stage.stage === 'SORT') {
       return toMongoAggregateSort(stage);
     }
     if (stage.stage === 'PAGINATE') {
-      return toMongoPaginatedFacet(stage);
+      const next = stages[index + 1];
+      const projection = next?.stage === 'PROJECT' ? toMongoProject(next) : undefined;
+      return toMongoPaginatedFacet(stage, projection);
+    }
```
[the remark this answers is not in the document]
btposey: Accepted by design. The renderer and pipeline are a matched pair — `PROJECT` following `PAGINATE` is a convention enforced by the library, not incidental caller ordering. The renderer is the right place to implement this fold rather than requiring callers to know about `$facet` internals.

ops/migrations/CAMS-793-trustee-case-aggregate-index.ts
```diff
@@ -0,0 +1,45 @@
...
+import { MongoClient } from 'mongodb';
+
+dotenv.config({ path: 'backend/.env' });
+
+async function ensureIndexes() {
+  const connectionString = process.env.MONGO_CONNECTION_STRING;
+  const databaseName = process.env.COSMOS_DATABASE_NAME;
+  if (!connectionString || !databaseName) return;
+
+  console.log('Creating compound index on trustee-case-appointments collection...');
```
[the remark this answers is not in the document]
btposey: ✅ Addressed — the migration script has been deleted entirely. Indexes belong in Bicep, not one-off migration scripts. The compound index `{ trusteeId, unassignedOn }` has been added to `cosmos-collections.bicep` so it is managed as infrastructure.

test/integration/trustee-case-filter/scripts/run-tests.ts
```diff
@@ -0,0 +1,749 @@
...
+
+  // -------------------------------------------------------------------------
+  // Test 1: No filters — all active appointments visible (>500 regression guard)
+  // Slot 510 has unassignedOn set so it is excluded.
+  // Slot 508 has movedToCaseId so it is excluded.
+  // Expected: 508 total (510 slots minus slot 510 excluded by unassigned appt, minus slot 508
+  // excluded by movedToCaseId filter in searchCases)
+  // -------------------------------------------------------------------------
+  console.log('Test 1: No filters — full result set (>500 regression guard)');
+  {
```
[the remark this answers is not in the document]
btposey: ✅ Added — Test 1b asserts `dateFiled DESC, caseId ASC` using the chapter 11 fixtures (slots 509, 504, 501) which have distinct filed dates making the order unambiguous. 32/32 passing.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -107,18 +123,100 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     }
   }
 
-  async getActiveByTrusteeId(trusteeId: string): Promise<CaseAppointment[]> {
+  async getCasesForTrustee(
```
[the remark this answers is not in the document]
btposey: ✅ Addressed — extracted `buildAppointmentMatch(trusteeId)` and `buildCaseConditions(predicate)` as private methods. `getCasesForTrustee` is now purely orchestration.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.test.ts
```diff
@@ -373,4 +402,221 @@ describe('TrusteeCaseAppointmentsMongoRepository', () => {
...
+
+    test('error thrown by aggregate — propagates to caller', async () => {
+      vi.spyOn(CollectionHumble.prototype, 'aggregate').mockRejectedValue(
+        new Error('mongo connection failed'),
+      );
+      const context = await createMockApplicationContext();
+      const repo = TrusteeCaseAppointmentsMongoRepository.getInstance(context);
+
+      await expect(repo.getCasesForTrustee(TRUSTEE_ID, basePredicate)).rejects.toThrow(
+        'mongo connection failed',
```
cruzjone-flexion, line 616: **Test asserts the raw mongo message, but the repository wraps errors with `getCamsErrorWithStack`**

The `catch` block in `getCasesForTrustee` always re-wraps errors into a CAMS error:

```typescript
throw getCamsErrorWithStack(originalError, MODULE_NAME, {
  message: `Failed to retrieve cases for trustee ${trusteeId}.`,
});
```

This test passes today only because `getCamsErrorWithStack` happens to embed the original message in the output, so `.toThrow("mongo connection failed")` still matches. But the actual contract visible to callers is a CAMS error, not a raw `Error`. If the wrapping behavior ever changes (e.g. a different error util that omits the cause message), the test stays green while silently testing the wrong thing.

Consider asserting the CAMS contract directly:

```typescript
await expect(repo.getCasesForTrustee(TRUSTEE_ID, basePredicate)).rejects.toMatchObject({
  isCamsError: true,
  message: expect.stringContaining(`Failed to retrieve cases for trustee ${TRUSTEE_ID}`),
});
```

btposey: Good instinct, but the test is actually correct given the error propagation mechanics here.

The call chain when `aggregate` rejects is:

1. `MongoCollectionAdapter.paginate()` catches the raw `Error("mongo connection failed")` — since `isCamsError` is false, its `handleError` wraps it into a new `UnknownError` with `message: "Query failed. mongo connection failed"`. That `UnknownError` **is** a `CamsError`.

2. That `CamsError` propagates to `getCasesForTrustee`'s catch block, which calls `getCamsErrorWithStack(originalError, ...)`. Since `isCamsError` is now **true**, `getCamsErrorWithStack` calls `addCamsStack` — which only appends to the stack and **returns the error unchanged**. The repository's own message (`"Failed to retrieve cases for trustee ..."`) is never applied.

So the actual thrown error has message `"Query failed. mongo connection failed"` — Cruz's suggested assertion `.toMatchObject({ isCamsError: true, message: stringContaining("Failed to retrieve cases for trustee") })` would **fail** in practice.

The current assertion `.toThrow("mongo connection failed")` is a correct substring match against what actually gets thrown. The repository catch clause's message string is effectively dead code for this code path — a follow-up could either remove it or restructure so the repository-level message is always applied.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -107,22 +128,97 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+
+    if (predicate.caseStatus === 'OPEN') {
+      conditions.push(
+        or(
+          doc('_case.closedDate').notExists(),
+          and(
+            doc('_case.closedDate').exists(),
+            doc('_case.reopenedDate').exists(),
+            doc('_case.reopenedDate').greaterThanOrEqual({ name: '_case.closedDate' }),
+          ),
```
mattstankey: **Duplicated domain rule.** `buildCaseConditions()` reimplements the OPEN/CLOSED case-status logic below as raw Mongo conditions. This is the same rule already expressed as a pure domain function (`isCaseClosed` in `common/src/cams/cases.ts`) and independently re-encoded again in `cases.mongo.repository.ts`'s `addConditions`. There's no shared helper tying these together, so a future change to closed/reopened semantics could be applied in one place and silently miss the other. Recommend extracting a shared `buildCaseStatusCondition(doc, status)` helper used by both Mongo repositories.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -37,6 +55,9 @@ class TrusteePartitionRepository extends BaseMongoRepository {
   adapter<T>() {
     return this.getAdapter<T>();
   }
+  collection() {
```
mattstankey: **Dead code.** `collection()` was added when this method built the aggregate array by hand and called `.aggregate(pipeline)` directly; the final refactor moved to the QueryPipeline/adapter path and this raw-driver accessor now has zero call sites in the file. It also reintroduces the kind of direct-driver escape hatch the humble-object/adapter pattern exists to prevent. Recommend deleting it now rather than leaving it as a "candidate for follow-up cleanup" — it's a two-line deletion with no cost to removing it today.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -1,21 +1,39 @@
...
+const CASES_COLLECTION = 'cases';
+
+const { using, and, or } = QueryBuilder;
+const { pipeline, match, join, sort, paginate, project, pick, omit, alias, descending, ascending } =
+  QueryPipeline;
+
+// Cast a dotted field name to FieldReference<never> — source<T>() generic inference
+// produces incompatible types when T has a string index signature or is never.
+const fr = (name: string, src?: string): FieldReference<never> =>
+  ({ name, source: src }) as unknown as FieldReference<never>;
```
mattstankey: **Type-safety gap in the new join/project API's flagship use case.** `fr()` casts a dotted field name to `FieldReference<never>` via `as unknown as`, and `buildCaseConditions()` similarly widens to `Record<string, unknown>` — both route around `QueryBuilder`'s normal compile-time field-name checking specifically to reference post-join dotted paths (e.g. `'_case.dateFiled'`), which is exactly the scenario the join/project extension was built to support. A typo like `'_case.datefiled'` would only surface at runtime against a live MongoDB collection, not at build time. Consider extending `FieldReference<T>`/`source<T>()` with a typed dotted-path accessor, or at minimum file a tracked follow-up so this workaround doesn't get copied as the standard pattern by the next repository.

backend/lib/query/query-pipeline.ts
```diff
@@ -188,23 +203,54 @@ function descending(field: FieldReference<never>): SortedField {
...
+  outer(): Join;
+};
+
 function join<Foreign = never>(foreign: FieldReference<Foreign>) {
   return {
     onto: <Local = never>(local: FieldReference<Local>) => {
       return {
-        as: <T = never>(alias: FieldReference<T>): Join => {
-          return {
+        as: <T = never>(alias: FieldReference<T>): JoinBuilder => {
```
mattstankey: **`join().as()` silently defaults to INNER, defeating the explicit-intent API.** The object returned by `as()` is already a structurally complete `Join` with `joinType: 'INNER'` baked in before `.inner()`/`.outer()` is ever called. Since `Stage[]` accepts anything structurally matching `Join`, a caller can write `join(f).onto(l).as(a)` — omitting the intent call entirely — and it compiles fine, silently producing an inner join. This defeats the stated purpose of the `.inner()/.outer()` rename (renaming `leftOuter()` → `outer()` to make intent explicit at the call site). Recommend making `as()` return a type that is *not* independently a valid `Join` (e.g. only `{ inner(): Join; outer(): Join }`, no spread of `base` at the top level) so `.inner()`/`.outer()` is structurally required.

backend/lib/query/query-pipeline.ts
```diff
@@ -108,11 +108,25 @@ export type FieldReference<T> = Field<T> & {
 
 type QueryFieldReference<T> = FieldReference<T> & ConditionFunctions<T>;
 
+export type JoinType = 'INNER' | 'OUTER';
```
mattstankey, line 125: **Unused exports kept alive by suppressing the linter instead of deleting them.** `JoinType`/`FieldMapping` have zero callers outside `query-pipeline.ts` itself (confirmed via grep — no other repository uses `join()`/`project()`/`pick()`/`omit()`/`alias()`, and `OUTER` has no caller anywhere except its own unit test). Rather than removing them, `query-pipeline.ts` was added as a knip entry point to silence the unused-export warning. Per this codebase's YAGNI convention, code that isn't used now should be removed now — "for future consumers" isn't a second use case, it's a prediction. Recommend deleting the unused exports (or the knip entry-point exemption) until a second repository actually needs fluent joins/projects.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
```diff
@@ -749,12 +771,20 @@ const mapCondition: { [key: string]: string } = {
 };
 
 function toMongoAggregate(pipeline: Pipeline): AggregateQuery {
-  return pipeline.stages.flatMap((stage) => {
+  const { stages } = pipeline;
+  return stages.flatMap((stage, index) => {
     if (stage.stage === 'SORT') {
       return toMongoAggregateSort(stage);
     }
     if (stage.stage === 'PAGINATE') {
```
mattstankey, line 779: **Fragile positional coupling between PAGINATE and PROJECT stages.** The "PROJECT-immediately-after-PAGINATE gets folded into `$facet.data`" behavior is encoded purely as array-index lookahead/lookbehind (lines 781, 786) — nothing in the `Pipeline`/`Stage` types expresses that these two stages are related. It works for the current call site, but nothing stops a future pipeline from inserting a stage between `paginate()` and `project()`; that would silently produce a top-level `$project` applied to the `{metadata, data}` facet envelope instead of each document, most likely yielding an empty/garbage `data` array with no clear error signal. Consider nesting the projection directly inside `Paginate` (e.g. `paginate(skip, limit, projectionStage?)`) so the relationship is structural rather than positional, or at minimum cross-reference this coupling in a comment at the `Paginate`/`Project` type definitions in `query-pipeline.ts`.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -107,22 +128,97 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
-        doc('trusteeId').equals(trusteeId),
-        doc('unassignedOn').equals(null),
-      );
-      return await this.trusteePartition.adapter<CaseAppointmentDocument>().find(query);
+      return await this.trusteePartition
+        .adapter<TrusteeCaseListItem>()
+        .paginate(
+          pipeline(
+            match(this.buildAppointmentMatch(trusteeId)),
+            join(fr('caseId', CASES_COLLECTION)).onto(fr('caseId')).as(fr('_case')).inner(),
```
mattstankey: **Possible missing index for post-join filter/sort fields, and no `allowDiskUse`.** The new compound index `{trusteeId, unassignedOn}` covers the first `$match` on appointments, but the second `$match` (on `_case.documentType`, `movedToCaseId`, `closedDate`/`reopenedDate`, chapters, filedDate range, divisionCodes) and the subsequent `$sort` on `_case.dateFiled` run against the unwound joined set with no supporting index on the `cases` side, and `allowDiskUse` isn't set anywhere in the codebase. For a trustee with several thousand appointments — the scale this PR's own description says is routine in staging — this risks exceeding MongoDB's default 100MB in-memory sort limit, causing a hard aggregation error (arguably worse than the silent truncation being fixed). Worth confirming via `explain()` at realistic scale, or adding a covering index on the `cases` side for the common filter/sort combination.

Relatedly: no logging/metrics around this aggregate call, unlike sibling repositories (e.g. `cases.mongo.repository.ts` logs at debug level around mutations) — a slow/failing aggregate here wouldn't be correlatable to a specific trustee's appointment count in production.

mattstankey (commented):
Automated review pass (pragmatic/YAGNI + architecture + security + SRE lenses). Core bug fix (removing the 500-case truncation) is sound — findings below are about the surrounding refactor. One additional note that isn't inline-anchorable because the file isn't touched by this PR's diff: **`trustee-cases.controller.ts:49-51`** validates `limit` only for being a finite positive number, with no upper ceiling before it reaches the aggregate's `$skip`/`$limit`. The 500-case hard cap this PR removes isn't replaced by any bounded max page size — worth adding a clamp (e.g. capping at some reasonable max) while this area is already being touched.

test/integration/trustee-case-filter/scripts/run-tests.ts
```diff
@@ -0,0 +1,771 @@
+/**
```
mattstankey, line 1: **Integration harness under-tests the scale that matters.** This seeds 510 appointments as the "large trustee" regression case — just past the old 500 cap — which validates correctness of the fix but doesn't exercise the performance characteristics (thousands of appointments) that the PR description itself says are routine in staging ("most trustees in staging routinely exceed 500 cases"). The suite passing gives confidence on correctness but not on whether the aggregate performs acceptably at real scale (see the index-coverage comment on the repository file). Consider adding a scale test with a few thousand seeded appointments, even if just to characterize latency rather than assert a hard threshold.

backend/lib/query/query-pipeline.ts
btposey: The common case is the implied `inner()` so it is implicitly default making `inner()` optional but leaves the API complete from an "expressive" point-of-view.

backend/lib/query/query-pipeline.ts
btposey: This is because the QueryBuilder and QueryPipeline are generic libraries that do not have all API features in use. This is an acceptable exception to the dead-code findings Knip provides.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
btposey: The renderer can expect a structured position from the QueryPipeline library that we assume we can depend on. If this is indeed too fragile or obvious edge cases can be identified then we can harden this code.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: We should add very specific logging if this exception class is encountered so we know how we can fix it if encountered.

test/integration/trustee-case-filter/scripts/run-tests.ts
btposey: The 500 record cap was arbitrary. I'm more interested in the filter functionality than I am the volume.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -107,22 +128,97 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+            match(and(...this.buildCaseConditions(predicate))),
+            sort(descending(fr('_case.dateFiled')), ascending(fr('_case.caseId'))),
+            paginate(predicate.offset, predicate.limit),
```
fmaddenflx, line 156: I'd like to understand how this works better.  We have a paginate inside of a pipeline, inside of a paginate.  How does that work exactly?

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Fixed. Extracted `buildCaseStatusCondition<T>(caseStatus, fieldPrefix?)` into `backend/lib/adapters/gateways/mongo/utils/case-status-conditions.ts`. Both `cases.mongo.repository` (`excludeClosedCases`/`includeOnlyClosedCases`) and `trustee-case-appointments.mongo.repository` (`buildCaseConditions`) now delegate to this single function. The optional `fieldPrefix` handles the post-join `_case.*` namespace. The rule now lives in one place.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Fixed. `collection()` deleted from `TrusteePartitionRepository`.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Fixed. Extended `source<T>(collectionName?, prefix?)` in `query-pipeline.ts` to accept an optional dotted-path prefix. When set, `source<SyncedCase>(CASES_COLLECTION, '_case').field('chapter')` produces a `QueryFieldReference<SyncedCase>` whose condition closures carry `leftOperand.name = '_case.chapter'` — fully typed against `SyncedCase` fields with the prefixed path rendered to Mongo. `fr()` and the `Record<string, unknown>` widening in `buildCaseConditions` are both gone. `buildAppointmentMatch` similarly uses `source<CaseAppointmentDocument>(TRUSTEE_COLLECTION)` for the appointment side. All 3149 backend tests pass.

backend/lib/query/query-pipeline.ts
fmaddenflx: Is it used in a test file?  If it's not used in a test, remove "export".

fmaddenflx: Just saw Brians comment.  I'm okay with that.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
mattstankey: Confirmed fixed — `buildCaseStatusCondition()` in `case-status-conditions.ts` is now shared between `cases.mongo.repository.ts` and `trustee-case-appointments.mongo.repository.ts`. Resolving.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
mattstankey: Confirmed — `collection()` removed from `TrusteePartitionRepository`. Resolving.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
mattstankey: Confirmed — `source(collectionName, prefix)` now supports typed dotted-path field references, and the unsafe `fr()` cast is gone. Resolving.

backend/lib/query/query-pipeline.ts
mattstankey: Understood — treating implicit INNER as the sensible default with `.inner()/.outer()` available for explicit opt-in is a reasonable tradeoff. Not blocking. Resolving.

backend/lib/query/query-pipeline.ts
mattstankey: Team consensus reached with fmaddenflx's follow-up — acceptable exception for a generic library API. Resolving.

backend/lib/adapters/gateways/mongo/utils/mongo-aggregate-renderer.ts
mattstankey: Understood — treating this as an intentional renderer/library contract rather than incidental caller ordering, consistent with how the equivalent Sourcery finding on this same coupling was resolved. Not blocking. Resolving.

test/integration/trustee-case-filter/scripts/run-tests.ts
mattstankey: Understood — filter correctness was the point of this fix, not volume/perf characterization. Not blocking this PR. Resolving.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
mattstankey: Author acknowledged intent to add targeted logging if this exception path is hit, but no index or logging change shipped in the latest commit. Leaving open — worth confirming before merge whether the index/allowDiskUse risk is accepted as-is or tracked as a follow-up.

btposey: ✅ Added — the catch block now inspects the error before rethrowing: if the error code is 292 or the message contains `$sort exceeded memory limit`, a `console.error` is emitted with the MODULE_NAME and trusteeId so operators can correlate the failure to a specific trustee's appointment count in production logs.

mattstankey (approved):

PR #2620, opened 2026-07-02
Add observability and class refactor to trustee migration

btposey:
# Purpose

Bring the ATS trustee migration dataflow up to current standards by adding Application Insights telemetry, converting the use case to class-based, and adding the standard `flushQueues` flag.

# Major Changes

- Converted `migrate-trustees` use case to a `MigrateTrusteesUseCase` class with constructor-bound `ApplicationContext`, matching the class-based pattern used by other migration dataflows. All existing named exports are preserved so tests require no changes.
- Extended `MigrationStartMessage` with `flushQueues?: boolean`. When set, all three queues (start, page, dlq) are drained to blob storage and the handler returns — useful for emergency queue inspection without processing.
- Instrumented all three handlers (`handleStart`, `handlePage`, `handleError`) with `context.observability.startTrace` / `completeDataflowTrace` on every exit path, including DLQ routing, early-return, and catch blocks.

# Testing/Validation

All 103 existing unit tests pass unchanged. The class-based wrapper delegates to the same underlying exported functions, so no test modifications were required. TypeScript type-check passes clean across all workspaces.

# Notes

The `MigrateTrusteesUseCase` class is a thin wrapper — it stores `context` in the constructor and delegates each method to the pre-existing exported functions. This keeps the test surface intact while giving the handler a clean class-based call site.

DLQ routing exit paths use `success: true` in the trace result (consistent with the `migrate-case-appointments` pattern) — routing an error to the DLQ is the handler completing its job correctly, not a handler-level failure. Only unexpected throws use `success: false`.

The `flushQueues` implementation uses `AzureWebJobsStorage` (not `AzureWebJobsDataflowsStorage`) consistent with how the existing queue output bindings in this dataflow are configured.

# Definition of Done:

## Summary by Sourcery

Refactor the trustee migration dataflow to use a class-based use case with bound application context, add observability traces to all handlers, and introduce an option to flush migration queues to blob storage for inspection.

New Features:
- Add a flushQueues flag to the trustee migration start message to allow draining start, page, and DLQ queues to blob storage without processing messages.

Enhancements:
- Wrap migrate-trustees use case functions in a MigrateTrusteesUseCase class so handlers can operate with a constructor-bound ApplicationContext while preserving existing function exports.
- Instrument the trustee migration start, page, and error handlers with Application Insights dataflow tracing on all exit paths, including DLQ routing and unexpected errors.
- Add a helper to dump Azure Storage Queue messages to blob storage as JSONL for migration queue inspection.

### DIFF
--- a/backend/function-apps/dataflows/migrations/migrate-trustees.ts
+++ b/backend/function-apps/dataflows/migrations/migrate-trustees.ts
@@ -1,4 +1,5 @@
 import { app, InvocationContext, output } from '@azure/functions';
+import { QueueServiceClient } from '@azure/storage-queue';
 
 import ApplicationContextCreator from '../../azure/application-context-creator';
 import { ApplicationContext } from '../../../lib/adapters/types/basic';
@@ -11,13 +12,15 @@ import {
   ensureContainersExist,
   StartMessage,
 } from '../dataflows-common';
-import * as MigrateTrusteesUseCase from '../../../lib/use-cases/dataflows/migrate-trustees';
+import MigrateTrusteesUseCase from '../../../lib/use-cases/dataflows/migrate-trustees';
 import * as MigrationStateService from '../../../lib/use-cases/dataflows/trustee-migration-state.service';
 import { buildQueueError, QueueError } from '../../../lib/use-cases/dataflows/queue-types';
 import { CamsError } from '../../../lib/common-errors/cams-error';
+import { completeDataflowTrace } from '../../../lib/use-cases/dataflows/dataflow-telemetry';
 import { STORAGE_QUEUE_CONNECTION } from '../../../lib/storage-queues';
 import ModuleNames from '../module-names';
 import { TrusteeMigrationStartEvent } from '@common/cams/dataflow-events';
+import factory from '../../../lib/factory';
 
 const MODULE_NAME = ModuleNames.MIGRATE_TRUSTEES;
 const PAGE_SIZE = 50; // Smaller page size for trustees with appointments
@@ -49,7 +52,45 @@ const HANDLE_PAGE = buildFunctionName(MODULE_NAME, 'handlePage');
 const HANDLE_ERROR = buildFunctionName(MODULE_NAME, 'handleError');
 const HTTP_TRIGGER = buildFunctionName(MODULE_NAME, 'httpTrigger');
 
-type MigrationStartMessage = StartMessage & TrusteeMigrationStartEvent;
+type MigrationStartMessage = StartMessage &
+  TrusteeMigrationStartEvent & {
+    flushQueues?: boolean;
+  };
+
+// Drains all messages from a named queue and writes them as a JSONL blob.
+async function dumpQueueToBlob(
+  objectStorage: ReturnType<typeof factory.getObjectStorageGateway>,
+  logger: { info: (module: string, msg: string) => void },
+  queueName: string,
+  blobName: string,
+  outputContainerName: string,
+): Promise<number> {
+  const connectionString = process.env.AzureWebJobsStorage;
+  if (!connectionString) {
+    logger.info(MODULE_NAME, `flushQueues: AzureWebJobsStorage not set — skipping ${queueName}`);
+    return 0;
+  }
+  const queueClient =
+    QueueServiceClient.fromConnectionString(connectionString).getQueueClient(queueName);
+  const lines: string[] = [];
+  let response = await queueClient.receiveMessages({ numberOfMessages: 32 });
+  while (response.receivedMessageItems.length > 0) {
+    for (const msg of response.receivedMessageItems) {
+      lines.push(Buffer.from(msg.messageText, 'base64').toString('utf-8'));
+    }
+    response = await queueClient.receiveMessages({ numberOfMessages: 32 });
+  }
+  if (lines.length > 0) {
+    await objectStorage.writeObject(outputContainerName, blobName, lines.join('\n'));
+    logger.info(
+      MODULE_NAME,
+      `flushQueues: wrote ${lines.length} messages to ${outputContainerName}/${blobName}`,
+    );
+  } else {
+    logger.info(MODULE_NAME, `flushQueues: ${queueName} is empty — no blob written`);
+  }
+  return lines.length;
+}
 
 /**
  * performDeleteAllIfRequested
@@ -61,6 +102,7 @@ type MigrationStartMessage = StartMessage & TrusteeMigrationStartEvent;
  */
 async function performDeleteAllIfRequested(
   start: MigrationStartMessage,
+  useCase: MigrateTrusteesUseCase,
   context: ApplicationContext,
   invocationContext: InvocationContext,
 ): Promise<boolean> {
@@ -75,7 +117,7 @@ async function performDeleteAllIfRequested(
     'deleteAll flag detected. Deleting all existing trustees and appointments.',
   );
 
-  const deleteResult = await MigrateTrusteesUseCase.deleteAllTrusteesAndAppointments(context);
+  const deleteResult = await useCase.deleteAllTrusteesAndAppointments();
   if (deleteResult.error) {
     invocationContext.extraOutputs.set(
       DLQ,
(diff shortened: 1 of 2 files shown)


backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -49,7 +52,45 @@ const HANDLE_PAGE = buildFunctionName(MODULE_NAME, 'handlePage');
...
+  }
+  const queueClient =
+    QueueServiceClient.fromConnectionString(connectionString).getQueueClient(queueName);
+  const lines: string[] = [];
+  let response = await queueClient.receiveMessages({ numberOfMessages: 32 });
+  while (response.receivedMessageItems.length > 0) {
+    for (const msg of response.receivedMessageItems) {
+      lines.push(Buffer.from(msg.messageText, 'base64').toString('utf-8'));
+    }
+    response = await queueClient.receiveMessages({ numberOfMessages: 32 });
```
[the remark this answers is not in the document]
btposey: ✅ Fixed. Added `deleteMessage` after each received message so queues are truly drained, not just peeked. The visibility timeout no longer causes messages to reappear.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -49,7 +52,45 @@ const HANDLE_PAGE = buildFunctionName(MODULE_NAME, 'handlePage');
...
+  }
+  if (lines.length > 0) {
+    await objectStorage.writeObject(outputContainerName, blobName, lines.join('\n'));
```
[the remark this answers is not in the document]
btposey: ✅ Fixed. Replaced single in-memory buffer with a batched write loop (`FLUSH_BATCH_SIZE = 1000`). Each full batch is written to a separate numbered blob (`blobName`, `blobName.1`, etc.) before clearing the buffer, keeping memory bounded regardless of queue depth.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -1192,3 +1192,43 @@ export async function deleteAllTrusteesAndAppointments(context: ApplicationConte
...
+ *
+ * Wraps the stateless exported functions as instance methods, with the
+ * ApplicationContext bound at construction time.  The handler imports this
+ * class as the default export so it can call instance methods without
+ * threading `context` through every call site.
+ *
+ * All underlying named exports remain intact so the test suite can continue
+ * to import and call them directly.
+ */
+class MigrateTrusteesUseCase {
```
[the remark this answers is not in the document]
btposey: Intentionally leaving the named exports in place. The `__testOnly` pattern is not one we want to adopt in CAMS — the tests import these functions directly by name and that is the established convention here. This is also a one-time migration, so the dual-surface tradeoff is acceptable.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -108,57 +150,132 @@ async function performDeleteAllIfRequested(
  * Initialize the trustee migration by reading existing state for resumability.
  * If already completed, skip. Otherwise, queue first/next CursorMessage with lastTrusteeId from state.
  * If deleteAll flag is present, delete all existing trustees and appointments before starting.
+ * If flushQueues flag is present, dump all queues to blob storage and return.
  */
 async function handleStart(start: MigrationStartMessage, invocationContext: InvocationContext) {
```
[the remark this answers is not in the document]
btposey: Good suggestion. Deferring for now — this handler is deliberately following an established pattern from another engineer on the team and it is not the right moment to introduce a divergence. Tracked as tech debt in beads issue `cams-km1e` (TECH-DEBT) for a future pass.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -108,57 +171,132 @@ async function performDeleteAllIfRequested(
...
+      return;
+    }
+
+    const deleteOk = await performDeleteAllIfRequested(start, useCase, context, invocationContext);
+    if (!deleteOk) {
+      // DLQ already populated inside performDeleteAllIfRequested
+      completeDataflowTrace(context.observability, trace, MODULE_NAME, 'handleStart', logger, {
+        documentsWritten: 0,
+        documentsFailed: 0,
+        success: true,
```
jamesobrooks: All DLQ-routing traces in `handleStart` and `handlePage` use `success: true`, including paths where a genuine failure caused the DLQ routing (deleteAll failed, state read failed, page fetch failed, etc.). This means all three failure-detection panels in the `dataflow-troubleshooting` workbook will miss these events — the panels filter on `success == "false"` or `documentsFailed > 0`, and neither condition is satisfied here.

`migrate-case-appointments` uses `success: false` for its fatal DLQ paths (e.g. deleteAll failure). Suggest aligning: `success: false` for DLQ paths that represent actual failures; `success: true` for intentional no-ops (already-completed, flushQueues mode).

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -108,57 +171,132 @@ async function performDeleteAllIfRequested(
...
+
+  try {
+    const useCase = new MigrateTrusteesUseCase(context);
+
+    if (start.flushQueues) {
+      logger.info(MODULE_NAME, 'flushQueues flag detected — dumping queues to blob storage.');
+      const objectStorage = factory.getObjectStorageGateway(context);
+      const outputContainerName = buildContainerName(MODULE_NAME, 'out');
+      const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
+      await dumpQueueToBlob(
```
jamesobrooks: The return values from all three `dumpQueueToBlob` calls are discarded, so `documentsWritten: 0` is always hardcoded in the trace regardless of how many messages were flushed. The counts do appear in correlated APPI trace logs (the `logger.info` inside `dumpQueueToBlob`), so they are not lost — but they are not in the custom event properties either. Nit: capture the return values and add them to `details`:

```typescript
const startCount = await dumpQueueToBlob(...);
const pageCount  = await dumpQueueToBlob(...);
const dlqCount   = await dumpQueueToBlob(...);
// then in completeDataflowTrace details:
details: { mode: 'flushQueues', startFlushed: String(startCount), pageFlushed: String(pageCount), dlqFlushed: String(dlqCount) }
```

backend/lib/use-cases/dataflows/migrate-trustees.ts
jamesobrooks, line 1207: Optional: this class pre-binds `context` at construction time to save passing it at three downstream call sites. The tradeoff is a dual public API — the same three operations are accessible both via the class (default export) and the underlying named exports, with the JSDoc comment here explaining why both exist. `performDeleteAllIfRequested` still takes `context` explicitly alongside `useCase`, so the context threading is not fully eliminated.

One alternative: drop the class, import the three named functions directly in the handler, and thread `context` at those three call sites. All three functions are already exported and `context` is in scope at each call site, so the change would be mechanical. But this is strictly a style preference given that constructor-bound use-case classes are a common pattern in this codebase — no action required.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -339,6 +533,12 @@ async function handleError(event: QueueError, invocationContext: InvocationConte
 
   // Error already in DLQ (this handler processes DLQ messages)
   // Just log for visibility - no further action needed
+  completeDataflowTrace(context.observability, trace, MODULE_NAME, 'handleError', logger, {
+    documentsWritten: 0,
+    documentsFailed: 0,
+    success: true,
+    details: { activityName: queueError.activityName },
```
jamesobrooks: Question: is omitting `queueError.error?.message` from the custom event properties intentional? The error message is logged via `logger.error` above and correlates to this invocation in APPI, so it is discoverable — just curious whether the team's KQL queries or alerts rely on the correlated log or on custom event properties for diagnosing DLQ items.

jamesobrooks (commented):
The observability instrumentation is thorough and consistent — every exit path across all three handlers has a `completeDataflowTrace` call, including unexpected-throw re-raises. The class refactor is clean and the `flushQueues` diagnostic feature is a useful operational addition. One issue to address before the next migration run is the `success` flag semantics on DLQ routing paths; the remaining items are optional suggestions.

---

**`dataflows-common.ts` line 136 (`buildStartQueueHttpTrigger`) — nit:** This function always enqueues `{}` and ignores the HTTP request body entirely (the inner closure does not bind the `request` parameter). Any flags passed in a POST body (`flushQueues`, `deleteAll`, `reset`, `importAll`) are silently discarded and the caller gets a 201 OK with a default migration run started. Direct queue injection is the correct path for operational flags — just worth noting here for future readers who might expect the body to be forwarded.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
btposey: ✅ Fixed. All genuine failure DLQ paths in `handleStart` and `handlePage` now use `success: false` with `error: <message>` — deleteAll failure, state read failure, page fetch failure, process failure, and state update failure. Intentional no-ops (already-completed, flushQueues, resume) remain `success: true`. This aligns with `migrate-case-appointments` and ensures the workbook failure panels catch these events.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
btposey: ✅ Fixed. Captured all three return values (`startFlushed`, `pageFlushed`, `dlqFlushed`) and added them to the `details` object. Also updated `documentsWritten` to reflect the total count flushed rather than hardcoding `0`.

backend/lib/use-cases/dataflows/migrate-trustees.ts
btposey: ✅ Addressed. Added a public `context` getter on `MigrateTrusteesUseCase` so the handler can access the bound context without re-threading it as a separate parameter. Removed the redundant `context` parameter from `performDeleteAllIfRequested` and its call site — the function now reads `useCase.context` directly. Also removed the now-unused `ApplicationContext` import from the handler file.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
btposey: ✅ Fixed. The omission was not intentional. Added `errorMessage: queueError.error.message` to the custom event `details` when present, so the error is available in both the correlated `logger.error` log and the App Insights custom event properties — no KQL join required.

jamesobrooks (approved):

PR #43, opened 2023-05-19
CAMS-19: Add SCA scan to pipeline

governmentSponsored:
# Dependency Checking 

As specified in the contract, we implement the use of static dependency checking of the repositories third party open-source libraries.

## Approach

To be able to run dependency security scans against application third party open-source libraries, we have integrated Veracode's SCA scanning tool before the application is built. It does two scans -- one for frontend and one for backend libraries -- and if the scans pass then the build process may commence.

## Local
**API Credentials Prerequisite:**
1. Login into https://web.analysiscenter.veracode.us/login
2. Navigate to Platform Home view
3. Go to API Credentials thru the menu options on the left side
4. Click Generate API Credentials
5. Copy the IF and Secret before navigating away
6. Setup API Credentials file locally. For OS X, 'vi ~/.veracode/credentials'

```
[default]
veracode_api_key_id = <<ID>>
veracode_api_key_secret = <<SECRETS>
```

**Running in VS Code**
1. Install extension for Veracode SCA
2. Open project folder at the root of project ~/repo/boss/gui or ~/repo/boss/function/node *as of 2023-05-18*
3. Start scan
    1. You should see findings thru the SCA extension
    2. Results can be viewed in dashboard
   	 1. Navigate to the following Software Composition Analysis > Agent-Based Scan > My Workspace

**Running via CLI**
1. Log into dashboard
2. Navigate to the following Software Composition Analysis > Agent-Based Scan > My Workspace > Manage Workspace > Agents
3. Create new Action by clicking on green Action button on right side of screen
    1. Choose appropriate os to setup the scanner
    2. Follow provided instructions
4. Start scan

```
srcclr scan ~/dev/repo/ustp/Bankruptcy-Oversight-Support-Systems/user-interface
srcclr scan ~/dev/repo/ustp/Bankruptcy-Oversight-Support-Systems/backend/functions
```
5. View result thru console output and/or thru the dashboard
## GitHub Actions

We've added two new `security-sca-scan-frontend` and `security-sca-scan-backend` jobs which scan both the React app and Node API. If errors exist, we fail the job, otherwise we pass.

## Further Improvement Opportunities

PR #19, opened 2023-04-20
111: Rename BOSS -> CAMS

jamesobrooks:

PR #50, opened 2023-05-26
CAMS-21 Veracode dast scanning

magile32:
**Description of changes**
## Goal 
Implement DAST scanning and pre-scanning for CICD workflow. 

## Additions
- Bicep updates to allow Veracode agent access to scan live web application and api
- Add job step for smoke testing deployed application
- The following are some supporting changes
  - Updates to Bicep and Deployment scripts to enable deployment of application from branch with new set of Azure resources
- Created new GHA workflow file to trigger Veracode DAST prescan and schedule a scan. Added python helper scripts.

## Approach
To implement DAST scanning and prescanning in the pipeline, the addition of a python script was necessary. Python supports the generation of HMAC tokens for authentication. The DAST prescan and scan were separated out into another GitHub Actions workflow which can be manually triggered at this time and the set frequency will be determined. 

New DAST Workflow: https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/workflows/veracode-dast-scan.yml

When the new GitHub Action is triggered, the workflow kicks off by setting up the agent with necessary tools and triggering a REST request to Veracode for a Dynamic Prescan of the application under the CAMS Flexion scan. Once this scan is triggered, the workflow waits for the prescan to finish (2-10 minutes). Upon finishing successful completion of the prescan, the workflow then schedules an immediate DAST scan of the web application through a separate rest call and finishes.

## Local
To run Tests locally

1. Install Veracode API signing python Library
        pip install veracode-api-signing
2. Within your terminal, cd into /ops/helper-scripts
3. run the python script
        python veracode-dast-scan.py

This script should return a 204 if it is successful

PR #2621, opened 2026-07-03
Add CAMS brand guidelines and UI issues documentation

katiecissell:
## Summary

- Adds `docs/BRAND_GUIDELINES.md` — a comprehensive design and content reference for the CAMS application, intended to guide both human developers and LLM assistants toward consistent UI implementation
- Adds `docs/BRAND_GUIDELINES_ISSUES.md` — 43 issues discovered during the brand guidelines audit, covering accessibility bugs, dead code, and component inconsistencies

## What's in the brand guidelines

The guidelines were built by auditing the actual codebase, so everything documented reflects what CAMS truly does rather than aspirational standards. Sections include:

- **Color system** — every color audited against actual usage, unused colors called out
- **Typography** — text casing rules, heading hierarchy with actual font sizes from theme.scss
- **Spacing** — two distinct systems (rem-based SCSS vs USWDS utility classes) clearly separated
- **Component patterns** — buttons, forms, tables, modals, loading states, feedback/messaging
- **Core UX principles** — escapability, recognition over recall, visibility of system state, etc.
- **Accessibility** — WCAG 2.1 AA requirements, ARIA patterns audited per component, landmark rules, screen reader patterns
- **Terminology** — preferred CAMS-specific language (Case Number not Case ID, Staff Assignment not Case Assignment, etc.)
- **Data formatting** — date, time, currency, number standards
- **Icons** — USWDS sprite icons and custom icons documented

## What's in the issues document

43 issues discovered during the audit, numbered and prioritized:
- **Critical accessibility bugs** (issues #19–#22): ComboBox errors/hints invisible to screen readers, Radio label association broken
- **Significant accessibility gaps** (issues #23–#26, #31–#32, #36–#38): aria-live on `<main>`, error messages not in aria-describedby, required field indicators not accessible
- **Dead code** (issues #2, #4, #5, #17): unused CSS variables, mixins, and class definitions
- **Component inconsistencies** (issues #8–#16): IconButton underuse, TextArea bypassed, gray color proliferation
- **Lower priority cleanups** (issues #27–#43)

## Test plan

- [ ] Read through `BRAND_GUIDELINES.md` and verify the guidelines accurately reflect your understanding of CAMS
- [ ] Review `BRAND_GUIDELINES_ISSUES.md` and confirm issue descriptions and priorities are accurate
- [ ] Flag any guidelines that conflict with existing team decisions or should be changed
- [ ] Flag any issues that are already fixed or not worth addressing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Add comprehensive CAMS brand and UI implementation guidelines and track a prioritized list of UI, accessibility, and design consistency issues found during the audit.

Documentation:
- Document CAMS brand, UX, accessibility, and UI implementation standards in a new BRAND_GUIDELINES.md reference.
- Record and prioritize 43 audit findings covering accessibility bugs, dead code, and component inconsistencies in BRAND_GUIDELINES_ISSUES.md.

### DIFF
--- a/docs/BRAND_GUIDELINES.md
+++ b/docs/BRAND_GUIDELINES.md
@@ -35,7 +35,7 @@ Users should not have to remember or type names, codes, or IDs from memory. When
 - Show "New York Southern District" not "USTP_CAMS_Region_02_Office_New_York"
 
 ### 3. Visibility of System State
-After any action (assignment created, role removed, order verified), the user must see immediate confirmation that the action succeeded. Failed actions must show a clear error explaining what went wrong and what to do next. Buttons must not do nothing silently.
+After any action (assignment created, role removed, order verified), the user must see immediate confirmation that the action succeeded. Failed actions must show a clear error explaining what went wrong and what to do next. Every enabled button click must produce a visible response.
 
 **Implementation**:
 - Success/error alerts at top of page after every action
@@ -979,7 +979,7 @@ The most broadly compatible pattern for WCAG 2.1 AA compliance combines three me
 </div>
 ```
 
-Note on `aria-errormessage`: The ARIA spec defines `aria-errormessage` for this purpose, but browser/AT support is still inconsistent as of 2025. **Do not rely on `aria-errormessage` alone** — always include the error ID in `aria-describedby` as well.
+Note on `aria-errormessage`: The ARIA spec defines `aria-errormessage` for this purpose, but browser/AT support is still inconsistent across major screen readers. **Do not rely on `aria-errormessage` alone** — always include the error ID in `aria-describedby` as well.
 
 **Required Fields**
 - Always set the HTML5 `required` attribute on the input element
@@ -1175,17 +1175,18 @@ Custom icons are defined in `lib/components/cams/RawSvgIcon.tsx` as inline SVG c
 
 ## Animation & Transitions
 
-### Transition Timing
-- Standard duration: [200ms / 300ms]
-- Easing: [ease-in-out / cubic-bezier()]
+CAMS does not define a custom animation system. USWDS default transitions apply throughout the application. Where custom animations are needed, standard web animation values are used.
 
-### When to Animate
-- Page transitions: [Yes/No]
-- Modal open/close: [Fade in / Slide down / etc.]
-- Hover states: [Smooth color transitions]
-- Loading states: [Spinner animation / Pulse / etc.]
+### Current Animations in CAMS
 
+- **Alert fade-in/out** (`Alert.scss`): `opacity 0.2s ease-in-out` — alerts fade in when shown and fade out when hidden
+- **Filter panel slide-down** (`TrusteeDistrictFilter.scss`): `slideDown 0.2s ease-out` — filter panel fades in and slides down 10px when expanded
+- **Loading spinner** (`LoadingSpinner.scss`): `spin 1s linear infinite` — continuous rotation
+- **Scroll-to-top button** (`ScrollToTopButton.scss`): `bounce 1s steps(5)` — button bounces up when it appears
 
+### Accessibility
+- Always respect `prefers-reduced-motion` — users who have requested reduced motion should not see decorative animations
+- Critical information must never depend on animation to be understood
 
 ---
 
@@ -1240,10 +1241,3 @@ When implementing features in CAMS:
 - Color alone cannot convey information
 - Focus indicators must always be visible
 
----
-
-## Document Maintenance
-
-**Last Updated**: 07/02/2026
-**Owner**: DOJ CAMS Team
-**Feedback**: Submit a pull request with proposed changes — the team will review and discuss before merging
--- a/docs/BRAND_GUIDELINES.md
+++ b/docs/BRAND_GUIDELINES.md
@@ -21,7 +21,7 @@ These interaction principles apply to every screen and component in CAMS. They a
 ### 1. Escapability
 Users must be able to back out of what they're doing without following through. Every modal, form, and multi-step flow must have a cancel or back-out path that returns the user to their prior state without side effects. A user who starts an action by mistake should never be trapped into completing it.
 
-**Implementation**: 
+**Implementation**:
 - Every modal has Cancel button + Escape key support
 - Clicking outside modal cancels action
 - Multi-step forms have Back/Cancel at each step
@@ -66,10 +66,10 @@ CAMS uses the U.S. Web Design System (USWDS) 3.13.0 as its foundation. All color
 
 ### Primary Interactive Colors (Actually Used)
 
-- **Primary Blue**: `#005ea2` 
+- **Primary Blue**: `#005ea2`
   - Main brand color for primary CTAs, links, interactive elements
   - Used in: Links (theme.scss), filters, buttons
-  
+
 - **Primary Dark**: `#1a4480`
   - Link hover states, active/selected filter text
   - Used in: Link hovers, active filter states
@@ -193,7 +193,7 @@ CAMS uses the U.S. Web Design System (USWDS) 3.13.0 as its foundation. All color
 These colors are defined in `styles/abstracts/_colors.scss` but not currently used:
 
 - `$secondary-darker` (`#8b0a03`) - similar to but not identical to actual hover state `#8b0a0a`
-- `$warning-text` (`#990000`) 
+- `$warning-text` (`#990000`)
 - `$warning-light` (`#ffffca`) - USWDS provides warning colors instead
 
 ---
@@ -216,7 +216,7 @@ This section provides a complete technical breakdown of all colors in the CAMS c
 
 **Actually Used:**
 1. `$primary` (`#005ea2`) - Links in theme.scss
-2. `$primary-dark` (`#1a4480`) - Link hover states, active filters  
+2. `$primary-dark` (`#1a4480`) - Link hover states, active filters
 3. `$navy` (`#162e51`) - Header background
 4. `$gold` (`#ffbe2e`) - Header icons, table borders
 5. `$eagle` (`#2e2e2a`) - App background, table headers
@@ -372,7 +372,7 @@ CAMS follows specific casing conventions throughout the application:
 ### Font Families
 - **Body/UI Font**: Source Sans Pro (primary), with fallbacks
   ```
-  'Source Sans Pro', 'Source Sans Pro Web', 'Helvetica Neue', 
+  'Source Sans Pro', 'Source Sans Pro Web', 'Helvetica Neue',
   'Helvetica', 'Roboto', 'Arial', sans-serif
   ```
   - Applied globally via the `*` selector in `theme.scss`, which overrides all USWDS component fonts
@@ -383,7 +383,7 @@ CAMS follows specific casing conventions throughout the application:
   ```
   - Used only in `.cams-logo-and-title .site-title` (the application header)
 
-- **Monospace (Code)**: 
+- **Monospace (Code)**:
   ```
   source-code-pro, Menlo, Monaco, Consolas, 'Courier New', monospace
   ```
@@ -547,12 +547,12 @@ All buttons use USWDS button components with the base class `usa-button`.
   - Class: `usa-button` (no modifier)
   - Style: Solid fill with primary blue background
   - Usage: Limit to 1-2 per view for the most important action
-  
+
 - **Secondary Button**: Alternative or cancel actions (e.g., "Cancel", "Go Back")
   - Class: `usa-button usa-button--secondary`
   - Style: Lighter fill, less prominent than primary
   - Usage: Supporting actions that are less critical than primary
-  
+
 - **Outline Button**: Secondary actions that need clear boundaries (e.g., "Edit", "View Details")
   - Class: `usa-button usa-button--outline`
   - Style: Transparent background with colored border
@@ -598,7 +598,7 @@ All buttons use USWDS button components with the base class `usa-button`.
 - Note: `Warm` and `Cool` colors are used in the **Tag** component (`bg-accent-cool`, `bg-accent-warm-dark`) but not in buttons
 
 #### Button States - Visibility of System State
-- **During submission**: 
+- **During submission**:
   - Disable the submit button (`disabled` attribute) OR
   - Show loading indicator (spinner icon or text like "Saving...")
   - Prevents double-submission
@@ -629,12 +629,12 @@ CAMS uses USWDS form components with custom validation and accessibility enhance
 #### Input States
 - **Default**: Standard USWDS input styling (`usa-input`)
 - **Focus**: Blue outline ring (USWDS focus state)
-- **Error**: 
+- **Error**:
   - Class: `usa-input-group--error` on wrapper
   - Attribute: `aria-invalid="true"` on input
   - Visual: Red border/styling per USWDS
   - Message: Error text appears below input (see Error Messages below)
-- **Disabled**: 
+- **Disabled**:
   - Attribute: `disabled` on input
   - Background: `#c9c9c9` (gray)
   - Text: `#454545` (muted gray)
@@ -696,7 +696,7 @@ CAMS uses both USWDS tables and custom `CamsTable` component with responsive des
 
 #### Table Structure
 - **USWDS Tables**: Standard HTML `<table>` with USWDS classes
-- **CamsTable Component**: Custom responsive table using `role="table"` 
+- **CamsTable Component**: Custom responsive table using `role="table"`
   - Class: `cams-table cams-table--responsive`
   - Must include either `caption` or `aria-label` for accessibility
   - Caption uses `cams-table__caption` class
@@ -839,7 +839,7 @@ Every authenticated screen in CAMS must include:
   - Contains CAMS/USTP wordmark or logo
   - Navigation links appropriate to user's role
   - Present on every authenticated screen including admin pages
-  
+
 - **Privacy Act Footer**: Required on every authenticated screen
   - Legal/compliance requirement for government applications
 
@@ -859,7 +859,7 @@ Every authenticated screen in CAMS must include:
   - **Hidden** (preferred - remove from DOM entirely), OR
   - **Visibly disabled** with explanation (when hiding would be confusing)
   - **Never**: Silently failing after user attempts the action
-  
+
 -
 
 **Implementation**: Check user roles on the frontend before rendering controls, and always validate permissions on the backend before executing actions.
@@ -1126,12 +1126,12 @@ CAMS uses **USWDS Icon Sprites** (`sprite.svg`)
 - **decorative**: Boolean, default `true` - whether icon is decorative only
 
 ### Accessibility
-- **Decorative icons** (default): 
+- **Decorative icons** (default):
   - `aria-hidden="true"`
   - No `role` attribute
   - `focusable={false}`
   - Use when icon accompanies text or is purely visual
-  
+
 - **Non-decorative icons** (`decorative={false}`):
   - `role="img"`
   - `aria-label="{name} icon"` or use `tooltip` for better description
@@ -1203,7 +1203,7 @@ When implementing features in CAMS:
 ### Implementation Guidelines
 1. **Default to established patterns**: Unless explicitly asked to create something new, match existing UI patterns
 2. **Consistency over novelty**: CAMS values consistency - match what exists even if you think there's a "better" way
-3. **Text casing**: 
+3. **Text casing**:
    - Remember: **Title Case** for headers (except in alert messages), buttons, table headers
    - **Sentence case** for labels, body text, error messages
 4. **Terminology precision**: Use exact CAMS terms (see Terminology section):
@@ -1211,13 +1211,13 @@ When implementing features in CAMS:
    - "Staff Assignment" not "Case Assignment"
    - "Verify" not "Approve"
    - "Case Filed Date" not "Filing Date" or "Petition Date"
-5. **Error messages**: 
+5. **Error messages**:
    - Clear verb phrases in sentence case
    - Short directives: no period (e.g., "Enter state", "Select trial city")
    - Full sentences: use periods (e.g., "Unable to save changes. Please try again.")
 6. **Dates**: Always format as MM/DD/YYYY (or MM/DD/YY when space-constrained)
 7. **Currency**: Always show two decimal places ($60.00)
-8. **Success/Error alerts**: 
+8. **Success/Error alerts**:
    - Position at top of page under header
    - Scroll user to top
    - Include both summary and next steps
@@ -1244,6 +1244,6 @@ When implementing features in CAMS:
 
 ## Document Maintenance
 
-**Last Updated**: 07/02/2026  
-**Owner**: DOJ CAMS Team  
+**Last Updated**: 07/02/2026
+**Owner**: DOJ CAMS Team
 **Feedback**: Submit a pull request with proposed changes — the team will review and discuss before merging
--- a/docs/BRAND_GUIDELINES.md
+++ b/docs/BRAND_GUIDELINES.md
@@ -21,7 +21,7 @@ These interaction principles apply to every screen and component in CAMS. They a
 ### 1. Escapability
 Users must be able to back out of what they're doing without following through. Every modal, form, and multi-step flow must have a cancel or back-out path that returns the user to their prior state without side effects. A user who starts an action by mistake should never be trapped into completing it.
 
-**Implementation**:
+**Implementation**: 
 - Every modal has Cancel button + Escape key support
 - Clicking outside modal cancels action
 - Multi-step forms have Back/Cancel at each step
@@ -35,7 +35,7 @@ Users should not have to remember or type names, codes, or IDs from memory. When
 - Show "New York Southern District" not "USTP_CAMS_Region_02_Office_New_York"
 
 ### 3. Visibility of System State
-After any action (assignment created, role removed, order verified), the user must see immediate confirmation that the action succeeded. Failed actions must show a clear error explaining what went wrong and what to do next. Every enabled button click must produce a visible response.
+After any action (assignment created, role removed, order verified), the user must see immediate confirmation that the action succeeded. Failed actions must show a clear error explaining what went wrong and what to do next. Buttons must not do nothing silently.
 
 **Implementation**:
 - Success/error alerts at top of page after every action
@@ -66,10 +66,10 @@ CAMS uses the U.S. Web Design System (USWDS) 3.13.0 as its foundation. All color
 
 ### Primary Interactive Colors (Actually Used)
 
-- **Primary Blue**: `#005ea2`
+- **Primary Blue**: `#005ea2` 
   - Main brand color for primary CTAs, links, interactive elements
   - Used in: Links (theme.scss), filters, buttons
-
+  
 - **Primary Dark**: `#1a4480`
   - Link hover states, active/selected filter text
   - Used in: Link hovers, active filter states
@@ -193,7 +193,7 @@ CAMS uses the U.S. Web Design System (USWDS) 3.13.0 as its foundation. All color
 These colors are defined in `styles/abstracts/_colors.scss` but not currently used:
 
 - `$secondary-darker` (`#8b0a03`) - similar to but not identical to actual hover state `#8b0a0a`
-- `$warning-text` (`#990000`)
+- `$warning-text` (`#990000`) 
 - `$warning-light` (`#ffffca`) - USWDS provides warning colors instead
 
 ---
@@ -216,7 +216,7 @@ This section provides a complete technical breakdown of all colors in the CAMS c
 
 **Actually Used:**
 1. `$primary` (`#005ea2`) - Links in theme.scss
-2. `$primary-dark` (`#1a4480`) - Link hover states, active filters
+2. `$primary-dark` (`#1a4480`) - Link hover states, active filters  
 3. `$navy` (`#162e51`) - Header background
 4. `$gold` (`#ffbe2e`) - Header icons, table borders
 5. `$eagle` (`#2e2e2a`) - App background, table headers
@@ -372,7 +372,7 @@ CAMS follows specific casing conventions throughout the application:
 ### Font Families
 - **Body/UI Font**: Source Sans Pro (primary), with fallbacks
   ```
-  'Source Sans Pro', 'Source Sans Pro Web', 'Helvetica Neue',
+  'Source Sans Pro', 'Source Sans Pro Web', 'Helvetica Neue', 
   'Helvetica', 'Roboto', 'Arial', sans-serif
   ```
   - Applied globally via the `*` selector in `theme.scss`, which overrides all USWDS component fonts
@@ -383,7 +383,7 @@ CAMS follows specific casing conventions throughout the application:
   ```
   - Used only in `.cams-logo-and-title .site-title` (the application header)
 
-- **Monospace (Code)**:
+- **Monospace (Code)**: 
   ```
   source-code-pro, Menlo, Monaco, Consolas, 'Courier New', monospace
   ```
@@ -547,12 +547,12 @@ All buttons use USWDS button components with the base class `usa-button`.
   - Class: `usa-button` (no modifier)
   - Style: Solid fill with primary blue background
   - Usage: Limit to 1-2 per view for the most important action
-
+  
 - **Secondary Button**: Alternative or cancel actions (e.g., "Cancel", "Go Back")
   - Class: `usa-button usa-button--secondary`
   - Style: Lighter fill, less prominent than primary
   - Usage: Supporting actions that are less critical than primary
-
+  
 - **Outline Button**: Secondary actions that need clear boundaries (e.g., "Edit", "View Details")
   - Class: `usa-button usa-button--outline`
   - Style: Transparent background with colored border
@@ -598,7 +598,7 @@ All buttons use USWDS button components with the base class `usa-button`.
 - Note: `Warm` and `Cool` colors are used in the **Tag** component (`bg-accent-cool`, `bg-accent-warm-dark`) but not in buttons
 
 #### Button States - Visibility of System State
-- **During submission**:
+- **During submission**: 
   - Disable the submit button (`disabled` attribute) OR
   - Show loading indicator (spinner icon or text like "Saving...")
   - Prevents double-submission
@@ -629,12 +629,12 @@ CAMS uses USWDS form components with custom validation and accessibility enhance
 #### Input States
 - **Default**: Standard USWDS input styling (`usa-input`)
 - **Focus**: Blue outline ring (USWDS focus state)
-- **Error**:
+- **Error**: 
   - Class: `usa-input-group--error` on wrapper
   - Attribute: `aria-invalid="true"` on input
   - Visual: Red border/styling per USWDS
   - Message: Error text appears below input (see Error Messages below)
-- **Disabled**:
+- **Disabled**: 
   - Attribute: `disabled` on input
   - Background: `#c9c9c9` (gray)
   - Text: `#454545` (muted gray)
@@ -696,7 +696,7 @@ CAMS uses both USWDS tables and custom `CamsTable` component with responsive des
 
 #### Table Structure
 - **USWDS Tables**: Standard HTML `<table>` with USWDS classes
-- **CamsTable Component**: Custom responsive table using `role="table"`
+- **CamsTable Component**: Custom responsive table using `role="table"` 
   - Class: `cams-table cams-table--responsive`
   - Must include either `caption` or `aria-label` for accessibility
   - Caption uses `cams-table__caption` class
@@ -839,7 +839,7 @@ Every authenticated screen in CAMS must include:
   - Contains CAMS/USTP wordmark or logo
   - Navigation links appropriate to user's role
   - Present on every authenticated screen including admin pages
-
+  
 - **Privacy Act Footer**: Required on every authenticated screen
   - Legal/compliance requirement for government applications
 
@@ -859,7 +859,7 @@ Every authenticated screen in CAMS must include:
   - **Hidden** (preferred - remove from DOM entirely), OR
   - **Visibly disabled** with explanation (when hiding would be confusing)
   - **Never**: Silently failing after user attempts the action
-
+  
 -
 
 **Implementation**: Check user roles on the frontend before rendering controls, and always validate permissions on the backend before executing actions.
@@ -979,7 +979,7 @@ The most broadly compatible pattern for WCAG 2.1 AA compliance combines three me
 </div>
 ```
 
-Note on `aria-errormessage`: The ARIA spec defines `aria-errormessage` for this purpose, but browser/AT support is still inconsistent across major screen readers. **Do not rely on `aria-errormessage` alone** — always include the error ID in `aria-describedby` as well.
+Note on `aria-errormessage`: The ARIA spec defines `aria-errormessage` for this purpose, but browser/AT support is still inconsistent as of 2025. **Do not rely on `aria-errormessage` alone** — always include the error ID in `aria-describedby` as well.
 
 **Required Fields**
 - Always set the HTML5 `required` attribute on the input element
@@ -1126,12 +1126,12 @@ CAMS uses **USWDS Icon Sprites** (`sprite.svg`)
 - **decorative**: Boolean, default `true` - whether icon is decorative only
 
 ### Accessibility
-- **Decorative icons** (default):
+- **Decorative icons** (default): 
   - `aria-hidden="true"`
   - No `role` attribute
   - `focusable={false}`
   - Use when icon accompanies text or is purely visual
-
+  
 - **Non-decorative icons** (`decorative={false}`):
   - `role="img"`
   - `aria-label="{name} icon"` or use `tooltip` for better description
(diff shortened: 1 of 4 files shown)


jalba-flexion (approved):

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1249 @@
...
+
+### 3. Visibility of System State
+After any action (assignment created, role removed, order verified), the user must see immediate confirmation that the action succeeded. Failed actions must show a clear error explaining what went wrong and what to do next. Buttons must not do nothing silently.
+
+**Implementation**:
+- Success/error alerts at top of page after every action
+- Loading indicators during async operations
+- Disabled buttons show loading state during submission
+
+### 4. Clarity of Scope
```
jharcey-flexion, line 68: Should we add the tab name to this?

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
+# CAMS Brand Guidelines
+
+This document provides clear guidelines for implementing UI components, styling, and content in the CAMS (Case Assignment & Management System) application. These guidelines ensure consistency across the application and help LLM assistants make appropriate styling decisions.
+
+## Application Context
+
+**CAMS (Bankruptcy Oversight Support Systems)** is an internal web application used by USTP (US Trustee Program) staff:
+- **Users**: Trial attorneys, office managers, paralegals, auditors, IT administrators
+- **Purpose**: Manage bankruptcy cases, staff assignments, court dockets, and user access
```
fmaddenflx: No sure if "Managing user access" is applicable or what it refers to exactly.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- **Environment**: Desktop-first responsive application for federal employees on government-issued workstations
+- **Minimum viewport**: 1024px wide, adapts to various desktop screen sizes
```
fmaddenflx: I don't believe this is true.  Are we really assuming that the minimum screen width is 1024px???

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+## Core UX Principles
+
+These interaction principles apply to every screen and component in CAMS. They are binding design constraints, not optional guidelines.
+
+### 1. Escapability
+Users must be able to back out of what they're doing without following through. Every modal, form, and multi-step flow must have a cancel or back-out path that returns the user to their prior state without side effects. A user who starts an action by mistake should never be trapped into completing it.
+
+**Implementation**:
+- Every modal has Cancel button + Escape key support
+- Clicking outside modal cancels action
```
fmaddenflx: This is fine, but not sure this is 100% always going to be true.  The basic difference between a "dialog" and a "modal" is that a modal requires a response.  A dialog is for performing optional actions.  A dialog may not darken the background screen and could possibly remain open while the user is doing other things. A modal on the other hand, darkens the background and prevents the user from doing other things until they take action on the modal.  A modal could be canceled, but closing when clicking outside of the modal is not necessarily always the desired action.

This may or may not apply to our application, but that is the basic vocabulary for such things traditionally.  I know that modal and dialog have become synonymous these days, but in reality they are slightly different concepts.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- Multi-step forms have Back/Cancel at each step
```
fmaddenflx: Also worth noting that long, complex, or multi-step forms should be avoided in dialogs and modals.  Content in dialog/modals should be kept to a minimum so that the user can easily view the content in the dialog/modal and so that they don't loose track of what they are doing.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- Use dropdowns for role/office selection with plain English labels
+- Show "New York Southern District" not "USTP_CAMS_Region_02_Office_New_York"
+
+### 3. Visibility of System State
+After any action (assignment created, role removed, order verified), the user must see immediate confirmation that the action succeeded. Failed actions must show a clear error explaining what went wrong and what to do next. Every enabled button click must produce a visible response.
+
+**Implementation**:
+- Success/error alerts at top of page after every action
+- Loading indicators during async operations
+- Disabled buttons show loading state during submission
```
fmaddenflx, line 66:
```suggestion
- Buttons become disabled after clicking until action has completed (for both visual feed back and to prevent multiple clicks firing the event multiple times).
- Disabled buttons show loading state during submission
```

docs/BRAND_GUIDELINES.md
fmaddenflx: Jon, could you provide more context to that?  What is it about tab name that should be stated here? Are we referring to a "browser tab" or are we referring to a tab in the application UI? Are we perhaps suggesting that a tab in the UI should be highlighted when it is active?

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- Plain language labels throughout
+
+### 5. Appropriate Access by Role
+The UI must not offer actions to users who lack permission to take them. Controls for disallowed actions are hidden or visibly disabled—not silently failing after submission.
+
+**Implementation**:
+- Check user roles before rendering admin links
+- Hide/disable controls for actions user can't perform
+- Always validate permissions on backend
+
```
fmaddenflx: I would suggest we don't "hide" them but rather that we don't put them into the DOM at all.  If an action should not be available to a user, that UI element should not be in the DOM at all, not just hidden from user view.  This is a more secure approach.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+---
+
+## Color System
```
fmaddenflx, line 98: Thinking ahead -- we may at some point want to provide the users with color themes for accessibility purposes.  If so, we should avoid hard-coding colors into any scss file except for our theme.scss.  All other places in our tsx where we build UI should reference theme.scss.  This would keep things organized well and allow us to swap out themes easily.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+### Text Colors (Actually Used)
+
+- **Primary Text (Ink)**: `#1b1b1b`
+  - Body text, default content
+  - Used in: Borders in trustees, verification components
+
+- **Eagle (Dark Text)**: `#2e2e2a`
+  - Used in: `$tableHeaderBackground` variable in `_tables.scss` (defined but that variable is never imported or applied), and `.App-header` background in `App.scss` (Create React App boilerplate, not the actual CAMS header)
+  - Note: Effectively unused in the live CAMS UI
+
```
fmaddenflx: Based on the name, and my cloudy imaginative memory, I seem to recall, many centuries ago, this was a color I grabbed from the eagle logo of the DOJ USTP.  It's a cool color name.  But if it's not being used, then it shouldn't be listed under the heading of "Text Colros (Actually Used)".

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+### Background Colors (Actually Used)
+
+- **White**: `#ffffff` - Page background, cards, panels
+- **Light Gray**: `#f0f0f0`
+  - RichTextEditor toolbar button hover and active states
+  - Case Detail Header border when in sticky/fixed position (see issue #6 — sticky header not currently working; feature may be removed)
+- **Near-White**: `#fefefe`
+  - Case Detail Header background when in sticky/fixed position (see issue #6 — sticky header not currently working; feature may be removed)
+- **Disabled Input**: `#c9c9c9` - Disabled ComboBox background (e.g., "New Court" field in Data Verification → Transfers while validation is processing)
```
fmaddenflx: Sounds like we need to standardize our disabled UI element color.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+### Border & Divider Colors (Actually Used)
+
+- **Light Gray**: `#ddd`
```
fmaddenflx: To be clear, #ddd is shorthand for #dddddd, not for #d0d0d0 or #d6d6d6.  Just incase anyone is wondering.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+  - `CaseDetailHeader.scss` box shadow on sticky header (non-functioning — see issue #6)
+  - `BankruptcySoftware.scss` list item border (dead code — see issue #4)
+  - `Table.scss` cell border inside `small-table` mixin (mixin defined but never included anywhere — see issue #17)
+  - Note: Effectively unused in the live CAMS UI
```
fmaddenflx: Once again, #ddd probably shouldn't be under the category of "Border & Divider Colors (Actually Used)" if it's not actually used.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+- **Black**: `#000000`
+  - Note item dividers (Case Notes and Trustee Notes panels)
+  - Loading spinner text
+  - TEST environment banner text
+  - Search result highlight text (Court Docket)
+  - Pagination button background
```
fmaddenflx: The pagination buttons are black?  I don't think they are. This confuses me.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+  - Note: `3px solid black` table row borders appear in `Table.scss` but are inside the unused `small-table` mixin (see issue #17)
+
+### Component-Specific Colors
+
+**Rich Text Editor** (hardcoded in component):
+- Toolbar Background: `#e9ecef`
+- Active Button Background: `#71767a` (with white text)
+- Hover State: `#f0f0f0`
+- Link Color: `#005ea6`
+- Muted Controls: `#565c65`
```
fmaddenflx, line 216: We should standardize our colors and keep them all in either _theme.scss or _colors.scss.  We should probably stop hard-coding them in components.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `$secondary-darker` (`#8b0a03`)
+- `$warning-text` (`#990000`)
+- `$warning-light` (`#ffffca`)
+- `$tableBodyBackgroundOddRows` (`#f6f6f6`) - defined in `_tables.scss` but never applied anywhere
+
+### All Hardcoded Hex Colors Found
+
+**Blues:**
+- `#005ea2` - Primary blue (also SCSS variable)
+- `#005ea6` - RichTextEditor links (slightly different)
```
fmaddenflx: Yeah we can probably just have RichTextEditor links use Primary.  There's no reason to have differing colors that are so close.  I probably wouldn't be able to tell the difference.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `#71767a` - RichTextEditor active button
+
+**Medium Grays:**
+- `#757575` - Secondary text, helper text
+- `#949494` - Additional gray shade
+- `#b0b0b0` - Additional gray shade
+
+**Light Grays:**
+- `#c6c6c6` - Additional gray shade
+- `#c6cace` - Additional gray shade
```
fmaddenflx: We can probably do away with one of these in favor of the other.  They are likely very close.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `#c9c9c9` - Disabled input background (also SCSS variable)
+- `#d0d0d0` - Additional gray shade
+- `#ddd` - Table borders (3-digit hex)
```
fmaddenflx: We should change this to #dddddd just to avoid any confusion.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `#f9f9f9` - Light background for cards
+- `#fefefe` - Near-white for headers
+
+**Whites:**
+- `#ffffff` - White backgrounds
+
+**Reds (Error/Danger):**
+- `#b50909` - Error red (also SCSS variable)
+- `#8b0a0a` - Error hover state (NOT the same as `$secondary-darker #8b0a03`)
+- `#8b0a03` - Defined in SCSS but unused
```
fmaddenflx: Delete the second one.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `#ffbe2e` - Gold (also SCSS variable)
+- `#e6c74c` - Yellow-20 (also SCSS variable)
+- `#ffd700` - Banner gold for TEST environment
+- `#ffffca` - Defined in SCSS but unused
+
+**Greens:**
+- `#ecf3ec` - Success green (also SCSS variable)
+
+**Other:**
+- `#61dafb` - React logo cyan (App.scss, likely not relevant)
```
fmaddenflx: React logo ?!!!??  DELETE!!! worthless collection of bytes...

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `#990000` - Defined in SCSS but unused
+
+### USWDS Theme Colors (Not in colors.scss)
+
+These come directly from USWDS:
+- `$theme-color-warning-lighter` (`#faf3d1`) - Warning alerts, Notes search highlighting
+- USWDS alert colors via `usa-alert--success`, `usa-alert--error`, `usa-alert--warning`, `usa-alert--info`
+- `$theme-color-secondary-dark` - Used in Notes component for remove button hover
+
```
fmaddenflx, line 273: We need to verify that color variables coming from USWDS scss are actually being properly imported and are usable.  I had trouble with this early on, which is why we're defining new variables for existing USWDS colors in _colors.scss and _theme.scss.  If USWDS color variables are properly imported, then we don't need to redefine them.

jharcey-flexion (commented):
Looks pretty good to me. I don't have specific knowledge of implementation, but I like how this is set up. We'll need to decide how to deal with the issues.

docs/BRAND_GUIDELINES.md
jharcey-flexion: Yes, I am wondering if we should include how we are naming the browser tabs here.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+## Typography
+
+### Text Casing Rules
+CAMS follows specific casing conventions throughout the application:
+
+- **Headers (H1, H2, H3, etc.)**: Title Case for all headers
+  - **Exception**: Use Sentence case for headers in page-level success/alert/error messages
+- **Field Labels**: Sentence case (e.g., "Case number", "Debtor name")
+  - **Exception**: Use Title Case for document type and document titles
```
fmaddenflx: I could use some clarification here as I'm not sure what we mean by "document type" and "document titles"

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- `padding-y-4` — vertical padding in modals
+
+### What to Use Where
+- **In `.scss` files**: Use rem values
+- **In `.tsx`/`.jsx` files**: Use USWDS utility classes (`margin-top-2`, etc.)
+- **Avoid**: Hardcoding pixel values in either context
+
+### Note on Table Spacing Variables
+`$tableCellPadding` (10px), `$tableBorderRadius` (5px), and `$tableBorderCollapse` (collapse) are defined in `_tables.scss` but **never used** anywhere in the application.
+
```
fmaddenflx, line 445: We should either use them or delete them.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+  - Style: No background, border, or padding - looks like plain text/icon
+  - Usage: Non-prominent actions, icon-only buttons in headers
+
+- **IconButton**: Dedicated component for unstyled buttons with a single icon
+  - Component: `IconButton` in `lib/components/IconButton.tsx`
+  - Wraps `Button` with `UswdsButtonStyle.Unstyled` and an `Icon` inside
+  - Props: All standard button props + `icon` (string name from USWDS sprite)
+  - Example: `<IconButton icon="close" aria-label="Close panel" />`
+  - **Prefer `IconButton` over manually composing an unstyled button + icon** — it is the canonical pattern
+  - Note: Much of the codebase currently composes unstyled buttons with icons directly rather than using `IconButton` — new code should use `IconButton`
```
fmaddenflx: This is interesting.  We should swap those out for IconButton's assuming the code is essentially the same.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- **Disabled**:
+  - Attribute: `disabled` on input
+  - Background: `#c9c9c9` (gray)
+  - Text: `#454545` (muted gray)
+  - Cursor: `not-allowed`
+- **Valid**: No special styling (CAMS doesn't use green checkmarks for valid fields)
+
+#### Error Messages (Field-Level)
+- **Position**: Directly below the input field with error
+- **Class**: `usa-input__error-message` (custom CAMS class — not the USWDS standard `usa-error-message`, though it follows USWDS naming conventions)
```
fmaddenflx: This is interesting.  Don't know why we're doing this but we should look into it.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+### Tables
+
+CAMS uses both USWDS tables and custom `CamsTable` component with responsive design.
+
+#### Table Headers
+- **Text casing**: Title Case
+- **Alignment**: Left-aligned by default; right-alignment used for specific columns where it aids layout (e.g., document number column in Court Docket)
+- **Font weight**: Bold (700) - typically in `<thead>` elements
+- **Sorting**: Show sort indicators when column is sortable
+- **Mobile**: Hidden on small screens (handled by responsive mixin)
```
fmaddenflx: It seems like this is all referring to actual HTML tables (ie USWDS table), but that's not clear. However, most if not all of what it says pertains only to <table> elements.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+#### Table Structure
+- **USWDS Tables**: Standard HTML `<table>` with USWDS classes
+- **CamsTable Component**: Custom responsive table using `role="table"`
+  - Class: `cams-table cams-table--responsive`
+  - Must include either `caption` or `aria-label` for accessibility
+  - Caption uses `cams-table__caption` class
+
+#### Table Rows
+- **Alternating rows**: Not used in CAMS — all table rows have a white background (note: `$tableBodyBackgroundOddRows` and `$tableBodyBackgroundEvenRows` are defined in `_tables.scss` but never applied — see issue #2)
```
fmaddenflx: Yeah this was something we used in the past but seem to have abandoned the last year or so.  We should probably remove these variables if we aren't going to use them to avoid confusion.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+  - Each cell becomes full-width block
+  - Shows column labels via `data-cell` attribute using `::before` pseudo-element
+  - Row borders come from USWDS defaults (`1px solid #1b1b1b`)
+
+- **`cams-table-flex-row`**: Enables horizontal flex-row column layout on larger screens
+  - Shows header row
+  - `1px solid #1b1b1b` border under the header row
+  - Cells display in a flex row with gaps
+
+Note: A separate `small-table` mixin exists in `Table.scss` with `3px solid black` row borders but is never included anywhere — see issue #17.
```
fmaddenflx: Yeah we should remove the mixin probably.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+- **Privacy Act Footer**: Required on every authenticated screen
+  - Legal/compliance requirement for government applications
+
+- **Browser Page Title**: Must include "CAMS" in the `<title>` tag
+  - Helps users identify the application in browser tabs
+
+### Route Structure
+- **Admin features**: All administrative features (user management, permissions, etc.) should be under `/admin/*` routes
+- **Consistency**: Maintain consistent URL patterns with existing features
```
fmaddenflx: This is completely lacking any information about normal App routing outside of admin.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+### Writing Style
+- **Style Guide**: Generally follows AP style
+- **Contractions**: Use contractions whenever possible to sound more natural and approachable
+- **Commas**: Always use serial/Oxford commas
+- **Tone**: Professional but clear and direct. Avoid overly formal or legalistic language in UI copy.
+
+---
+
+## Terminology & Preferred Language
```
fmaddenflx, line 949: There's a lot of great information in this section.  I'm noticing though a pattern of the documentation repeating itself several times.  Perhaps we can tune that?  It kind of gives me the feeling that I want to tune out or ignore information because I've heard it belabored before.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+  {errorMessage}
+</div>
+```
+
+Note on `aria-errormessage`: The ARIA spec defines `aria-errormessage` for this purpose, but browser/AT support is still inconsistent across major screen readers. **Do not rely on `aria-errormessage` alone** — always include the error ID in `aria-describedby` as well.
+
+**Required Fields**
+- Always set the HTML5 `required` attribute on the input element
+- Also set `aria-required="true"` for explicit AT communication (some AT handle `required` natively; `aria-required` is belt-and-suspenders)
+- Visual required indicator (asterisk) must be in actual text content or a visually-hidden span — **CSS `::after` pseudo-content is not reliably exposed to screen readers**
```
fmaddenflx: I was not aware of this about ::after pseudo-content!  This is good to know and we'll need to modify how we are displaying the * on the components!

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+| `Input` | ✓ | ✓ | ✗ | ✓ (custom live region) | ✓ | ✗ | ✗ (CSS only) |
+| `ComboBox` | ✓ | ✗ | ✗ | ✗ | ✗ | ✗ | Partial (* in label) |
+| `Radio` | ✗ (points to hidden input) | N/A | N/A | N/A | N/A | ✗ | ✗ |
+| `RadioGroup` | ✓ fieldset/legend | N/A | N/A | N/A | N/A | Partial | Partial (unreliable) |
+| `Checkbox` | ✓ content | N/A | N/A | N/A | ✗ | ✗ | ✗ |
+| `TextArea` | ✓ | ✓ | ✗ | ✗ | ✓ | ✗ | ✗ (empty span — bug) |
+| `DatePicker` | ✓ | ✓ | ✓ | ✓ | ✓ | ✗ | ✗ |
+| `DateRangePicker` | ✓ (via DatePicker) | ✓ | ✓ | ✓ | ✓ | ✗ | ✗ |
+
+**`DatePicker` is the most complete implementation** and should be used as the reference pattern for other components. Several components have known gaps — see issues #19–#25.
```
fmaddenflx: Really good to know!

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+#### Live Regions
+- **Form validation**: Use `aria-live="polite"` on the error message element itself (DatePicker pattern) — more reliable than a separate live region div
+- **Alerts**: Use appropriate `role` on alert components:
+  - `role="alert"` with `aria-live="assertive"` for errors (USWDS default for `usa-alert--error`)
+  - `role="status"` with `aria-live="polite"` for success/info/warning
+- **Live region text**: Only update on genuine state changes to avoid over-announcement
+- **`aria-atomic="true"`**: Use on live regions that should be read in full when updated
+
+#### Tables
```
fmaddenflx, line 1128: Is it me or is this entire section a repeat from earlier?!!?

fmaddenflx: Guess it goes into more detail here but makes the previous information totally redundant and makes me instinctively want to skip this section.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+  role="rowgroup"  (body — CamsTableBody)
+    role="row"  (CamsTableRow)
+      role="cell"  (CamsTableCell)
+```
+
+**Labeling — required for all tables**
+- Every `CamsTable` must have either `caption` or `aria-label` prop (enforced by TypeScript)
+- Do not pass both `aria-label` and `caption` — `aria-labelledby` (from caption) will override `aria-label`. Use one or the other.
+- When `caption` is provided: rendered as a visible `<div>` and associated via `aria-labelledby`
+- When `aria-label` is provided: passed directly to the `role="table"` div
```
fmaddenflx: hmmm.  We should enforce this with code, so that if caption is provided, then aria-label is turned off and not put in the DOM.  Or perhaps it doesn't matter because it will be overridden by caption anyway?  Maybe there's some way to catch it in build so that the developer is aware of the problem and takes steps to ensure the correct text is displayed?

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+**`data-cell` attribute — required for responsive layout**
+- Every `CamsTableCell` in a multi-column table **must** have a `data-cell` attribute
+- In mobile/stacked view, the header is visually hidden and `data-cell` provides the only visible column label via CSS `::before` pseudo-element
+- Without `data-cell`, column context is completely lost in stacked layout — WCAG 1.3.1 failure
+- For action columns with no meaningful label, use `data-cell=""` (empty string suppresses the label)
+- Example: `<CamsTableCell data-cell="Case number">091-23-12345</CamsTableCell>`
```
fmaddenflx, line 1168: It should be noted that from my recollection, the CamsTable component is designed to add the data-cell attribute with the contents of the column headers automatically, so though this should be noted, it is not necessary for the developer to worry about it, as long as they are using the CamsTable component.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+**Known gaps in CamsTable (see issues #27–#29)**
+- No `role="rowheader"` support — cannot identify row-scoping cells
+- No `aria-sort` support on `CamsTableHeaderCell` — blocks accessible sort
+- Several existing tables are missing `data-cell` attributes
```
fmaddenflx, line 1174: This is interesting and we should fix all of these.  The last one contradicts what I said earlier about data-cell.  We better check into that.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+**Rules for new development:**
+- `<main>`: One per page, always via `<MainContent>` wrapper
+- `<nav>`: Always include `aria-label` when more than one `<nav>` exists on a page
+- `<header>` and `<footer>`: Do NOT add `role="banner"` or `role="contentinfo"` — they are implicit on top-level elements. These explicit roles are redundant and create noise.
+- `<section>`: Only use when the section has a meaningful name. Add `aria-label` or `aria-labelledby` to make it a navigable `role="region"` landmark. Otherwise use `<div>` for purely structural grouping.
+- `<header>` nested inside another `<header>`: Not allowed — use `<div>` instead
+- `<header role="banner">` inside `<main>`: Not allowed — banner must be top-level
+
+
```
fmaddenflx: Just noting where I left off on 7/9 (this is a LONG document).  I'll pick up here tomorrow.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- **Icon-only buttons**: Must have `aria-label` on the **button**, not just on the icon — the `IconButton` component does not enforce this via the type system, so callers must supply it
+- **`focusable={false}`**: Required to prevent legacy browsers (IE/Edge) from focusing SVGs via keyboard. Always set on decorative SVGs.
+- Do not add `aria-label` to an SVG without also adding `role="img"` — `aria-label` alone on an SVG is not reliably announced by all screen readers
+
+#### Visually Hidden Text
+- **Canonical class**: Use `.usa-sr-only` (USWDS ships this class) — do not use `.screen-reader-only` for new code
+- The visually-hidden pattern uses the WCAG-approved clipping technique: `position: absolute; width: 1px; height: 1px; overflow: hidden; clip: rect(0 0 0 0); white-space: nowrap` — this keeps content in the DOM and accessible to screen readers
+- **Never use `display: none` or `visibility: hidden`** to hide content that should still be accessible to screen readers — both remove it from the accessibility tree entirely
+- `visibility: hidden` is appropriate only when the content should be hidden from both sighted users and screen readers
```
fmaddenflx: **_Correction here_**:  `visibility: hidden` puts a blank space in the DOM and includes padding and margin.  If you actually want something completely absent that doesn't affect the DOM in anyway, you must use `display: none` instead.

fmaddenflx: Or better yet, leave the element out of the DOM in the first place.  If we don't need it and don't want the screen reader to see it, don't include it at all.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+---
+
+## Icons
+
+### Icon Library
+CAMS uses **USWDS Icon Sprites** (`sprite.svg`)
+- Icons are loaded from `/assets/styles/img/sprite.svg#[icon-name]`
+- Uses SVG `<use>` pattern for performance and consistency
+- Component: `Icon` component in `lib/components/uswds/Icon.tsx`
```
fmaddenflx: This is not true 100% of the time.  Sometimes we use dedicated icon files, though less common.  USWDS may not always have an icon that we need.

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+
+### Icon Component Usage
+```tsx
+<Icon name="icon-name" />
+<Icon name="close" tooltip="Close this panel" decorative={false} />
+```
+
+#### Icon Props
+- **name** (required): String name of icon from sprite (e.g., "close", "info", "warning")
```
fmaddenflx, line 1284: We should note that those names can be found at the uswds website, which maintains a list of icons.  https://designsystem.digital.gov/components/icon/

docs/BRAND_GUIDELINES.md
```diff
@@ -0,0 +1,1242 @@
...
+- **tooltip**: Optional tooltip text (rendered as `<title>` element)
+- **className**: Additional CSS classes
+- **focusable**: Boolean, default `false` - whether icon can receive keyboard focus
+- **decorative**: Boolean, default `true` - whether icon is decorative only
```
fmaddenflx, line 1288: I'm wondering if these are actually somewhat mutually exclusive.  We should look into that.  I'm not sure what we'd do about that though, as we do need to have both of these attributes settable separately.  Maybe simply noting it in here and in the code would be useful enough.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #48 in BRAND_GUIDELINES_ISSUES.md to investigate why this deviates from the USWDS standard `usa-error-message` class. No doc change needed here since the guidelines already flag it as a custom class.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — added a scoping note to Table Headers clarifying which rules apply to USWDS `<table>` vs `CamsTable`, and clarified the Mobile bullet is CamsTable-specific.

docs/BRAND_GUIDELINES.md
fmaddenflx: This is already tracked in issue #2 (Dead Code — Unused Color Variables), which covers removing `$tableBodyBackgroundOddRows`/`$tableBodyBackgroundEvenRows`. No further doc change needed.

docs/BRAND_GUIDELINES.md
fmaddenflx: Already tracked in issue #17 (Dead CSS — small-table Mixin Defined But Never Included). No further doc change needed.

docs/BRAND_GUIDELINES.md
fmaddenflx: Removed the Route Structure section entirely rather than patch it — it only covered admin routing, and a short blurb about the rest of the app's routing would go stale quickly.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed by the repetition-reduction pass in 92bbbdf54 and 19ce625fb (period rules, escapability, LLM notes, icon accessibility, etc. consolidated with cross-references instead of being restated).

docs/BRAND_GUIDELINES.md
fmaddenflx: Acknowledged — no action needed, this was an FYI note.

docs/BRAND_GUIDELINES.md
fmaddenflx: Acknowledged — no action needed.

docs/BRAND_GUIDELINES.md
fmaddenflx: Found and fixed one genuine duplication — Table Structure previously restated the caption/aria-label labeling rule that's covered in full under Accessibility Requirements → ARIA Patterns → Tables. Trimmed it to a cross-reference. The Buttons/Live Regions ARIA subsections don't duplicate anything else, so left those as-is.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #49 — proposing a discriminated-union prop type or lint rule so passing both `caption` and `aria-label` is a build-time error instead of a silent no-op.

docs/BRAND_GUIDELINES.md
fmaddenflx: Checked the actual `CamsTable`/`CamsTableCell` source — `data-cell` is not auto-derived from headers, it's a manual prop with no computed default. Added an explicit note to the guidelines and filed issue #50 to investigate auto-deriving it so developers don't have to keep cell and header text in sync manually.

docs/BRAND_GUIDELINES.md
fmaddenflx: Resolved together with the thread above — confirmed via code that data-cell must be set manually, so the "Known gaps" note about missing data-cell attributes isn't actually a contradiction, it's accurate. Filed issue #50 to consider auto-deriving data-cell going forward.

docs/BRAND_GUIDELINES.md
fmaddenflx: No action needed — just a placeholder note.

docs/BRAND_GUIDELINES.md
fmaddenflx: Corrected — clarified that `visibility: hidden` retains layout space (padding/margin) while `display: none` removes it entirely, and added a preference for not rendering the element at all when it should be hidden from everyone.

docs/BRAND_GUIDELINES.md
fmaddenflx: Qualified the claim and cross-referenced the existing Custom Icons (non-USWDS) section for the `RawSvgIcon` exception.

docs/BRAND_GUIDELINES.md
fmaddenflx: Added a link to the USWDS icon list (https://designsystem.digital.gov/components/icon/) in the Icon Library section.

docs/BRAND_GUIDELINES.md
fmaddenflx: Checked the `Icon` component — `focusable`/`decorative` are independent in code, but functionally `focusable=true` is never meaningful in CAMS since interactivity is always delegated to a wrapping button/link. Also confirmed USWDS's own icon markup always hardcodes `focusable="false"` and never exposes it as a variable. Filed issue #51 proposing removal of the prop.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — added a "Browser Page Title" subsection under Layout & Navigation → Required Page Elements documenting the `DocumentTitle` component and the fixed tab-title format (`{name} | U.S. Trustee Program - Case Management System (CAMS)`).

docs/BRAND_GUIDELINES.md
fmaddenflx: Fixed — removed the fixed 1024px minimum viewport claim. Now documents that CAMS has no fixed minimum viewport and remains usable down to 320px, with a mobile-first authoring note for SCSS.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — added a "Modal vs. Dialog Terminology" subsection explaining the WAI-ARIA distinction (dialog vs. modal vs. alertdialog) and clarifying that CAMS's `Modal` component is always a blocking modal dialog today, while noting a non-blocking dialog variant would be a valid future addition.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — added a bullet under Modal Content Guidelines: avoid long/complex/multi-step forms in a modal, quoting USWDS's own modal guidance on this.

docs/BRAND_GUIDELINES.md
fmaddenflx: Applied — the suggested bullet about buttons disabling on click (for visual feedback and to prevent duplicate submissions) was added.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — reworded to "not rendered (removed from the DOM entirely) or visibly disabled," and expanded the Implementation bullets to explain why not-rendering is preferred (avoids leaking control existence, closes off client-side bypass bugs).

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — added a "Theming Readiness" bullet under Color Usage Guidelines: avoid hardcoding colors outside theme.scss so future accessibility color themes can be added without reworking scattered hardcoded values.

docs/BRAND_GUIDELINES.md
fmaddenflx: Fixed — removed the unused "Eagle" color entry from Text Colors, and removed the "(Actually Used)" qualifier from all color subsection headings so unused colors aren't implied to be in use.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #45 (Disabled UI Element Styling Is Not Standardized) to track standardizing the disabled-state color.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #46 (Replace 3-Digit Hex Shorthand #ddd with Full #dddddd) to remove the shorthand ambiguity.

docs/BRAND_GUIDELINES.md
fmaddenflx: Fixed — removed the "(Actually Used)" qualifier from the Border & Divider Colors heading and split out a separate "(Not Actually Used)" heading, so #ddd is no longer implied to be in active use.

docs/BRAND_GUIDELINES.md
fmaddenflx: Fixed — clarified that only the current/active pagination button has a black background; the other (previous/next, non-current page) pagination buttons don't.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — removed the full hardcoded-hex-colors audit list and instead filed dedicated issues for the specific hardcoding problems it surfaced: issue #7 (RichTextEditor), #8 (ToggleButton), #10 (ComboBox), #11 (RichTextEditor palette), #12 (gray proliferation).

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #7 (RichTextEditor Link Color Nearly Identical to $primary But Hardcoded Differently) — proposes using `$primary` instead of the near-duplicate `#005ea6`.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #12 (Gray Proliferation — No Standardized Gray Scale) to consolidate the near-duplicate gray shades.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #46 (Replace 3-Digit Hex Shorthand #ddd with Full #dddddd) — same fix as the other #ddd thread above.

docs/BRAND_GUIDELINES.md
fmaddenflx: Addressed — removed the full hardcoded-hex-colors audit list that had this duplicate red entry; issue #9 (Error Hover Color Inconsistency) tracks consolidating `#8b0a0a`/`#8b0a03` to the single correct value.

docs/BRAND_GUIDELINES.md
fmaddenflx: Deleted — removed the entire hardcoded-hex-colors audit section, React logo entry included.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #47 (Verify USWDS SCSS Color Variables Are Properly Imported Before Redefining Them) to investigate whether the custom redefinitions are still necessary.

docs/BRAND_GUIDELINES.md
fmaddenflx: Already tracked in issue #2 (Dead Code — Unused Color Variables), which lists `$tableCellPadding`, `$tableBorderRadius`, and `$tableBorderCollapse` among the variables needing a use-or-remove decision.

docs/BRAND_GUIDELINES.md
fmaddenflx: Filed as issue #14 (Inconsistent Icon Button Pattern — IconButton Component Underused) to track migrating manual unstyled-button-plus-icon compositions over to `IconButton`.

docs/BRAND_GUIDELINES.md
fmaddenflx: Fixed — removed ", and user access" from the Purpose bullet.

docs/BRAND_GUIDELINES.md
fmaddenflx: Fixed — removed the "Use Title Case for document type and document titles" exception bullet.

PR #2622, opened 2026-07-03
CAMS-792 Add Case Status column to Trustee Case List table

mattstankey:
# Purpose

Display the case status (Open/Closed) on the Trustee Case List table so trustees can quickly see the current state of each case without navigating to the individual case.

# Major Changes

- Added Case Status column to the Trustee Case List table (component, data mapping, and test coverage)
- Fixed a `$lookup foreignField` bug with prefixed source in the MongoDB aggregate pipeline
- Adjusted Case Status column width (`max-width: 120px`) to prevent header text from wrapping or being clipped
- Raised the responsive breakpoint to `1200px` so the table switches to stacked/mobile view before columns compress enough to wrap

# Testing/Validation

Implemented using TDD. Test suite updated with the following quality improvements identified during spec review:

- Merged overlapping column header tests into a single ordered assertion
- Replaced fragile positional cell-index lookup with stable `data-cell` attribute selector
- Removed `CaseNumber` formatting assertions (wrong component boundary — `CaseNumber` owns that contract)
- Captured `onFilterChange` spy in `rerender` instead of inline `vi.fn()`
- Added test documenting that any non-CLOSED `caseStatus` renders as "Open"

All 23 tests pass.

# Notes

The `Case Status` column header was wrapping at the previous `max-width: 90px`. Increasing to `120px` gives the header enough room. The responsive breakpoint was raised to `1200px` (from `880px`) because the sidebar + column minimums require more horizontal space before the table can lay out without compression.

# Definition of Done:

## Summary by Sourcery

Add case status information to trustee case list data and UI and align related data pipelines, fixtures, and tests.

New Features:
- Expose case status on trustee case list items and render a Case Status column in the trustee case list table.

Bug Fixes:
- Correct MongoDB aggregate lookup foreignField handling when the foreign field name includes a prefixed source alias.

Enhancements:
- Derive case status from closed and reopened dates in the trustee case appointments repository without exposing those dates in the API response.
- Adjust trustee case list table responsive layout and column sizing to accommodate the new Case Status column and prevent header wrapping.
- Refine trustee-case-related seed scenarios and expectations to omit unassignedOn for active appointments and update integration test configuration for local trustee-case-filter runs.
- Improve trustee case list tests for header ordering, stable cell selection, and explicit case status rendering semantics.

Tests:
- Add unit tests for case status derivation logic in the trustee case appointments repository and for Case Status rendering and behavior in the trustee case list UI.
- Update controller, use case, and scenario tests to account for the new caseStatus field in trustee case list items.


mattstankey:
Addressed @sourcery-ai suggestion #1: added a guard in `toMongoLookup` that throws a descriptive error if stripping the alias prefix from `join.foreign.name` produces an empty string, along with a comment explaining why the prefix is stripped. Added a unit test covering this case.

Skipping suggestion #2: the `1200px`/`1201px` breakpoints are specific to this table only, so centralizing them in `device-screens` tokens would add indirection without real benefit.

btposey (approved):

PR #2624, opened 2026-07-03
CAMS-768 Fix reply-to email for change notifications

mattstankey:
# Bug

When a user saves changes to a trustee's contact information, the automated notification email was not setting a Reply-To address. Replies went to the ACS do-not-reply sender address instead of the user who made the change.

# Purpose

Ensure that recipients of trustee change notification emails can reply directly to the USTP staff member who made the change.

# Major Changes

* Added `email` field to the combined user object in `getPrivilegedIdentityUser` — the field was being received from Okta but silently dropped when building the session user
* Added `email` scope to the Okta auth request in `OktaProvider` so the userinfo endpoint returns the email claim

# Testing/Validation

Tested manually end-to-end: saved trustee contact info changes while authenticated via Okta, confirmed the notification email arrived with a Reply-To header matching the authenticated user's email address.

# Notes

Root cause was a two-layer drop: (1) the `email` scope was not being requested from Okta so the claim was absent from the userinfo response, and (2) even when present, `getPrivilegedIdentityUser` was building the `combined` user object without forwarding the `email` field from `idpUser`. Both needed to be fixed for Reply-To to populate.

The session is cached in Cosmos DB keyed by access token, so testing required clearing the `user-session-cache` collection after deploying the backend fix to force a fresh session build.

# Definition of Done:

## Summary by Sourcery

Fix trustee change notification emails to include the author’s Reply-To address, expand notification routing to support multiple recipients per category, and refine contact change rendering and admin UI headings.

New Features:
- Allow notification routing records to define multiple recipient email addresses per routing key and manage them via the admin UI.
- Include the authenticated user’s email as the Reply-To address on trustee change notifications sourced from Okta.
- Provide more granular chapter-specific oversight routing definitions for trustee notifications.

Bug Fixes:
- Ensure trustee change notifications use the USTP staff member’s email in the Reply-To header instead of the no-reply sender address.
- Prevent trustee update flows from failing when notification dispatch errors by handling send failures asynchronously and logging them.

Enhancements:
- Improve formatting of public and internal contact changes in notification emails, including stacked, labeled fields and clearer address rendering.
- Format trustee change timestamps in notifications as human-readable UTC date-times in both HTML and plaintext versions.
- Normalize trustee contact form payloads (e.g., address2, website, companyName, phone extension) for consistent backend representation.
- Adjust admin screens (banks, bankruptcy software, notification routing) to use visible, semantically correct headings and tighter spacing.
- Refine notification recipient de-duplication to work across multiple addresses per routing record.

Tests:
- Update and expand unit and UI tests across notification routing, trustee change notifications, contact forms, and templates to cover multi-recipient routing, Reply-To behavior, new formatting, and validation rules.
- Refresh seed data, mocks, and API tests to align with the new notification routing model and chapter-specific oversight records.


backend/lib/controllers/admin/notification-routing.controller.ts
```diff
@@ -66,8 +66,8 @@ export class NotificationRoutingController implements CamsController {
     await this.repository.createRoutingAuditRecord({
       documentType: 'AUDIT_NOTIFICATION_ROUTING',
       routingRecordId: routingId,
-      before: existing?.recipientAddress ?? '',
-      after: input.recipientAddress,
+      before: existing?.recipientAddresses.join(', ') ?? '',
+      after: input.recipientAddresses.join(', '),
```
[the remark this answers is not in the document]
mattstankey: ✅ Fixed — changed to `existing?.recipientAddresses?.join(', ') ?? ''` with optional chaining on both `existing` and `recipientAddresses`.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -117,18 +133,30 @@ export function NotificationRouting() {
...
-                value={emails[def.id] ?? ''}
-                onChange={(e) => handleEmailChange(def.id, e.target.value)}
-              />
+              <label className="usa-label">{def.displayName}</label>
+              {(emails[def.id] ?? ['']).map((addr, index) => (
+                <input
+                  key={index}
+                  className="usa-input"
+                  style={{ marginTop: index > 0 ? '20px' : undefined }}
+                  id={index === 0 ? `routing-email-${def.id}` : `routing-email-${def.id}-${index}`}
```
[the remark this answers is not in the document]
mattstankey: ✅ Fixed — added `htmlFor={`routing-email-${def.id}`}` to the label so it points to the first input. Also moved the inline `marginTop` styles to SCSS classes (`.routing-email-additional` and `.routing-add-email-button`) in `AdminScreen.scss`.

mattstankey:
All four Sourcery suggestions addressed in commit 7e06b7ef2:

- **`notification-routing.controller.ts`** — added `?.` on `recipientAddresses` to guard the `join` call when `existing` is undefined
- **`NotificationRouting.tsx`** — added `htmlFor` to the label pointing to `routing-email-${def.id}`
- **`NotificationRouting.tsx` + `AdminScreen.scss`** — replaced inline `marginTop` styles with `.routing-email-additional` and `.routing-add-email-button` SCSS classes
- **`OktaProvider.tsx`** — now merges any env-configured scopes with the required `['openid', 'profile', 'email']` instead of overwriting

jalba-flexion (approved):

mattstankey:
@sourcery-ai review

jalba-flexion (approved):

bityogi (approved):

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -517,7 +516,8 @@ export class TrusteesUseCase {
```
cruzjone-flexion, line 518: **Is fire-and-forget safe here for the notification dispatch?**

The dispatch changed from awaited to fire-and-forget. In Azure Functions, the runtime can terminate the invocation after the HTTP response is sent — unawaited async work may be killed before it completes, silently dropping notifications.

```typescript
// Before
await this.dispatchChangeNotification(context, changeSet, trusteeId);

// After
this.dispatchChangeNotification(context, changeSet, trusteeId).catch(() => {});
```

Was this a deliberate trade-off to avoid latency on the save, or is there something in the runtime that guarantees in-flight async work completes? The tests were updated with `vi.waitFor` precisely because the timing is now non-deterministic — wondering if the same race could happen in prod.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -40,17 +41,28 @@ export function NotificationRouting() {
...
     setSaveSuccess(false);
   }
 
   async function handleSave() {
     const errors: string[] = [];
     for (const def of NOTIFICATION_ROUTING_DEFINITIONS) {
-      const email = emails[def.id]?.trim();
-      if (email && !EMAIL_REGEX.test(email)) {
-        errors.push(`${def.displayName}: invalid email address.`);
+      const addrs = (emails[def.id] ?? ['']).map((a) => a.trim()).filter(Boolean);
```
cruzjone-flexion, line 68: `key={index}` on these dynamic inputs will cause React to bleed state (focus, cursor position) into the wrong input when one is removed from the middle of the list.

```tsx
{(emails[def.id] ?? ['']).map((addr, index) => (
  <input
    key={index}
    ...
  />
))}
```

Low impact today since there's no remove UI, but the moment removal is added this will cause bugs. Consider a stable key — e.g. a UUID generated when the address is added, or the address value itself as a fallback.

user-interface/src/login/providers/okta/OktaProvider.tsx
```diff
@@ -13,7 +13,11 @@ export function OktaProvider(props: OktaProviderProps) {
     const config = getLoginConfiguration<EnvLoginConfig>();
     const { protocol, host } = window.location;
     config.redirectUri = `${protocol}//${host}${LOGIN_CONTINUE_PATH}`;
-    const oktaAuth = new OktaAuth(config);
+    const requiredScopes = ['openid', 'profile', 'email'];
+    const configuredScopes = (config as Record<string, unknown>).scopes;
+    const baseScopes = Array.isArray(configuredScopes) ? (configuredScopes as string[]) : [];
```
cruzjone-flexion: The double cast here is a smell — `config` is cast to `Record<string, unknown>` to extract `scopes`, then the value is cast again to `string[]` without verifying the elements are strings.

```typescript
const configuredScopes = (config as Record<string, unknown>).scopes;
const baseScopes = Array.isArray(configuredScopes) ? (configuredScopes as string[]) : [];
```

If `EnvLoginConfig` already types `scopes` as `string[]`, use it directly. If not, add a `.filter((s): s is string => typeof s === 'string')` after the `Array.isArray` check to eliminate the unsafe cast and handle malformed config values explicitly.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -40,17 +41,28 @@ export function NotificationRouting() {
...
-    setEmails((prev) => ({ ...prev, [id]: value }));
+  function handleEmailChange(defId: string, index: number, value: string) {
+    setEmails((prev) => {
+      const updated = [...(prev[defId] ?? [''])];
+      updated[index] = value;
+      return { ...prev, [defId]: updated };
+    });
+    setSaveSuccess(false);
+  }
+
```
cruzjone-flexion, line 57: If a user deletes all addresses and clicks save, `addrs` will be empty and the API call fires with `recipientAddresses: []` — the backend rejects this with a `BadRequestError`, but the frontend surfaces it as a generic save error rather than a helpful validation message.

```typescript
const addrs = (emails[def.id] ?? ['']).map((a) => a.trim()).filter(Boolean);
// ...
if (!unchanged) {
  const response = await Api2.updateNotificationRouting(def.id, {
    recipientAddresses: addrs,  // can be []
  });
```

The empty-array case is easy to catch in the same validation loop that already checks for invalid email format — push an error like ``${def.displayName}: at least one email address is required.`` before reaching the API call.

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -529,40 +529,50 @@ export class TrusteesUseCase {
...
+        after: publicDiff.after,
         category: 'profile',
         section: 'appointment',
+        stackValues: true,
       });
     }
 
-    if (!deepEqual(before.internal, after.internal)) {
+    const beforeInternalNorm = normalizeForUndefined(before.internal);
+    const afterInternalNorm = normalizeForUndefined(after.internal);
```
cruzjone-flexion, line 548: Replacing `deepEqual` with `formatContactInfo` string comparison for change detection means any field not included in `formatContactInfo` (e.g. a future addition to `ContactInformation`) will be silently ignored — no history record written, no notification sent.

```typescript
const publicChanged = formatContactInfo(before.public) !== formatContactInfo(after.public);
if (publicChanged) { ... }
```

Either keep `deepEqual` as the change gate and use `formatChangedContactInfo` only for building the display diff, or add a comment here making it explicit that `formatContactInfo` is the canonical change-detection surface and new fields must be added there to participate in notifications.

backend/lib/use-cases/trustees/trustees.ts
mattstankey: ✅ Deliberate trade-off to keep save latency low — fire-and-forget so the HTTP response returns as soon as the DB write completes. `dispatchChangeNotification` already has its own `try/catch` that logs and traces failures, so the outer `.catch` was purely defensive. The Azure termination risk is accepted here (notifications are informational, not transactional). Updated the outer `.catch` to log via `context.logger.error` rather than silently swallowing, so any unexpected outer rejection is observable.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
mattstankey: ✅ Fixed — added a parallel `emailKeys` state that stores a `crypto.randomUUID()` per entry, assigned on load and when an address is added. The render now uses `emailKeys[def.id]?.[index]` as `key` instead of `index`, giving each input a stable identity independent of its position.

user-interface/src/login/providers/okta/OktaProvider.tsx
mattstankey: ✅ Fixed — `EnvLoginConfig` does not type `scopes`, so the cast to `Record<string, unknown>` is still needed to extract it, but replaced `configuredScopes as string[]` with `.filter((s): s is string => typeof s === 'string')` to eliminate the unsafe cast and handle any malformed config values explicitly.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
mattstankey: ✅ Fixed — added a validation check in `handleSave` that triggers when a routing record already exists in the backend and all addresses have been cleared, surfacing `"${def.displayName}: at least one email address is required."` before the API call. New/unconfigured defs are skipped silently (no record in state = not yet set up). Added a test covering the clear-existing-record case.

backend/lib/use-cases/trustees/trustees.ts
mattstankey: Good catch. Added a comment at the `publicChanged` detection site making it explicit that `formatContactInfo` is the canonical change-detection surface and any new `ContactInformation` field must be added there to participate in history and notifications. The `internalChanged` line has a cross-reference to the same note.

cruzjone-flexion (approved):

PR #2628, opened 2026-07-06
CAMS-768 Improve notification routing UX and validation

cruzjone-flexion:
# Bug

The notification routing admin panel had three UX issues:
1. Email validation errors appeared as a single block above all sections instead of inline under the failing input
2. Attempting to save with all emails cleared in a section returned an error — there was no way to remove all recipients from a routing slot
3. After saving, blank inputs were not removed from the UI, leaving stale empty fields visible

# Purpose

Make the notification routing admin form more responsive and forgiving: surface errors where they occur, allow clearing a section's recipients, and reflect saved state accurately after every save.

# Major Changes

- Email validation errors now display inline under each failing input with the message "Must be a valid email address" and apply `usa-input--error` styling to the offending field
- Removed the frontend guard that blocked saving an empty recipient list; empty inputs are filtered out before the API call
- Removed the backend `recipientAddresses must be a non-empty array` validation — an empty array is now accepted as a valid "clear all" operation
- After a successful save, `loadData()` is called to re-render from server state, ensuring blank inputs that were filtered out disappear from the UI
- Updated and added tests across the frontend component and backend controller to match the new behavior and cover previously untested logic paths

# Testing/Validation

Implemented using TDD. All 3,242 unit tests pass across common, backend, and user-interface workspaces. Backend coverage held at 95.7%. Manually verified all three scenarios in the browser.

# Notes

The backend change is intentional: an empty `recipientAddresses` array now means "no recipients configured for this routing slot," which is a valid state. The frontend still filters out blank input strings before calling the API, so an empty input and no input are treated identically on save.

New backend tests cover the previously untested `nonStrings` guard (non-string values in the array), a non-array typed value, the audit `before` field when a prior record exists, and the response body shape on a successful PUT.

# Definition of Done:

## Summary by Sourcery

Improve notification routing admin behavior by making email validation inline, allowing sections to be cleared, and syncing the UI with saved backend state.

New Features:
- Display per-input email validation errors inline in the notification routing admin form with appropriate error styling.
- Allow saving a routing section with an empty recipient list to represent a cleared configuration state.
- Reload notification routing data from the server after successful saves so the UI reflects the persisted addresses.

Bug Fixes:
- Resolve the inability to save when all email addresses in a section are cleared.
- Eliminate stale blank email input fields that previously remained visible after saving.
- Ensure the backend accepts empty recipient address arrays while still rejecting non-array and non-string values.

Enhancements:
- Return the updated routing record in the controller response body for successful PUT requests.
- Capture prior recipient addresses in the routing audit "before" field when updating an existing record.

Tests:
- Expand frontend tests to cover inline email validation, clearing all addresses for a section, and reloading data after a successful save.
- Add backend tests for empty recipient arrays, non-array request values, non-string entries in recipient lists, mixed valid and invalid recipient values, and audit behavior on updates.

### DIFF
--- a/user-interface/src/admin/notification-routing/NotificationRouting.tsx
+++ b/user-interface/src/admin/notification-routing/NotificationRouting.tsx
@@ -16,7 +16,8 @@ export function NotificationRouting() {
   const [loadError, setLoadError] = useState<string | null>(null);
   const [emails, setEmails] = useState<Record<string, string[]>>({});
   const [emailKeys, setEmailKeys] = useState<Record<string, string[]>>({});
-  const [formErrors, setFormErrors] = useState<string[]>([]);
+  const [emailErrors, setEmailErrors] = useState<Record<string, (string | null)[]>>({});
+  const [apiError, setApiError] = useState<string | null>(null);
   const [saveSuccess, setSaveSuccess] = useState(false);
 
   async function loadData() {
@@ -35,6 +36,7 @@ export function NotificationRouting() {
       }
       setEmails(emailMap);
       setEmailKeys(keyMap);
+      setEmailErrors({});
       setLoadError(null);
     } catch (error) {
       setLoadError((error as Error).message);
@@ -52,6 +54,11 @@ export function NotificationRouting() {
       updated[index] = value;
       return { ...prev, [defId]: updated };
     });
+    setEmailErrors((prev) => {
+      const defErrors = [...(prev[defId] ?? [])];
+      defErrors[index] = null;
+      return { ...prev, [defId]: defErrors };
+    });
     setSaveSuccess(false);
   }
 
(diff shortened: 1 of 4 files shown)


user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -91,20 +100,15 @@ export function NotificationRouting() {
...
-          const updated = (response as { data: NotificationRoutingRecord }).data;
-          setRecords((prev) => {
-            const exists = prev.some((r) => r.id === updated.id);
-            if (exists) return prev.map((r) => (r.id === updated.id ? updated : r));
-            return [...prev, updated];
-          });
         }
       }
       setSaveSuccess(true);
+      await loadData();
```
mattstankey: Partial multi-section save failure leaves stale/inconsistent UI state.

If a user edits recipients for two or more routing definitions in one save action, the `for` loop inside this `try` block awaits `Api2.updateNotificationRouting` sequentially. If an earlier call (e.g. `chapter-7-oversight`) succeeds and a later call (e.g. `chapter-11-oversight`) throws, the exception propagates immediately — skipping `setSaveSuccess(true)` and `await loadData()` on line 108-109 — and lands in the `catch` block below, which only sets `apiError`.

The earlier update is durably persisted server-side, but the UI never refreshes to reflect it, and `records`/`emails` state stays stale relative to the server. The user only sees a generic error with no indication that part of the save actually succeeded. The previous implementation called `setRecords` inline after each successful update, so a partial success was at least reflected locally — this is a regression relative to that behavior.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -148,17 +150,28 @@ export function NotificationRouting() {
...
+                    id={
+                      index === 0 ? `routing-email-${def.id}` : `routing-email-${def.id}-${index}`
+                    }
+                    data-testid={
+                      index === 0 ? `routing-email-${def.id}` : `routing-email-${def.id}-${index}`
+                    }
+                    type="email"
+                    value={addr}
+                    onChange={(e) => handleEmailChange(def.id, index, e.target.value)}
+                  />
```
mattstankey: New inline error markup breaks screen-reader association — deviates from established CAMS convention.

The error `<span>` here has no `id`, and the `<input>` above has no `aria-invalid` or `aria-describedby`/`aria-errormessage` pointing at it. The shared `Input.tsx` component (and `TextArea.tsx`, `DatePicker.tsx`, `ComboBox.tsx`, etc.) consistently wire up `aria-invalid` + `aria-describedby`/`aria-errormessage` to an `id`d error element for exactly this scenario.

Since this form renders a raw `<input>`/`<span>` pair instead of reusing the shared component, a screen-reader user editing an invalid email address gets no announcement that the field is invalid or why. Given CLAUDE.md's "WCAG 2.1 AA compliance required" / "Accessibility is a requirement, not optional" mandate, this looks like a real accessibility regression worth fixing before merge — either reuse the shared `Input` component or add `id` on the error span plus `aria-invalid`/`aria-describedby` on the input.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
bityogi: Fixed in a9189157b. The save loop no longer aborts on the first failure — each section attempt is now wrapped individually, `loadData()` runs after all sections have been attempted (so successful saves are reflected regardless of later failures), and the error message now names the specific section(s) that failed (e.g. "Failed to save: Chapter 11 Oversight."). Added a test covering the partial-failure case.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
bityogi: Already fixed in ff12394d2 (pushed shortly after this comment). The error `<span>` now has an `id`, and the `<input>` has both `aria-invalid` and `aria-describedby` pointing at that error element id, matching the pattern used elsewhere in CAMS. I kept the raw input/span markup here rather than switching to the shared `Input` component, since this form manages per-row dynamic email lists that dont map cleanly onto that components API — happy to revisit if you feel strongly about reuse.

jalba-flexion (approved):

PR #2630, opened 2026-07-06
HOT FIX - Trustees, Trustee case list performance and data integrity

btposey:
# Bug

The Trustee Case List screen was hitting 6-second API timeouts and ~1490 RU/request for trustees with 7000+ appointments. Profiling confirmed the root cause: the `getCasesForTrustee` aggregate pipeline joined all appointment documents to the `cases` collection before pagination — 7598 cross-collection reads per page request. Separately, the dual-write pattern in `TrusteeCaseAppointmentsMongoRepository` silently swallowed secondary partition write failures, causing document count divergence between `case-trustee-appointments` and `trustee-case-appointments` after the production migration run.

# Purpose

Restore Trustee Case List to sub-500ms response times and establish a reliable, self-healing data integrity model for the two-partition appointment document architecture.

# Major Changes

- **Pipeline restructured**: `getCasesForTrustee` now filters, sorts, and paginates on the appointment collection before a ~25-document enrichment `$lookup` — eliminating the 7000+ doc join. All filters (`caseStatus`, `chapter`, `courtDivisionCode`, date ranges) apply pre-paginate against denormalized appointment fields.

- **Denormalized case fields on appointments**: `dateFiled`, `chapter`, `courtDivisionCode`, and `caseStatus` (computed via `isCaseClosed`) are written to appointment documents at migration time. Source: ACMS CMMDB column projections and a LEFT JOIN to CMMKE for OCO (case reopening) events — no Cosmos lookup required.

- **Dual-write hard-fail**: Secondary partition writes now throw on failure instead of swallowing errors. Natural key for upsert simplified to `(documentType, caseId, trusteeId, assignedOn)` — removing `source` so ACMS and DXTR writes for the same appointment merge into a single document.

- **sync-cases integration**: `updateCaseFields()` called after `syncDxtrCase()` to keep denormalized fields current and transition `source` ownership to `'dxtr'`, signalling the migration backfill to skip those documents.

- **Self-healing migration**: `migrate-case-appointments` extended with a self-polling reindex phase (60s visibility delay, creates compound index, drops old) and a `heal` intent that performs per-trustee partition parity repair and legacy field detection.

- **Cosmos DB compound index**: `(trusteeId, unassignedOn, dateFiled, caseStatus)` declared in Bicep; applied at runtime by the reindex intent since Cosmos ignores Bicep index changes on existing collections.

- **Bug fixes caught in final review**: WHERE clause appearing after GROUP BY in both CMMAP queries (T-SQL syntax error preventing all appointment ingestion); `courtDivisionCode` zero-padding; `updateCaseFields` second failure swallowed with no rethrow.

- **Supporting improvements**: 429 log noise suppressed (INFO on first retry only); `SAFE_THRESHOLD_MS` extracted to shared constant at 56 minutes; `WRITE_BATCH_SIZE` reduced to 100 to accommodate larger per-record payload.

# Testing/Validation

Implemented using TDD throughout. 3228 unit tests passing across `common` and `backend` workspaces.

Integration test harnesses updated:
- `migrate-case-appointments`: closed case fixture added, `closedDate`/`caseStatus='CLOSED'`/`reopenedDate` from CMMKE all asserted end-to-end. `run-upgrade` verifies old-shape docs updated in-place by re-migration. `run-heal` verifies partition divergence repaired by the heal intent.
- `trustee-case-filter`: all 14 pipeline tests updated with denormalized fields on appointment fixtures; slot 511 (no `dateFiled`) confirms the `$exists true` guard excludes un-migrated docs.

# Notes

The migration window (during which un-migrated docs are excluded from results) is expected to last 24–48 hours on production data volumes. Users are informed out-of-band. The `dateFiled $exists true` guard self-heals — as the migration progresses, cases appear in trustee lists without any deployment change.

The `source` field on appointment documents now encodes ownership: `'acms'` = safe to overwrite by migration, `'dxtr'` = owned by sync-cases, skip. This eliminates the need for timestamp comparison logic to prevent stale overwrites.

`deleteAllBySource` removed — the migration is fully idempotent via natural key upserts; a fresh start resets the cursor to null and overwrites existing documents.

# Definition of Done:


mattstankey:
## Multi-agent review summary (orchestrated: pragmatic/YAGNI, architecture/OeSA, DevSecOps, SRE)

The core fix (aggregation pipeline restructure + ACMS SQL subquery restructure) is sound and well-tested. The following issues should be resolved before merge:

### 1. PR description doesn't match the code
The description says "the `source` field on appointment documents now encodes ownership," but the diff removes `source` entirely with no equivalent replacement for the reconciliation strategy it describes. Please confirm the working tree reflects the intended final state and update the description — if `source`-based ownership really was descoped, the data-consistency story for the 3 independent writers of denormalized fields (ACMS migration, DXTR sync, `heal`) needs a documented replacement.

### 2. Dual-write hard-fail has unhandled blast radius outside the migration path
`upsert`/`updateCaseAppointment` now throw on secondary-partition write failure (correct fix), but only `migrate-case-appointments.ts` safely downgrades this to a per-record failure. Two other call sites don't:
- `sync-trustee-appointments.ts`: the exception has no `mismatchReason`, so it falls into the generic "unclassified sync error" branch and gets routed to DLQ indistinguishably from a real trustee-matching failure.
- `trustee-match-verification.use-case.ts`: a user-initiated "approve verification" action can now surface an unhandled 500 with no indication of partial write state (case partition may have already succeeded).

Please wrap these call sites with error handling that distinguishes "partition write failed" from other failure modes, and/or add a distinguishable log/metric.

### 3. Migration bookkeeping fields leak across the public API boundary
`acmsProfessionalId` and `reason` were added to the shared `CaseAppointment` type for internal migration diagnostics, but `case-trustee-appointment.controller.ts` returns `CaseAppointment` unfiltered to the frontend. This is a Good-Fences violation — recommend a narrower response DTO, or move these fields to an internal-only type that doesn't cross into `common/`.

### 4. `updateCaseFields` can silently go stale with zero telemetry
In `export-and-load-case.ts`, after 2 failed attempts `updateAppointmentFieldsWithRetry` logs a `warn` and gives up — no metric, no alert, no dead-letter path. Since the new fast-path `getCasesForTrustee` query sorts/filters directly on these denormalized fields, a silent failure here means a case can show stale `caseStatus`/`dateFiled` in the Trustee Case List indefinitely with no way to detect it. Recommend adding a metric on the final give-up path at minimum.

### 5. Scope creep: `migrate-trustees` integration harness is unrelated to this PR
`test/integration/migrate-trustees/` (756-line harness + 4 support files) tests the address-normalization fix from CAMS-794, which is already merged and unrelated to this PR's pagination/dual-write scope. Recommend splitting this into its own PR so this one stays reviewable.

### Also: `as any` cast bypasses the repository interface
`migrate-case-appointments.ts`'s `readHealState`/`updateHealState` call `factory.getRuntimeStateRepository(context) as any` to smuggle `HealCaseAppointmentsState` past the type system, because `'HEAL_CASE_APPOINTMENTS_STATE'` was never added to `RuntimeStateDocumentType` in `gateways.types.ts`. Recommend widening that union and calling `getRuntimeStateRepository<HealCaseAppointmentsState>(context)` directly instead of casting — removes two `eslint-disable` escapes along with it.

---
*Posted via multi-agent orchestrated review (4 parallel specialist passes). Happy to expand on any finding.*

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -147,46 +139,63 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
-        );
+      const prePaginateMatch = toMongoQuery(this.buildPrePaginateMatch(trusteeId, predicate));
+
+      const mongoAggregate = [
+        { $match: prePaginateMatch },
+        { $sort: { dateFiled: -1, caseId: 1 } },
+        {
+          $facet: {
+            metadata: [{ $count: 'total' }],
+            data: [
```
jamesobrooks, line 150: **[High] Pagination `$skip`/`$limit` runs before the post-lookup filter, while `metadata.total` counts pre-filter.**

In this `$facet`, `data` applies `$skip`/`$limit` (lines 151-152) *before* the `$lookup`/`$unwind`/`$match` that drops cases with no `SYNCED_CASE` doc or a set `movedToCaseId`. Meanwhile `metadata: [{ $count: 'total' }]` counts the stream immediately after the pre-paginate `$match`/`$sort`, before that same drop happens. Net effect: a dropped row's page slot is permanently lost (it won't reappear on a later page) while `total` still counts it — wrong pagination/totals in the exact feature this PR fixes.

Confirmed against the integration fixture itself: `test/integration/trustee-case-filter/scripts/run-tests.ts` hand-computes an expected total of 508 for its no-filter case, but tracing the actual pipeline against that fixture's seed data (slot 508 has `movedToCaseId` set but passes the pre-paginate `$match`) shows the pipeline would produce 509 — the fixture's own expected value doesn't match what the shipped code computes.

Suggest moving the `movedToCaseId`/`SYNCED_CASE` guard into the pre-paginate `$match` (denormalizing whatever signal is needed onto the appointment doc, consistent with this PR's overall denormalization approach) so both `metadata` and `data` see the same filtered set.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -133,6 +124,7 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
         doc('documentType').equals('CASE_APPOINTMENT'),
         doc('caseId').equals(caseId),
         doc('unassignedOn').notExists(),
+        doc('trusteeId').notEqual(SENTINEL_TRUSTEE_ID),
```
jamesobrooks, line 127: **[High] Sentinel exclusion here breaks `applyResolvedTrustee`'s close-then-create sequence.**

`getActiveByCaseId` now excludes `SENTINEL_TRUSTEE_ID` docs. But `sync-trustee-appointments.ts:175` calls exactly this method to find the prior active appointment to soft-close before creating the new one for a resolved trustee. If the only active appointment on a case is a sentinel (unmapped-trustee, written by the ACMS migration path), this returns `null`, the soft-close is skipped, and `upsert` creates a *second* active appointment — the sentinel is orphaned rather than closed. Downstream, `case-trustee-appointment.use-case.ts:34` (`all.find((a) => !a.unassignedOn)`) picks whichever unclosed doc Mongo returns first, which can nondeterministically surface the orphaned sentinel as "current" instead of the real trustee.

This is reachable: ACMS-migrated cases with an unmapped trustee later resolved via DXTR name-matching (`sync-trustee-appointments.ts`) hit exactly this path, and no test exercises "existing active appointment is a sentinel."

Suggest `applyResolvedTrustee` use a sentinel-inclusive lookup to find-and-close a prior sentinel appointment before upserting.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -22,13 +28,54 @@ import factory from '../../../lib/factory';
...
+})();
 
-// Records per write queue message. Azure Storage Queue limit is 64KB base64-encoded,
-// which is ~48KB raw. ResolvedAcmsRecord serializes to ~200 bytes worst-case —
-// 150 records ≈ 30KB raw / 40KB encoded, leaving 8KB headroom.
-const WRITE_BATCH_SIZE = 150;
+export function isAcmsTimeoutError(error: unknown): boolean {
+  if (!(error instanceof Error)) return false;
+  const err = error as { originalError?: { code?: string } };
+  return err.originalError?.code === 'ETIMEOUT';
```
jamesobrooks: **[High] `.code` is unreachable here — the ACMS timeout retry path never fires, and neither does the fallback.**

`error.originalError.code` is checked, but tracing the wrap chain: `abstract-mssql-client.ts`'s `executeQuery` catch calls `getCamsError(unknownError, moduleName)` with no `message` arg, producing an `UnknownError` whose `CamsError` constructor sets `this.originalError = util.inspect(options.originalError)` — a **string**, not the original error object — and `this.message = options.message || 'Unknown CAMS Error'`, discarding the real driver message. So on a genuine ACMS `ETIMEOUT`:
- `originalError.originalError` is a string → `.code` is `undefined` → this check is always `false`.
- The fallback at line 398/403 (`errMsg.includes('Timeout')`) also always `false`, since `errMsg` is the generic `'Unknown Error'`, not the real driver message.

Both paths to detecting the timeout that this PR was written to retry are dead in production. The unit tests (`migrate-case-appointments.test.ts`) mock `readPage` to reject with a hand-crafted `{ originalError: { code: 'ETIMEOUT' } }` shape that bypasses the real `CamsError`/`UnknownError` wrapping entirely, so they can't catch this.

Fix needs either: preserve the real error/message through `getCamsError`/`CamsError` construction, or inspect the raw mssql error for the timeout signature before it gets wrapped (inside `abstract-mssql-client.ts` / `acms.gateway.ts`), plus a test using the real wrapped-error shape.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -308,11 +329,14 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
         .adapter<CaseAppointmentDocument>()
         .replaceOne(query, updatedDocument);
     } catch (secondaryError) {
-      console.error(
+      this.context.logger.error(
         MODULE_NAME,
         `Dual-write update to trustee partition failed for appointment ${appointment.id}:`,
         secondaryError,
       );
+      throw getCamsErrorWithStack(secondaryError, MODULE_NAME, {
```
jamesobrooks, line 353: **[High] This hard-fail, combined with `sync-trustee-appointments.ts`'s close-then-create sequence, can leave a case with zero active trustee appointments.**

`updateCaseAppointment`'s secondary-partition write now throws here instead of swallowing the error (previous behavior). `applyResolvedTrustee` in `sync-trustee-appointments.ts` calls this method to soft-close the prior appointment (line ~182) *before* creating the new one via `upsert` (line ~218), with no compensating transaction between the two calls. If this throw fires during the close step, the primary (case) partition has already succeeded in closing the old appointment, but the new one is never created — leaving the case with **no active trustee appointment**.

The resulting error has no `mismatchReason`, so `processAppointments`'s catch (in `sync-trustee-appointments.ts`) routes it to the generic DLQ, which has no registered consumer (confirmed: only `HANDLE_START`/`HANDLE_PAGE`/`HANDLE_PAGE_POISON` are registered as queue triggers in the function app). The new `heal()` intent doesn't cover this gap either — it only detects "active in case partition, missing in trustee partition"; here the case-partition doc is already closed and no new doc exists, so heal's active-doc filter excludes it entirely.

Notably, this PR's own `dataflows-rate-limit.ts` treats Cosmos 429s during high-load migration as expected/retriable — but that retry-with-backoff mechanism can't engage here because the error is caught and routed to the DLQ *inside* `processAppointments`, never rethrown to the page-level handler that has the retry wrapper.

Suggest reordering `applyResolvedTrustee` to create-before-close (fails safe: over-assigned/detectable rather than silently unassigned), or wrapping the sequence with a bounded retry on secondary-partition-only failures.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -349,28 +382,246 @@ async function incrementMetric(
...
+  context: ApplicationContext,
+): Promise<{ status: 'ready' | 'needs-polling' }> {
+  const repo = factory.getTrusteeCaseAppointmentsRepository(context);
+
+  const newIndexExists = await repo.checkIndexExists(NEW_COMPOUND_INDEX_NAME);
+
+  if (!newIndexExists) {
+    // Attempt to create the index; ignore "already building" errors
+    try {
+      await repo.createCompoundIndex();
```
jamesobrooks, line 438: **[High] `reindexPhase` polls forever on persistent index-build failure, and structurally evades the poison-queue safety net.**

Any error from `createCompoundIndex()` — not just "build in progress" — is caught, logged at `INFO` only (line 447), and converted to `{ status: 'needs-polling' }` unconditionally. There's no attempt counter or elapsed-time guard anywhere in this function. The caller (`function-apps/dataflows/migrations/migrate-case-appointments.ts:307-333`) re-enqueues a **brand-new** message (`sendMessage(JSON.stringify({}), 60)`) each time rather than nacking the current one — so Azure's `maxDequeueCount: 5` poison-queue mechanism never engages, since each retry's message starts at `dequeueCount: 0`.

The same file already has the correct pattern for this class of problem: the ACMS-timeout path tracks `timeoutRetryCount` against `ACMS_TIMEOUT_RETRY_LIMIT` and routes to DLQ on exhaustion. This path doesn't reuse it. If `createCompoundIndex()` fails persistently (RU throttling on the exact collection this PR is trying to relieve, or an unsupported index shape on the Cosmos tier), the migration silently stalls at Phase 1 forever with no alert.

Suggest tracking attempt count/elapsed time and escalating to WARN/ERROR + DLQ after a bound, matching the existing ACMS-timeout pattern in this same file.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -6,12 +6,18 @@ import factory from '../../factory';
...
   AcmsCaseAppointmentRawRecord,
   formatCaseId,
   formatAcmsProfessionalId,
 } from '../gateways.types';
-import { CaseAppointmentInput } from '@common/cams/trustee-appointments';
+import { CaseAppointment, CaseAppointmentInput } from '@common/cams/trustee-appointments';
+import { SAFE_THRESHOLD_MS } from './migrate-case-appointments-constants';
+
+// Name of the compound index added by this migration (replacing the old 2-field index)
+const NEW_COMPOUND_INDEX_NAME = 'trusteeId_1_unassignedOn_1_dateFiled_1_caseStatus_1';
```
jamesobrooks, line 19: **[High] Dependency Rule: hardcoded MongoDB auto-generated index name in the use case, plus index-admin methods added to the shared repository interface.**

This string is Mongo's driver-derived index-naming convention (`field_direction_...`), reproduced here as a magic literal with no compiler-enforced link to the actual `collection.createIndex({...})` call in `trustee-case-appointments.mongo.repository.ts:490-495` that produces it — if that field list ever changes, this constant silently goes stale and `reindexPhase` polls forever with no type error.

Separately, `checkIndexExists`/`createCompoundIndex`/`dropIndex`/`replaceOneInTrusteePartition` (`gateways.types.ts` ~537-544) were added to `TrusteeCaseAppointmentsRepository` — the same interface ordinary business use cases (case assignment, trustee case listing) already depend on via the factory — rather than a migration-scoped interface. No other repository interface in the codebase exposes index-management or partition-named methods; this widens the blast radius of Mongo-specific concepts beyond just this migration module.

Suggest: derive the index name from (or co-locate it with) the actual `createIndex()` call instead of duplicating it as a literal, and move index-lifecycle methods to a migration-scoped interface/humble object rather than the shared business-facing one.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -295,13 +307,16 @@ async function writePage(
...
 async function writeRecord(
   record: ResolvedAcmsRecord,
   appointmentsRepo: ReturnType<typeof factory.getTrusteeCaseAppointmentsRepository>,
 ): Promise<WriteRecordResult> {
-  if (!record.trusteeId) {
-    return { kind: 'failure', failure: { record, reason: 'trustee-not-found' } };
-  }
+  // When no trustee mapping exists, write a sentinel document so the appointment
+  // is queryable and healable later by acmsProfessionalId once the mapping is added.
+  const trusteeId = record.trusteeId ?? SENTINEL_TRUSTEE_ID;
```
jamesobrooks, line 316: **[Medium-High] `deleteAllBySource` removal leaves no path to clean up records ACMS no longer returns.**

`deleteAllBySource` was removed from both the repository implementation and the `TrusteeCaseAppointmentsRepository` interface; the "fresh start" path (`function-apps/.../migrate-case-appointments.ts` handleStart) now just resets `lastId: null` with no deletion. Since re-import is upsert-only by natural key `(documentType, caseId, trusteeId, assignedOn)`, any Cosmos document whose corresponding ACMS record has since disappeared, been corrected, or is now excluded by this same PR's own WHERE-clause fix is never touched — it's permanently orphaned. `heal()` doesn't cover this either; it only repairs cross-partition parity between existing docs, never compares against ACMS's authoritative set, and has no delete path.

The PR description's "fully idempotent via natural key upserts" framing conflates idempotency (no duplicates for records ACMS still returns) with completeness (removing stale records ACMS no longer returns) — these are different properties, and only the first is actually provided.

Worth at least a follow-up plan for detecting/purging stale records, since this PR's own SQL fix changes which rows ACMS returns going forward, making this more likely to matter, not less.

backend/lib/use-cases/dataflows/export-and-load-case.ts
```diff
@@ -71,11 +125,39 @@ async function exportAndLoad(
         );
       }
 
-      await repo.syncDxtrCase(createAuditRecord<SyncedCase>(syncedCase));
+      // Get existing case before sync to detect field changes
+      let existingCase: DxtrCase | undefined;
+      try {
+        const existing = await casesRepo.getSyncedCase(event.caseId);
+        existingCase = existing;
+      } catch {
```
jamesobrooks, line 133: **[Medium] Bare `catch {}` treats transient errors the same as "case doesn't exist," with no logging.**

Unlike the `isNotFoundError(originalError)` check used elsewhere in this same file for a similar `getSyncedCase` call (~line 158) with logging on every other catch block in the file, this one swallows *any* error unconditionally and silently. `getSyncedCase`'s repository implementation wraps generic failures as a distinct `UnknownError` from a `NotFoundError`, so the two are actually distinguishable here — the file just isn't distinguishing them.

Consequence: a transient Mongo/network error skips the denormalized-field diff for that run, and — since `syncDxtrCase` still writes the new field values to the synced-case doc regardless — the *next* run's comparison won't detect a diff either, making this a silent, non-self-healing divergence rather than a transient one.

Suggest binding the error and checking `isNotFoundError` before treating it as "doesn't exist yet," with a `warn` log on anything else, matching the pattern at line ~158.

common/src/cams/trustee-appointments.ts
```diff
@@ -189,7 +189,13 @@ export type CaseAppointmentInput = {
   assignedOn: string;
   appointedDate?: string;
   unassignedOn?: string;
-  source?: 'dxtr' | 'acms';
+  dateFiled?: string;
+  chapter?: string;
+  courtDivisionCode?: string;
+  closedDate?: string;
+  reopenedDate?: string;
+  acmsProfessionalId?: string;
```
jamesobrooks: **[Medium] Migration bookkeeping fields leak through the public case-detail history API with no DTO/whitelist step.**

`acmsProfessionalId` (written unconditionally) and `reason` (written as `'trustee-not-found'` for sentinel/unmapped appointments) are added here to the shared `CaseAppointment`/`CaseAppointmentInput` domain type. `getCaseTrusteeAppointmentHistory` (feature-flagged behind `trustee-appointment-history-enabled`) calls `repo.getByCaseId`, which — unlike `getActiveByCaseId` — has no sentinel filter, and the use case (`case-trustee-appointment.use-case.ts:33-34`) passes the raw docs straight through to the controller with no field projection. Both fields serialize verbatim into the API response once that endpoint is exercised.

Note the sibling `getCasesForTrustee` endpoint already solves this correctly via a purpose-built `TrusteeCaseListItem` whitelist projection — worth reusing that pattern here rather than exposing internal migration bookkeeping through the shared domain type.

backend/lib/use-cases/gateways.types.ts
```diff
@@ -652,6 +674,17 @@ export type MigrateCaseAppointmentsState = RuntimeState & {
   status: 'IN_PROGRESS' | 'COMPLETED' | 'FAILED';
 };
 
+export type HealCaseAppointmentsState = {
```
jamesobrooks: **[Medium] `HealCaseAppointmentsState` doesn't extend `RuntimeState`, forcing `as any` casts that silence type-checking.**

Every sibling state type in this file (`MigrateCaseAppointmentsState`, `TrusteeAppointmentsSyncState`, etc.) extends `RuntimeState`, and `'HEAL_CASE_APPOINTMENTS_STATE'` isn't in the `RuntimeStateDocumentType` union. Since `factory.getRuntimeStateRepository<T extends RuntimeState>` requires that bound, `migrate-case-appointments.ts:390,415` work around it with `factory.getRuntimeStateRepository(context) as any` + `eslint-disable-next-line @typescript-eslint/no-explicit-any`, silencing type-checking on both `read` and `upsert` calls.

The fix is mechanical: add `RuntimeState &` to this type and append the literal to the union — no field conflicts, no behavioral change, removes both lint suppressions.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
```diff
@@ -22,13 +28,54 @@ import factory from '../../../lib/factory';
 
 const MODULE_NAME = ModuleNames.MIGRATE_CASE_APPOINTMENTS;
 
-// Rows fetched from ACMS per handleStart continuation invocation.
-const FETCH_SIZE = 10000;
+const FETCH_SIZE = (() => {
+  const raw = process.env.MIGRATE_CASE_APPOINTMENTS_FETCH_SIZE;
```
jamesobrooks, line 32: **[Low] Env-var int-parsing IIFE duplicated 4x (here, line 51, line 62, and `acms.gateway.ts:20`).**

Same "parse, validate finite/positive, warn-and-fallback" shape repeated with only the var name/default changed (this one also has an extra max-clamp branch). Worth extracting a shared `parseEnvInt(name, default, { max? })` helper, e.g. in `migrate-case-appointments-constants.ts` since 3 of 4 sites already import from there.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -295,13 +307,16 @@ async function writePage(
   return { successCount, failures, remaining: [], recommendedVisibilitySeconds: 0 };
 }
 
+const SENTINEL_TRUSTEE_ID = '00000000-0000-0000-0000-000000000000';
```
jamesobrooks: **[Low] `SENTINEL_TRUSTEE_ID` redefined here independently of `trustee-case-appointments.mongo.repository.ts:26` (plus a third copy in the integration test harness).**

No shared constant currently exists in `common/`; both files already import from `common/src/cams/trustee-appointments.ts`, which would be a natural home. Risk of silent drift if one copy is ever changed without the other.

backend/lib/use-cases/dataflows/migrate-case-appointments-constants.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import {
+  SAFE_THRESHOLD_MS,
+  DEFAULT_FETCH_SIZE,
+  MAX_FETCH_SIZE,
+  WRITE_BATCH_SIZE,
+} from './migrate-case-appointments-constants';
+
+describe('migrate-case-appointments constants', () => {
+  test('SAFE_THRESHOLD_MS equals 56 minutes (4-minute buffer before 60-min timeout)', () => {
+    expect(SAFE_THRESHOLD_MS).toBe(3_360_000);
```
jamesobrooks: **[Low] `expect(CONST).toBe(<same literal it's assigned>)` tests add no regression value.**

This and the two below (`DEFAULT_FETCH_SIZE`/`MAX_FETCH_SIZE`) just restate the constant's own definition — they only fail if someone edits the same line they're pinning, at which point they're already looking at it. They don't check against the actual external constraints these values encode (e.g., the Azure Function timeout budget for `SAFE_THRESHOLD_MS`). The file's own `DEFAULT_FETCH_SIZE <= MAX_FETCH_SIZE` and byte-size-fit tests are the right model — real invariants, not restatements. Suggest dropping these three.

.github/workflows/reusable-deploy.yml
```diff
@@ -114,6 +114,8 @@ jobs:
           _val=$(az keyvault secret show --vault-name "$KV" --name MSSQL-REQUEST-TIMEOUT --query value -o tsv)
           echo "::add-mask::${_val}"
           echo "MSSQL_REQUEST_TIMEOUT=${_val}" >> "$GITHUB_ENV"
+          _val=$(az keyvault secret show --vault-name "$KV" --name MIGRATE-CASE-APPOINTMENTS-FETCH-SIZE --query value -o tsv)
```
jamesobrooks, line 117: **[Low] Missing `::add-mask::` on this Key Vault secret lookup — every other secret retrieval in this step has it.**

All 14 other `az keyvault secret show` calls in this "Get secrets" step are immediately followed by `echo "::add-mask::${_val}"` (including a conditional variant at line 100). This is the sole exception. The value itself (a batch/page-size integer) is low-sensitivity, so no immediate risk, but it breaks an otherwise 100%-consistent convention and sets a precedent that could get copy-pasted for something sensitive later.

backend/lib/humble-objects/mongo-humble.ts
```diff
@@ -106,7 +118,7 @@ export class DocumentClient implements Closable {
   protected client: MongoClient;
 
   constructor(connectionString: string) {
-    this.client = new MongoClient(connectionString);
+    this.client = new MongoClient(connectionString, { ignoreUndefined: true });
```
jamesobrooks, line 121: **[Low] `ignoreUndefined: true` applied globally to the shared `MongoClient` (24+ repositories) rather than scoped to the write path that needed it.**

This flips BSON serialization behavior for `undefined` fields (previously written as `null`, now omitted) app-wide, when a narrower per-operation option was available (e.g., scoped to the `trustees.mongo.repository.ts` `replaceOne` call that motivated this). No test in the repo pins down or guards this app-wide contract. An exhaustive check found no current repository relying on the old undefined-becomes-null behavior, and it's actually load-bearing for this PR's own `$exists`-based migration guard — so this isn't an active bug, but it's worth a scoped fix or at least a regression test so a future partial-update-with-null-intent doesn't silently break.

backend/lib/adapters/gateways/acms/acms.gateway.ts
```diff
@@ -306,48 +320,70 @@ export class AcmsGatewayImpl extends AbstractMssqlClient implements AcmsGateway
...
+        x.id, x.CASE_DIV, x.CASE_YEAR, x.CASE_NUMBER,
+        x.GROUP_DESIGNATOR, x.PROF_CODE, x.APPT_DATE, x.DISP_DATE,
+        x.CASE_FILED_DATE, x.CURR_CASE_CHAPT,
+        x.CLOSED_BY_COURT_DATE, x.CLOSED_BY_UST_DATE
+      ORDER BY x.id`;
 
     try {
-      const { results } = await this.executeQuery<RawRecord>(context, query, input);
-      return (results as mssql.IResult<RawRecord>).recordset;
+      const { results } = await this.executeQuery<AcmsCaseAppointmentRecord>(context, query, input);
```
jamesobrooks: **[Low] `getCmmapAppointments` never received the new `ACMS_REQUEST_TIMEOUT_MS` override that its sibling `getCmmapAppointmentsRaw` (line 473) got, despite running the same heavy CMMAP+CMMDB+CMMKE join.**

Currently low-risk since `getCmmapAppointments` has no production callers (only `getCmmapAppointmentsRaw` is wired into the migration dataflow) — it's exercised only by its own unit test. Worth either applying the same timeout override for consistency, or removing the now-unused method.

jamesobrooks (changes requested):
Ran a multi-agent adversarial review (pragmatic/architecture/devsecops/SRE passes, each self-refuting then independently skeptic-checked, with a second skeptic pass on the highest-severity items). Also independently verified 8 candidate findings surfaced by a separate review tool against the live code. 16 findings survived verification; details inline. Six are High severity and land directly in the code paths this hotfix exists to restore (trustee case list pagination, ACMS timeout retry, dual-write integrity) — recommending changes before merge, particularly #1-#6 below.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
btposey: The "deleteAll" consistently failed due to timeouts. We are deleting records by dropping the collections out of band. We loose any response from the driver when we deleteAll and the function hits it's hard timeout limit.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
btposey: I say we wait for up to 30 minutes then throw so the poison queue takes over.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: To note there is a "heal" mode on the migration that is supposed to go back and correct these missed writes on the assumption they are transient in nature.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Fixed. Changed to outer join (preserveNullAndEmptyArrays: true) and replaced the post-join $match with a conditional $project. When _case is null or has movedToCaseId set, the row returns caseTitle: 'Case not available' and empty courtDivisionName so the count stays accurate and the gap is visible to users.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Addressed. The sync path now retries the soft-close once, then routes to DLQ with full context (caseId, old/new trusteeId, error detail) and continues to create the new appointment regardless — the case is never left with zero active trustee. The stale appointment is corrected on next heal run.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
btposey: ✅ Fixed. Removed the dead ETIMEOUT code check. The wrap chain through getCamsError means the code is never preserved on originalError. isAcmsTimeoutError now uses only the string match that actually fires in production.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Fixed. applyResolvedTrustee now wraps updateCaseAppointment with retry-once-then-DLQ. On persistent failure, logged at ERROR to App Insights with caseId and trusteeIds, a SoftCloseWriteFailed DLQ entry is created, and the new appointment upsert still runs.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
btposey: ✅ Acknowledged. reindexPhase will be updated to wait up to 30 minutes then throw so the Azure Functions poison queue safety net takes over after 5 attempts. Being tracked as a follow-up.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
btposey: ✅ Acknowledged. The hardcoded MongoDB index name string and index-admin methods on the shared interface are a known layering issue. Deferring to a follow-up refactor — the index name is stable for now and the impact is contained to the migration dataflow.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
btposey: The deleteAllBySource consistently failed due to driver timeouts at scale. Collections are dropped out-of-band when a clean slate is needed. The heal intent handles repair of any partition divergence that accumulates.

backend/lib/use-cases/dataflows/export-and-load-case.ts
btposey: ✅ Fixed. The bare catch now distinguishes NotFoundError (expected on first sync) from transient errors, logging a warn for non-NotFound failures so they are visible in App Insights.

common/src/cams/trustee-appointments.ts
btposey: ✅ Fixed. acmsProfessionalId and reason removed from the public CaseAppointment and CaseAppointmentInput types in common/. They now live only on the internal CaseAppointmentDocument type in the repository layer and cannot leak through the API.

backend/lib/use-cases/gateways.types.ts
btposey: ✅ Fixed. Added 'HEAL_CASE_APPOINTMENTS_STATE' to the RuntimeStateDocumentType union. HealCaseAppointmentsState now extends RuntimeState properly. The as-any casts and eslint-disable comments are removed.

backend/function-apps/dataflows/migrations/migrate-case-appointments.ts
btposey: ✅ Acknowledged. The IIFE pattern is duplicated 4x. Deferring extraction to a follow-up to keep this PR focused on the correctness issues.

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
btposey: ✅ Fixed. SENTINEL_TRUSTEE_ID extracted to migrate-case-appointments-constants.ts and imported in both the repository and the use case.

backend/lib/use-cases/dataflows/migrate-case-appointments-constants.test.ts
btposey: ✅ Fixed. Removed the constant-equality tests that just restate the definition. Kept the behavioral tests that verify relationships between constants.

.github/workflows/reusable-deploy.yml
btposey: ✅ Fixed. Added the missing ::add-mask:: directive to match all other secret lookups in the same step.

backend/lib/humble-objects/mongo-humble.ts
btposey: ✅ Acknowledged. The ignoreUndefined: true is applied globally to the shared MongoClient. Deferring scope-narrowing to a follow-up — the global setting is conservative and doesn't change observable behavior for existing documents.

backend/lib/adapters/gateways/acms/acms.gateway.ts
btposey: ✅ Fixed. Applied ACMS_REQUEST_TIMEOUT_MS to getCmmapAppointments (was only on the raw variant). Both methods now use the same configurable timeout.

jamesobrooks (approved):

PR #2644, opened 2026-07-08
HOT FIX - Fix sort composite index for getCasesForTrustee

btposey:
# Bug

The Trustee Case List was returning a 500 error in staging and production with:
> `The order by query does not have a corresponding composite index that it can be served from.`

Cosmos DB for MongoDB requires that ORDER BY composite indexes in aggregation pipelines include the shard key as the leading field. The sort index `(dateFiled DESC, caseId ASC)` was missing `trusteeId` as the leading key.

# Purpose

Unblock the Trustee Case List screen in staging/production. The index was created via the `reindex` intent but with the wrong key shape, causing Cosmos to reject the sort.

# Major Changes

- Changed sort index from `(dateFiled: -1, caseId: 1)` to `(trusteeId: 1, dateFiled: -1, caseId: 1)` in the repository `createCompoundIndex` method
- Updated Bicep declaration to match
- Updated integration test harnesses to create and assert the correct 3-field index

# Testing/Validation

- Integration test harness (`trustee-case-filter`) seeds and asserts the correct sort index exists before executing any sort queries
- `reindex` intent must be re-triggered in staging to replace the incorrect 2-field index with the correct 3-field index

# Notes

Cosmos DB's requirement: when an aggregation pipeline includes a `$sort` on non-partition-key fields, the composite index for that sort must be prefixed by the shard key (`trusteeId`). Native MongoDB does not enforce this, which is why the issue only surfaces in Cosmos.

# Definition of Done:

## Summary by Sourcery

Ensure trustee case list queries use a Cosmos-compliant composite index for sorting case appointments.

Bug Fixes:
- Fix trustee case appointments sort index to include trusteeId so ORDER BY dateFiled DESC, caseId ASC can be served without errors in Cosmos DB.

Enhancements:
- Align repository, integration harnesses, and Cosmos Bicep configuration to use a consistent three-field sort index on trustee case appointments.

Tests:
- Update integration test harnesses to create and assert the presence of the trusteeId-prefixed sort index for trustee case appointments.

### DIFF
--- a/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
@@ -510,8 +510,11 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     // Drop the old 2-field sort index if it exists from a previous reindex run
     try {
       await collection.dropIndex('dateFiled_-1_caseId_1');
-    } catch {
-      // Index doesn't exist — nothing to drop
+    } catch (dropError) {
+      const msg = dropError instanceof Error ? dropError.message : String(dropError);
+      if (!msg.includes('index not found') && !msg.includes('IndexNotFound')) {
+        this.context.logger.warn(MODULE_NAME, `Failed to drop old sort index: ${msg}`);
+      }
     }
   }
 
--- a/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
@@ -507,7 +507,7 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     const collection = this.getTrusteeCollection();
     await collection.createIndex({ trusteeId: 1, unassignedOn: 1, dateFiled: 1, caseStatus: 1 });
     // Sort index to serve ORDER BY dateFiled DESC, caseId ASC in getCasesForTrustee
-    await collection.createIndex({ dateFiled: -1, caseId: 1 });
+    await collection.createIndex({ trusteeId: 1, dateFiled: -1, caseId: 1 });
   }
 
   async dropIndex(indexName: string): Promise<void> {
--- a/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
@@ -506,8 +506,13 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
   async createCompoundIndex(): Promise<void> {
     const collection = this.getTrusteeCollection();
     await collection.createIndex({ trusteeId: 1, unassignedOn: 1, dateFiled: 1, caseStatus: 1 });
-    // Sort index to serve ORDER BY dateFiled DESC, caseId ASC in getCasesForTrustee
     await collection.createIndex({ trusteeId: 1, dateFiled: -1, caseId: 1 });
+    // Drop the old 2-field sort index if it exists from a previous reindex run
+    try {
+      await collection.dropIndex('dateFiled_-1_caseId_1');
+    } catch {
+      // Index doesn't exist — nothing to drop
+    }
   }
 
   async dropIndex(indexName: string): Promise<void> {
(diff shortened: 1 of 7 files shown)


backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -507,7 +507,7 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     const collection = this.getTrusteeCollection();
     await collection.createIndex({ trusteeId: 1, unassignedOn: 1, dateFiled: 1, caseStatus: 1 });
     // Sort index to serve ORDER BY dateFiled DESC, caseId ASC in getCasesForTrustee
-    await collection.createIndex({ dateFiled: -1, caseId: 1 });
+    await collection.createIndex({ trusteeId: 1, dateFiled: -1, caseId: 1 });
```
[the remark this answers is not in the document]
btposey: ✅ Fixed both. Added a  call (with a silent catch if the index doesn't exist) so staging won't carry the redundant 2-field index after reindex runs. Also replaced the key-count/field-by-field assertion with deep equality on the expected key shape: `JSON.stringify(idx.key) === JSON.stringify({ trusteeId: 1, dateFiled: -1, caseId: 1 })`.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -506,8 +506,13 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
   async createCompoundIndex(): Promise<void> {
     const collection = this.getTrusteeCollection();
     await collection.createIndex({ trusteeId: 1, unassignedOn: 1, dateFiled: 1, caseStatus: 1 });
-    // Sort index to serve ORDER BY dateFiled DESC, caseId ASC in getCasesForTrustee
-    await collection.createIndex({ dateFiled: -1, caseId: 1 });
+    await collection.createIndex({ trusteeId: 1, dateFiled: -1, caseId: 1 });
+    // Drop the old 2-field sort index if it exists from a previous reindex run
+    try {
+      await collection.dropIndex('dateFiled_-1_caseId_1');
+    } catch {
```
jamesobrooks: Nit: this bare `catch {}` swallows *every* error from `dropIndex`, not just "index doesn't exist" — a real failure here (e.g. insufficient permission to drop indexes on Cosmos) would be silently discarded with no log line, contrary to what the comment claims. The class already has `checkIndexExists(indexName)` a few lines up, used by the `reindexPhase` use case for the same "drop old index" concern — consider using that guard here instead (`if (await this.checkIndexExists('dateFiled_-1_caseId_1')) await collection.dropIndex(...)`), or at least log the caught error so an SRE debugging a Cosmos index issue isn't left with no signal. Not a blocker for this hotfix.

jamesobrooks (commented):
Reviewed via pragmatic-reviewer + software-architect-review passes with adversarial verification. Core fix is correct and minimal — `getCasesForTrustee` sorts on `{ dateFiled: -1, caseId: 1 }` while filtering by `trusteeId`, and Cosmos requires the shard key as the leading composite-index field, matching the diagnosis in the PR description. Bicep and test-harness updates are consistent with the new 3-field index. Two low-risk nits inline; nothing blocking for a hotfix.

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
```diff
@@ -753,7 +753,7 @@ async function assertHappyPath(db: ReturnType<MongoClient['db']>) {
     }
 
     const hasSortIndex = indexKeys.some(
-      (k) => k === JSON.stringify({ dateFiled: -1, caseId: 1 }),
+      (k) => k === JSON.stringify({ trusteeId: 1, dateFiled: -1, caseId: 1 }),
     );
     if (hasSortIndex) {
       pass('sort index (dateFiled DESC, caseId ASC) exists');
```
jamesobrooks: Nit: the assertion above was correctly updated to require `{ trusteeId: 1, dateFiled: -1, caseId: 1 }`, but these `pass`/`fail` messages (lines 759 and 761) still say "sort index (dateFiled DESC, caseId ASC)" with no mention of `trusteeId`. An operator reading harness output on failure won't know trusteeId is now part of the required key. The sibling harness (`test/integration/trustee-case-filter/scripts/run-tests.ts`) already updated its equivalent message to include trusteeId — worth matching that here.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: Fixed. Now catches the error and only silently ignores 'index not found' / 'IndexNotFound' — any other error logs a warn so permission failures surface in App Insights.

test/integration/migrate-case-appointments/scripts/migrate-case-appointments-harness.ts
btposey: Fixed. Updated both messages to 'sort index (trusteeId ASC, dateFiled DESC, caseId ASC)'.

jamesobrooks (approved):

PR #2648, opened 2026-07-09
CAMS-797 Suppress notifications for internal contact, software, and bank changes

cruzjone-flexion:
# Purpose

The Office of Oversight receives trustee change notification emails for every field change, but internal contact information, software vendor, and bank assignments are outside the scope of what they track. This change stops those fields from triggering notifications while preserving their audit history.

# Major Changes

- Removed `fields.push()` calls for internal contact, software vendor, and banks in `TrusteesUseCase.recordAuditHistory()` — these fields no longer appear in the `TrusteeChangeSet` passed to the notification dispatcher
- Audit history records (`AUDIT_INTERNAL_CONTACT`, `AUDIT_SOFTWARE`, `AUDIT_BANKS`) continue to be written unchanged
- Suppressed-only saves produce an empty `fields` array, which the existing `fields.length === 0` guard already uses to skip dispatch
- Mixed saves (suppressed + non-suppressed fields in the same update) still dispatch — the notification reflects only the non-suppressed fields

# Testing/Validation

Implemented using TDD. Six new tests added to the existing notification dispatch describe block in `trustees.test.ts`:
- Suppressed-only saves for each of the three field types produce no notification
- A save with all suppressed fields produces no notification
- A mixed save (name + internal contact) dispatches and the email body contains the name change but not the internal contact change
- Audit history is still written for suppressed fields even when no notification fires

All 3281 backend and common tests pass.

# Notes

Assistant changes (create/update/delete via `TrusteeAssistantsUseCase`) were already suppressed — that use case has no notification dispatch code. No changes to `NotificationRoutingRepository`, routing records, or recipient lists were required.

The suppression is hardcoded rather than config-driven. The suppressed types represent a stable business rule (the Office of Oversight's scope), not an operational knob, so a deploy-tracked code change is preferable to a live Cosmos config that could drift or be accidentally deleted.

# Definition of Done:

## Summary by Sourcery

Suppress trustee change notification emails for internal contact, software vendor, and bank assignment updates while preserving their audit history.

Bug Fixes:
- Prevent notifications from being dispatched when only internal contact, software vendor, or bank fields change.
- Ensure mixed updates still dispatch notifications that reflect only non-suppressed field changes.

Enhancements:
- Stop including internal contact, software, and bank fields in trustee change sets used for notification dispatch.
- Verify via tests that audit history continues to be recorded for suppressed fields even when notifications are not sent.
- Expand trustee notification tests to cover suppression scenarios and mixed field updates.

### DIFF
--- a/backend/lib/use-cases/trustees/trustees.test.ts
+++ b/backend/lib/use-cases/trustees/trustees.test.ts
@@ -14,6 +14,7 @@ import { CourtDivisionDetails } from '@common/cams/courts';
 import { MockNotificationGateway } from '../../testing/mock-gateways/mock-notification.gateway';
 import { AppointmentChapterType } from '@common/cams/trustees';
 import { ContactInformation } from '@common/cams/contact';
+import { BankruptcySoftwareProfile } from '@common/cams/bankruptcy-software';
 
 describe('TrusteesUseCase tests', () => {
   let context: ApplicationContext;
@@ -1754,7 +1755,7 @@ describe('TrusteesUseCase tests', () => {
       expect(addressComparison?.before ?? '').toBe('');
     });
 
-    test('software removal emits empty after value in change set', async () => {
+    test('software removal writes audit history but does not add to notification fields', async () => {
       const oldSoftwareId = 'sw-old';
       const trusteeWithSoftware = MockData.getTrustee({ softwareId: oldSoftwareId });
       vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(trusteeWithSoftware);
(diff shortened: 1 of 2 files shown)


cruzjone-flexion:
Thanks for the review, @sourcery-ai.

**Suggestion 1 (centralize suppression rule):** Considered and declined. The suppression is structural — we removed the `fields.push()` calls rather than filtering a list at runtime. A named constant would only add value if the suppression were data-driven, which we explicitly decided against (stable business rule, deploy-tracked change is preferable to a live config). The PR description documents the intent for reviewers.

**Suggestion 2 (assert on change set instead of HTML):** Considered and declined. The HTML assertion is intentional — we're verifying what the recipient actually receives in the email, not just that the change set is correct. Asserting only on `fields` length would leave a gap where a template bug could still render suppressed content. The label "Internal Contact" is a stable production constant, so if it ever changes, that's a real behavioral change worth catching.

mattstankey (approved):

PR #2660, opened 2026-07-09
CAMS-760 Fix and harden deploy identity KV secret access

jamesobrooks:
# Bug

The Main-Gov deploy failed with `ForbiddenByRbac` on
`migrate-case-appointments-fetch-size`. CAMS-760 added a new
`az keyvault secret show --name MIGRATE-CASE-APPOINTMENTS-FETCH-SIZE`
call to `reusable-deploy.yml`, but the corresponding per-secret RBAC
grant in `setup-deploy-federated-credential.sh`'s `KV_SECRETS` array
was never added, so the deploy identity had no access to the secret.

# Purpose

Fix the immediate access gap, then close the structural hole that
allowed it: two independently-maintained lists (a workflow's secret
reads, and a runbook's granted secrets) with nothing keeping them in
sync, repeated across 9 separate runbook scripts and a third
hardcoded orchestrator call-list.

# Major Changes

* Added the missing secret to `KV_SECRETS` in `setup-deploy-federated-credential.sh`
* Consolidated all 9 `setup-*-federated-credential.sh` runbooks and their
  shared `_oidc-helpers.sh` into `ops/scripts/utility/federated-credentials/`
* Made `setup-all-federated-credentials.sh` discover runbooks via glob
  instead of a hardcoded per-script call list
* Added a standardized `# KV-Workflows: <file>` tag above each runbook's
  `KV_SECRETS`, co-locating the workflow association with the data it describes
* Added a `check-kv-secret-parity.sh` pre-commit hook that cross-checks each
  runbook's `KV_SECRETS` against the `az keyvault secret show --name X`
  references in its tagged workflow(s), failing the commit on any gap, and
  prints a non-blocking reminder when a workflow references a brand-new
  secret name (since no local check can confirm the secret exists in Azure)

# Testing/Validation

* Ran `setup-deploy-federated-credential.sh` with `TARGET=main` against the
  live Main-Gov Key Vault and confirmed the new secret's role assignment
  was created
* Manually validated the pre-commit hook's failure path by temporarily
  removing a `KV_SECRETS` entry and confirming the commit was blocked with
  the correct missing-secret message, then restored and confirmed a clean pass
* Verified the hook's "new secret" reminder logic against the actual
  CAMS-760 commit history (diffing the workflow file before/after the
  fetch-size secret was added) and confirmed it correctly flags exactly
  that secret
* `pre-commit run --all-files` and `shellcheck` pass on all touched scripts

# Notes

The Main-Gov deploy failure was masked by a generic Key Vault 403 —
Azure returns the same `ForbiddenByRbac` error whether the caller lacks
RBAC or the secret doesn't exist at all (to prevent enumeration by
unauthorized callers), so a live preflight check alone would not fully
distinguish the two cases; this fix focuses on preventing the
list-drift class of bug, which was the actual root cause here.

Per-secret (rather than vault-scoped) RBAC is intentional: the vault
also holds secrets no deploy workflow should ever read (e.g.
`MONGO-CONNECTION-STRING`, `ACMS-MSSQL-PASS`), so broadening to
vault-scope access was ruled out as the wrong tradeoff.

Companion PR in `ustp-cams-fdp` updates the `cams-gha-security` skill
with a matching checklist and fixes stale paths from the directory move.

# Definition of Done:

## Summary by Sourcery

Fix Key Vault RBAC gap for deploy identity and harden federated credential runbooks to keep workflow secret access in sync.

Bug Fixes:
- Grant the deploy federated credential access to the MIGRATE-CASE-APPOINTMENTS-FETCH-SIZE Key Vault secret used by the deployment workflow.

Enhancements:
- Move all federated-credential setup runbooks and shared helpers into a dedicated ops/scripts/utility/federated-credentials directory.
- Update the federated credential orchestration script to automatically run all setup-*-federated-credential.sh scripts via glob discovery.
- Annotate each federated credential runbook with a KV-Workflows tag that documents the associated GitHub workflow files and their expected secrets.
- Add a pre-commit hook that enforces parity between each runbook's KV_SECRETS list and the Key Vault secrets referenced by its tagged workflows, and reminds about newly added secret names.

CI:
- Extend pre-commit configuration to run the Key Vault secret/runbook parity check on federated credential scripts and GitHub workflow files.


mattstankey:
**Bug:** `.pre-commit-hooks/check-kv-secret-parity.sh` uses `mapfile` (lines with `mapfile -t declared_secrets` and `mapfile -t required_secrets`), which requires Bash ≥ 4. macOS ships Bash 3.2 by default (`/bin/bash --version` → `3.2.57(1)-release`), and the hook's `language: script` entry runs with whatever `bash` resolves first on `PATH` — not necessarily a Homebrew-installed 4+/5+ bash.

Verified locally: running the hook under macOS's default `/bin/bash` fails immediately with `mapfile: command not found` (exit 127), before any of the actual parity logic runs. On any contributor machine without a newer Bash installed and prioritized on `PATH`, this would hard-fail on every commit touching a runbook or workflow file — fully blocking commits, not just the intended "gap detected" failure path.

Suggest replacing the two `mapfile` calls with a portable `while read` loop into an array (the `added` variable in `remind_if_new_secret` already does this pattern a few lines below), or explicitly requiring/documenting a Bash 4+ dependency if `mapfile` stays.

Worth confirming on a clean macOS shell (no Homebrew bash on `PATH`) before merging, since the PR's testing notes mention `pre-commit run --all-files` passing, which likely ran against a shell where a newer bash was already first on `PATH`.

.pre-commit-hooks/check-kv-secret-parity.sh
```diff
@@ -0,0 +1,98 @@
...
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+RUNBOOK_DIR="$REPO_ROOT/ops/scripts/utility/federated-credentials"
+WORKFLOW_DIR="$REPO_ROOT/.github/workflows"
+
+exit_code=0
+
+extract_kv_secrets() {
+  sed -n '/^KV_SECRETS=(/,/)/p' "$1" | grep -oE '"[A-Z0-9-]+"' | tr -d '"'
```
fmaddenflx: I had to look up the sed range selector because I wasn't familiar with it.  The Regex syntax for sed here is using a range selector such as /start/,/end/, which is what the /,/ does.  That threw me for a bit.

fmaddenflx (approved):

.pre-commit-hooks/check-kv-secret-parity.sh
jamesobrooks: Yep — `/pattern1/,/pattern2/p` in `sed -n` is the address-range form: it prints every line from the first line matching `pattern1` through the first subsequent line matching `pattern2` (inclusive). Here that means "print from the `KV_SECRETS=(` line through the next line containing `)`", which handles both the single-line arrays (open+close paren on one line, matching both patterns at once) and the multi-line arrays used elsewhere in these runbooks. Happy to add an inline comment calling that out if it would help future readers.

jamesobrooks:
Addressed in 48d7807:

- **@mattstankey** — confirmed: `mapfile` requires Bash ≥4, and the hook's `#!/usr/bin/env bash` shebang isn't guaranteed to resolve to a modern bash (macOS ships 3.2 by default). Replaced both `mapfile` calls with portable `while read` loops and verified the hook runs clean under `/bin/bash` 3.2 directly, including the missing-secret failure path.
- **@sourcery-ai** — broadened the secret-name pattern from `[A-Z0-9-]` to `[A-Za-z0-9_-]` so a future lowercase or underscored secret name won't be silently excluded from the parity check.

Leaving the `nullglob`/`[[ -f ]]` guard suggestion on `setup-all-federated-credentials.sh` as-is — the glob always has matches today, and the loop body already handles a non-matching glob defensively enough for the current invariant.

PR #2665, opened 2026-07-10
CAMS-786 Upgrade Node runtime from 22 to 24 LTS

cruzjone-flexion:
# Purpose

Stay current with the Node.js LTS lifecycle by upgrading from Node 22 to Node 24 across all in-repo version pins. Node 24 LTS is now Generally Available on Azure Functions, removing the previous blocker for this upgrade.

# Major Changes

- `.nvmrc` bumped from `v22.22.2` to `v24.4.0`
- CI Docker base image updated from `node:22.22.2-bookworm-slim` to `node:24.4.0-bookworm-slim`
- `linuxFxVersionMap` updated from `NODE|22` to `NODE|24` in all three Bicep deployment files (`backend-api-deploy.bicep`, `dataflows-resource-deploy.bicep`, `frontend-webapp-deploy.bicep`), including their inline version comments
- Added `ignoreDeprecations: "6.0"` to `test/e2e/tsconfig.json` (mirroring the same fix already applied to `test/uswds`) — npm 11's `npx` resolution under Node 24 surfaced TS5101/TS5107 deprecation errors (`moduleResolution=node10`, `baseUrl`) that were previously silently masked
- Excluded `scripts/harvest-mongo.ts` from `test/e2e/tsconfig.json` — this file is already dead code (marked `@deprecated`, `process.exit(1)` at the top) but its import of a removed backend path was pulling ~600 unrelated errors into typecheck once the deprecation errors above stopped short-circuiting the compiler
- Fixed ~13 genuine strict-null-check errors in `test/e2e` that had never actually been type-checked before (auth-setup.ts, urlQueryString.ts, transfer-orders.spec.ts, harvest-sqlserver.ts, load-cosmos-data.ts) — these were previously masked because tsc aborted on the config-level deprecation errors before checking file bodies
- ~~Pinned the `accessibility-test` CI job to Node 22~~ — no longer needed. The Playwright ESM loader bug (`context.conditions?.includes is not a function`) was already fixed upstream in `@playwright/test@1.61.1` (a stable release), which this repo picked up via an unrelated Renovate bump before this branch's Node 22 pin was even written. Verified on a throwaway branch that reverting `reusable-accessibility.yml` back to `node-version-file: '.nvmrc'` (Node 24) passes cleanly — [72 tests passed](https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/29123889829/job/86465025715). See PR comment for details.

# Testing/Validation

No application logic changed. All GitHub Actions workflows already reference `.nvmrc` via `node-version-file` — no other workflow YAML edits were needed. Azure Functions extension bundle v4 is compatible with the Node 24 runtime; no `host.json` changes were needed.

CI failures surfaced by the Node 24 upgrade and their resolutions:
- `typecheck / TypeScript`: fixed by the tsconfig/type-error changes above — confirmed passing on the latest push.
- `accessibility-test / playwright-accessibility-test`: originally worked around by pinning that job to Node 22; superseded once we confirmed `@playwright/test@1.61.1` (already in the lockfile) resolves the issue on Node 24 directly — job now uses `node-version-file: '.nvmrc'` like every other workflow.

# Notes

**Out-of-repo follow-ups required before CI/CD runs on Node 24 (manual, post-merge):**

| System | Variable | Location | New Value |
|---|---|---|---|
| GitHub Actions | `NODE_VERSION` | Repository → Settings → Variables → Actions | `24.4.0` |
| Azure DevOps | `NODE_API_VERSION` | ADO Library variable group | `24.4.0` |

`NODE_VERSION` is consumed by `continuous-deployment.yml`, `sub-build.yml`, `sub-deploy.yml`, and `update-dependencies.yml`. `NODE_API_VERSION` is consumed by the `NodeTool@0` install step in the ADO-Mirror pipeline templates.

# Definition of Done:

## Summary by Sourcery

Upgrade project tooling and deployment configuration to use Node.js 24 LTS instead of Node.js 22.

Enhancements:
- Align Azure Functions runtime mapping to Node 24 in backend, dataflows, and frontend Bicep deployment templates.
- Update local Node version pin and CI Docker build-and-test image to Node 24 LTS.
- Fix TypeScript deprecation errors and previously-masked strict-null-check errors in `test/e2e` surfaced by npm 11 under Node 24.
- Pin the accessibility-test CI job to Node 22 to work around an upstream Playwright/Node 24 ESM loader incompatibility.

Build:
- Refresh CI Docker base image to a Node 24 LTS variant for build-and-test workflows.

Deployment:
- Adjust linuxFxVersionMap values to deploy functions on the Node 24 runtime across all relevant cloud deployment templates.


mattstankey:
Addressed one of the two Sourcery suggestions above in a96cc450b: dropped the hardcoded `(currently v24.x.x)` version string from the Node NOTE comments in all three Bicep files, so they won't go stale on future Node bumps.

Not implementing the version-centralization suggestion in this PR — it would require threading the Node version through as a deploy-time parameter from CI into Bicep (touching the CI/CD pipelines, not just these files), which is a larger, separate infra change better scoped as a follow-up.

test/e2e/playwright/auth-setup.ts
```diff
@@ -4,7 +4,11 @@ import { test } from './fixture/urlQueryString';
 require('dotenv').config({ quiet: true });
 
 const authFile = 'playwright/.auth/user.json';
-const { OKTA_USER_NAME, OKTA_PASSWORD, TARGET_HOST } = process.env;
+const { OKTA_USER_NAME, OKTA_PASSWORD } = process.env;
+if (!process.env.TARGET_HOST) {
+  throw new Error('TARGET_HOST environment variable is required for E2E tests');
+}
+const TARGET_HOST: string = process.env.TARGET_HOST;
```
jamesobrooks, line 11: What's the reasoning behind this change?

test/e2e/playwright/auth-setup.ts
```diff
@@ -74,7 +84,7 @@ async function oktaLogin(page: Page) {
 }
 
 function usingAuthenticationProvider() {
-  let loginFunction;
+  let loginFunction = mockLogin;
```
jamesobrooks: Why do we need this? It seems like the same result as before, but with a perceived preference for mock login, which I'm not sure should be preferred.

jamesobrooks (commented):
I left some questions where I'm wondering why we made some changes. I'm also wondering why there are TypeScript v7 related changes and the new `vite-env.d.ts` file, which seems to only include a TypeScript directive.

test/e2e/playwright/auth-setup.ts
mattstankey: This guard+reassignment lets TypeScript narrow `TARGET_HOST` from `string | undefined` to `string` for the rest of the file.

The original code destructured `const { TARGET_HOST } = process.env`, which is `string | undefined`. Adding `if (!TARGET_HOST) throw ...` right after doesnt narrow the destructured variable inside the `mockLogin`/`oktaLogin` closures defined later in the file — TypeScript cannot prove those closures only run after the guard executes. Re-declaring `TARGET_HOST` as an explicitly typed `const` after the check fixes that, so the rest of the file can use it as a plain `string` without further guards.

This was needed to get `tsc --noEmit` passing in `test/e2e` — see the tsconfig fix in 67b47dc11 for context on why this file was never actually typechecked before.

test/e2e/playwright/auth-setup.ts
mattstankey: This is not a preference change — its fixing a TS18048 "possibly undefined" typecheck error, and picking a default that matches existing runtime behavior.

The original `let loginFunction;` had no initializer, so if `provider.toLowerCase()` didnt match any of the three `case` branches, `loginFunction` would stay `undefined` and `tsc` flagged the `await login(page)` call at line 17 as possibly invoking `undefined`.

I defaulted it to `mockLogin` because thats consistent with the line directly above it: `process.env.CAMS_LOGIN_PROVIDER ?? 'mock'` already treats `'mock'` as the runtime fallback whenever `CAMS_LOGIN_PROVIDER` is unset. So this only makes the type-level default agree with the fallback that was already there — it doesnt change which login function actually gets used for any given `CAMS_LOGIN_PROVIDER` value.

Happy to use a different fallback (e.g. `noOp`, or throw for unrecognized providers) if youd prefer — let me know.

mattstankey: Good catch — you were right to push on this. I dug in further: `CAMS_LOGIN_PROVIDER` is also set when running against deployed environments (per `docs/operations/e2e-tests.md`), so a typo'd or misconfigured value there is a real failure mode, not just a TS-exhaustiveness technicality. Defaulting to `mockLogin` would have silently attempted mock login against a live Okta-secured environment instead of failing with a clear error about the bad config.

Fixed in dd21b806f: now throws `Unrecognized CAMS_LOGIN_PROVIDER: "<value>"` for any value that does not match `none`/`okta`/`mock`, instead of silently falling back to `mockLogin`. Still satisfies the typechecker since every path either returns or throws.

test/e2e/playwright/auth-setup.ts
mattstankey: Verified this empirically to make sure I wasnt just asserting it — three minimal repros with `tsc --noEmit --strict`:

```ts
// 1. Destructured const + guard: FAILS to narrow
const { TARGET_HOST } = process.env;
if (!TARGET_HOST) { throw new Error("required"); }
function useIt() { console.log(TARGET_HOST.length); }
// error TS18048: TARGET_HOST is possibly undefined

// 2. Non-destructured const + guard: still FAILS
const TARGET_HOST = process.env.TARGET_HOST;
if (!TARGET_HOST) { throw new Error("required"); }
function useIt() { console.log(TARGET_HOST.length); }
// error TS18048: TARGET_HOST is possibly undefined

// 3. Guard on process.env.X directly, then re-declare: PASSES
if (!process.env.TARGET_HOST) { throw new Error("required"); }
const TARGET_HOST: string = process.env.TARGET_HOST;
function useIt() { console.log(TARGET_HOST.length); }
// no error
```

So the failure mode isnt about destructuring specifically — its that `tsc` wont carry a narrowed type across a function boundary for any `const` derived from `process.env`, no matter how its declared. The only pattern that survives into the closures (`mockLogin`/`oktaLogin`) is guarding the `process.env.X` property access itself and then giving the re-declared local an explicit type, which is what the current code does.

mattstankey: Another followup to this after pushing back on the change:

(`TARGET_HOST` reassignment): I re-verified this was actually necessary rather than just re-asserting it — ran three isolated `tsc --strict` repros. Confirmed TypeScript won't carry a narrowed process.env type across a function boundary regardless of whether it's destructured or not; the only pattern that survives into the `mockLogin/oktaLogin` closures is guarding `process.env.TARGET_HOST` directly and re-declaring the local with an explicit type. Posted that evidence on the thread — this one holds up, no change needed.

jamesobrooks (dismissed):

jamesobrooks:
Dropped the Node 22 pin on the \`accessibility-test\` job in fb81997a2.

While reviewing this PR, we found that the Playwright ESM loader bug that motivated the pin (\`context.conditions?.includes is not a function\` under Node 24) was already fixed upstream in \`@playwright/test@1.61.1\` — a stable release, not the \`1.62.0-alpha\` this PR's original comment cited. This repo already had 1.61.1 installed via an unrelated Renovate bump (commit ec7d94082) that landed on \`main\` before this branch's Node 22 pin was written, so the workaround was stale before it even merged.

Confirmed on a throwaway branch (\`test-playwright\`) that reverting \`reusable-accessibility.yml\` to \`node-version-file: '.nvmrc'\` (Node 24) passes cleanly: https://github.com/US-Trustee-Program/Bankruptcy-Oversight-Support-Systems/actions/runs/29123889829/job/86465025715 (72 tests passed).

Reference for the upstream bug/fix: https://github.com/microsoft/playwright/issues/41311

jamesobrooks (approved):

PR #2669, opened 2026-07-13
Manage trustee-case-appointments indexes outside of bicep

btposey:
# Purpose

Cosmos DB for MongoDB's Bicep/ARM resource can't express the mixed-direction ORDER BY sort index (`dateFiled DESC, caseId ASC`) required by `getCasesForTrustee`. The prior out-of-band mongosh script covered only that one index, leaving the rest of the collection's indexes declared in Bicep — but that split still forces Bicep to drop and rebuild the out-of-band index on every single deploy, since the mongosh step recreates it right after each deploy and Bicep rediscovers it as an undeclared addition the next time around.

# Major Changes

* Omit the `indexes` property entirely from the `trustee-case-appointments` collection in `cosmos-collections.bicep`, rather than declaring some indexes and excluding just the sort index. Empirical testing (see the new ADR) showed this is the only way to avoid a recurring rebuild — an *empty* index array still triggers full declarative replacement, but an *omitted* property leaves existing indexes untouched.
* Replace the old `create-trustee-case-appointments-sort-index.mongosh.js` with `index-trustee-case-appointments.js`, a small idempotent Node script (MongoDB driver, no mongosh dependency) that owns all of the collection's non-default indexes and is invoked inline from `az-cosmos-deploy.sh`, which both the GHA and ADO pipelines share.
* Add a new ADR, `CosmosMongoIndexManagement`, documenting the pattern and the experimentation that justified it, so it doesn't need to be re-litigated for the next collection that hits this same ARM limitation.
* Add CI lint coverage for `ops/`, which is not an npm workspace and was previously unchecked by `npm run lint --workspaces`.
* Update comments in the `trustee-case-filter` and `migrate-case-appointments` integration test harnesses to reflect the new script name/location (no logic changes — both harnesses already seed their own indexes directly via the driver and were never dependent on Bicep).

# Testing/Validation

* Verified empirically against a disposable Cosmos scratch database and account: three candidate Bicep encodings of the mixed-direction index were each deployed and inspected — none produced the required index, confirming the ARM limitation directly rather than by inference.
* Verified the new architecture (omit the `indexes` property, script owns everything) is a true no-op on repeat runs: Cosmos's own `createIndexes` response reports `"note": "all indexes already exist"` with unchanged before/after index counts, for both the filtering index and the sort index itself.
* Smoke-tested the actual `index-trustee-case-appointments.js` script (not just the test harness) against a real disposable Cosmos database: first run creates both indexes, second identical run reports both as no-ops.
* Ran the `trustee-case-filter` integration test end-to-end (local MongoDB via Podman) — all 38 assertions pass, including the two that directly assert on the sort index's existence and correctness.
* Confirmed the compiled ARM template for `trustee-case-appointments` has no `indexes` property after the Bicep change.
* Full validation suite passed: format, lint (including the new `ops/` lint step), typecheck, all unit tests (common/backend/user-interface), backend coverage, knip, and npm audit (one pre-existing low-severity finding, unrelated).

# Notes

The scratch-database experimentation (Bicep fixtures, the probing script, and full write-up) is preserved under `backend/_experiments/bicep-index-probe/` for reuse if this question needs revisiting, but is git-ignored and not part of this diff.

This branch's ADO pipeline integration relies on `az-cosmos-deploy.sh` being the shared script both GHA and ADO invoke (confirmed via `docs/operations/deployment.md`) — the ADO YAML itself was not reviewed as part of this change.

# Definition of Done:

## Summary by Sourcery

Manage trustee-case-appointments collection indexes outside Bicep using an idempotent deployment script and document the Cosmos Mongo index management pattern.

Enhancements:
- Omit the indexes property from the trustee-case-appointments Cosmos Mongo collection so ARM no longer reconciles or rebuilds its indexes on deploy.
- Add an idempotent Node-based index-trustee-case-appointments script to own the collection's non-default filtering and mixed-direction sort indexes, invoked from the shared az-cosmos-deploy pipeline script.
- Update integration test harness comments to reflect that trustee-case-appointments indexes are managed out-of-band by the new script rather than via Bicep.
- Extend ESLint configuration to lint the new deployment script file in ops/ and add a dedicated npm lint:ops command.
- Document the Cosmos Mongo API index management approach and experimentation in a new architecture decision record and sidebar entry.

CI:
- Update the reusable lint GitHub workflow to run ESLint against ops/ via the new lint:ops command.


ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
```diff
@@ -0,0 +1,81 @@
...
+  },
+];
+
+async function main() {
+  const [connectionString, databaseName] = process.argv.slice(2);
+  if (!connectionString || !databaseName) {
+    console.error(
+      'Usage: index-trustee-case-appointments.js "<mongo-connection-string>" <databaseName>',
+    );
+    process.exit(2);
```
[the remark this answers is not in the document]
btposey: ✅ Addressed. The connection string is now read from the `MONGO_CONNECTION_STRING` environment variable when only one CLI argument (`<databaseName>`) is supplied, avoiding exposure in shell history and `ps` output — this is how `az-cosmos-deploy.sh` now invokes it. The two-argument CLI form (`<connection-string> <databaseName>`) is still accepted for backward compatibility with manual/ad-hoc runs.

Also hardened response handling per your other suggestion: the script now checks `response.ok === 1` and confirms the index actually exists via `indexExists()` before treating the `createIndexes` call as successful, rather than trusting `numIndexesBefore`/`numIndexesAfter` or the `note` field alone.

docs/architecture/decision-records/CosmosMongoIndexManagement.md
```diff
@@ -0,0 +1,79 @@
...
+
+This is a hard limitation of the ARM resource, not a Bicep authoring gap: no encoding of the index
+specification (embedded direction markers, prefix conventions, or index option overrides) is
+accepted by the resource provider. It was confirmed empirically, by deploying a disposable scratch
+collection through each candidate encoding and inspecting what Cosmos actually materialized, that
+none succeed — the resource silently either misinterprets the input as a literal (and incorrect)
+field name or ignores it outright.
+
+The same experimentation also settled a related question: what happens when an index a collection
+resource is expected to manage is left out of its declared index list. ARM's declarative
```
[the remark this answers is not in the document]
btposey: ✅ Fixed. Updated to "what happens when an index that a collection resource is expected to manage is left out of its declared index list."

fmaddenflx (approved):

ops/scripts/pipeline/az-cosmos-deploy.sh
```diff
@@ -120,3 +120,18 @@ az deployment group create -w -g "${resourceGroup}" -f ./ops/cloud-deployment/us
...
+# that Cosmos DB Mongo API's Bicep/ARM `keys` array cannot express (see the
+# NOTE in cosmos-collections.bicep). index-trustee-case-appointments.js owns
+# that collection's non-default indexes and is idempotent -- safe to run on
+# every deploy. Runs against both the main database and the e2e database,
+# since cosmos-collections.bicep provisions trustee-case-appointments in both.
+# The connection string is passed via MONGO_CONNECTION_STRING rather than a
+# CLI argument so it never appears in `ps` output or shell history on the CI
+# runner.
+MONGO_CONNECTION_STRING=$(az keyvault secret show --vault-name "${keyVaultName}" --name MONGO-CONNECTION-STRING --query value -o tsv)
+export MONGO_CONNECTION_STRING
```
jamesobrooks, line 135: `MONGO_CONNECTION_STRING` is fetched from Key Vault and exported here without `::add-mask::`, unlike every other Key Vault value pulled in this same job in `reusable-infrastructure-deploy.yml` (resource-group names, workspace IDs, etc. are all explicitly masked before use, even though they're less sensitive than a Mongo connection string). GHA's automatic secret redaction only covers values sourced from the `secrets:` context, not arbitrary command output like this `az keyvault secret show` result — so if a future debugging change ever echoes this variable (e.g. `set -x`), it wouldn't be redacted.

Suggest:
```bash
MONGO_CONNECTION_STRING=$(az keyvault secret show --vault-name "${keyVaultName}" --name MONGO-CONNECTION-STRING --query value -o tsv)
echo "::add-mask::${MONGO_CONNECTION_STRING}"
export MONGO_CONNECTION_STRING
```

ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  const client = new MongoClient(connectionString);
+  try {
+    await client.connect();
+    const db = client.db(databaseName);
+
+    for (const spec of TARGET_INDEXES) {
+      const response = await db.command({
```
jamesobrooks: This `db.command({ createIndexes: ... })` call has no `maxTimeMS` and no retry/backoff. Two related gaps:

- `RunCommandOperation` (what backs `db.command()`) doesn't declare `Aspect.RETRYABLE`, so the driver's `retryWrites`/`retryReads` never kick in here — a transient Cosmos RU-throttle response (code 16500) fails the whole deploy immediately instead of retrying.
- `socketTimeoutMS` defaults to `0` (unbounded) once a socket is open, so a hang after a successful connection has no bound at this layer — it'd rely entirely on GitHub Actions' default job timeout with no proactive signal.

This operation moved from a human-supervised manual script to unattended CI in this PR, which is what changes the risk calculus here. Suggest adding `maxTimeMS` to the command and wrapping it in a small bounded retry that specifically catches the Cosmos throttling error code.

ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
```diff
@@ -0,0 +1,111 @@
...
+// full declarative replace, dropping anything not listed. Since the ORDER BY
+// dateFiled DESC, caseId ASC sort used by getCasesForTrustee needs a mixed
+// ascending/descending composite index, which cannot be expressed in Bicep
+// at all, the `indexes` property is omitted ENTIRELY from
+// trusteeCaseAppointmentsCollection in cosmos-collections.bicep so ARM never
+// reconciles indexes on this collection. This script owns all of that
+// collection's non-default indexes instead. `_id` and the `trusteeId` shard
+// key are auto-indexed by Cosmos and are not managed here.
+//
+// This was verified empirically (see backend/_experiments/bicep-index-probe)
```
jamesobrooks: `backend/_experiments/bicep-index-probe` doesn't exist anywhere in the repo (checked `git log --all --diff-filter=A` — it was never committed) and isn't gitignored despite the PR description saying so. Since the empirical claim itself is already corroborated in the PR description and the ADR's prose, can we just drop this specific path reference here (and the matching one in `cosmos-collections.bicep`) rather than pointing at something that doesn't resolve?

ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
```diff
@@ -0,0 +1,111 @@
...
+  {
+    name: 'dateFiled_-1_caseId_1',
+    key: { dateFiled: -1, caseId: 1 },
+  },
+];
+
+async function main() {
+  const args = process.argv.slice(2);
+  // One argument: `<databaseName>`, connection string must come from the env var.
+  // Two arguments: `<connection-string> <databaseName>` (CLI form).
```
jamesobrooks: Confirming intent here rather than flagging a bug: the two-argument CLI form re-opens the shell-history/`ps` exposure that the env-var switch was specifically added to close (per the commit message), and the only real caller (`az-cosmos-deploy.sh`) exclusively uses the one-arg env-var form. I see from the review thread this was a deliberate choice to support manual/ad-hoc runs, per Sourcery's suggestion — just want to confirm the team is still good with that trade-off long-term, since as written it reads like a security mitigation that's opted back out of for convenience. If so, might be worth saying that explicitly in the comment ("kept for manual/ad-hoc operator convenience" rather than "for backward compatibility") so a future reader doesn't assume it's leftover, unjustified code.

jamesobrooks (commented):
Adversarial multi-agent review (pragmatic, architecture, devsecops, SRE reviewers + independent skeptics on every surviving finding). Summary and full disposition table shared out-of-band; inline comments below cover the items worth addressing before/shortly after merge.

ops/scripts/pipeline/az-cosmos-deploy.sh
btposey: ✅ Fixed in 3020c7759. Now masks `MONGO_CONNECTION_STRING` with `::add-mask::` immediately after fetching it from Key Vault, before exporting, matching every other secret fetch in this job.

ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
btposey: ✅ Fixed in 3020c7759. Added `maxTimeMS` on the `createIndexes` command and wrapped it in a bounded retry (5 attempts, exponential backoff) that specifically catches Cosmos error code 16500 (RU throttling) and re-throws anything else immediately. Confirmed `RunCommandOperation` is not `Aspect.RETRYABLE` in the driver source, so this was a real gap -- verified the retry logic with an isolated test (succeeds after N throttled attempts; fails fast on non-throttle errors).

ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
btposey: ✅ Fixed in 3020c7759. Youre right -- that directory was local-only scratch work and was never going to be committed. Dropped the dangling reference here and the matching one in cosmos-collections.bicep; pointed both at the ADR instead, which already carries the rationale.

ops/cloud-deployment/lib/cosmos/mongo/index-trustee-case-appointments.js
btposey: Good catch, and fair pushback -- removed the two-argument form entirely in 3020c7759 rather than re-label it. You were right that keeping it read like an unjustified opt-out of the exact mitigation the commit was for. The script now only accepts MONGO_CONNECTION_STRING; az-cosmos-deploy.sh (the only real caller) already only used that form.

PR #2672, opened 2026-07-14
Remove redundant and always-pass tests across user-interface

jalba-flexion:
# Purpose

Eliminate test suite noise: duplicate tests, always-passing assertions, and tautological tests that give false confidence in CI without covering any real code paths.

# Major Changes

* Removed ~40 tests across 27 files spanning admin, case-detail, data-verification, lib, login, search, staff-assignment, and trustees subsystems
* Categories of removals: verbatim copy-paste blocks, strict-subset tests, tautological tests (asserting test fixture behavior), always-pass assertions (querying elements guaranteed to be present regardless of the condition under test), and tests that never render the component they claim to test
* One always-failing test assertion fixed: `DataVerificationScreenAlert` now actually asserts the transfer option is absent when the flag is off, instead of asserting an unrelated always-present button
* Two `CaseDetailTrusteePanel` zoomInfo tests strengthened to assert the presence/absence of actual zoomInfo content, not just the card wrapper

# Testing/Validation

Full test suite passes (229 test files, 3277 tests). Coverage is unchanged: statements 97.46% (was 97.47%), branches 92.65% (was 92.67%), functions 96.72% (unchanged), lines 97.96% (was 97.97%). The near-zero delta confirms the removed tests contributed no real coverage.

# Notes

Notable removals by category:

**Verbatim copy-paste:** `SoftwareVendorContactInfoForm.test.tsx` had 11 tests duplicated in full (lines 228–313 copied word-for-word to 315–400).

**Tautological / never-renders-component:** `SessionEnd.test.tsx` had a test that manually called `navigatorMock.navigateTo(path)` then asserted `navigateTo` was called with that path — asserting the behavior of the test mock itself. `TrusteeAppointmentForm.helpers.test.ts` had 4 tests running `Array.some()` inline with hardcoded literals — no production code involved.

**Always-pass assertion:** `DataVerificationScreenAlert.test.tsx` tested "transfer option absent when flag off" by asserting an expand button that's always present regardless of feature flags. Now asserts the transfer option is actually absent.

**Disconnected component tests:** Two `SearchScreen` debtor-name tests called `validateFormData` with a hardcoded literal after doing `userEvent.type` on the component — the component state was never read. Replaced with direct unit test calls.

# Definition of Done:

## Summary by Sourcery

Clean up and tighten UI-related tests by removing redundant or tautological coverage and strengthening a few behavior-focused assertions.

Bug Fixes:
- Correct DataVerificationScreenAlert test to assert the absence of the transfer option when the feature flag is off, rather than relying on an unrelated always-present control.

Enhancements:
- Strengthen CaseDetailTrusteePanel zoomInfo tests to validate actual zoom information content and absence/presence behavior, not just container rendering.

Tests:
- Remove numerous redundant, tautological, or trivially true tests across admin, login, data-verification, search, staff-assignment, trustees, and shared UI utilities to reduce noise without affecting coverage.
- Refocus debtor name validation tests in SearchScreen to exercise the validation logic directly via validateFormData instead of indirectly through disconnected UI interactions.
- Tidy several utility and component tests (e.g., clipboard, debounce hook, court sorting, zoomInfo formatting, trustee filters, header navigation) by dropping edge-case assertions that did not add meaningful coverage and adjusting a header selector expectation.


btposey (approved):

cruzjone-flexion (approved):

mattstankey:
Two coverage concerns worth a second look before merging — both are edge-case branches on security/session-expiry logic that don't clearly fit the PR's stated removal categories (duplicate / tautological / always-pass / disconnected):

1. **`okta-library.test.ts` — removed `isTokenCloseToExpiry` "no expiry field" test.** `isTokenCloseToExpiry` (`okta-library.ts:64-72`) has `if (!session || !session.expires) return true`. The file keeps a "no session" test but the *only* test for the "session exists but `expires` is undefined" branch was deleted outright — no duplicate of it remains elsewhere. This looks like a real branch-coverage loss on a path that guards against malformed/incomplete session state.

2. **`OktaSession.test.tsx` — removed "JWT missing `exp`" test**, keeping only "JWT missing `iss`". Source: `if (!oktaJwt.payload.iss || !oktaJwt.payload.exp)` (`OktaSession.tsx:71`). Both operands hit the identical `return <AccessDenied />`, so this one's more defensible than #1, but it's still a coverage reduction on an auth-guard branch that isn't called out in the PR description.

Could you confirm these two were intentional? If so, a one-line note in the PR description tying them to one of the existing removal categories would close the loop for future readers.

jalba-flexion:
@mattstankey Yes, confirmed intentional — both restored in commit e7a04e3f6.

1. **`isTokenCloseToExpiry` "no expiry field" test** — restored. You're right that this is a distinct branch of the `!session || !session.expires` condition that wasn't covered elsewhere; the remaining "no session" test only exercised the first operand.
2. **`OktaSession` "JWT missing `exp`" test** — restored as well, alongside "missing `iss`", to keep both operands of that auth-guard `||` condition covered even though they converge on the same `AccessDenied` outcome.

Everything else that was removed (clipboard edge cases, zoomInfo boundary strings, the `okta-library` `isActive()` tests) was verified against the underlying implementation — no branch-coverage loss there, just duplicate coverage of trivial pass-throughs or logic that moved to `session-timer.ts` and is already tested there.

PR #2691, opened 2026-07-15
CAMS-816 Backfill trustee professional IDs via inverse ACMS pass

### LINKED ISSUE #2687: Backfill trustee-professional-ids by matching from ACMS professional records outward
## Overview

The `trustee-professional-ids` collection maps CAMS trustees to their ACMS professional IDs. It is currently populated **only** as a side effect of MIGRATE-TRUSTEES: `upsertProfessionalIds` (`migrate-trustees.ts`, ~line 707) starts from each ATS trustee record and queries ACMS by exact `(firstName, lastName, state)` match against CMMPR (`PROF_TYPE = 'TR' OR NULL`) — see `acms.gateway.ts` `getTrusteeProfessionalIds` (~line 265).

This matching is **one-directional (ATS → ACMS)**: it only ever visits ACMS professional records reachable from an ATS trustee. ACMS professional records that have no corresponding ATS trustee (or whose ATS name/state didn't align) are never visited, so their case appointments cannot be resolved to a CAMS trustee and are stamped with the sentinel `trusteeId` — which is why sentinel case appointments remain in CAMS.

This task adds the **inverse (ACMS → CAMS)** pass — starting from the full set of ACMS professional records and finding CAMS trustees still missing a mapping — so the mapping set is comprehensive. It directly increases the mappings available to the sentinel-heal work.

## Current Behavior

- `trustee-professional-ids` is populated one CAMS trustee at a time, looking outward from ATS to ACMS.
- Per `/cams-gateways`: ACMS professional IDs use a compound key `(GROUP_DESIGNATOR, PROF_CODE)` formatted `"{GROUP_DESIGNATOR}-{PROF_CODE:5}"` (e.g. `"NY-00063"`). **One CAMS trustee can hold multiple ACMS professional IDs** (across groups); model the relationship as one-CAMS-trustee-to-many-ACMS-IDs, not 1:1. Never filter by `PROF_CODE` alone — always use the compound key.

## Required Behavior

**1. Add a `heal` intent to the MIGRATE-TRUSTEES start message.** Extend `MigrationStartMessage` (`migrate-trustees.ts`) with a `heal` flag, handled in `handleStart` alongside the existing `flushQueues` / `deleteAll` / `reset` intents. The `heal` intent triggers the professional-ID backfill logic (below) rather than a full trustee migration.

**2. New CMMPR query for the full professional-record set.** Add a new ACMS gateway query (distinct from `getTrusteeProfessionalIds`, which filters by name/state) that pulls the **full set** of trustee professional records from CMMPR filtered by `PROF_TYPE = 'TR'`, independent of ATS, keyed on the compound `(GROUP_DESIGNATOR, PROF_CODE)`.

**3. Backfill missing mappings.** For each ACMS professional record **without** an existing `trustee-professional-ids` entry (check via `repo.findByAcmsProfessionalId`), attempt to match it to a CAMS trustee (name + state; same matching rules as today, improved if warranted). Create new `trustee-professional-ids` mappings for matches the original ATS-driven pass missed. `repo.createProfessionalId(camsTrusteeId, acmsProfessionalId, user)` is already idempotent (checks for the existing `(camsTrusteeId, acmsProfessionalId)` pair before inserting).

**4. Report unmatched via a dedicated queue.** Route ACMS professional records that still could not be matched to a CAMS trustee to a **dedicated unmatched-professional-ids queue** (mirroring the existing `FAILED_APPOINTMENTS` queue pattern in the MIGRATE-TRUSTEES handler). The report is then retrieved by draining that queue to a blob file via the existing `flushQueues` intent — the same mechanism used to produce `flush-failed-appointments-*.jsonl`.

**5. Idempotency.** Re-running the `heal` intent creates no duplicate mappings and produces no undesired side effects; a subsequent run converges to the same state as the first successful run.

## Acceptance Criteria

- [ ] `heal` intent added to `MigrationStartMessage` and handled in `handleStart`
- [ ] New CMMPR query returns the full ACMS trustee professional-record set filtered by `PROF_TYPE = 'TR'`, using the compound `(GROUP_DESIGNATOR, PROF_CODE)` key (never `PROF_CODE` alone)
- [ ] Backfill matches ACMS professionals → CAMS trustees for records missing a mapping
- [ ] New mappings created in `trustee-professional-ids` for matches found this way
- [ ] One-CAMS-trustee-to-many-ACMS-IDs relationship handled (not assumed 1:1)
- [ ] Unmatched ACMS professional records are routed to a dedicated queue and retrievable via `flushQueues` (as with failed appointments)
- [ ] Idempotent — no duplicate mappings created for professionals already mapped, and re-runs have no undesired side effects

bityogi:
# Purpose

Make the `trustee-professional-ids` mapping set comprehensive so ACMS case appointments can be resolved to CAMS trustees, reducing the sentinel-stamped appointments that remain because the mapping was only ever built one-directionally (ATS → ACMS). Closes #2687.

# Major Changes

* Added a `heal` intent to the MIGRATE-TRUSTEES start message, handled in `handleStart` alongside the existing `flushQueues` / `deleteAll` / `reset` intents.
* Added `getAllTrusteeProfessionalRecords` to the ACMS gateway: pulls the full CMMPR trustee professional-record set filtered by `PROF_TYPE = 'TR'`, keyed on the compound `(GROUP_DESIGNATOR, PROF_CODE)` professional ID (never `PROF_CODE` alone).
* Added the `backfillProfessionalIds` use case: for each ACMS professional record without an existing mapping (`findByAcmsProfessionalId`), match it to a CAMS trustee by name + state and create the mapping idempotently. Handles the one-CAMS-trustee-to-many-ACMS-IDs relationship (records processed independently).
* Routed ACMS professional records that still can't be matched to a dedicated `unmatched-professional-ids` queue, drained to a blob via the existing `flushQueues` intent (mirrors the `failed-appointments` pattern; produces `flush-unmatched-professional-ids-*.jsonl`).
* Extracted `runFlushQueues` / `runHeal` / `runDeleteAll` helpers from `handleStart` to keep it within the cognitive-complexity limit.

# Testing/Validation

Implemented using TDD. Added unit coverage for:
* `backfillProfessionalIds` — create-on-match, skip-already-mapped (idempotency), unmatched routing with reason codes (`NO_TRUSTEE_MATCH`, `INCOMPLETE_NAME_OR_STATE`, `LOOKUP_FAILED`, `CREATE_FAILED`), one-to-many mapping, mixed-page counter aggregation, and gateway failure.
* `getAllTrusteeProfessionalRecords` — full set returned, compound-key/`PROF_TYPE` filter, extended request timeout, empty set, error wrapping.
* `handleStart` heal intent — backfill invoked, unmatched records routed to the queue, backfill failure routed to DLQ with preserved error, and success/failure telemetry. Updated the `flushQueues` tests for the new fifth queue.

Full backend suite: 3314 tests pass; backend coverage 95.75% statements / 90.67% branches (`migrate-trustees.ts` at 98%). Typecheck and lint clean on changed files.

# Notes

* Idempotent by design: already-mapped ACMS records are skipped (no lookup), and `createProfessionalId` de-dupes the `(camsTrusteeId, acmsProfessionalId)` pair, so re-running `heal` converges to the same state.
* Matching reuses the existing name+state rules (`findTrusteeByNameAndState`), consistent with the original ATS-driven pass.
* The `unmatched-professional-ids` queue is registered as an extra output on `handleStart` and is now also drained by `flushQueues`.
* Unrelated: the BDD suite currently fails to resolve `path-to-regexp` (via `express@5 → router`) on `main`; this is pre-existing and not touched by this branch.

# Definition of Done:

## Summary by Sourcery

Add a heal intent to trustee migration to backfill trustee-professional-ids from ACMS, route unmatched records for review, and expand integration coverage for migration and Zoom CSV import flows.

New Features:
- Introduce heal intent on MIGRATE-TRUSTEES start to trigger an ACMS-driven trustee-professional-ids backfill.
- Expose an ACMS gateway operation to retrieve all trustee professional records keyed by compound professional ID.
- Add an import-zoom-csv integration harness and supporting scripts to validate Zoom CSV ingestion end-to-end.

Enhancements:
- Extend migrate-trustees handleStart to support professional-ID backfill and unmatched-professional-ids output while refactoring intent handling for clarity.
- Augment migration integration harness to cover active trustee filtering, migration state lifecycle, CAMS-772 archive-date overrides, and heal backfill scenarios.

Tests:
- Add unit tests for the professional-ID backfill use case, including idempotency, error handling, and aggregate counters.
- Add tests for the new ACMS gateway professional-record fetch behaviour and migrate-trustees heal/flushQueues handling.
- Introduce new SQL seed fixtures and integration scripts to drive trustee migration and Zoom CSV import scenarios end-to-end.

Chores:
- Remove legacy trustee migration testing scripts now superseded by the new integration harnesses.


btposey:
## Review: architecture, YAGNI/pragmatic, and test-quality pass

Ran three parallel specialist reviews against this PR (architecture, pragmatic/YAGNI, test quality). Summary below, with follow-up work tracked as beads for traceability. Filing as follow-up beads rather than blocking here since the mapping logic, idempotency design, and error routing in the core `backfillProfessionalIds` implementation are sound — the main concern is the `heal` intent's execution model at production scale.

### High priority

**`heal` intent needs to be paginated before running against production ACMS volume** — `backfillProfessionalIds()` (`migrate-trustees.ts:1148`) pulls the entire CMMPR professional-record set into memory and processes it serially inside a single synchronous `handleStart` invocation (`runHeal`, `migrate-trustees.ts:180`). Every other backfill in this codebase (`backfill-phonetic-tokens.ts`, `backfill-case-appointment-dates.ts`, `backfill-trustee-appointments-downstream.ts`, and others) uses `CursorMessage`-driven pagination with `MigrationStateService`-tracked resumability specifically to avoid the 1-hour Function timeout and survive partial failures. This diverges from that established pattern with no cursor, no persisted progress, and no recovery if it times out mid-scan.

Agreed redesign with @btposey: a reader step does one ACMS fetch (~10k records), chunks the result set in memory, and enqueues one page message per chunk to a new `heal-page` queue. A new `handleHealPage` handler processes each chunk with its own smaller retry ceiling (429-aware backoff) and proper continuation/resume via `MigrationStateService`. Unmatched records route to a new `heal-unmatched-professional-ids` queue, kept separate from the `unmatched-professional-ids` queue already added in this PR.

Tracked as:
- `cams-2cc4` — reader/chunking half (new `heal-page` queue, rewritten `runHeal`)
- `cams-l3e8` — `handleHealPage` handler (continuation, retry, unmatched routing)
- `cams-zxws` — trim `runFlushQueues` to failure/DLQ/unmatched queues only (drops START/PAGE from the flush list, adds `heal-unmatched-professional-ids`)

**`dumpQueueToBlob` is a duplicated helper with a divergent bugfix** — the copy in this PR (`migrate-trustees.ts:67-116`) calls `queueClient.deleteMessage` after reading each message; the pre-existing original in `migrate-case-appointments.ts:153-183` does not, meaning its queue never actually drains on flush. This PR silently fixes the bug only in its own copy. Recommend extracting one shared helper and fixing the original's missing `deleteMessage` call as part of that extraction.

Tracked as: `cams-xubc`

### Medium priority — test gaps

Test quality review found the core matching/idempotency/reason-code logic is well covered (all 4 reason codes, idempotency, mixed-page aggregation, and gateway failure are genuinely asserted, not incidental — good work). Gaps found:
- `vi.restoreAllMocks()` in `afterEach` instead of `beforeEach` (`migrate-trustees.test.ts:63-65`) — can leak mock state if a prior test throws
- No test for an empty ACMS record set, or for whitespace-only name/state fields that only become empty after `.trim()`
- No test for duplicate `acmsProfessionalId` within one page
- Heal handler tests never assert `scanned`/`alreadyMapped` in trace details or `documentsFailed` on the partial-unmatched path
- `acms.gateway.test.ts:637-639` asserts on literal SQL substrings rather than behavior — brittle to harmless query reformatting

Tracked as: `cams-j3ya` (deliberately sequenced after the heal redesign lands, since `backfillProfessionalIds` is being restructured and some of these tests will move/change shape)

### Cleared as non-issues (pragmatic/YAGNI pass)

The four `UnmatchedProfessionalId` reason codes, the `runFlushQueues`/`runHeal`/`runDeleteAll` extraction from `handleStart`, and the new `import-zoom-csv-harness.ts` integration coverage were all reviewed for over-engineering/YAGNI violations and found justified — each maps to genuine, distinguishable, actionable behavior rather than speculative generalization.

backend/lib/use-cases/dataflows/migrate-trustees.ts
```diff
@@ -1104,6 +1104,170 @@ async function writeUnmatchedProfessionalIds(
...
+ * missing a mapping, making the mapping set comprehensive.
+ *
+ * Idempotent: records that already have a `trustee-professional-ids` entry are
+ * skipped, and `repo.createProfessionalId` de-dupes the `(camsTrusteeId,
+ * acmsProfessionalId)` pair, so re-runs converge to the same state.
+ *
+ * ACMS records that still cannot be matched to a CAMS trustee are returned in
+ * `unmatched` for the handler to route to the dedicated unmatched queue.
+ */
+export async function backfillProfessionalIds(
```
btposey: **Architecture finding (High):** this pulls the entire ACMS CMMPR professional-record set into memory and processes it serially in a single pass, with two awaited DB round-trips per record and no batching. Six other backfills in this codebase (`backfill-phonetic-tokens.ts`, `backfill-case-appointment-dates.ts`, `backfill-trustee-appointments-downstream.ts`, etc.) use `CursorMessage`-driven pagination with `MigrationStateService`-tracked resumability for exactly this workload shape, to avoid the 1-hour Function timeout and survive partial failures.

Agreed redesign: a reader step does one ACMS fetch (~10k records is an acceptable single-fetch size), chunks the result set in memory, and enqueues one page message per chunk to a new `heal-page` queue — no per-page ACMS re-query. Tracked in **cams-2cc4** (reader/chunking) and **cams-l3e8** (page handler with continuation, retry, unmatched routing).

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -112,13 +119,185 @@ function requiresDeleteAll(start: MigrationStartMessage): boolean {
...
+  });
+}
+
+/**
+ * heal intent: backfill trustee-professional-ids by matching outward from the
+ * full ACMS professional-record set to CAMS trustees. Unmatched records are
+ * routed to the dedicated unmatched-professional-ids queue (retrieved via the
+ * flushQueues intent, mirroring the failed-appointments pattern).
+ */
+async function runHeal(
```
btposey, line 160: **Architecture finding (High):** `runHeal` runs the entire backfill synchronously within one `handleStart` invocation — no cursor, no persisted progress, no requeue. If this times out partway through a ~10k-record scan, all progress is lost with no resume point.

Agreed redesign (see **cams-2cc4**, **cams-l3e8**): `runHeal` becomes a reader that chunks the ACMS result set and enqueues page messages; a new `handleHealPage` handler processes each chunk with its own retry ceiling (1 retry, 429-aware backoff — distinct from the existing `MAX_RETRY_ATTEMPTS`) and tracks progress via `MigrationStateService` for proper continuation/resume. Unmatched records route to a new `heal-unmatched-professional-ids` queue, kept separate from `unmatched-professional-ids`.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
```diff
@@ -112,13 +119,185 @@ function requiresDeleteAll(start: MigrationStartMessage): boolean {
...
+ * unmatched-professional-ids queue produced by the heal intent.
+ */
+async function runFlushQueues(context: ApplicationContext, trace: HandleStartTrace): Promise<void> {
+  const { logger } = context;
+  logger.info(MODULE_NAME, 'flushQueues flag detected — dumping queues to blob storage.');
+  const objectStorage = factory.getObjectStorageGateway(context);
+  const outputContainerName = buildContainerName(MODULE_NAME, 'out');
+  const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
+
+  const queuesToFlush = [
```
btposey, line 110: **Design change (per discussion with @btposey):** `queuesToFlush` should drop the START and PAGE entries entirely — a healthy running migration's start/page queues should be empty/transient, and `flushQueues` should be purely a failure/DLQ/unmatched inspection tool going forward (for both the ATS migration and heal, not just heal-specific scoping). The new `heal-unmatched-professional-ids` queue (from **cams-l3e8**) should be added to this trimmed list; the new `heal-page` continuation queue should NOT be added (paging queues are excluded by design). Also worth noting: `dumpQueueToBlob` (defined above this function) duplicates `migrate-case-appointments.ts`'s helper of the same name, and the two have diverged — this copy calls `deleteMessage` after reading, the original doesn't, so the original's queue never actually drains on flush. Recommend extracting one shared helper into `dataflows-common.ts` and fixing that bug in the original. Tracked in **cams-zxws** (flush trim) and **cams-xubc** (dumpQueueToBlob dedup).

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
```diff
@@ -615,6 +615,60 @@ describe('ACMS gateway tests', () => {
...
+      });
+
+      const context = await createMockApplicationContext();
+      const gateway = new AcmsGatewayImpl(context);
+      const result = await gateway.getAllTrusteeProfessionalRecords(context);
+
+      expect(result).toEqual(dbResults);
+
+      // The compound (GROUP_DESIGNATOR, PROF_CODE) key must be used — never PROF_CODE alone.
+      const query = spy.mock.calls[0][1] as string;
```
btposey, line 641: **Test quality finding (High):** asserting on literal SQL substrings (`toContain('GROUP_DESIGNATOR')`, `"PROF_TYPE = 'TR'"`) couples this test to query text formatting rather than behavior — it'll break on harmless SQL reformatting, and doesn't actually verify filtering since the mock returns canned data regardless of the query. Tracked along with other test gaps in **cams-j3ya** (sequenced after the heal redesign lands, since this code path is being restructured).

btposey (commented):
Automated architecture/pragmatic/test-quality review — see summary comment for full context.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
bityogi: Done in 701016ad. `runHeal` is now a reader (one ACMS fetch → chunk in memory → one `heal-page` message per chunk, records embedded), and a new `handleHealPage` handler processes each chunk with in-place 429 backoff + escape-hatch re-enqueue and its own retry ceiling (distinct from `MAX_RETRY_ATTEMPTS`). Unmatched → new `heal-unmatched-professional-ids` queue.

One divergence from the bead as written: heal progress is tracked via **flat `heal*` fields on `TRUSTEE_MIGRATION_STATE`** rather than a new `HEAL_PROFESSIONAL_IDS_STATE` documentType. Reason: dataflows `host.json` sets queue `batchSize: 10`, so the enqueued pages run concurrently, and `atomicIncrement`/`atomicDecrement` (the only race-safe way to update counters) operate on top-level fields only. Flat fields keep blast radius zero on the shared repo method used by migrate-case-appointments/trustees. Completion is driven by an atomic `healRecordsRemaining` countdown.

backend/function-apps/dataflows/migrations/migrate-trustees.ts
bityogi: Done in 701016ad. `queuesToFlush` is trimmed to failure/DLQ/unmatched only — `dlq`, `failed-appointments`, `unmatched-professional-ids`, and the new `heal-unmatched-professional-ids`. START and PAGE are dropped; `heal-page` is deliberately excluded (paging queues are never flushed).

On the `dumpQueueToBlob` divergence you flagged: extracted a single shared helper into `dataflows-common.ts`, parameterized on connection string + queue/blob/container, and fixed the migrate-case-appointments copy that never called `deleteMessage` — so its flush now actually drains the queue. Added a regression test asserting `deleteMessage` fires on that path.

backend/lib/adapters/gateways/acms/acms.gateway.test.ts
bityogi: Partially addressed here, remainder tracked in cams-j3ya (which you sequenced to land after this redesign). In this commit I fixed the near-useless timeout assertion (`toBeGreaterThan(0)` → asserts the actual configured `ACMS_REQUEST_TIMEOUT_MS` is threaded through) and folded the use-case test-gap findings into the restructured code: `vi.restoreAllMocks()` moved to `beforeEach`, empty-set case, whitespace-only trim, and duplicate `acmsProfessionalId` within one chunk.

The literal-SQL-substring assertions here (#6) and the `test.each` dedup (#8) I intentionally left for cams-j3ya, since `getAllTrusteeProfessionalRecords` itself was not restructured by this change and those are better done as a focused pass. Integration coverage now also exercises the real CMMPR query end-to-end (Step 9 of the migrate-trustees harness), which verifies filtering behaviorally rather than by query text.

backend/lib/use-cases/dataflows/migrate-trustees.ts
bityogi: Done in 701016ad (cams-2cc4). The monolithic `backfillProfessionalIds` is replaced by a reader + page-processor split: `readAllTrusteeProfessionalRecords` does the single ACMS fetch (~10k acceptable), `runHeal` chunks the result in memory and enqueues one `heal-page` message per chunk (records embedded, no per-page re-query), and `backfillProfessionalIdsPage` runs the per-record matching for a chunk with 429 in-place backoff + escape hatch. See the companion reply on `runHeal`/`handleHealPage` for the state-tracking divergence (flat `heal*` fields on `TRUSTEE_MIGRATION_STATE` instead of a new documentType, driven by concurrent page handlers under `batchSize: 10`).

mattstankey:
## Review notes on the `heal` paging design

Ran a multi-angle review (architecture, reliability, security, simplicity) against this diff. Security and OeSA boundary compliance came back clean (no injection risk, no new secrets, Good-Fences respected on queue DTOs). Given the paged reader→queue→handler design is the settled architecture here (per the CAMS-816 architecture review that drove the last commit), the actionable items are about making that design correct and consistent with the rest of the dataflow's paging conventions — not about reconsidering it.

### Should fix before merge

1. **Counter race in `recordHealPageResult`** (`trustee-migration-state.service.ts:79-108, 354-396`) — the atomic `$inc` on `healCreated`/`healAlreadyMapped`/`healUnmatched`/`healRecordsRemaining` is immediately followed by `updateMigrationState`, which does a non-atomic read-then-`replaceOne` of the whole document. With `batchSize: 10` concurrent page handlers, one handler's full-document write can revert another's atomic increment mid-flight. `migrate-case-appointments.ts`'s `incrementMetric` already avoids this exact pattern with an explicit warning comment — this regresses that lesson. Recommend a targeted `$set` for just `healStatus`/`healLastUpdatedAt` instead of round-tripping the full document, or derive `COMPLETED` purely from the atomic increment's return value.

2. **Silent record loss on re-enqueue failure** (`migrate-trustees.ts:1054-1101, 1165-1167`) — `handleHealPage` doesn't check `reEnqueueHealPage`'s boolean return. If the re-enqueue send fails, the deferred records are dropped with no error, no DLQ entry, and no counter update — they just vanish and the run appears stuck `IN_PROGRESS` forever with no diagnosable cause.

3. **No jitter on `computeHealBackoffMs`** (`migrate-trustees.ts` use-case, `HEAL_BASE_DELAY_MS`/backoff calc) — up to 10 concurrent pages compute backoff deterministically from the same formula, so a Cosmos 429 under load causes synchronized retry-storms across pages. `reEnqueueHealPage` already adds jitter elsewhere in this same file — worth applying the same treatment here.

4. **No poison-queue handler for `heal-page`** (`migrate-trustees.ts:1230-1236`) — every other queue handler in this dataflow family (`handle-missed-division-changes.ts`, `division-change-cleanup.ts`, `sync-cases.ts`, `sync-trustee-case-appointments.ts`) registers a `-poison` handler for messages that exceed `maxDequeueCount`. `HANDLE_HEAL_PAGE` doesn't, so an unexpected error in `backfillProfessionalIdsPage` silently strands messages in `heal-page-poison` with nothing draining, logging, or alerting on it.

5. **Dead `unmatched-professional-ids` queue** (`migrate-trustees.ts:642-645, 724-727`) — declared and drained in `runFlushQueues`, but nothing in the diff ever writes to it; all unmatched records actually go to `heal-unmatched-professional-ids`. The PR description says records route to the former, but the shipped code routes to the latter. Recommend deleting the unused queue.

### Worth a look

6. **`migrate-case-appointments.ts` still drains live `START`/`PAGE` queues via `flushQueues`** — this PR correctly excludes `START`/`PAGE` from `runFlushQueues` for `migrate-trustees` (since draining them mid-run via the shared `dumpQueueToBlob` would delete live continuation messages and silently halt an in-progress migration), but the sibling dataflow still flushes those same queue types with the newly-shared helper. Looks like an oversight rather than an intentional difference.

7. **No resumable cursor for the heal reader** — the rest of this dataflow's paging uses a cursor-based `CursorMessage` (`{ lastId, importAll? }`) that's naturally resumable on restart. The heal reader does a single upfront full-table fetch and embeds records directly into `HealPageMessage`, which is reasonable for the current ~10k-row scale but has no equivalent resumability if `runHeal` itself fails partway through enqueuing. Worth a short comment noting this as a deliberate scale-based tradeoff (or adding one) so a future maintainer extending the `PAGE` convention elsewhere doesn't assume resumability applies here too.

### Scope note

`test/integration/import-zoom-csv/` (596-line new harness + Azurite/env scaffolding) exercises `importZoomCsv`, which is unrelated to the professional-ID backfill this PR is scoped to. Consider splitting it into its own PR against whichever ticket owns that use case, so review of the `heal` logic isn't diluted and the Zoom CSV harness gets review attention of its own.

---
No blocking security findings. `dumpQueueToBlob` extraction and the `runFlushQueues`/`runHeal`/`runDeleteAll` split out of `handleStart` are solid, well-scoped refactors — nice cleanup.

btposey:
## Follow-up review: commit `701016adb` ("paginate heal intent for resilience")

Reviewed the paginated redesign against the concerns raised in the earlier review round. The reader/writer split, escape-hatch retry logic, `dumpQueueToBlob` dedup, and `runFlushQueues` trim all landed correctly and are well-tested. Remaining gaps below, ranked by relevance to observable backfill completion (the actual success criterion for this work):

### 1. Hard crashes bypass the escape hatch's own requeue mechanism

`handleHealPage`'s escape hatch (`processRecordWithRetry`/`healShouldEscape`/`reEnqueueHealPage`) correctly re-enqueues `remaining` records with a visibility delay when a 429 backoff would exceed the wall-clock budget — that path is sound. But the outer `catch` block (`migrate-trustees.ts`, function-apps layer) handles a different failure mode: an actual thrown exception (not a 429, something that crashes `backfillProfessionalIdsPage` or a `MigrationStateService` call outright). On that path, it traces `success: false` and then `throw err`, handing the whole message back to Azure's platform-level redelivery/poison-queue mechanism instead of calling `reEnqueueHealPage` with the same records. That means there are two different recovery mechanisms for two different failure types — the well-built one (escape hatch, tracked via `remaining`, visibility delay + jitter) and an opaque platform fallback with no equivalent tracking. Worth asking: should the catch block route through the same `reEnqueueHealPage` mechanism instead of relying on `throw` + Azure's default retry/poison-queue behavior, so there's one resilience mechanism instead of two?

### 2. Non-atomic write can revert concurrently-applied atomic counters

`recordHealPageResult` (`trustee-migration-state.service.ts`) correctly increments `healCreated`/`healAlreadyMapped`/`healUnmatched`/`healRecordsRemaining` via atomic `$inc`, but immediately follows with `updateMigrationState` — a full-document read → spread → `replaceOne` — to stamp `healLastUpdatedAt`/`healStatus`. Under concurrent `handleHealPage` invocations (`batchSize: 10`), one handler's stale read can silently overwrite another handler's already-committed increment on `replaceOne`. Worst case this manifests as `healStatus` never flipping to `COMPLETED` even though all records finished — which bears directly on knowing when the backfill is actually done. Recommend folding the timestamp into the same atomic operation (Mongo supports combining `$inc` and `$set` in one `findOneAndUpdate`), and making the `COMPLETED` transition its own atomic conditional update rather than routing through the read-modify-write helper.

### 3. Escape-hatch record slicing has no test proving correctness past the first record

Only one escape-hatch test exists (`migrate-trustees.test.ts`, use-case layer), and it triggers on record 0 of the chunk, so `remaining = records.slice(0)` trivially equals the entire input. Manual trace of the loop logic looks correct (the triggering record is included in `remaining` and not yet counted toward `created`/`alreadyMapped`/`unmatched`), but nothing exercises the case where N records succeed before the escape hatch fires mid-chunk, which is the only scenario that could actually reveal an off-by-one dropping or duplicating a record.

### 4. Question for the implementer: was a bulk read considered for the professional-ID idempotency check?

`processAcmsRecord` still does a per-record `findByAcmsProfessionalId` exact-match query against Cosmos for each of the ~10k ACMS records, same as before pagination — pagination solved the timeout/resumability problem but not the RU-cost problem. This repository already exposes `findAll()`, so a single bulk read into an in-memory `Set<acmsProfessionalId>` per heal run could replace thousands of per-record exact-match queries with one, directly cutting RU cost. (For contrast: `findTrusteeByNameAndState`, the other per-record lookup, is a regex-based scatter-gather query with fuzzy composite-name matching that genuinely resists this kind of caching — so this question is specific to the professional-ID check, not the trustee-matching lookup.) Was a bulk-read approach considered and set aside for a reason (e.g., staleness risk if new mappings are created mid-run by another process), or is this open for a follow-up?

### 5. Prefer deleting the false part of a comment over correcting it

The flat `heal*`-fields comment on `TrusteeMigrationState` claims MongoDB "only operates on top-level fields" for atomic `$inc` — this is incorrect; MongoDB natively supports dotted-path atomic increments. The actual constraint is this codebase's own `atomicIncrement` repository helper's TypeScript signature (`keyof T & string`), not MongoDB itself. That specific claim should just be deleted rather than reworded — a false technical claim is worse than no comment. The remaining part of the comment (documenting that escape-hatch-deferred records are intentionally not double-counted in `healRecordsRemaining` until their re-enqueued page runs) is worth keeping as-is — that's a genuinely non-obvious invariant that isn't derivable from the field names or types alone, unlike the MongoDB claim.

btposey (changes requested):
See comments above.

mattstankey (approved):

btposey:
## Follow-up: closed remaining gaps from the CHANGES_REQUESTED review

Went back through the 5 "should fix before merge" findings from my 2026-07-17T21:05 review round. 2 were already landed by the later heal-redesign commits (record-loss fix, dead-queue removal). Of the remaining 3, plus a related design change, here's what was done and why — closing this out rather than leaving it open-ended:

**1. Jitter on `computeHealBackoffMs`** (`migrate-trustees.ts` use-case layer)
Was purely deterministic (`2^(attempt+1) * baseDelayMs`), so concurrent `heal-page` handlers hitting a 429 at the same moment would all compute the identical delay and retry in lockstep — a synchronized retry storm against Cosmos. Now applies full jitter (uniform random in `[0, computedDelay)`), matching the jitter already used on the escape-hatch re-enqueue path. Exported the function and added a direct unit test proving non-determinism (50 samples, >1 distinct value) plus a ceiling-bound test.

**2. In-memory professionalId→Trustee mapping cache**
Per Brian's direction: rather than the reader-built/embedded-per-page design, `backfillProfessionalIdsPage` now does one `findAll()` bulk read at the start of each page instead of one Cosmos point-read (`findByAcmsProfessionalId`) per record — cutting per-page Cosmos calls by ~100x (the `HEAL_PAGE_SIZE`). Safe because ACMS professional records never share an `acmsProfessionalId` across the result set, so a page-start snapshot can't go stale from that page's own writes; the in-page cache is also updated immediately after each `created` outcome so a defensive same-page duplicate ID still resolves correctly. All existing `backfillProfessionalIdsPage` tests updated to mock `findAll()` instead of the now-removed per-record calls (101 → 108 passing after also adding new coverage).

**3. Outer `catch` in `handleHealPage` now routes through the same escape-hatch mechanism**
Previously, a hard exception (as opposed to the handled `pageResult.error` path) fell straight to `throw err`, handing the whole message to Azure's opaque platform-level redelivery/poison-queue behavior — a second, untracked recovery mechanism alongside the well-built escape hatch. Now the outer catch calls `reEnqueueHealPage` with the full original page (re-processing is idempotent — already-mapped records are skipped via the cache, `createProfessionalId` de-dupes on write) and only falls back to `throw` if the re-enqueue itself fails. Extracted into `handleHealPageCrash` to keep `handleHealPage`'s cognitive complexity from creeping past the lint threshold. Two new tests cover both the successful-reenqueue and re-enqueue-failure-falls-back-to-throw paths.

**4. Escape-hatch test coverage past the first record**
The only existing escape-hatch test triggered on index 0, so `remaining: records.slice(i)` trivially equaled the whole input — never proving correctness when N records succeed before the hatch fires. Added a test with one record created successfully before a second record's every retry attempt hits a 429, asserting `created: 1` and a non-trivial `remaining` slice starting at index 1.

**Deliberately not done — counter race in `recordHealPageResult`:** Brian's decided this migration won't use `MigrationStateService`/runtime-state tracking for heal at all going forward, so the atomic-`$inc`-then-read-modify-write race in `updateMigrationState` is moot for this flow and any fix would target code being removed. Left as-is; will be addressed as part of that broader change if/when it lands.

**Deliberately not done — poison-queue handler for `heal-page`:** considered and explicitly declined. This is a one-time migration being monitored manually, not a recurring dataflow; a poison handler adds a new queue + error path + tests for a failure mode that, if it occurs, will be visibly investigated by hand rather than silently ignored. Can be added later if `heal-page-poison` actually accumulates messages in practice.

**Verification:** full backend suite (3393 tests via `npm test`), typecheck, and lint all pass clean — no new errors, only pre-existing warnings (the `sonarjs/pseudo-random` warning on the new jitter code is expected/intentional).

btposey (approved):

PR #2709, opened 2026-07-17
CAMS-809 Add DXTR-backed trustee petition match integration test

mattstankey:
# Purpose

Add a real-database integration test for `SyncTrusteeCaseAppointmentsUseCase`, covering the DXTR read path (`getTrusteeAppointments` / `getTrusteePetitionEvents`) and the professional-id match + case-appointment write path, none of which any existing integration test in this repo actually exercises against a real DXTR instance.

# Major Changes

* Added `test/integration/trustee-petition-match/`, scaffolded on the `migrate-trustees` pattern: hand-crafted DXTR DDL (`AO_CS_DIV`/`AO_CS`/`AO_PY`/`AO_TX` — no DXTR DDL existed anywhere in this repo before) covering only the columns the two gateway queries actually use, fixture seed data for one case with both an appointment and a petition transaction, and a harness with `seed-schema`/`seed-sql`/`seed-cosmos`/`run`/`clean` commands
* Fixed two production bugs found while getting the test to actually run against a real database:
  * `CasesDxtrGateway` used `FORMAT(...)` and `AT TIME ZONE`, both of which require CLR support unavailable on some SQL Server editions (e.g. Azure SQL Edge, the only arm64-native SQL image, used by every other local integration test in this repo); replaced with `CONVERT(..., 126)` and dropped the no-op `AT TIME ZONE` annotations (both sides were already naive UTC values, so this is behavior-neutral)
  * `TrusteeCaseAppointmentsMongoRepository.getActiveByCaseId` threw instead of returning `null` when no active appointment exists — contradicting its own declared `Promise<CaseAppointment | null>` type and breaking every first-time trustee-case appointment in production. Unit tests never caught this because they mock this method directly (`mockResolvedValue(null)`), bypassing the real adapter's throw-on-not-found behavior

# Testing/Validation

* Full pipeline verified end-to-end locally against a real Azure SQL Edge + MongoDB container: schema creation, fixture seeding, DXTR read (both event types), professional-id fast-path matching, case-appointment write, and cleanup all pass
* `cases.dxtr.gateway.test.ts` (85 tests) and `trustee-case-appointments.mongo.repository.test.ts` (62 tests) pass after the production fixes; full `npm test` across `common`/`backend`/`user-interface` (3343 + 3250 tests) passes
* Ran `/cams-spec-review` on the one modified test file — the change on this branch (an assertion updated to match the new SQL) is correctly aligned; other findings reported are pre-existing and out of scope for this PR
* `npm run typecheck`, `npm run lint` (scoped), `npm run coverage:backend`, `npm audit`, and `npm run knip` all pass

# Notes

The two production fixes are in a separate commit (`e77192ca4`) from the test scaffolding (`942ff5c58`) so they can be reviewed/reverted independently — they were found by, but are logically distinct from, the new test.

# Definition of Done:

## Summary by Sourcery

Introduce a new trustee petition match integration harness that seeds minimal DXTR schema and Cosmos fixtures to validate SyncTrusteeCaseAppointmentsUseCase end-to-end, while fixing DXTR gateway SQL compatibility issues and correcting trustee appointment repository not-found handling.

New Features:
- Add a DXTR-backed integration test harness for SyncTrusteeCaseAppointmentsUseCase that exercises trustee appointment and petition events end-to-end against real SQL and Cosmos.

Bug Fixes:
- Adjust CasesDxtrGateway datetime formatting and comparison to avoid CLR-only SQL functions and ensure compatibility with Azure SQL Edge.
- Update TrusteeCaseAppointmentsMongoRepository.getActiveByCaseId to return null on not-found errors instead of throwing, aligning behavior with its declared return type.

Enhancements:
- Extend DXTR gateway tests to assert the updated SQL predicate used for transaction recency filtering.

Build:
- Add npm scripts to run the new trustee-petition-match integration harness in local and Azure environments.

### DIFF
--- a/backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.test.ts
+++ b/backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.test.ts
@@ -1513,9 +1513,7 @@ describe('Test DXTR Gateway', () => {
       expect(result).toEqual(['081-20-10508', '081-21-12345']);
       expect(querySpy).toHaveBeenCalledWith(
         expect.anything(),
-        expect.stringContaining(
-          "TX.TX_DATE AT TIME ZONE 'UTC' > C.LAST_UPDATE_DATE AT TIME ZONE 'UTC'",
-        ),
+        expect.stringContaining('TX.TX_DATE > C.LAST_UPDATE_DATE'),
         expect.arrayContaining([
           expect.objectContaining({ name: 'cutoffDate', value: '2018-01-01' }),
         ]),
--- a/backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.ts
+++ b/backend/lib/adapters/gateways/dxtr/cases.dxtr.gateway.ts
@@ -631,10 +631,10 @@ class CasesDxtrGateway extends AbstractMssqlClient implements CasesInterface {
     const casesQuery = `
       SELECT
         CONCAT(CS_DIV.CS_DIV_ACMS, '-', C.CASE_ID) AS caseId,
-        FORMAT(C.LAST_UPDATE_DATE AT TIME ZONE 'UTC', 'yyyy-MM-ddTHH:mm:ss.fff') + 'Z' AS latestSyncDate
+        CONVERT(VARCHAR(23), C.LAST_UPDATE_DATE, 126) + 'Z' AS latestSyncDate
       FROM AO_CS C
       JOIN AO_CS_DIV AS CS_DIV ON C.CS_DIV = CS_DIV.CS_DIV
-      WHERE C.LAST_UPDATE_DATE AT TIME ZONE 'UTC' > @casesStart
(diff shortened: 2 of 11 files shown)


mattstankey:
Addressed both suggestions in 8d66f3fc5:

1. `seed-schema` now reads `MSSQL_DATABASE_DXTR` (defaulting to `DXTR_INT`) for both the `CREATE DATABASE` statement and the schema-apply connection, instead of hard-coding `DXTR_INT` — matches every other command in the harness.
2. The `TRUSTEE_APPOINTMENTS_SYNC_STATE`/`TRUSTEE_PETITION_SYNC_STATE` docs can't be scoped to a single case — they're dataflow-wide singleton watermarks with no caseId field. Rather than leave that implicit, added an explicit code comment and a README "Warning: isolated databases only" section stating this harness must only be run against an isolated local/test Cosmos database.

Re-verified the full pipeline end-to-end locally after both changes.

btposey (approved):

PR #2711, opened 2026-07-20
CAMS-814 Constrain trustee case list division filter

mattstankey:
# Purpose

The Trustee Case List's District/Division filter defaulted its options and selection to the logged-in user's own office divisions, not the trustee's. A reviewer covering a trustee outside their own office divisions saw the wrong (or no) filter options.

# Major Changes

* Added `GET /trustees/{trusteeId}/divisions`, returning the distinct districts/divisions a trustee has case appointments in, across all case statuses (never filtered by open/closed) — new repository aggregate, use-case method, controller, and route.
* Added an opt-in `divisionCodeAllowList` prop on the shared `DistrictDivisionComboBox` that restricts its option set (including correct handling of the per-district "select all" pseudo-option) and defaults selection to the full allow list. The prop is additive only — the component's existing behavior is unchanged when it isn't passed, so the other two consumers (Trustee District Filter, Search Screen) are unaffected.
* Wired `TrusteeCaseList` → `TrusteeCaseListFilter` → `TrusteeCaseListFilterView` to fetch the trustee's divisions on mount and pass them through; the combo box is not rendered until the divisions have resolved (avoiding a flash of unfiltered options), and falls back to an empty (non-erroring) state if the trustee has no cases or the fetch fails.
* Added seed data (`cams-814-single-division`, `cams-814-division-closed`) for manual verification of a single-division trustee and a trustee whose one division has all cases closed.

# Testing/Validation

Implemented using TDD (tests written alongside each change, run continuously). Backend: 3359 tests passing, including new repository/use-case/controller test coverage for the distinct-divisions query (asserting no status filter is applied). Frontend: 3259 tests passing (+3 skipped), including new coverage for the allow-list filtering/defaulting logic, the per-district "All" edge case, the fetch-then-render gating, and the error-fallback path. Confirmed the two other `DistrictDivisionComboBox` consumers' existing tests pass unmodified (no regression). `npm run typecheck`, `npm run build` (backend + frontend), and `npm run lint` all pass repo-wide.

Ran `/cams-spec-review` against all 5 changed test files. No critical or high-severity issues found in the new test code; most flagged items were pre-existing conventions in these files (e.g. an established `{ selector: 'li span' }` pattern, existing pipeline-structure assertions) unrelated to this change. One genuine gap in new code was found and fixed: added a test for the `getTrusteeCaseDivisions` rejection path.

# Notes

Slice 1 of a 3-slice feature (see the linked spec). Remaining slices: zero-case trustee empty state (largely already covered by this slice's fetch-gating behavior, per implementation notes) and verifying the trustee-scoped default survives return navigation without clobbering a manual selection.

The seed data changes were not run against a live database in this session (no local DXTR/Mongo available) — worth a manual `npm run seed:local` check before merge.

# Definition of Done:

## Summary by Sourcery

Constrain the Trustee Case List district/division filter to the trustee’s own case divisions via a new backend divisions endpoint and updated filter wiring, ensuring correct options and defaults for cross-office reviewers.

New Features:
- Expose GET /trustees/{trusteeId}/divisions to return the trustee’s distinct case division codes.
- Add an opt-in divisionCodeAllowList capability to DistrictDivisionComboBox to limit options and default selections to a provided set.
- Wire TrusteeCaseListFilter to fetch trustee-specific case divisions by trusteeId and gate rendering of the district combo box on those results.

Enhancements:
- Extend TrusteeCasesUseCase and Mongo repository with a distinct-division query for trustees, without case-status filtering.
- Update TrusteeCaseListFilter types and view to support trustee-scoped division allow lists and non-rendered empty states.
- Add mock and API client support for trustee case divisions, consistent with existing trustee endpoints.

Tests:
- Add backend tests covering the distinct-division aggregate, error handling, and use-case delegation.
- Add frontend tests for divisionCodeAllowList behavior, trustee division fetching, gating/empty/error states, and rerendering on trusteeId changes.
- Extend trustee-case-list seed scenario data with single-division and all-closed-division trustees for manual and automated verification.

### DIFF
--- a/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
@@ -228,6 +228,28 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     }
   }
 
+  async getDistinctDivisionsForTrustee(trusteeId: string): Promise<string[]> {
+    try {
+      const doc = using<CaseAppointmentDocument>();
+      const match = toMongoQuery(and(doc('trusteeId').equals(trusteeId)));
+
+      const mongoAggregate = [
+        { $match: match },
+        { $group: { _id: null, divisions: { $addToSet: '$courtDivisionCode' } } },
+      ];
+
+      const collection = this.trusteePartition.collection<CaseAppointmentDocument>();
+      const cursor = await collection.aggregate(mongoAggregate);
+      const result = await cursor.next();
+
+      return (result?.divisions ?? []).filter((code: string | undefined) => !!code);
+    } catch (originalError) {
+      throw getCamsErrorWithStack(originalError, MODULE_NAME, {
+        message: `Failed to retrieve distinct divisions for trustee ${trusteeId}.`,
+      });
+    }
+  }
+
   private buildPrePaginateMatch(trusteeId: string, predicate: TrusteeCasesSearchPredicate) {
     const conditions: ConditionOrConjunction<CaseAppointmentDocument>[] = [
       apptDoc.field('trusteeId').equals(trusteeId),
--- a/user-interface/src/lib/components/DistrictDivisionComboBox.tsx
+++ b/user-interface/src/lib/components/DistrictDivisionComboBox.tsx
@@ -108,31 +108,29 @@ const DistrictDivisionComboBox_ = (
             ) as ComboOption[])
           : nationalOptions;
 
+        const applyDefaults = (defaults: ComboOption[]) => {
+          if (defaults.length > 0) {
+            const codes = encodeDivisionCodes(defaults, allCourts);
+            setSelectedDivisions(defaults);
+            previousSelectionsRef.current = defaults;
+            onDivisionCodesChange?.(codes);
+            onSelectionsChange?.(defaults);
+          }
+        };
+
         let defaults: ComboOption[] = [];
         if (initialDivisionCodes?.length) {
           defaults = allOptions.filter((opt) => {
             const [, code] = opt.value.split('|');
             return code !== 'ALL' && initialDivisionCodes.includes(code);
           });
-          if (defaults.length > 0) {
-            const codes = encodeDivisionCodes(defaults, allCourts);
-            setSelectedDivisions(defaults);
-            previousSelectionsRef.current = defaults;
-            onDivisionCodesChange?.(codes);
-            onSelectionsChange?.(defaults);
-          }
+          applyDefaults(defaults);
         } else if (divisionCodeAllowList) {
           defaults = allOptions.filter((opt) => {
             const [, code] = opt.value.split('|');
             return code !== 'ALL';
           });
-          if (defaults.length > 0) {
-            const codes = encodeDivisionCodes(defaults, allCourts);
-            setSelectedDivisions(defaults);
-            previousSelectionsRef.current = defaults;
-            onDivisionCodesChange?.(codes);
-            onSelectionsChange?.(defaults);
-          }
+          applyDefaults(defaults);
         } else {
           const userCodes = getUserDivisionCodes(LocalStorage.getSession());
           if (userCodes.size > 0) {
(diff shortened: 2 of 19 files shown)


mattstankey:
Thanks for the review, @sourcery-ai.

- ✅ Fixed: `TrusteeCaseListFilter` now resets `divisionCodeAllowList` to `undefined` at the start of the `trusteeId` effect, so the district combo box hides while divisions are refetched instead of briefly showing the previous trustee's divisions (commit 773e9f7b8).
- Skipped for now: memoizing `divisionsByCourtId` in `filterOptionsToAllowList` — the national court list is small and this isn't in a hot render loop, so we're treating it as a non-issue for this PR.

user-interface/src/lib/components/DistrictDivisionComboBox.tsx
```diff
@@ -65,35 +99,46 @@ const DistrictDivisionComboBox_ = (
         const allCourts = r.data;
         setCourts(allCourts);
         onCourtsLoaded?.(allCourts);
-        const allOptions = getDistrictDivisionComboOptions(allCourts) as ComboOption[];
+        const nationalOptions = getDistrictDivisionComboOptions(allCourts) as ComboOption[];
```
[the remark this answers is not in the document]
mattstankey: ✅ Fixed! Added a `parseDivisionOptionValue` helper to centralize the `value.split('|')` parsing, and split the default-selection logic into named helpers (`computeInitialDivisionDefaults`, `computeAllowListDefaults`, `computeUserOfficeDefaults`) that each express one scenario in the initial-codes → allow-list → user-office priority order (commit 8492397946e739e1a17862f77b3fc778366fe28c).

mattstankey:
Re: the remaining high-level comment about `divisionCodeAllowList` being `[]` for zero-division/error cases while the view gates on `!== undefined`.

No code change needed here — `DistrictDivisionComboBox` already guards against this: when the (possibly allow-list-filtered) option set resolves to zero options, it returns `null` (see the `divisionComboOptions.length === 0` check), so no combo box is rendered in the zero-division/error case. The `!== undefined` gate in `TrusteeCaseListFilterView` is only meant to avoid rendering during the initial loading state before the allow list resolves; once it resolves to `[]`, the child component itself hides. Verified this is covered by the existing zero-division/error tests in `TrusteeCaseListFilter.test.tsx`.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -228,6 +228,28 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
 
+  async getDistinctDivisionsForTrustee(trusteeId: string): Promise<string[]> {
+    try {
+      const doc = using<CaseAppointmentDocument>();
+      const match = toMongoQuery(and(doc('trusteeId').equals(trusteeId)));
+
+      const mongoAggregate = [
+        { $match: match },
+        { $group: { _id: null, divisions: { $addToSet: '$courtDivisionCode' } } },
+      ];
```
btposey, line 239: I think this is good. I was considering just using the division+chapter appointments, but this returns all possible divisions for all appointed cases regardless.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
```diff
@@ -58,10 +58,14 @@ export interface TrusteeCaseListFilterViewModel extends TrusteeCaseListFilterHan
   filterAnnouncement: string;
   selectedDivisions: ComboOption[];
   initialDivisionCodes?: string[];
+  // undefined until the trustee's case divisions have been fetched; the district
+  // combo box is not rendered until this is resolved (possibly to an empty array).
```
btposey:
```suggestion
```

btposey (approved):
Approved with a single nit-pick.

user-interface/src/trustees/panels/filters/trusteeCaseListFilter.types.ts
mattstankey: ✅ Fixed! Removed the comment (commit a757db070).

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
mattstankey: Thanks for the context — agreed, this is intentional. Getting all divisions across all appointed cases (regardless of chapter/status) is what we want here so the filter reflects every division the trustee has ever had a case in.

PR #2733, opened 2026-07-21
CAMS-824 Add typed phone numbers for trustee contacts

fmaddenflx:
# Purpose

Trustees, internal contacts, and staff assistants previously supported only a single phone number. This adds typed phone numbers (direct/cell/home) for trustee internal contacts and staff assistants, and unifies phone display into a single reusable component.

# Major Changes

- Consolidated single/multiple phone display logic into `FormattedContact` via a generic `phones` prop, replacing duplicated logic that previously lived in `ContactInformationCard` and `TrusteeDetailAuditHistory`.
- Added typed phone number support (direct/cell/home) for trustee staff contacts, gated behind the `TRUSTEE_TYPED_PHONES` feature flag, mirroring the existing internal-contact behavior.
- Added front-end and back-end validation for typed phone rows (no duplicate types, valid number/extension format).
- Added a migration script (`CAMS-824-staff-contact-phone-types.js`) to backfill staff contact phone data.
- Fixed CSS alignment/spacing bugs in the trustee staff card and mobile/desktop form button groups.
- Fixed two latent shared-validator bugs (`arrayOf` silently dropping spec-based validator errors; `noDuplicatePhoneTypes` double-nesting its reason map) that also affected the already-shipped internal-contact phones.
- Fixed four latent error-message bugs in the mongo trustees repository where the error message was only set inside `camsStackInfo` and never surfaced at the top level, so failures reported a generic "Unknown Error" instead of a descriptive one.

# Testing/Validation

Implemented using TDD throughout. Ran a full TDD-quality spec review (`/cams-spec-review`) across all 14 test files changed on this branch and addressed every HIGH/MEDIUM finding: added missing repository/component test coverage, removed boundary-crossing and assertion-free tests, replaced disallowed selectors with `data-testid`, and consolidated near-duplicate tests. Full validation suite (typecheck, unit tests, eslint, knip, npm audit) passes clean across the `common`, `backend`, and `user-interface` workspaces (3,345+ backend/common tests, 3,290 user-interface tests).

# Notes

- Chose to extend `FormattedContact`'s `phones` prop rather than build a separate `TypedPhoneDisplay` component, since single/multiple-phone rendering was already unified there earlier in this branch.
- 3 pre-existing high-severity `npm audit` findings (`brace-expansion`, `fast-uri`, `immutable` — all transitive dependencies) are unrelated to this branch and were not addressed here.
- Filed a follow-up issue for pre-existing flaky tests discovered in the dataflows migration test suite during validation (unrelated to this branch's changes).

# Definition of Done:


common/src/cams/trustees-validators.ts
```diff
@@ -111,13 +111,39 @@ export const contactInformationSpec: ValidationSpec<ContactInformation> = {
...
 };
 
-export const staffContactInformationSpec: ValidationSpec<ContactInformation> = {
-  ...contactInformationSpec,
-  address: [V.optional(V.spec(addressSpec))],
+export const typedPhoneNumberSpec: ValidationSpec<TypedPhoneNumber> = {
+  number: [phoneNumber],
+  extension: [phoneExtension],
+  type: [V.checkFirst(V.minLength(1, 'Phone type is required'))],
+};
```
cruzjone-flexion, line 118: **`staffContactSpec` duplicates `internalContactSpec` identically**

`TrusteeStaffContact` is declared as a direct alias for `TrusteeInternalContact` (`export type TrusteeStaffContact = TrusteeInternalContact`), yet two structurally identical specs were created for them. If one spec is ever updated without the other, validation will silently diverge between internal contacts and staff contacts.

```typescript
// current
export const staffContactSpec: ValidationSpec<TrusteeStaffContact> = {
  phones: [V.optional(V.arrayOf(V.spec(typedPhoneNumberSpec))), noDuplicatePhoneTypes],
  email: [V.optional(V.nullable(email))],
};
```

Since the types are identical, `staffContactSpec` can just reference the existing one:

```typescript
export const staffContactSpec: ValidationSpec<TrusteeStaffContact> = internalContactSpec;
```

If they ever need to diverge, the type alias should become a distinct type at that point.

user-interface/src/trustees/forms/TrusteeInternalContactForm.tsx
```diff
@@ -94,7 +104,10 @@ function TrusteeInternalContactForm(props: Readonly<TrusteeInternalContactFormPr
...
               }
             : null,
-        phone: formData.phone ? { number: formData.phone, extension: formData.extension } : null,
+        phones:
+          formData.phones.filter((p) => p.number.trim()).length > 0
+            ? formData.phones.filter((p) => p.number.trim())
+            : undefined,
         email: formData.email ?? null,
       },
     } as Partial<TrusteeInput>;
```
cruzjone-flexion, line 113: **`phones` filter predicate runs twice — assign once**

The `phones` filtering in the payload builder evaluates the same predicate twice: once to test length, then again to produce the filtered array. `TrusteeStaffForm` handles this correctly with a single assignment.

```typescript
// current
phones:
  formData.phones.filter((p) => p.number.trim()).length > 0
    ? formData.phones.filter((p) => p.number.trim())
    : undefined,
```

```typescript
// suggested
const filteredPhones = formData.phones.filter((p) => p.number.trim());
phones: filteredPhones.length > 0 ? filteredPhones : undefined,
```

common/src/cams/trustees-validators.ts
cruzjone-flexion: ✅ Fixed. Rather than aliasing one spec to the other (which knip flags as duplicate exports), I renamed both to a single `trusteeContactSpec` shared by both internal and staff contacts. Updated all callers (`trustees.ts`, `staffInputSpec`) and collapsed the duplicate test blocks.

user-interface/src/trustees/forms/TrusteeInternalContactForm.tsx
cruzjone-flexion: ✅ Fixed. Assigned to `filteredPhones` once and used it in the ternary.

PR #2734, opened 2026-07-22
CAMS-817 heal re-resolves sentinel case appointments

bityogi:
# Purpose

Sentinel `CASE_APPOINTMENT` documents (stamped with `SENTINEL_TRUSTEE_ID` when an ACMS appointment's `acmsProfessionalId` didn't resolve to a CAMS trustee at migration time) were stuck with the placeholder `trusteeId` indefinitely. As professional-ID gaps get backfilled (#2687), these docs need a healing step to become real appointments.

# Major Changes

* Extended `heal()` so that, in addition to its existing partition-parity repair, it re-resolves sentinel documents against `trustee-professional-ids`:
  * Scans each batch for docs with `trusteeId === SENTINEL_TRUSTEE_ID`
  * Looks up `acmsProfessionalId` via `findByAcmsProfessionalId`
  * **Exactly one** match → rewrites `trusteeId` in both partitions and clears the sentinel `reason`
  * **Zero / multiple / missing-id** → left untouched and counted for visibility
* Added a new repository method `resolveSentinelTrusteeId` on `TrusteeCaseAppointmentsRepository`. Because `trusteeId` is the trustee-partition key, resolution is a cross-partition *move* (delete sentinel row + upsert resolved doc), not an in-place replace. The case partition is replaced in place, preserving `_id`.
* Added optional `sentinelResolvedCount` / `sentinelUnresolvedCount` counters to `HealCaseAppointmentsState`, persisted per batch and surfaced in the completion log.

# Testing/Validation

Implemented using TDD (RED → GREEN → REFACTOR). `resolveSentinelDocs` is exported and unit-tested directly across the resolve / no-match / multiple-match / missing-id / historical paths, with assertions on the returned counts and resolved-id set. Repository tests cover the cross-partition move (delete targets sentinel, resolved doc written to both partitions), the crash-safe write ordering (trustee before case), `NotFoundError` tolerance, and the non-`NotFoundError` failure branch. `heal`-level integration tests verify the parity pass skips just-resolved docs and that counts persist to state. Verified the new count assertions bite via a mutation check. Full backend suite: 3402 tests pass; typecheck, lint, and knip clean; coverage 96.29%.

# Notes

* **Idempotency & resumability:** resolution runs inside `heal()`'s existing paginated, cursor-resumable loop and respects `SAFE_THRESHOLD_MS`. Resolved docs no longer carry the sentinel, so re-runs skip them. The trustee partition is written *before* the case partition, so an interrupted run leaves the sentinel visible to the case-partition scan and re-resolves cleanly without double-writing.
* Docs resolved within a batch are excluded from the parity-repair pass to avoid recreating a stale sentinel row in the trustee partition.
* Depends on #2687 (already merged to main) for there to be meaningfully more mappings to heal against.

# Definition of Done:

## Summary by Sourcery

Add healing of sentinel CASE_APPOINTMENT documents during the migrate-case-appointments heal process and track resolution metrics.

New Features:
- Introduce resolveSentinelDocs to re-resolve sentinel case appointments using trustee-professional-id mappings within heal batches.

Bug Fixes:
- Ensure partition-parity repair skips just-resolved sentinel documents to avoid recreating stale sentinel rows in the trustee partition.

Enhancements:
- Extend the heal workflow to process sentinel documents before partition parity checks and persist per-batch sentinel resolution counts in HealCaseAppointmentsState.
- Add ScannedCaseAppointment and CaseAppointmentDocument shapes to surface sentinel-specific fields while keeping the public API type constrained.

Tests:
- Add unit and integration tests for resolveSentinelDocs, heal sentinel integration, and the new resolveSentinelTrusteeId repository behavior, including crash-safety and error handling scenarios.

Chores:
- Update TrusteeCaseAppointmentsRepository, its Mongo implementation, and the mock repository to support resolveSentinelTrusteeId with backward-compatible heal state changes.

### DIFF
--- a/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
@@ -519,6 +519,75 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     }
   }
 
+  /**
+   * resolveSentinelTrusteeId — rewrites a sentinel appointment (trusteeId ===
+   * SENTINEL_TRUSTEE_ID) to a now-resolved trustee across both partitions.
+   *
+   * The trustee partition is keyed by trusteeId, so a resolved doc belongs in a
+   * different partition than the sentinel — this is a move, not an in-place
+   * replace. We delete the sentinel-partition row and upsert the resolved doc.
+   * The case partition is keyed by caseId, so the resolved doc replaces the
+   * sentinel row in place (its _id is preserved by the natural-key match).
+   *
+   * Trustee partition is written FIRST: if the process crashes before the case
+   * partition replace, the sentinel remains visible to heal's scan (which reads
+   * the case partition) and a re-run resolves it again idempotently. A missing
+   * sentinel row in the trustee partition (already deleted by a prior partial
+   * run) is tolerated.
+   */
+  async resolveSentinelTrusteeId(
+    sentinelKey: { caseId: string; assignedOn: string },
+    resolvedDocument: CaseAppointmentDocument,
+  ): Promise<void> {
+    const doc = using<CaseAppointmentDocument>();
+    try {
+      // 1. Trustee partition: remove the stale sentinel-partition row.
+      const sentinelTrusteeQuery = and(
+        doc('documentType').equals('CASE_APPOINTMENT'),
+        doc('caseId').equals(sentinelKey.caseId),
+        doc('trusteeId').equals(SENTINEL_TRUSTEE_ID),
+        doc('assignedOn').equals(sentinelKey.assignedOn),
+      );
+      try {
+        await this.trusteePartition
+          .adapter<CaseAppointmentDocument>()
+          .deleteOne(sentinelTrusteeQuery);
+      } catch (deleteError) {
+        // Idempotent: a prior partial run may have already removed it.
+        if (!isNotFoundError(deleteError)) {
+          throw deleteError;
+        }
+      }
+
+      // 2. Trustee partition: upsert the resolved doc into its new partition.
+      const resolvedNaturalKey = and(
+        doc('documentType').equals('CASE_APPOINTMENT'),
+        doc('caseId').equals(resolvedDocument.caseId),
+        doc('trusteeId').equals(resolvedDocument.trusteeId),
+        doc('assignedOn').equals(resolvedDocument.assignedOn),
+      );
+      await this.trusteePartition
+        .adapter<CaseAppointmentDocument>()
+        .replaceOne(resolvedNaturalKey, resolvedDocument, true);
+
+      // 3. Case partition: replace the sentinel row in place (caseId partition is
+      // stable, so the existing _id is preserved by the natural-key match).
+      const sentinelCaseQuery = and(
+        doc('documentType').equals('CASE_APPOINTMENT'),
+        doc('caseId').equals(sentinelKey.caseId),
+        doc('trusteeId').equals(SENTINEL_TRUSTEE_ID),
+        doc('assignedOn').equals(sentinelKey.assignedOn),
+      );
+      await this.casePartition
+        .adapter<CaseAppointmentDocument>()
+        .replaceOne(sentinelCaseQuery, resolvedDocument, true);
+    } catch (originalError) {
+      throw getCamsErrorWithStack(originalError, MODULE_NAME, {
+        message: `Failed to resolve sentinel appointment for case ${sentinelKey.caseId}.`,
+      });
+    }
+  }
+
   private async findByCursor<T>(
     query: ConditionOrConjunction<T>,
     options: { limit: number; sortField: keyof T; sortDirection: 'ASCENDING' | 'DESCENDING' },
--- a/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
+++ b/backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
@@ -32,6 +32,9 @@ const TRUSTEE_COLLECTION = 'trustee-case-appointments';
 
 const CASES_COLLECTION = 'cases';
 
+// documentType discriminator for appointment docs in both partitions.
+const CASE_APPOINTMENT_DOC_TYPE = 'CASE_APPOINTMENT' as const;
+
 const { using, and } = QueryBuilder;
 const { source } = QueryPipeline;
 
@@ -108,7 +111,7 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     try {
       const doc = using<CaseAppointmentDocument>();
       const query = and(
-        doc('documentType').equals('CASE_APPOINTMENT'),
+        doc('documentType').equals(CASE_APPOINTMENT_DOC_TYPE),
         doc('caseId').equals(caseId),
       );
       return await this.casePartition.adapter<CaseAppointmentDocument>().find(query);
@@ -123,7 +126,7 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
     try {
       const doc = using<CaseAppointmentDocument>();
       const query = and(
-        doc('documentType').equals('CASE_APPOINTMENT'),
+        doc('documentType').equals(CASE_APPOINTMENT_DOC_TYPE),
         doc('caseId').equals(caseId),
         doc('unassignedOn').notExists(),
         doc('trusteeId').notEqual(SENTINEL_TRUSTEE_ID),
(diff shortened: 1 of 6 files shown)


bityogi:
Thanks @sourcery-ai — both suggestions had merit and are addressed in 78e51d22b:

1. **Batch lookups** — `resolveSentinelDocs` now memoizes `findByAcmsProfessionalId` per heal batch (`lookupCache` + `resolvedTrusteeByAcmsId`). Sentinels for the same unmapped trustee cluster across many cases, so repeated `acmsProfessionalId`s now share a single Cosmos read. Added a unit test asserting one lookup is shared across two same-id sentinels.

2. **Log noise** — replaced the per-document `logger.info` calls with a single aggregated batch-level summary listing all unresolved sentinels, keeping per-doc diagnostic detail without the per-record flood.

mattstankey:
A couple of minor notes from review, neither blocking:

1. **Mock repository typed as `any`** (`mock-mongo.repository.ts:231`): `resolveSentinelTrusteeId(_sentinelKey: any, _resolvedDocument: any)` loses the type safety the interface (`gateways.types.ts:784-787`) otherwise provides. Worth typing against the same shape used in the interface, if neighboring mock methods in this file don't already follow the `any` convention.

2. **No batch-level failure isolation** in `resolveSentinelDocs`/`heal()`: if `resolveSentinelTrusteeId` throws on doc N in a batch (e.g. a genuine non-`NotFoundError` write failure), the whole `heal()` invocation aborts, and resolutions completed earlier in that batch aren't reflected in `totalSentinelResolved`/persisted state since `updateHealState` only runs after the batch completes. On retry, the cursor restarts at the same `lastId` and re-attempts in the same order, so a single bad doc could stall healing of an otherwise-healthy batch. This mirrors the existing repair loop's error-handling posture, so it's consistent rather than a regression — just flagging in case it's worth a follow-up issue.

bityogi:
Thanks @mattstankey — both had merit; addressed in 6ccbf8e10:

1. **Mock typed as `any`** — typed `resolveSentinelTrusteeId`'s params in `MockMongoRepository` against the interface shape (`{ caseId, assignedOn }` and `CaseAppointment & { documentType: 'CASE_APPOINTMENT' }`).

2. **Batch-level failure isolation** — good catch, and I decided to fix it now rather than defer. Both heal passes now isolate per-doc write failures in try/catch:
   - **Sentinel resolution**: a failed `resolveSentinelTrusteeId` counts the doc as unresolved and continues; the doc keeps its sentinel and is retried on a later run.
   - **Partition-parity repair**: a failed `replaceOneInTrusteePartition` is logged and the doc is left diverged for a later run.

   This means one poison doc can no longer abort the invocation or strand progress already made in the batch — directly serving the resumable/idempotent criteria. Since the gap was shared with the pre-existing repair loop, I fixed both paths together for consistency. Added unit + integration tests for each (mutation-verified they bite).

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -519,6 +519,75 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+   */
+  async resolveSentinelTrusteeId(
+    sentinelKey: { caseId: string; assignedOn: string },
+    resolvedDocument: CaseAppointmentDocument,
+  ): Promise<void> {
+    const doc = using<CaseAppointmentDocument>();
+    try {
+      // 1. Trustee partition: remove the stale sentinel-partition row.
+      const sentinelTrusteeQuery = and(
+        doc('documentType').equals('CASE_APPOINTMENT'),
```
jalba-flexion: CASE_APPOINTMENT can be a constant

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -519,6 +519,75 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+      } catch (deleteError) {
+        // Idempotent: a prior partial run may have already removed it.
+        if (!isNotFoundError(deleteError)) {
+          throw deleteError;
+        }
+      }
+
+      // 2. Trustee partition: upsert the resolved doc into its new partition.
+      const resolvedNaturalKey = and(
+        doc('documentType').equals('CASE_APPOINTMENT'),
```
jalba-flexion: Use same Constant

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -519,6 +519,75 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+        doc('assignedOn').equals(resolvedDocument.assignedOn),
+      );
+      await this.trusteePartition
+        .adapter<CaseAppointmentDocument>()
+        .replaceOne(resolvedNaturalKey, resolvedDocument, true);
+
+      // 3. Case partition: replace the sentinel row in place (caseId partition is
+      // stable, so the existing _id is preserved by the natural-key match).
+      const sentinelCaseQuery = and(
+        doc('documentType').equals('CASE_APPOINTMENT'),
```
jalba-flexion: Same

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
bityogi: ✅ Done in 9dff17559 — extracted a file-level `CASE_APPOINTMENT_DOC_TYPE` constant and replaced all 16 value usages of the literal across this repository (the literal type annotation on `CaseAppointmentDocument` stays as-is). Thanks for the catch!

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
bityogi: ✅ Done in 9dff17559 — extracted a file-level `CASE_APPOINTMENT_DOC_TYPE` constant and replaced all 16 value usages of the literal across this repository (the literal type annotation on `CaseAppointmentDocument` stays as-is). Thanks for the catch!

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
bityogi: ✅ Done in 9dff17559 — extracted a file-level `CASE_APPOINTMENT_DOC_TYPE` constant and replaced all 16 value usages of the literal across this repository (the literal type annotation on `CaseAppointmentDocument` stays as-is). Thanks for the catch!

jalba-flexion (approved):

backend/lib/use-cases/dataflows/migrate-case-appointments.ts
```diff
@@ -558,11 +558,130 @@ async function updateHealState(
...
+  // trustee cluster (one trustee → many cases), so the same acmsProfessionalId
+  // repeats within a batch — reuse the result to avoid redundant Cosmos reads.
+  const lookupCache = new Map<string, number>();
+  // For uniquely-matched acmsProfessionalIds, remember the resolved trusteeId so
+  // repeated docs reuse it without a second lookup.
+  const resolvedTrusteeByAcmsId = new Map<string, string>();
+  // Accumulate unresolved detail and log once per batch to avoid per-doc log spam.
+  const unresolvedDetails: string[] = [];
+
+  const sentinelDocs = batch.filter((doc) => doc.trusteeId === SENTINEL_TRUSTEE_ID);
```
btposey, line 606: This approach is unbounded and unsafe at scale. `resolveSentinelDocs` iterates the batch document-by-document, issuing one `findByAcmsProfessionalId` lookup and one `resolveSentinelTrusteeId` write per sentinel doc. We have no upper bound on how many sentinel-stamped case appointments exist, so this will not scale.

A better approach: use an aggregation pipeline to pull the **unique set** of `acmsProfessionalId`s across all sentinel-stamped (`trusteeId === SENTINEL_TRUSTEE_ID`) case appointments, rather than scanning documents one at a time. Join that set against `trustee-professional-ids` to build the `acmsProfessionalId → trusteeId` mapping. Then, for each mapped professional ID, reassign **all** case appointments carrying that sentinel ID and matching professional ID to the resolved `trusteeId` in one pass.

Note that we have two collections to update — `trustee-case-appointments` and `case-trustee-appointments` — since they're both partitions of the same logical case-appointment data. Once you've computed the resolution for one collection, you already have the answer for the other; there's no need to re-run the matching/lookup logic twice.

---

**AI prompt (for Prem):**

> In `backend/lib/use-cases/dataflows/migrate-case-appointments.ts`, the `resolveSentinelDocs` function (around line 584-673) currently resolves sentinel-stamped case appointments (`trusteeId === SENTINEL_TRUSTEE_ID`) one document at a time within a fixed-size batch, doing a per-document `findByAcmsProfessionalId` lookup and a per-document `resolveSentinelTrusteeId` write. This does not scale — the number of sentinel appointments is unbounded and this approach processes them serially per batch cursor.
>
> Redesign this to be set-based instead of per-document:
> 1. Use a MongoDB aggregation pipeline to compute the **distinct set** of `acmsProfessionalId` values across all documents where `trusteeId === SENTINEL_TRUSTEE_ID`, rather than paging through documents and checking each one individually.
> 2. Join that distinct set against the `trustee-professional-ids` collection to resolve each `acmsProfessionalId` to its unique `camsTrusteeId` (preserve existing behavior: skip/log any professional ID that maps to zero or multiple trustees — do not resolve ambiguous mappings).
> 3. For each professional ID that resolves to exactly one trustee, issue a single reassignment operation (e.g., `updateMany`) that rewrites `trusteeId` (and clears the sentinel `reason` field) for **all** documents matching `{ trusteeId: SENTINEL_TRUSTEE_ID, acmsProfessionalId: <id> }` — instead of individually resolving each document.
> 4. There are two collections holding case-appointment data that both need this update: `trustee-case-appointments` and `case-trustee-appointments`. They are two partitions of the same logical data, so the acmsProfessionalId → trusteeId resolution only needs to be computed once (per steps 1-2) and then applied to both collections — do not re-run the aggregation/lookup logic separately for each collection.
>
> Preserve the existing counting/logging behavior for unresolved cases (zero matches, multiple matches, missing acmsProfessionalId) so operational visibility doesn't regress. Check existing tests for `resolveSentinelDocs` and update them to reflect the new aggregate-based approach.

btposey (changes requested):

btposey:
Blocked until CAMS-809 and CAMS-816 have added more professional IDs.

btposey:
This ticket has been superseded. We're going a different direction.

PR #2736, opened 2026-07-22
CAMS-824 Add typed phone numbers for trustee contacts

cruzjone-flexion:
# Purpose

Trustees, internal contacts, and staff assistants previously supported only a single phone number. This adds typed phone numbers (direct/cell/home) for trustee internal contacts and staff assistants, and unifies phone display into a single reusable component.

# Major Changes

- Introduced typed phone model (`PhoneType`, `TypedPhoneNumber`, `PHONE_TYPES`) and renamed `TrusteeInternalContact` to `TrusteeContact` (shared type for both internal contacts and staff), dropping the redundant `TrusteeStaffContact` alias.
- Consolidated single/multiple phone display logic into `FormattedContact` via a generic `phones` prop, replacing duplicated logic that previously lived in `ContactInformationCard` and `TrusteeDetailAuditHistory`.
- Added typed phone number support (direct/cell/home) for trustee internal contacts (Slice 1) and staff contacts (Slice 2), both gated behind the `TRUSTEE_TYPED_PHONES` feature flag.
- Added a single shared `trusteeContactSpec` validator (replacing the previous `internalContactSpec`/`staffContactSpec` pair) for typed phone validation (no duplicate types, valid number/extension format).
- Added two migration scripts (`CAMS-824-internal-contact-phone-types.js`, `CAMS-824-staff-contact-phone-types.js`) to backfill existing records — these are manual deploy steps run once per environment via mongosh.
- Fixed CSS alignment/spacing bugs in the trustee staff card and mobile/desktop form button groups.
- Fixed two latent shared-validator bugs (`arrayOf` silently dropping spec-based validator errors; `noDuplicatePhoneTypes` double-nesting its reason map).
- Fixed four latent error-message bugs in the mongo trustees repository where the message was only set inside `camsStackInfo` and never surfaced at the top level, so failures reported a generic "Unknown Error" instead of a descriptive one.

# Testing/Validation

Implemented using TDD throughout. Ran a full TDD-quality spec review (`/cams-spec-review`) across all 14 test files changed on this branch and addressed every HIGH/MEDIUM finding: added missing repository/component test coverage, removed boundary-crossing and assertion-free tests, replaced disallowed selectors with `data-testid`, and consolidated near-duplicate tests. Full validation suite (typecheck, unit tests, eslint, knip, npm audit) passes clean across the `common`, `backend`, and `user-interface` workspaces (3,345+ backend/common tests, 3,289 user-interface tests).

# Notes

- Chose to extend `FormattedContact`'s `phones` prop rather than build a separate `TypedPhoneDisplay` component, since single/multiple-phone rendering was already unified there earlier in this branch.
- The `contact`/`phones` prop split at `FormattedContact` call sites (spreading `phones: undefined` off the contact object) is an intentional scoped design decision for this PR. Cleaning it up properly requires migrating `ContactInformation.phone` to `phones` across debtor attorneys, trustee match verification, and other domains — that is deferred to a follow-up slice.
- `AUDIT_INTERNAL_CONTACT` and `AUDIT_STAFF` snapshot documents are intentionally not touched by the migration scripts — audit history is immutable. `TrusteeDetailAuditHistory` retains a legacy `phone` fallback for pre-migration snapshots permanently.
- 3 pre-existing high-severity `npm audit` findings (`brace-expansion`, `fast-uri`, `immutable` — all transitive dependencies) are unrelated to this branch and were not addressed here.
- Filed a follow-up issue for pre-existing flaky tests discovered in the dataflows migration test suite during validation (unrelated to this branch's changes).

# Definition of Done:

- [x] Code refactored for clarity: Developers can understand the work simply by reviewing the code
- [x] Dependency rule followed: More important code doesn't directly depend on less important code
- [x] Development debt eliminated: UX and code aligns to the team's latest understanding of the domain
- [x] No regressions: Changes do not cause regression in related or unrelated areas of the application


common/src/cams/trustees-validators.ts
```diff
@@ -111,13 +111,39 @@ export const contactInformationSpec: ValidationSpec<ContactInformation> = {
...
+  return { valid: true };
+};
+
+export const internalContactSpec: ValidationSpec<TrusteeInternalContact> = {
+  address: [V.optional(V.nullable(V.spec(addressSpec)))],
+  phones: [V.optional(V.arrayOf(V.spec(typedPhoneNumberSpec))), noDuplicatePhoneTypes],
+  email: [V.optional(V.nullable(email))],
+};
+
+export const staffContactSpec: ValidationSpec<TrusteeStaffContact> = {
```
cruzjone-flexion: **`staffContactSpec` duplicates `internalContactSpec` identically**

`TrusteeStaffContact` is declared as a direct alias for `TrusteeInternalContact` (`export type TrusteeStaffContact = TrusteeInternalContact`), yet two structurally identical specs were created for them. If one spec is ever updated without the other, validation will silently diverge between internal contacts and staff contacts.

```typescript
// current
export const staffContactSpec: ValidationSpec<TrusteeStaffContact> = {
  phones: [V.optional(V.arrayOf(V.spec(typedPhoneNumberSpec))), noDuplicatePhoneTypes],
  email: [V.optional(V.nullable(email))],
};
```

Since the types are identical, `staffContactSpec` can just reference the existing one:

```typescript
export const staffContactSpec: ValidationSpec<TrusteeStaffContact> = internalContactSpec;
```

If they ever need to diverge, the type alias should become a distinct type at that point.

user-interface/src/trustees/forms/TrusteeInternalContactForm.tsx
```diff
@@ -94,7 +104,10 @@ function TrusteeInternalContactForm(props: Readonly<TrusteeInternalContactFormPr
                 countryCode: 'US',
               }
             : null,
-        phone: formData.phone ? { number: formData.phone, extension: formData.extension } : null,
+        phones:
+          formData.phones.filter((p) => p.number.trim()).length > 0
+            ? formData.phones.filter((p) => p.number.trim())
```
cruzjone-flexion: **`phones` filter predicate runs twice — assign once**

The `phones` filtering in the payload builder evaluates the same predicate twice: once to test length, then again to produce the filtered array. `TrusteeStaffForm` handles this correctly with a single assignment.

```typescript
// current
phones:
  formData.phones.filter((p) => p.number.trim()).length > 0
    ? formData.phones.filter((p) => p.number.trim())
    : undefined,
```

```typescript
// suggested
const filteredPhones = formData.phones.filter((p) => p.number.trim());
phones: filteredPhones.length > 0 ? filteredPhones : undefined,
```

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts
```diff
@@ -425,11 +426,6 @@ export function transformAppointmentRecord(
   // Map district to court ID
   const courtId = getCourtId(atsAppointment.DISTRICT);
 
-  // Validate chapter type for CAMS
-  if (!['7', '11', '11-subchapter-v', '12', '13'].includes(chapter)) {
-    throw new Error(`Invalid chapter for CAMS: ${chapter}`);
-  }
-
```
fmaddenflx, line 432: Why did we remove this?

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -178,17 +178,9 @@ describe('TrusteesMongoRepository', () => {
...
-
-      await expect(repository.listTrustees()).rejects.toThrow();
+      vi.spyOn(MongoCollectionAdapter.prototype, 'find').mockRejectedValue(error);
 
-      expect(mockAdapter).toHaveBeenCalledWith({
-        condition: 'EQUALS',
-        leftOperand: { name: 'documentType' },
-        rightOperand: 'TRUSTEE',
-      });
+      await expect(repository.listTrustees()).rejects.toThrow('Failed to retrieve trustees list.');
```
fmaddenflx, line 183: Why this change?

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -609,27 +597,11 @@ describe('TrusteesMongoRepository', () => {
...
-            leftOperand: { name: 'documentType' },
-            rightOperand: 'TRUSTEE',
-          },
-          {
-            condition: 'EQUALS',
-            leftOperand: { name: 'trusteeId' },
-            rightOperand: id,
-          },
-        ],
-      });
```
fmaddenflx, line 632: Why remove this?  It seems like a good asertion.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -789,36 +759,11 @@ describe('TrusteesMongoRepository', () => {
...
-              rightOperand: trusteeId,
-            },
-          ],
-        },
-        expect.objectContaining({
-          ...updatedTrusteeInput,
-          updatedOn: expect.stringMatching(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/),
-          updatedBy: mockUser,
-        }),
-      );
```
fmaddenflx, line 821: This also looks like a good assertion.  Why remove it?

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -948,27 +891,11 @@ describe('TrusteesMongoRepository', () => {
...
-            leftOperand: { name: 'documentType' },
-            rightOperand: '^AUDIT_',
-          },
-          {
-            condition: 'EQUALS',
-            leftOperand: { name: 'trusteeId' },
-            rightOperand: trusteeId,
-          },
-        ],
-      });
```
fmaddenflx, line 971: Ditto

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
```diff
@@ -1054,32 +981,11 @@ describe('TrusteesMongoRepository', () => {
...
-            leftOperand: { name: 'trusteeId' },
-            rightOperand: trusteeId,
-          },
-          {
-            condition: 'EXISTS',
-            leftOperand: { name: 'unassignedOn' },
-            rightOperand: false,
-          },
-        ],
-      });
```
fmaddenflx, line 1082: Why remove it?

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -735,7 +732,7 @@ function patchNestedObject(obj: Record<string, unknown>): Record<string, unknown
   return hasValidProperties ? result : undefined;
 }
 
-function formatAddress(c: Partial<ContactInformation>): string {
+function formatAddress(c: { address?: Partial<Address> | null }): string {
```
fmaddenflx, line 761: I don't understand why we're doing this.  Why would someone call formatAddress and pass it null?  Wouldn't it be better to just not call formatAddress if no address exists?

fmaddenflx: Never mind.  I see how it's being used.

common/src/cams/trustee-staff.ts
```diff
@@ -1,17 +1,19 @@
 import { Auditable } from './auditable';
 import { Identifiable } from './document';
-import { ContactInformation } from './contact';
+import { TrusteeInternalContact } from './trustees';
+
+export type TrusteeStaffContact = TrusteeInternalContact;
```
fmaddenflx: Why don't we just rename TrusteeInternalContact to something more generic.  That way we don't have to create aliases. This could be a source for confusion or misleading an engineer that might not realize that a change to TrusteeInternalContact will affect TrusteeStaffContact.

common/src/cams/trustees-validators.test.ts
```diff
@@ -798,9 +783,159 @@ describe('trustees-validators', () => {
...
+      const contact = {
+        phones: [
+          { number: '555-123-4567', type: 'direct' },
+          { number: '555-987-6543', type: 'direct' },
+        ],
+      };
+
+      const result = validateObject(TV.staffContactSpec, contact);
+      expect(result.reasonMap?.phones?.reasons).toContain('Each phone type may only be used once.');
+    });
```
fmaddenflx: These 3 tests can be combined.  We also don't need to test that it will accept only one phone number. The first two tests are nearly identical.  It doesn't provide any benefit to also test for 1.

common/src/cams/trustees-validators.test.ts
```diff
@@ -798,9 +783,159 @@ describe('trustees-validators', () => {
...
+      const contact = {
+        phones: [
+          { number: '555-123-4567', type: 'direct' },
+          { number: '555-987-6543', type: 'cell' },
+        ],
+      };
+
+      const result = validateObject(TV.internalContactSpec, contact);
+      expect(result.reasonMap?.phones).toBeUndefined();
+    });
```
fmaddenflx: Once again the check for only 1 doesn't provide any benefit.  Also because internalContactSpec and staffContactSpec are aliases, testing both doesn't provide any benefit.  If we define a more general name for internalContactSpec, then we can use that in both situations and we won't need an alias at all.  Then all the duplicate tests can go away.

common/src/cams/trustees-validators.ts
```diff
@@ -111,13 +111,39 @@ export const contactInformationSpec: ValidationSpec<ContactInformation> = {
...
+  address: [V.optional(V.nullable(V.spec(addressSpec)))],
+  phones: [V.optional(V.arrayOf(V.spec(typedPhoneNumberSpec))), noDuplicatePhoneTypes],
+  email: [V.optional(V.nullable(email))],
 };
```
fmaddenflx: Here again:  These are duplicates and not needed if we come up with a better generic type name for internalContactSpec.

common/src/cams/trustees.ts
```diff
@@ -10,6 +10,19 @@ import { AbstractTrusteeHistory } from './trustee-history-base';
 import { TrusteeUpcomingKeyDatesHistory } from './trustee-upcoming-key-dates';
 import type { TrusteeAppointment } from './trustee-appointments';
 
+export type PhoneType = 'direct' | 'cell' | 'home';
+export type TypedPhoneNumber = PhoneNumber & { type: PhoneType };
+
+export const PHONE_TYPES: PhoneType[] = ['direct', 'cell', 'home'];
```
fmaddenflx: Isn't there a better way to define PHONE_TYPES using PhoneType rather than restating the types on line 13 and 16?

common/src/cams/trustees.ts
```diff
@@ -10,6 +10,19 @@ import { AbstractTrusteeHistory } from './trustee-history-base';
...
+export const PHONE_TYPE_LABELS: Record<PhoneType, string> = {
+  direct: 'Direct',
+  cell: 'Cell',
+  home: 'Home',
+};
+export type TrusteeInternalContact = Omit<Partial<ContactInformation>, 'phone'> & {
+  phones?: TypedPhoneNumber[];
+};
```
fmaddenflx: Couldn't we just modify ContactInformation to take a phones property instead of a phone property?  Or is that too much work right now since we want to get this merged?

user-interface/src/lib/components/cams/TypedPhoneList/TypedPhoneList.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    expect(getPhoneNumberInput('home')).toHaveValue('');
+    expect(getPhoneExtensionInput('home')).toHaveValue('');
+  });
+
+  test('does not render dropdown, dividers, or add/remove buttons', () => {
+    setup([{ type: 'direct', number: '555-000-0000' }]);
+    expect(screen.queryByRole('combobox')).not.toBeInTheDocument();
+    expect(screen.queryByRole('button')).not.toBeInTheDocument();
+    expect(document.querySelector('.typed-phone-list__divider')).not.toBeInTheDocument();
+  });
```
fmaddenflx: LOL, we probably don't need to do this.  It looks like a test to ensure that we changed the UX from the direction we were previously going.

mattstankey:
## Automated Review Summary (4 parallel specialist passes: pragmatic/YAGNI, architecture/OeSA, security, product)

### 🔴 Blocking — phones disappear for un-migrated records, regardless of flag state
The new display code (`ContactInformationCard.tsx`, `TrusteeStaffCard.tsx`) only reads `contact.phones` (the new typed array) with no fallback to the legacy singular `contact.phone` field. Both migration scripts (`ops/migrations/CAMS-824-*.js`) are manual, engineer-run mongosh scripts — not run automatically on deploy. Any trustee/staff record that hasn't had the script run against it in a given environment will show **no phone number at all** the moment this code ships, independent of `TRUSTEE_TYPED_PHONES`. This isn't called out anywhere as a required deploy-sequencing step.

Also worth reconciling before merge:
- The PR description says internal-contact typed phones were "already-shipped" prior to this PR, but the diff shows `PhoneType`/`TypedPhoneNumber`/`PHONE_TYPES`/`TrusteeInternalContact` are all introduced in this same PR.
- All 4 Definition of Done checkboxes are unchecked despite the Testing/Validation section claiming full TDD and a clean validation suite — worth ticking or explaining the gap.

### 🟠 High — `FormattedContact`'s `contact`/`phones` prop split (flagged independently by two reviews)
Every caller with a `phones` array does:
```tsx
contact={{ ...internalContact, phones: undefined } as ContactWithPartialPhoneAndAddress}
phones={phones}
```
(e.g. `ContactInformationCard.tsx:60-64`, `TrusteeStaffCard.tsx:66-70`, `TrusteeDetailAuditHistory.tsx:117-121,414-419`) — a type-unsafe cast repeated at 4+ call sites, signaling the component boundary wasn't actually widened to accept the new domain type. Suggest: give `FormattedContact` a union `contact` type (or accept `phones` on the domain object directly) and normalize `phones ?? (phone ? [phone] : [])` internally, deleting the separate prop and the casts.

Related: the flag-gated phone-selection logic (`typedPhonesEnabled ? contact.phones : (directPhone ? [directPhone] : undefined)`) is reimplemented independently in ~5 places (`ContactInformationCard.tsx:19-27`, `TrusteeStaffCard.tsx:29-34`, `TrusteeDetailAuditHistory.tsx:110-119,381-390`, plus form-side branches) — worth one shared selector so there's a single place to clean up when the flag is retired.

### 🟡 Medium — Security
- `type` on `typedPhoneNumberSpec` (`common/src/cams/trustees-validators.ts:1117-1121`) only validates non-empty, not constrained to the `direct|cell|home` allow-list at runtime (the TS union isn't enforced), and the `phones` array has no max-length cap — a document-bloat vector against Cosmos's document size limit. Suggest adding a `V.oneOf(PHONE_TYPES, ...)` validator and a max-count check.
- Migration scripts (`ops/migrations/CAMS-824-internal-contact-phone-types.js`, `...staff-contact-phone-types.js`) have no dry-run/confirm gate before the destructive `$unset` of the legacy phone field. Not exploitable (no untrusted input, idempotent, correctly scoped), but worth a `--dry-run`/`--confirm` flag given they mutate PII in place.

### 🟡 Medium — Duplication / YAGNI
- The two migration scripts are identical except for `documentType`, field prefix, and print label — collapse into one function parameterized by `{ documentType, fieldPath }`.
- `PhoneRowErrors` is keyed by array index (`trusteeForms.utils.ts:37-64`, `TypedPhoneList.tsx:8-16`) rather than `PhoneType`, an implicit positional contract that only holds because both sides independently iterate `PHONE_TYPES` in the same order (flagged by both the pragmatic and architecture passes). Suggest `Record<PhoneType, PhoneRowErrors>` instead.
- `TrusteeStaffContact = TrusteeInternalContact` (`common/src/cams/trustee-staff.ts:1-9`) is a pure re-export alias with no distinct meaning yet — consider using `TrusteeInternalContact` directly until staff contact actually diverges.

---
No conflicts between the four reviews — the `FormattedContact` prop-split and the index-keyed errors were each independently raised by two separate passes, which raises confidence they're worth addressing. Security and architecture found no Critical/High-severity blockers on their own axes; the deploy-sequencing gap from the product review is the one item I'd treat as merge-blocking.

user-interface/src/lib/components/cams/FormattedContact.test.tsx
```diff
@@ -234,43 +236,43 @@ describe('FormattedAddress component', () => {
...
-          extension: '',
-        },
       };
 
-      renderComponent({ contact: phoneWithEmptyExtension, testIdPrefix: 'empty-ext' });
+      renderComponent({
+        contact: phoneWithEmptyExtension,
+        phones: [{ number: '555-444-3333', extension: '' }],
+        testIdPrefix: 'empty-ext',
+      });
```
fmaddenflx, line 253: If all of our FormattedContact tests are using `Omit<ContactInformation, 'phone'>` then it seems to me that we should change ContactInformation to use phones rather than phone.

user-interface/src/trustees/forms/TrusteeContactForm.scss
```diff
@@ -113,3 +149,21 @@
     }
   }
 }
+
+// usa-button-group (USWDS core) stacks vertically only below its own 480px
+// breakpoint, not this form's 599px field-stacking breakpoint. Match that
+// exact cutoff so this margin doesn't apply while the group is still laid
+// out as a row with align-items: stretch (which produced uneven button
+// heights when this margin leaked into that range).
```
fmaddenflx:
```suggestion
```

user-interface/src/trustees/forms/TrusteeContactForm.scss
```diff
@@ -60,6 +76,20 @@
   box-sizing: border-box;
 }
 
+// USWDS's own .usa-button-group switches to flex-direction: row with
+// align-items: stretch at its own 480px breakpoint. This form's button
+// group is a plain div of <button> elements (not the <li>-wrapped
+// usa-button-group__item markup USWDS expects), so stretch ends up
+// resizing Save/Delete/Cancel to uneven heights instead of matching
+// them. Override back to each button's own content height once row
+// layout kicks in; column (stacked) mode below 480px is unaffected.
```
fmaddenflx:
```suggestion
```

user-interface/src/trustees/forms/trusteeForms.utils.ts
```diff
@@ -43,3 +48,37 @@ export function normalizeFormData<T extends Record<string, unknown>>(
     }),
   ) as NormalizedFormData<T>;
 }
+
+/**
+ * Validates each phone in a fixed one-row-per-type TypedPhoneList array, skipping rows
+ * the user hasn't touched (both number and extension still blank) so an untouched Cell
+ * or Home row doesn't show a "must be a valid phone number" error before it's ever used.
+ */
```
fmaddenflx, line 61:
```suggestion
```

user-interface/src/trustees/forms/TrusteeStaffForm.test.tsx
```diff
@@ -24,7 +28,27 @@ vi.mock('react-router-dom', async () => {
   };
 });
 
+// Mocked only to capture the `openProps.onDelete` callback so its success/error
+// handling can be tested directly, without driving the real modal's dialog UI.
+// Renders a stub preserving the real component's testid convention so the existing
+// presence/absence tests below continue to work unchanged.
```
fmaddenflx:
```suggestion
```

user-interface/src/trustees/forms/TrusteeStaffForm.test.tsx
```diff
@@ -173,6 +198,76 @@ describe('TrusteeStaffForm', () => {
...
 
+  describe('TRUSTEE_TYPED_PHONES enabled', () => {
+    beforeEach(() => {
+      mockUseFeatureFlags.mockReturnValue({
+        [TRUSTEE_MANAGEMENT]: true,
+        [TRUSTEE_TYPED_PHONES]: true,
+      });
+    });
+
+    test('should render TypedPhoneList instead of the flat phone/extension inputs', () => {
```
fmaddenflx:
```suggestion
    test('should render TypedPhoneList', () => {
```

user-interface/src/trustees/forms/TrusteeStaffForm.test.tsx
```diff
@@ -173,6 +198,76 @@ describe('TrusteeStaffForm', () => {
     });
   });
 
+  describe('TRUSTEE_TYPED_PHONES enabled', () => {
```
fmaddenflx, line 213: Are we also testing for when the flag is off?

backend/lib/adapters/gateways/ats/cleansing/ats-mappings.ts
cruzjone-flexion: The chapter validation at this point was dead code — `parseChapterAndType()` (line ~37) already throws `"Unknown chapter code"` for any input not in `STANDARD_CHAPTERS = ['7', '11', '12', '13']`, and `applyAppointmentOverrides()` handles the `'11-subchapter-v'` transformation. By the time execution reaches this point, `chapter` can only be one of those five valid values, so this guard could never fire.

user-interface/src/trustees/panels/TrusteeDetailProfile.test.tsx
```diff
@@ -15,16 +43,13 @@ const mockTrustee: Trustee = {
...
       address1: '123 Main St',
-      address2: 'c/o John Smith',
-      address3: 'Ch 7',
       city: 'Anytown',
       state: 'NY',
       zipCode: '12345',
       countryCode: 'US',
     },
-    phone: { number: '555-123-4567', extension: '1234' },
+    phone: { number: '555-123-4567' },
```
fmaddenflx, line 51: Why remove the extension from the public data?

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
cruzjone-flexion: Two things happening here: (1) The `toHaveBeenCalledWith` assertion was asserting on the raw QueryBuilder output shape — an internal implementation detail of the adapter layer. That's a boundary-crossing unit test: the repository test shouldn't own the query AST produced by the adapter's fluent API. Removing it keeps the test focused on the repository's observable behavior.

(2) The error message assertion was added to cover a latent bug fixed in this PR: `listTrustees` was only setting `message` inside `camsStackInfo`, not at the top level of `getCamsErrorWithStack`, so failures surfaced as a generic "Unknown Error". The new assertion verifies the descriptive message now propagates correctly.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
cruzjone-flexion: Same reasoning as the `listTrustees` error test above: (1) The `toHaveBeenCalledWith` assertion was coupling the repository test to the raw query AST emitted by the QueryBuilder — an internal implementation detail of the adapter layer that the repository test shouldn't own. Removing it keeps the test focused on the repository's observable behavior.

(2) The replacement assertion verifies a latent bug fix: `read` was only setting `message` inside `camsStackInfo`, not at the top level of `getCamsErrorWithStack`, so failures surfaced as "Unknown Error". The test now confirms the descriptive message propagates correctly.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
cruzjone-flexion: Same reasoning as the other error-path tests: (1) The `toHaveBeenCalledWith` assertion was coupling the repository test to the raw QueryBuilder AST — an internal adapter implementation detail the repository test shouldn't own. Removing it keeps the test focused on observable behavior.

(2) The replacement assertion covers the same latent bug fixed across all four methods: `listTrusteeHistory` was only setting `message` inside `camsStackInfo`, not at the top level of `getCamsErrorWithStack`, so failures surfaced as "Unknown Error". The test now confirms the descriptive message propagates correctly.

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
cruzjone-flexion: Same reasoning as the other error-path tests: the `toHaveBeenCalledWith` assertion was coupling the repository test to the raw QueryBuilder AST — an internal adapter detail the repository test shouldn't own — and the replacement assertion covers the latent bug where `message` was only set inside `camsStackInfo`, causing failures to surface as "Unknown Error".

backend/lib/adapters/gateways/mongo/trustees.mongo.repository.test.ts
cruzjone-flexion: Same reasoning: the `toHaveBeenCalledWith` assertion was coupling the repository test to adapter internals (the exact shape passed to `insertOne`), which the repository test shouldn't own. The replacement assertion covers the latent bug where `message` was only set inside `camsStackInfo`, causing failures to surface as "Unknown Error".

common/src/cams/trustees-validators.test.ts
cruzjone-flexion: Both points are resolved: the single-phone test was redundant and has been removed, and `internalContactSpec`/`staffContactSpec` were already unified into a single `trusteeContactSpec` earlier in this branch — so there are no duplicate spec tests to worry about.

common/src/cams/trustees-validators.ts
cruzjone-flexion: Also resolved by the earlier unification — `internalContactSpec` and `staffContactSpec` were merged into a single `trusteeContactSpec` (typed against `TrusteeContact`), so the duplicate alias is gone.

common/src/cams/trustees.ts
cruzjone-flexion: TypeScript can't derive a runtime array from a union type — the direction only works the other way (const array → derived type). The literals have to be restated. Tightened it with `as const satisfies PhoneType[]` so the compiler catches any drift between the union and the array at build time.

common/src/cams/trustees.ts
cruzjone-flexion: Yes, that's the right long-term direction, but the blast radius is too wide for this PR — `ContactInformation.phone` is used across debtor attorneys, trustee match verification, ATS public contact mapping, and several UI components that are all out of scope for CAMS-824. Worth a dedicated slice rather than pulling those unrelated domains into this branch.

user-interface/src/trustees/forms/TrusteeStaffForm.test.tsx
cruzjone-flexion: Yes — the flag-off path is covered by the existing tests outside the `TRUSTEE_TYPED_PHONES enabled` describe block. The outer `beforeEach` sets `TRUSTEE_TYPED_PHONES: false`, and the "should render all form fields" test asserts `staff-phone` and `staff-extension` are present.

user-interface/src/trustees/panels/TrusteeDetailProfile.test.tsx
cruzjone-flexion: The extension was removed as part of the test refactor that introduced child component mocking. The new tests stub `TrusteeOverviewCard`, `ContactInformationCard`, and `TrusteeStaffCard`, so `TrusteeDetailProfile.test.tsx` no longer exercises phone/extension rendering — that's `ContactInformationCard`'s contract, covered in its own test file. The mock data was simplified to only what the stub-based tests actually need.

user-interface/src/lib/components/cams/FormattedContact.scss
```diff
@@ -13,6 +13,17 @@
     padding-bottom: 0.75rem;
   }
 
+  .phone {
+    display: inline-flex;
```
jalba-flexion: `.phone` is `display: inline-flex`, an inline-level box. The `.phones` wrapper it's nested under (see `formatPhones` in `FormattedContact.tsx`) has no `flex-direction: column` (or any flex display at all), so when a contact has 2+ typed phones, the `.phone` rows flow side-by-side in the same inline formatting context instead of stacking — e.g. `555-111-2222(Direct)555-333-4444(Cell)` running together with no separator. The `margin-bottom`/`&:last-child` rules only make sense for stacked block boxes, so I think this should be `display: flex` rather than `inline-flex`.

user-interface/src/trustees/forms/TrusteeContactForm.tsx
```diff
@@ -152,13 +179,18 @@ function TrusteeInternalContactForm(props: Readonly<TrusteeInternalContactFormPr
...
     const results = validateObject(trusteeInternalSpec, formData);
-
-    if (!results.valid && results.reasonMap) {
-      setFieldErrors(
-        Object.fromEntries(
-          Object.entries(results.reasonMap).map(([k, v]) => [k, v?.reasons]),
-        ) as FieldErrors,
-      );
+    const hasTypedPhoneRowErrors =
+      typedPhonesEnabled && Object.keys(validateTypedPhones(formData.phones)).length > 0;
```
jalba-flexion, line 187: When `TRUSTEE_TYPED_PHONES` is off, this only runs `trusteeInternalSpec` plus the typed-phone check (which no-ops since `typedPhonesEnabled` is false) — there's no equivalent of `TrusteeStaffForm.tsx`'s `validateDirectPhoneFields` fallback for the legacy single-phone UI. Repro: flag off, type an extension with no phone number for a trustee's internal contact, save. No validation error fires, and `mapPayload`'s `phones.filter(p => p.number.trim())` silently drops the row — the extension is lost with no feedback. Might be worth adding the same `validateDirectPhoneFields`-style check `TrusteeStaffForm` has.

common/src/cams/trustees-validators.ts
```diff
@@ -111,13 +111,33 @@ export const contactInformationSpec: ValidationSpec<ContactInformation> = {
...
+  extension: [phoneExtension],
+  type: [V.checkFirst(V.minLength(1, 'Phone type is required'))],
+};
+
+export const noDuplicatePhoneTypes: ValidatorFunction = (obj): ValidatorResult => {
+  // Wired everywhere as a field validator under a spec's `phones` key
+  // (`phones: [..., noDuplicatePhoneTypes]`), so it always receives the phones
+  // array itself, never a parent object.
+  const phones = (obj as TypedPhoneNumber[] | undefined) ?? [];
+  const types = phones.map((p) => p.type);
```
jalba-flexion: The `obj as TypedPhoneNumber[] | undefined` cast is erased at runtime — if a caller sends `phones` as a non-array truthy value (e.g. malformed JSON like `{"phones": "x"}`), `?? []` doesn't kick in (the value is truthy), so `phones` stays a string and `.map` throws an unhandled `TypeError`. Since `validateEach` (`validation.ts:42`) runs every validator in the array unconditionally with no short-circuit, this fires right alongside `arrayOf`'s own "Value is not an array" result. The thrown exception isn't a `CamsError`, so it surfaces as a 500 "Unknown Error" instead of a clean 400 validation failure. Might be worth guarding with `Array.isArray(obj) ? obj : []` instead of the cast.

user-interface/src/trustees/panels/TrusteeDetailAuditHistory.tsx
```diff
@@ -104,6 +105,57 @@ function ShowTrusteeContactHistory(props: ShowTrusteeContactHistoryProps) {
...
+      phone?: { number?: string; extension?: string };
+    };
+    let phones: FormattedPhone[] | undefined = snapshot.phones?.length
+      ? snapshot.phones
+      : undefined;
+    if (!phones && legacySnapshot.phone?.number) {
+      phones = [legacySnapshot.phone];
+    }
+    return (
+      <FormattedContact
```
jalba-flexion, line 130: This view doesn't reference `TRUSTEE_TYPED_PHONES` anywhere in the file, so Change History always renders every phone type on a snapshot even when the flag is off and the live profile card (`ContactInformationCard`) is intentionally hiding cell/home numbers. Might be intentional (history is immutable and arguably shouldn't retroactively hide data), but flagging since it creates a visible inconsistency between a trustee's live card and its own audit trail — worth confirming with product whether that's the intended behavior. Same pattern applies to `renderContact` below (~line 394).

jalba-flexion (commented):
Second-pass review focused on finding issues not already raised in the existing comments/automated review. A few things below.

One more note that doesn't land on a changed line: `common/src/cams/test-utilities/mock-data.ts`'s `getChapter13Trustee` (~line 647) still assigns a legacy `.phone`-shaped object to `Trustee.internal`, which is typed `TrusteeInternalContact` (`phones` array only, no `.phone`). It's unused elsewhere in the repo today, so it's latent, but any future test that calls it and inspects `internal.phones` will silently get no phone data — sibling mocks (`getTrusteeStaff`, `getTrusteeHistory`) were already updated for typed phones and this one looks like it was missed.

ops/migrations/CAMS-824-internal-contact-phone-types.js
```diff
@@ -0,0 +1,68 @@
...
+    documentType: 'TRUSTEE',
+    'internal.phone': { $exists: true },
+    'internal.phones': { $exists: false },
+  });
+
+  let updated = 0;
+
+  cursor.forEach(function (doc) {
+    const phone = doc.internal && doc.internal.phone;
+    if (!phone || !phone.number) {
```
jalba-flexion: The `countDocuments`/`find` query only checks that `internal.phone` exists, not whether it has a usable `number`. A record with `phone: { number: '' }` (or `phone: {}`) matches the query, hits this `return`, and is skipped with no log — no `$set`/`$unset` happens. Every subsequent run will report it again as "Found N document(s) to migrate" forever, with nothing indicating why it never converges. Might be worth logging the skipped `_id`s so an operator can see there's a subset needing manual attention. Same pattern in `CAMS-824-staff-contact-phone-types.js` (~line 47).

user-interface/src/trustees/forms/TrusteeContactForm.scss
```diff
@@ -36,6 +36,22 @@
             max-width: 150px;
             width: 100%;
           }
+
+          // TypedPhoneList fields size to their own content (see
+          // TypedPhoneList.scss); these entries only exist to win out over
+          // the blanket 385px default above within this form's field-group.
+          .usa-form-group.typed-phone-list__type-select {
```
jalba-flexion: `.usa-form-group.typed-phone-list__type-select` doesn't match anything — `TypedPhoneList.tsx` only ever renders `typed-phone-list__phone-input`/`typed-phone-list__extension-input` classNames (`type-select` doesn't appear anywhere else in the repo). This rule and the two other blocks referencing it below (lines 104, 138) look like dead leftovers from an earlier design that had a type-selector dropdown.

user-interface/test/accessibility/trustee-staff.test.ts
```diff
@@ -41,7 +41,7 @@ test.describe('Trustee Staff', () => {
     await expect(trusteeProfilePage.locator('#staff-city')).toBeVisible();
     await expect(trusteeProfilePage.locator('#staff-state')).toBeVisible();
     await expect(trusteeProfilePage.locator('#staff-zip')).toBeVisible();
-    await expect(trusteeProfilePage.locator('#staff-phone')).toBeVisible();
+    await expect(trusteeProfilePage.locator('#phone-group')).toBeVisible();
```
jalba-flexion, line 44: The only change here is the locator fix (`#staff-phone` → `#phone-group`) — it still only exercises the single/default phone state. Since this PR adds multi-row add/remove and duplicate-type validation-error UI, might be worth an a11y pass that also renders 2+ phone rows and a validation-error state, in case either introduces an unlabeled input or a focus-management issue the happy-path scan wouldn't catch.

common/src/cams/trustees.ts
jamesobrooks: There seems to be a bit of contradiction here, out of scope for CAMS-824 and worth a dedicated slice. I'd call this tech debt and suggest addressing it within the story, but yes, probably in it's own slice and accompanying branch.

common/src/cams/validators.ts
```diff
@@ -95,6 +96,24 @@ const optional = (...validators: ValidatorFunction[]): ValidatorFunction =>
  * @param {...ValidatorFunction[]} validators - Variable number of validator functions to apply to each array element
  * @returns {ValidatorFunction} A validator function that validates each element in an array and collects all validation errors
  */
+/**
+ * Collects every reason string out of a ValidatorResult, whether it reports
+ * failures as a flat `reasons` array (simple validators) or a nested
+ * `reasonMap` (e.g. a validator built from `spec(...)`).
+ */
```
jamesobrooks, line 99: Why are we inserting a block comment between the JSDOC and the function?

common/src/cams/validators.test.ts
```diff
@@ -766,7 +814,7 @@ describe('validators', () => {
       expect(result).toEqual(VALID);
     });
 
-    test('should stop collecting errors appropriately for large arrays with many failures', () => {
+    test('should collect all errors for a large array, not just a subset', () => {
```
jamesobrooks, line 817: I'm curious about the axis of change here. What prompted the shift? What constitutes a "large" array? Is there any point at which we stop? For example, an array of 100,000 with 50,000 errors could produce a reasons array that breaks because the reasons array is bigger than the source array (on-disk, not in terms of index length). What is the effect the user sees with this change?

ops/migrations/CAMS-824-internal-contact-phone-types.js
jamesobrooks, line 1: Is this a new pattern we're implementing? Why does this not follow the dataflows pattern of migration? When/how does this script get run?

user-interface/src/case-detail/panels/CaseDetailTrusteePanel.test.tsx
```diff
@@ -15,14 +17,23 @@ vi.mock('./useCaseAppointment', () => ({
...
+vi.mock('@/trustees/panels/TrusteeOverviewCard', () => ({
+  default: vi.fn(() => <div data-testid="mock-trustee-overview-card" />),
+}));
+
+vi.mock('@/trustees/panels/ContactInformationCard', () => ({
+  default: vi.fn(() => <div data-testid="mock-contact-information-card" />),
+}));
+
+vi.mock('@/trustees/panels/MeetingOfCreditorsInfoCard', () => ({
+  default: vi.fn(() => <div data-testid="mock-meeting-of-creditors-info-card" />),
```
jamesobrooks: This seems to follow the established pattern within the file, as does the use of `vi.mocked` below, but this is atypical for our tests. Is there some reason we have to use `vi.mock` and `vi.mocked` over our typical `vi.spyOn` with `.mockImplementation`?

user-interface/src/trustees/forms/TrusteeContactForm.tsx
fmaddenflx, line 1: Why did we rename this module.  Is this form shared between multiple screens?

user-interface/src/trustees/panels/TrusteeDetailAuditHistory.tsx
fmaddenflx: Glad you spotted this. Thanks!

cruzjone-flexion:
Thanks for the thorough review, Matt. Responses inline:

**🔴 Blocking — phones disappear for un-migrated records**

The two migration scripts (`CAMS-824-internal-contact-phone-types.js` and `CAMS-824-staff-contact-phone-types.js`) are manual deploy steps run once per environment via mongosh before the new code goes live. Once they run, all live `TRUSTEE` and `TRUSTEE_STAFF` records will have the new `phones` array — there's no window where un-migrated records exist alongside the new display code if the deploy sequence is followed.

`TrusteeDetailAuditHistory` retains a legacy `phone` fallback intentionally and permanently — `AUDIT_INTERNAL_CONTACT` and `AUDIT_STAFF` snapshot documents are immutable history and are explicitly excluded from the migration scripts. Pre-migration snapshots will always have the old `phone` field.

Updated the PR description to clarify the migration scripts are manual deploy steps and document the audit history exception.

**🟠 High — `FormattedContact` `contact`/`phones` prop split**

The cast at each call site is an intentional scoped design for this PR. Cleaning it up properly requires migrating `ContactInformation.phone` → `phones` across debtor attorneys, trustee match verification, and other domains that are out of scope here — deferred to a follow-up slice. Added a note to the PR description. The flag-gated phone-selection logic is in 2 display components (not 5) and is temporary — it goes away when `TRUSTEE_TYPED_PHONES` is retired.

**🟡 Medium — `type` not validated against allow-list, no max-length on `phones` array**

Acknowledged. The phone type/quantity constraints are actively being revisited as we have upcoming changes to the phone type model. We'll address the allow-list validation and any max-count logic as part of that work rather than hardcoding constraints that are about to change.

**🟡 Medium — `PhoneRowErrors` keyed by index not `PhoneType`**

Acknowledged. The UI for phone rows is being modified soon as part of the phone type changes, which will likely change this shape entirely. We'll address the typing as part of that work.

**🟡 Medium — Two migration scripts nearly identical**

They're intentionally separate one-time scripts — each is self-contained and independently auditable per document type. They'll be deleted after running in all environments.

**PR description accuracy + DOD checkboxes**

Updated the PR description to accurately reflect that the typed phone model (`PhoneType`, `TypedPhoneNumber`, `TrusteeContact`) is introduced in this PR (Slices 1 and 2 are both here). DOD checkboxes are now ticked.

user-interface/src/lib/components/cams/FormattedContact.scss
fmaddenflx: Good catch — fixed. `.phone` is now `display: flex` instead of `inline-flex`, so each phone row is a block-level flex container and stacks vertically with the existing `margin-bottom` spacing, instead of flowing inline with its siblings.

user-interface/src/trustees/forms/TrusteeContactForm.tsx
fmaddenflx: Good catch — fixed. Added the same `validateDirectPhoneFields` fallback `TrusteeStaffForm` has (moved it into `trusteeForms.utils.ts` so both forms share one implementation instead of duplicating it), wired `handleDirectPhoneChange`/`handleDirectExtensionChange` into debounced validation, and added the missing `errorMessage` props on the legacy phone/extension inputs — turns out those weren't wired to `fieldErrors` at all, so even a computed error couldn't have displayed. Added a regression test covering the extension-without-number case.

common/src/cams/trustees-validators.ts
fmaddenflx: Good catch — fixed. Replaced the cast with `Array.isArray(obj) ? obj : []`, so a non-array truthy `phones` value now falls through to an empty array in `noDuplicatePhoneTypes` instead of hitting `.map()` on a non-array and throwing. `arrayOf`'s own 'Value is not an array' reason still fires as expected. Added a regression test asserting `validateObject` doesn't throw on a non-array `phones` value.

user-interface/src/trustees/panels/TrusteeDetailAuditHistory.tsx
fmaddenflx: Confirmed this is intentional, not an oversight — Change History is meant to show the complete audit record regardless of feature-flag state. Gating it behind `TRUSTEE_TYPED_PHONES` would mean hiding real historical data from an audit trail based on a temporary UI rollout flag, which we don't want; audit history should never hide information. Leaving this as-is.

ops/migrations/CAMS-824-internal-contact-phone-types.js
fmaddenflx: Good catch — fixed, but differently than logging the skip: the script no longer skips these records at all. Every document matching the query now gets migrated — if `internal.phone` has no usable `number`, we still `$unset` the legacy field and `$set` `internal.phones` to `[]` instead of leaving the record untouched. That way the query condition (`internal.phone` exists, `internal.phones` doesn't) no longer matches after one run, so the count actually converges to zero instead of reporting the same N forever. Applied the same fix to `CAMS-824-staff-contact-phone-types.js`.

user-interface/src/trustees/forms/TrusteeContactForm.scss
fmaddenflx: Good catch — removed. Deleted the three `.usa-form-group.typed-phone-list__type-select` rules (base + the two media queries). Confirmed `TypedPhoneList.tsx` never renders that class, and there's no other trace of the old type-selector-dropdown design left in the codebase.

user-interface/test/accessibility/trustee-staff.test.ts
fmaddenflx: Good flag, but the premise has actually shifted underneath this comment: `TypedPhoneList` was simplified after this review to a fixed row per phone type (Direct/Cell/Home, no add/remove, no type selector), so duplicate-type entries can no longer occur through the UI at all — that specific a11y scenario doesn't apply to the current component.

That said, we're about to revert back toward the original add/remove-style UI shortly, which will bring the duplicate-type scenario back. Deferring this a11y coverage until that lands so we're not writing tests against a UI shape that's about to change again.

common/src/cams/validators.ts
fmaddenflx: Good catch — that was a refactoring artifact. When `collectReasons` got extracted as its own helper, `arrayOf`'s original JSDoc got left stranded above the new function instead of moving down with `arrayOf` itself. Moved it back to sit directly above `arrayOf`, and while in there, filled out `collectReasons`'s own comment with `@param`/`@returns` tags to match the file's convention.

common/src/cams/validators.test.ts
fmaddenflx: Good questions, worked through by diffing against main's merge-base:

The pre-existing test title was 'should stop collecting errors appropriately for large arrays with many failures' — but the test body (unchanged in this diff) has always just asserted `result.reasons` has length 1000 for a 1000-element all-invalid array. It never tested any stopping/truncation. The rename to 'should collect all errors for a large array, not just a subset' is a description correction, not a behavior change.

The actual behavioral change is in `arrayOf`'s per-element error extraction:
```diff
 value.forEach((element, index) => {
   const result = validateEach(validators, element);
-  if (!result.valid && result.reasons) {
-    const reasons = result.reasons.map((r) => `Element at index ${index}: ${r}`);
-    allErrors.push(...reasons);
-  }
+  const reasons = collectReasons(result).map((r) => `Element at index ${index}: ${r}`);
+  allErrors.push(...reasons);
 });
```
Old code only handled flat `result.reasons`. If a nested element validator produced a `reasonMap` instead (i.e. `V.spec(...)` — exactly what `V.arrayOf(V.spec(typedPhoneNumberSpec))` uses for typed-phone validation), `result.reasons` was `undefined`, the `if` was false, and that element's errors were silently dropped — the array element would wrongly be treated as valid. `collectReasons` (new helper) handles both shapes, so spec()-based errors now surface correctly. Two new tests ('surfaces errors from a spec()-based validator...', 'returns valid when every spec()-based element passes') cover this directly.

On the large-array-threshold questions specifically: I checked main's merge-base version of `arrayOf` directly, and the surrounding loop (`value.forEach`, no early exit, no length cap) is byte-for-byte identical before and after this PR — only the per-element reason-extraction expression changed (as above). So:
- 'What constitutes large?' — nothing does; there's no size-based branching in either version. 1000 is just the test's chosen size to demonstrate no truncation.
- 'Is there a point where we stop?' — no, in both versions `arrayOf` collects one reason per invalid element unconditionally, unbounded.
- 100k array / 50k errors 'breaking' — no such condition exists in either version. `allErrors` is a plain in-memory array built with `.push(...)`; no serialization step, no comparison against the source array's size.

User-visible effect of the actual change: for any validator using `arrayOf(spec(...))` (like typed phones), invalid array elements now correctly produce a validation error instead of being silently accepted as valid.

user-interface/src/case-detail/panels/CaseDetailTrusteePanel.test.tsx
fmaddenflx: Good catch — converted the whole file to `vi.spyOn`, no `vi.mock` left. There's no real technical reason `vi.spyOn` can't stub these; I initially assumed it wouldn't work for a rendered child component's default export (confirmed empirically that a naive one-time top-level `vi.spyOn(Module, 'default').mockImplementation(...)` gets silently undone by this file's own `beforeEach`'s `vi.restoreAllMocks()`/`vi.resetAllMocks()`, so the real component rendered instead of the stub). The fix is to re-arm the spy inside `beforeEach` after those calls, same pattern already established in `Login.test.tsx` (component defaults) and `SearchResults.test.tsx`/`BankruptcySoftware.test.tsx` (`getAppInsights`). Converted `useTrustee`/`useCaseAppointment`/`TrusteeOverviewCard`/`ContactInformationCard`/`MeetingOfCreditorsInfoCard`/`getAppInsights` all to `vi.spyOn`, assigned to `let` spy variables re-armed each `beforeEach` — all 20 tests (and the full case-detail suite, 222 tests) still pass.

fmaddenflx (approved):

PR #2744, opened 2026-07-23
CAMS-824 Expand phone types, sort phones, and UI polish

fmaddenflx:
# Purpose

Round out the multi-phone contact work for trustees and staff: broaden the phone-type options to match what USTP staff actually need to record, ensure phone lists render in a stable, predictable order, and clean up a few UI rough edges in the new phone entry component.

# Major Changes

* Replaced the 3-option phone type combo box (Direct/Home/Cell) with six options (Direct, Fax, Home, Office, Personal Mobile, Work Mobile), updating the shared `PhoneType`/`PHONE_TYPES`/`PHONE_TYPE_LABELS` definitions in `common` so `PhoneEntryList` no longer duplicates its own options array
* Added `sortTypedPhoneNumbers`/`sortTrusteePhoneNumbers` in `common` and wired sorting into the single point trustee data loads from the API (`TrusteeDetailScreen`), so phones are always ordered by type, then number, then extension — applied independently to the internal contact and to every staff member, so it correctly handles trustees with multiple staff
* Polished `PhoneEntryList` mobile layout: fixed input/error alignment, standardized divider spacing, and introduced a `trustee-form-button-group` class so the Cancel button is full-width and centered on mobile

# Testing/Validation

* Added unit tests in `common` for the new sort functions, including multi-staff scenarios
* Added an integration test on `TrusteeDetailScreen` confirming phones for the internal contact and for two independent staff members are each sorted correctly after loading from the API
* Updated existing test fixtures across `common` and `user-interface` that referenced the retired `Cell` phone type
* `npm run typecheck` and the `common`/`backend`/`user-interface` test suites all pass; no new test failures were introduced by this branch (a handful of pre-existing, unrelated failures — e.g. a flaky session-timer test and a couple of FormattedContact ordering assertions tied to an existing testid bug — were confirmed present on the base branch before these changes as well)

# Notes

Sorting happens once at data-load time (not during in-progress editing), so rows don't reorder under the user while they're mid-edit in the form.

# Definition of Done:

## Summary by Sourcery

Expand trustee phone handling with richer types, canonical sorting, and an updated multi-phone entry UI for internal and staff contacts.

New Features:
- Introduce additional phone types (fax, office, personal mobile, work mobile) in shared trustee phone definitions and UI.
- Add a reusable PhoneEntryList component to manage typed phone numbers, supporting both new multi-phone flows and legacy single-phone inputs.

Enhancements:
- Sort trustee phone numbers in a consistent order (by type, number, and extension) when loading trustee details, applied to internal contacts and staff.
- Simplify trustee contact/staff form initialization and validation to work with the new phone entry component and shared phone type definitions.
- Refine trustee contact and staff form layouts, including button group behavior and phone input alignment across breakpoints.
- Update FormattedContact and cards to display phone labels and ordering based on the expanded phone type set.

Tests:
- Add unit tests for phone sorting utilities, including multi-staff scenarios.
- Add an integration test on TrusteeDetailScreen to verify sorted phone rendering for internal and staff contacts.
- Update existing tests and fixtures across common and UI to reflect the revised phone types and new phone entry behavior.


mattstankey:
## Review findings: High & Medium issues

Automated multi-agent review (architecture, pragmatic/YAGNI, product) surfaced the following. (A separate Critical finding — 20 failing tests on this branch, verified by running the suite — was reported separately; this comment covers High/Medium only.)

### High

1. **No migration path for existing `'cell'` phone records.** `PhoneType` in `common/src/cams/trustees.ts` drops `'cell'` outright with no alias to `personalMobile`/`workMobile`. Any existing record with `type: 'cell'` will render `(undefined)` in `FormattedContact` (no matching key in `PHONE_TYPE_LABELS`) and won't match any `<option>` in the new 6-option `<select>` in `PhoneEntryList`, risking a silent, uninformed type change on next save.

2. **Duplicate-phone-type restriction removed with no stated rationale and no UI safeguard.** `noDuplicatePhoneTypes` (`common/src/cams/trustees-validators.ts`) is now unwired from `trusteeContactSpec`/`trusteeInternalSpec`/`trusteeStaffSpec` but not deleted — it's orphaned, misleading code. Functionally, `PhoneEntryList.handleAdd()` always inserts a fresh `direct` row, so a user can end up with two "Direct" entries with no visual distinction or warning.

3. **Sort-order ownership is inconsistent across fetch paths (good-fences violation).** `sortTrusteePhoneNumbers` is applied only in `TrusteeDetailScreen.tsx`'s fetch path. A second, independent fetch path (`useTrustee.ts` → `CaseDetailTrusteePanel` → `ContactInformationCard` → `FormattedContact`) is never sorted, so the same component/data renders in a different order depending on which screen loaded it.

### Medium

4. **Type-label suppression via field-stripping is fragile.** `ContactInformationCard.tsx`/`TrusteeStaffCard.tsx` strip the `type` field from the phone object to hide the label in legacy (flag-off) mode, coupling "has a `type`" to "show a label" implicitly. Recommend an explicit `showTypeLabels?: boolean` prop on `FormattedContact` instead.

5. **Sort-by-type is alphabetical, not usage-priority.** Current order (`direct, fax, home, office, personalMobile, workMobile`) puts Fax ahead of both mobile types as an artifact of alphabetization rather than a deliberate "most commonly used first" design — worth confirming this is the intended UX.

Full review also found `TypedPhoneList` left as dead code alongside its replacement `PhoneEntryList` (no production imports remain), which is a contributing cause of the test failures noted separately.

user-interface/src/lib/components/cams/PhoneEntryList/PhoneEntryList.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+export type PhoneEntryListProps = {
+  phones: TypedPhoneNumber[];
+  onChange: (phones: TypedPhoneNumber[]) => void;
+  errors?: Record<number, PhoneRowErrors>;
+  typedPhonesEnabled: boolean;
+  legacyPhoneErrors?: { phone?: string[]; extension?: string[] };
+};
+
+export default function PhoneEntryList(props: Readonly<PhoneEntryListProps>) {
```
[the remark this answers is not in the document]
fmaddenflx: ✅ Done — split into two single-purpose components in commit 33fae72da:

- `PhoneEntryList` now only renders the dynamic typed-phone rows (dropped `typedPhonesEnabled`/`legacyPhoneErrors` entirely).
- New `DirectPhoneFields` (`user-interface/src/lib/components/cams/DirectPhoneFields/DirectPhoneFields.tsx`) owns the legacy single phone + extension inputs.
- `TrusteeContactForm`/`TrusteeStaffForm` now choose which to render based on the feature flag, instead of pushing that branching into `PhoneEntryList`.

While extracting `DirectPhoneFields`, I also fixed the silent-drop issue flagged in your overall comment #2: it now creates a `direct` entry if one does not already exist in the array, rather than silently no-opping.

fmaddenflx:
Responding to the overall Sourcery feedback (commit 33fae72da):

1. **`formatPhones` doesn't sort the multi-phone branch** — rather than duplicating sort logic inside `FormattedContact` (a presentational component), I fixed this at the root: `useTrustee.ts` (the case-detail page's trustee-fetch path) now runs the fetched trustee through `sortTrusteePhoneNumbers`, matching `TrusteeDetailScreen.tsx`'s existing fetch path. This was actually the same underlying gap a human reviewer flagged independently (inconsistent sort-order ownership across fetch paths), so fixing it there closes both at once without giving `FormattedContact` a second responsibility.

2. **Silent drop when no `direct` phone exists** — addressed directly; see my reply on the inline comment.

3. **Fragile type-label suppression via field-stripping** — added an explicit `showTypeLabels?: boolean` prop to `FormattedContact` (defaults to `true`). `ContactInformationCard`/`TrusteeStaffCard` no longer strip `type` off the phone object — they pass phones through intact and set `showTypeLabels={typedPhonesEnabled}` instead.

Full suite passes with zero new failures.

fmaddenflx:
Thanks for the thorough review. Status on each item (commit 33fae72da unless noted):

**High**

1. **No migration path for `'cell'`** — confirmed with the PR author: not an issue, production doesn't currently have any records with `type: 'cell'`. Leaving as-is.
2. **Orphaned `noDuplicatePhoneTypes`** — already fixed in commit `7f8715677` (deleted the dead validator and its stale test that asserted the old restriction). Verified no references remain anywhere in the codebase.
3. **Inconsistent sort-order ownership across fetch paths** — fixed. `useTrustee.ts` (case-detail's trustee-fetch path, feeding `CaseDetailTrusteePanel` → `ContactInformationCard`) now runs the fetched trustee through `sortTrusteePhoneNumbers`, matching `TrusteeDetailScreen.tsx`. Added a test confirming the wiring, mirroring the existing `TrusteeDetailScreen` test.

**Medium**

4. **Type-label suppression via field-stripping** — fixed with an explicit `showTypeLabels?: boolean` prop on `FormattedContact` (default `true`). `ContactInformationCard`/`TrusteeStaffCard` now pass phones through with `type` intact and set `showTypeLabels={typedPhonesEnabled}` explicitly, rather than stripping the field to get the same visual effect.
5. **Alphabetical-looking sort order** — confirmed with the PR author: this order was dictated by Katie (UX), not an accident of alphabetization. Leaving as-is.

**Dead code note**

`TypedPhoneList` — confirmed zero production imports remained (only its own file + its own test referenced it). Deleted the component, its test, and its scss.

Also split `PhoneEntryList` into two single-purpose components per a related Sourcery suggestion (typed-rows-only `PhoneEntryList` + new `DirectPhoneFields` for the legacy single-phone case), and had `DirectPhoneFields` create a `direct` entry if one doesn't exist yet instead of silently dropping edits — that overlapped with a gap Sourcery flagged in the same area.

Full common/backend/user-interface suite passes with zero new failures.

user-interface/src/trustees/panels/TrusteeStaffCard.tsx
```diff
@@ -28,14 +28,10 @@ export default function TrusteeStaffCard({
   const handleClick = isEmpty ? onAdd : onEdit;
   const buttonLabel = isEmpty
     ? 'Add trustee staff member'
-    : `Edit trustee staff member ${staffMember?.name ?? ''}`;
+    : `Edit trustee staff member ${staffMember.name}`;
```
cruzjone-flexion: **`staffMember.name` drops the null fallback, producing a broken `aria-label`**

The change from `staffMember?.name ?? ''` to `staffMember.name` removes both the optional chain and the nullish fallback. The rest of the function still uses `staffMember?.contact?.phones` (optional chaining kept), so `name` can be `undefined` — and when it is, the button's `aria-label` becomes `"Edit trustee staff member undefined"`.

```tsx
// current
: `Edit trustee staff member ${staffMember.name}`;

// fix
: `Edit trustee staff member ${staffMember.name ?? ''}`;
```

Restoring the `?? ''` fallback keeps the label meaningful and matches the original intent.

cruzjone-flexion: ✅ Fixed in ecf7f1967 — restored the `?? ''` fallback.

cruzjone-flexion (approved):

PR #2745, opened 2026-07-23
CAMS-810 Pass authenticated user context to LaunchDarkly

mattstankey:
# Purpose

LaunchDarkly evaluated every flag with a single hardcoded, anonymous context on the backend and no context at all on the frontend, so segments and targeting rules couldn't distinguish one user from another. This slice makes both SDKs evaluate flags using the real authenticated user's identity, unblocking a phased pilot rollout (Detailees segment -> Working Group -> all users) for the trustee-management flag.

# Major Changes

* Added a shared `buildLaunchDarklyContext(user: CamsUser)` mapping function in `common/src/feature-flags.ts` so backend and frontend build an identical LaunchDarkly context (key, name, email, roles, office group designators) from one place.
* Restructured backend feature-flag fetching (`backend/lib/adapters/utils/feature-flag.ts`, both `application-context-creator.ts` files) so authenticated API requests get exactly one LD evaluation using the real resolved user, via a `skipFeatureFlags` option on `getApplicationContext()`. Dataflow, healthcheck, and mock-oauth2 call sites are untouched and keep today's anonymous-context behavior.
* Added a mount-time `identify()` call in `user-interface/src/App.tsx` so the frontend LD client also evaluates flags against the real session user.
* Documented the resulting context attributes and added a runbook for the Detailees pilot segment/targeting-rule setup in `docs/contributing/feature-flags.md`.

# Testing/Validation

Implemented using TDD (Red-Green-Refactor) across all three affected workspaces. Full test suites pass: backend (229 files / 3410 tests), user-interface (229 files / 3265 tests), common. Root-level lint, typecheck, and knip are clean. `cams-unit-test-reviewer` was run against all new/modified test files (`feature-flag.test.ts`, both `application-context-creator.test.ts` files, `feature-flags.test.ts`, `App.test.tsx`) and every finding (missing auth-failure-branch coverage, mock-lifecycle convention drift, tautological assertions, an untested `withLDProvider` branch, etc.) was fixed in a follow-up commit — no CRITICAL/HIGH issues remain.

# Notes

* `CamsUser.id` already equals the Okta `sub` claim (`backend/lib/adapters/gateways/okta/okta-gateway.ts`), so no separate key-mapping work was needed.
* The frontend `useEffect` depends on `[ldClient]` rather than `[]`, since `withLDProvider` initializes asynchronously and `ldClient` can be `undefined` on first render — this is more correct than suppressing the exhaustive-deps lint rule.
* This PR covers Slice 1 of a 2-slice feature. Slice 2 (creating the LaunchDarkly "Detailees" segment and a targeting rule on the `trustee-management` pilot flag) is manual dashboard configuration in the GovCloud US environment, documented as a runbook but out of scope for this PR.
* npm audit shows 4 pre-existing high/low vulnerabilities (brace-expansion, dompurify, fast-uri, immutable) — confirmed unrelated to this change (identical on `main`).

# Definition of Done:

## Summary by Sourcery

Update backend and frontend to evaluate LaunchDarkly feature flags using the authenticated user context while preserving existing anonymous behavior for non-user flows.

New Features:
- Introduce a shared buildLaunchDarklyContext helper and LaunchDarklyContext type to derive a consistent user context for feature flag evaluations across services.

Enhancements:
- Change backend feature-flag retrieval to accept an optional user and use the shared context mapping, wiring authenticated request contexts to pass the resolved user.
- Adjust Express and Azure application context creators to defer feature-flag evaluation until after session resolution and to support an option to skip eager flag fetching.
- Add a frontend mount-time identify call using the LaunchDarkly React client and the shared context mapping so flags are evaluated per logged-in user.
- Expand tests for backend context creation, feature-flag utilities, and the App component to cover user-aware LaunchDarkly behavior and configuration edge cases.
- Refine logging and observability behavior in the Azure context creator and logger helper.

Documentation:
- Document the shared LaunchDarkly user context shape and add a runbook for configuring user- and segment-based targeting in LaunchDarkly GovCloud.

Tests:
- Add comprehensive tests for Express application-context creation, including feature-flag wiring, request translation, and session error handling.
- Extend feature-flag tests in common and backend packages to validate context mapping, anonymous vs user-specific evaluation, and fallback behavior without an SDK key.
- Enhance App component tests to cover LaunchDarkly provider configuration and identify behavior for various session and client states.

### DIFF
--- a/backend/express/application-context-creator.test.ts
+++ b/backend/express/application-context-creator.test.ts
@@ -1,24 +1,38 @@
 import { vi } from 'vitest';
 import { Request } from 'express';
 import MockData from '@common/cams/test-utilities/mock-data';
+import { ApplicationContext } from '../lib/adapters/types/basic';
 import * as FeatureFlags from '../lib/adapters/utils/feature-flag';
+import { testFeatureFlags } from '@common/feature-flags';
+import { createMockApplicationContext } from '../lib/testing/testing-utilities';
 import ContextCreator from './application-context-creator';
 
-function createMockExpressRequest(headers: Record<string, string> = {}): Request {
+type MockExpressRequestOverrides = {
+  method?: string;
+  secure?: boolean;
+  host?: string;
+  query?: Record<string, string>;
+  params?: Record<string, string>;
+  body?: unknown;
+  headers?: Record<string, string | string[]>;
+};
+
+function createMockExpressRequest(overrides: MockExpressRequestOverrides = {}): Request {
+  const host = overrides.host ?? 'localhost:7071';
   return {
-    method: 'GET',
+    method: overrides.method ?? 'GET',
     url: '/cases',
-    secure: false,
-    get: (field: string): string | undefined => (field === 'host' ? 'localhost:7071' : undefined),
-    query: {},
-    params: {},
-    body: undefined,
-    headers: { authorization: 'Bearer ' + MockData.getJwt(), ...headers },
+    secure: overrides.secure ?? false,
+    get: (field: string): string | undefined => (field === 'host' ? host : undefined),
+    query: overrides.query ?? {},
+    params: overrides.params ?? {},
+    body: overrides.body,
+    headers: { authorization: 'Bearer ' + MockData.getJwt(), ...overrides.headers },
   } as Partial<Request> as Request;
 }
 
 describe('Express Application Context Creator', () => {
-  afterEach(() => {
+  beforeEach(() => {
     vi.restoreAllMocks();
   });
 
@@ -30,6 +44,7 @@ describe('Express Application Context Creator', () => {
 
     expect(featureFlagsSpy).toHaveBeenCalledTimes(1);
     expect(featureFlagsSpy).toHaveBeenCalledWith(context.config, context.session.user);
+    expect(context.featureFlags).toEqual(testFeatureFlags);
   });
 
   test('should eagerly fetch anonymous feature flags when getApplicationContext is called directly with no opts', async () => {
@@ -41,6 +56,93 @@ describe('Express Application Context Creator', () => {
 
     expect(featureFlagsSpy).toHaveBeenCalledTimes(1);
     expect(featureFlagsSpy).toHaveBeenCalledWith(context.config);
-    expect(context.featureFlags).toBeInstanceOf(Object);
+    expect(context.featureFlags).toEqual(testFeatureFlags);
+  });
+
+  test('should skip the eager feature-flag fetch when skipFeatureFlags is true', async () => {
+    const featureFlagsSpy = vi.spyOn(FeatureFlags, 'getFeatureFlags');
+    const request = createMockExpressRequest();
+    const logger = ContextCreator.getLogger('test-request-id');
+
+    const context = await ContextCreator.getApplicationContext(request, logger, 'test-request-id', {
+      skipFeatureFlags: true,
+    });
+
+    expect(featureFlagsSpy).not.toHaveBeenCalled();
+    expect(context.featureFlags).toEqual({});
+  });
+
+  describe('expressToCamsHttpRequest', () => {
+    test('joins array-valued headers and builds a secure URL', () => {
+      const request = createMockExpressRequest({
+        method: 'POST',
+        secure: true,
+        host: 'example.com',
+        query: { foo: 'bar' },
+        params: { id: '123' },
+        body: { some: 'body' },
+        headers: { 'x-multi': ['a', 'b'], authorization: 'Bearer token' },
+      });
+
+      const camsRequest = ContextCreator.expressToCamsHttpRequest(request);
+
+      expect(camsRequest).toEqual({
+        method: 'POST',
+        url: 'https://example.com/cases',
+        headers: { 'x-multi': 'a, b', authorization: 'Bearer token' },
+        query: { foo: 'bar' },
+        params: { id: '123' },
+        body: { some: 'body' },
+      });
+    });
+
+    test('falls back to localhost:7071 when no host header is present', () => {
+      const request = createMockExpressRequest({ headers: { authorization: 'Bearer token' } });
+
+      const camsRequest = ContextCreator.expressToCamsHttpRequest(request);
+
+      expect(camsRequest.url).toEqual('http://localhost:7071/cases');
+    });
+
+    test('wraps an unexpected error via getCamsError', () => {
+      const request = {
+        get headers(): Request['headers'] {
+          throw new Error('boom');
+        },
+      } as Partial<Request> as Request;
+
+      expect(() => ContextCreator.expressToCamsHttpRequest(request)).toThrow();
+    });
+  });
+
+  describe('getApplicationContextSession', () => {
+    let context: ApplicationContext;
+
+    beforeEach(async () => {
+      context = await createMockApplicationContext();
+    });
+
+    test('should throw an UnauthorizedError if authorization header is missing', async () => {
+      delete context.request.headers.authorization;
+      await expect(ContextCreator.getApplicationContextSession(context)).rejects.toThrow(
+        'Authorization header missing.',
+      );
+    });
+
+    test('should throw an UnauthorizedError if authorization header is not a bearer token', async () => {
+      context.request.headers.authorization = 'shouldthrowError';
+
+      await expect(ContextCreator.getApplicationContextSession(context)).rejects.toThrow(
+        'Bearer token not found in authorization header',
+      );
+    });
+
+    test('should throw an UnauthorizedError if authorization header contains Bearer with malformed token', async () => {
+      context.request.headers.authorization = 'Bearer some-text-that-is-not-possibly-a-valid-jwt';
+
+      await expect(ContextCreator.getApplicationContextSession(context)).rejects.toThrow(
+        'Malformed Bearer token in authorization header',
+      );
+    });
   });
 });
--- a/backend/express/application-context-creator.test.ts
+++ b/backend/express/application-context-creator.test.ts
@@ -1,0 +1,46 @@
+import { vi } from 'vitest';
+import { Request } from 'express';
+import MockData from '@common/cams/test-utilities/mock-data';
+import * as FeatureFlags from '../lib/adapters/utils/feature-flag';
+import ContextCreator from './application-context-creator';
+
+function createMockExpressRequest(headers: Record<string, string> = {}): Request {
+  return {
+    method: 'GET',
+    url: '/cases',
+    secure: false,
+    get: (field: string): string | undefined => (field === 'host' ? 'localhost:7071' : undefined),
+    query: {},
+    params: {},
+    body: undefined,
+    headers: { authorization: 'Bearer ' + MockData.getJwt(), ...headers },
+  } as Partial<Request> as Request;
+}
+
+describe('Express Application Context Creator', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  test('should call getFeatureFlags exactly once with the resolved session user', async () => {
+    const featureFlagsSpy = vi.spyOn(FeatureFlags, 'getFeatureFlags');
+    const request = createMockExpressRequest();
+
+    const context = await ContextCreator.applicationContextCreator(request);
+
+    expect(featureFlagsSpy).toHaveBeenCalledTimes(1);
+    expect(featureFlagsSpy).toHaveBeenCalledWith(context.config, context.session.user);
+  });
+
+  test('should eagerly fetch anonymous feature flags when getApplicationContext is called directly with no opts', async () => {
+    const featureFlagsSpy = vi.spyOn(FeatureFlags, 'getFeatureFlags');
+    const request = createMockExpressRequest();
+    const logger = ContextCreator.getLogger('test-request-id');
+
+    const context = await ContextCreator.getApplicationContext(request, logger, 'test-request-id');
+
+    expect(featureFlagsSpy).toHaveBeenCalledTimes(1);
+    expect(featureFlagsSpy).toHaveBeenCalledWith(context.config);
+    expect(context.featureFlags).toBeInstanceOf(Object);
+  });
+});
--- a/backend/express/application-context-creator.test.ts
+++ b/backend/express/application-context-creator.test.ts
@@ -5,6 +5,7 @@ import { ApplicationContext } from '../lib/adapters/types/basic';
 import * as FeatureFlags from '../lib/adapters/utils/feature-flag';
 import { testFeatureFlags } from '@common/feature-flags';
 import { createMockApplicationContext } from '../lib/testing/testing-utilities';
+import { CamsError } from '../lib/common-errors/cams-error';
 import ContextCreator from './application-context-creator';
 
 type MockExpressRequestOverrides = {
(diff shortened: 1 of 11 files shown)


docs/contributing/feature-flags.md
```diff
@@ -11,3 +11,63 @@ settings > Client-side SDK availability
...
+| `email`                  | `CamsUser.email`                                                | May be `undefined`.                                                                                                                 |
+| `roles`                  | `CamsUser.roles`                                                | e.g. `TrialAttorney`, `TrusteeAdmin`, `SuperUser` — see `common/src/cams/roles.ts` for the full `CamsRole` set. May be `undefined`. |
+| `officeGroupDesignators` | `getGroupDesignators(user)` (flattened from `CamsUser.offices`) | Always an array; empty if the user has no offices.                                                                                  |
+
+**Backend:** `getFeatureFlags(config, user?)` in `backend/lib/adapters/utils/feature-flag.ts` sends
+`buildLaunchDarklyContext(user)` when a resolved `CamsUser` is available (i.e. for authenticated API
+requests, via `applicationContextCreator()`). Dataflow, healthcheck, and mock-oauth2 invocations
```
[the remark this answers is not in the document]
mattstankey: Leaving this as-is — "healthcheck" here is referring to the actual code module by name (`backend/function-apps/api/healthcheck/healthcheck.function.ts`), not used as a common noun in prose. Changing it to "health check" would make the doc slightly less accurate, since it would no longer match the real folder/file name a reader might search for.

user-interface/src/App.test.tsx
```diff
@@ -112,8 +123,35 @@ describe('App', () => {
...
       default: () => ({ featureFlagClientId: undefined }),
     }));
+    vi.mocked(LaunchDarklyReactClientSdk.withLDProvider).mockClear();
+
+    const { default: AppComponent } = await import('./App');
+
+    expect(AppComponent).toBeDefined();
+    expect(LaunchDarklyReactClientSdk.withLDProvider).not.toHaveBeenCalled();
+
+    vi.doUnmock('@/configuration/appConfiguration');
```
diabagatekelly: **Mock-lifecycle leak risk:** `vi.doUnmock`/`vi.resetModules` cleanup happens at the end of the test body instead of in `beforeEach`/`afterEach`. If an assertion earlier in this test throws (e.g. line 130 or 131), this cleanup never runs, and the mocked `appConfiguration` module leaks into every subsequent test via the module registry — `vi.restoreAllMocks()` in the shared `beforeEach` (line 42) does not undo `vi.doMock` state.

Suggest moving `vi.doUnmock('@/configuration/appConfiguration')` + `vi.resetModules()` into the shared `beforeEach` (safe to call unconditionally) so cleanup runs regardless of test outcome.

user-interface/src/App.test.tsx
```diff
@@ -112,8 +123,35 @@ describe('App', () => {
...
+      reactOptions: { useCamelCaseFlagKeys: false },
+      options: {
+        baseUrl: 'https://clientsdk.launchdarkly.us',
+        streamUrl: 'https://clientstream.launchdarkly.us',
+        eventsUrl: 'https://events.launchdarkly.us',
+      },
+    });
     expect(AppComponent).toBeDefined();
 
     vi.doUnmock('@/configuration/appConfiguration');
```
diabagatekelly: Same cleanup-placement issue as the test above (line 133) — `vi.doUnmock`/`vi.resetModules` here also only runs if the preceding assertions (lines 146-155) all pass.

backend/lib/adapters/utils/feature-flag.ts
```diff
@@ -12,11 +16,10 @@ export async function getFeatureFlags(config: ApplicationConfiguration): Promise
...
   });
   await client.waitForInitialization();
-  const state = await client.allFlagsState({
-    kind: 'user',
-    key: 'feature-flag-migration',
-    anonymous: true,
-  });
+  const context = user
+    ? buildLaunchDarklyContext(user)
+    : { kind: 'user' as const, key: 'feature-flag-migration', anonymous: true };
```
diabagatekelly: **DRY / type-safety:** this anonymous-context literal is the one piece of context-building logic that escaped the new shared `buildLaunchDarklyContext` — it's also restated verbatim in `docs/contributing/feature-flags.md:36`, so any change now needs edits in two places. It's also not part of the exported `LaunchDarklyContext` type, so the "single authoritative shape" this PR introduces doesn't actually cover this branch.

Suggest hoisting this to a named const next to `buildLaunchDarklyContext` in `common/src/feature-flags.ts` (e.g. `ANONYMOUS_FEATURE_FLAG_CONTEXT`), and referencing it here, in the doc, and in the test that currently hardcodes this literal (`feature-flag.test.ts:44-48`).

Separately (unrelated to this hunk, just flagging while in this file): the three GovCloud LaunchDarkly URIs a few lines up (`baseUri`/`streamUri`/`eventsUri`, line 14) are hardcoded here and duplicated on the frontend (`App.tsx:93-95`, pre-existing). The SDK key is already externalized via `config.featureFlagKey` — the same pattern could cover these URIs via `ApplicationConfiguration` so both sides stay in lockstep if CAMS ever points at a different LD instance. Not urgent.

docs/contributing/feature-flags.md
```diff
@@ -11,3 +11,63 @@ settings > Client-side SDK availability
...
+have no user session, so they call `getFeatureFlags(config)` with no user and get today's hardcoded
+anonymous context (`{ kind: 'user', key: 'feature-flag-migration', anonymous: true }`) — this keeps
```
diabagatekelly: This restates the anonymous-context literal from `feature-flag.ts:21` verbatim — worth updating together if that literal gets hoisted to a shared const (see comment on `feature-flag.ts:21`).

common/src/feature-flags.ts
```diff
@@ -28,3 +30,23 @@ export const testFeatureFlags: FeatureFlagSet = {
   'downstream-trustee-appointments-enabled': true,
   'trustee-change-notification-enabled': true,
 };
+
+export type LaunchDarklyContext = {
+  kind: 'user';
+  key: string;
+  name?: string;
+  email?: string;
+  roles?: string[];
```
diabagatekelly: **Type safety:** `roles` is typed as `string[]`, which widens away the domain's `CamsRoleType`. `CamsUser.roles` is already `CamsRoleType[]`, and the doc explicitly enumerates role values here for targeting-rule authors — using `string[]` discards that type information at exactly the boundary where it'd be most useful, and would silently accept arbitrary strings on future edits.

Suggest `roles?: CamsRoleType[]` (import from `./cams/roles`) — should be a no-op assignment-wise given `CamsUser.roles`'s existing type.

backend/lib/adapters/utils/feature-flag.test.ts
```diff
@@ -26,4 +33,40 @@ describe('Tests for feature flags', () => {
...
+
+  test('calls allFlagsState with the real user context when a user is provided', async () => {
+    const context = await createMockApplicationContext({
+      env: { FEATURE_FLAG_SDK_KEY: 'fake-key' },
+    });
+    const user = MockData.getCamsUser();
+
+    await getFeatureFlags(context.config, user);
+
+    expect(allFlagsState).toHaveBeenCalledWith(buildLaunchDarklyContext(user));
```
diabagatekelly: **Self-referential assertion:** the expected value here is built by calling `buildLaunchDarklyContext(user)` — the exact same production function `getFeatureFlags` calls internally. Since `buildLaunchDarklyContext` is pure, this test can't distinguish "the real user was routed to a correct mapper" from "the real user was routed to a subtly-broken mapper," because both sides of the equality run through the identical code path. (A regression inside `buildLaunchDarklyContext` itself would still be caught by `common/src/feature-flags.test.ts`, but that's an indirect safety net for a bug that could be introduced right here.)

Suggest asserting against a hand-built literal, or at minimum checking `allFlagsState.mock.calls[0][0].key === user.id` directly.

user-interface/src/App.test.tsx
```diff
@@ -143,4 +181,42 @@ describe('App', () => {
...
+      const identify = vi.fn();
+      vi.mocked(LaunchDarklyReactClientSdk.useLDClient).mockReturnValue({
+        identify,
+        waitForInitialization: vi.fn().mockResolvedValue(undefined),
+      } as Partial<MockLDClient> as MockLDClient);
+
+      renderWithoutProps();
+
+      await waitFor(() => {
+        expect(identify).toHaveBeenCalledWith(buildLaunchDarklyContext(session.user));
```
diabagatekelly: Same self-referential-assertion pattern as `backend/lib/adapters/utils/feature-flag.test.ts:59` — the expected value is built via `buildLaunchDarklyContext(session.user)`, the same function `App.tsx` calls, so this test can't independently verify that call site. Suggest building the expected context object by hand here too.

docs/contributing/feature-flags.md
diabagatekelly, line 34: Sourcery's existing nitpick on this line ("healthcheck" → "health check") is still outstanding — cheap fix, worth picking up in the same pass as the other doc edit above.

backend/express/application-context-creator.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    });
+
+    test('should throw an UnauthorizedError if authorization header contains Bearer with malformed token', async () => {
+      context.request.headers.authorization = 'Bearer some-text-that-is-not-possibly-a-valid-jwt';
+
+      await expect(ContextCreator.getApplicationContextSession(context)).rejects.toThrow(
+        'Malformed Bearer token in authorization header',
+      );
+    });
+  });
```
diabagatekelly, line 148: **Coverage parity:** this suite has no `describe('getLogger', ...)` block asserting the request-ID-prefixed `console.log` wrapping behavior (`application-context-creator.ts:22-28`), unlike the Azure counterpart (`function-apps/azure/application-context-creator.test.ts:195-207`). The wrapper is incidentally exercised via `sanitizeDeep`'s internal `logger.debug` calls, so statement coverage is likely fine, but nothing explicitly asserts the prefixing behavior itself.

Suggest adding a test here mirroring the Azure one for consistency.

diabagatekelly (commented):
Deep review synthesis (automated code-review + tests-audit + cams-spec-review audits, cross-checked manually). No blockers — this is a solid, well-tested refactor. Inline notes below cover the cheap, high-value fixes worth picking up before merge, plus a couple of lower-priority follow-ups.

backend/express/application-context-creator.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    });
+
+    test('wraps an unexpected error via getCamsError', () => {
+      const request = {
+        get headers(): Request['headers'] {
+          throw new Error('boom');
+        },
+      } as Partial<Request> as Request;
+
+      expect(() => ContextCreator.expressToCamsHttpRequest(request)).toThrow();
```
diabagatekelly: This asserts only a generic `.toThrow()`. The implementation's catch block (`application-context-creator.ts:54-56`) wraps the error via `getCamsError(originalError, MODULE_NAME)`, but this assertion would pass identically even if that wrapping were removed entirely.

Suggest asserting on the thrown error's shape (e.g. `instanceof CamsError` or `expect.objectContaining({...})`).

user-interface/src/App.tsx
```diff
@@ -30,6 +32,14 @@ export const GlobalAlertContext = createContext<React.RefObject<GlobalAlertRef |
 function App() {
   const { reactPlugin } = getAppInsights();
   const globalAlertRef = useRef<GlobalAlertRef>(null);
+  const ldClient = useLDClient();
+
+  useEffect(() => {
+    const session = LocalStorage.getSession();
+    if (session?.user && ldClient) {
+      ldClient.identify(buildLaunchDarklyContext(session.user));
```
diabagatekelly, line 43: This effect depends only on `[ldClient]` and reads the session imperatively inside the effect, so it only ever calls `identify()` once, on mount. I checked `Session.tsx` — it gates rendering of `App` behind `getMe()` fully resolving the session, so this isn't a live bug today (App always mounts with the final session already in `LocalStorage`). But if a user is ever re-identified mid-session (re-login as a different user, role/office change without a full remount), LD would keep evaluating flags against the original identity. Worth a short comment noting this is a known scope limitation for this slice, or a follow-up if that scenario becomes reachable.

Separately (unrelated to this hunk, just flagging while in this file): the hardcoded GovCloud LaunchDarkly URIs a bit further down (`options.baseUrl`/`streamUrl`/`eventsUrl`, line 93) are duplicated on the backend (`backend/lib/adapters/utils/feature-flag.ts:14`) — see the comment there for the suggested fix.

user-interface/src/App.test.tsx
diabagatekelly: ✅ Fixed — moved `vi.doUnmock`/`vi.resetModules` into a shared `afterEach` (unconditional, so it runs even if the assertions above it throw).

user-interface/src/App.test.tsx
diabagatekelly: ✅ Fixed — same `afterEach` cleanup now covers this test too.

backend/lib/adapters/utils/feature-flag.ts
diabagatekelly: ✅ Fixed — hoisted to `ANONYMOUS_FEATURE_FLAG_CONTEXT` in `common/src/feature-flags.ts`, so this and the anonymous-context literal in the docs now share one source of truth.

docs/contributing/feature-flags.md
diabagatekelly: ✅ Fixed — updated to reference `ANONYMOUS_FEATURE_FLAG_CONTEXT` instead of restating the literal.

common/src/feature-flags.ts
diabagatekelly: ✅ Fixed — `roles` is now typed as `CamsRoleType[]`.

backend/lib/adapters/utils/feature-flag.test.ts
diabagatekelly: ✅ Fixed — the expected context object is now hand-built from the mock user's fields instead of calling `buildLaunchDarklyContext` itself.

user-interface/src/App.test.tsx
diabagatekelly: ✅ Fixed — same hand-built-expected-object fix applied here.

docs/contributing/feature-flags.md
diabagatekelly: Withdrawing this one — Matt already addressed the underlying Sourcery nitpick on this line (explained "healthcheck" refers to the actual module name, not a typo), which applies equally to my duplicate note. No change needed.

backend/express/application-context-creator.test.ts
diabagatekelly: ✅ Fixed — added a `describe('getLogger', ...)` block mirroring the Azure suite, asserting the request-ID-prefixed `console.log` wrapping.

backend/express/application-context-creator.test.ts
diabagatekelly: ✅ Fixed — now asserts `expect.any(CamsError)` so the test would actually fail if `getCamsError`'s wrapping were removed.

user-interface/src/App.tsx
diabagatekelly: Added a short comment documenting this as a known scope limitation for this slice. Left the GovCloud-URI externalization as a follow-up rather than a refactor here, per scope discussion.

jalba-flexion (approved):

PR #2746, opened 2026-07-23
CAMS-824 Limit trustee/staff phones to 20, track additions

fmaddenflx:
# Purpose

Cap the number of phone numbers a trustee (or a single staff member) can accumulate, and give the team visibility into how often phone numbers are actually being added, without silently letting an unbounded list grow in either the UI or the API.

# Major Changes

* Added `MAX_PHONE_NUMBERS = 20` to `common/src/cams/trustees.ts` as the single source of truth for the limit
* `trusteeContactSpec.phones` now rejects arrays over 20 entries server-side — since `staffInputSpec.contact` wraps this same spec, one change enforces the limit on both the trustee internal-contact PATCH and the staff POST/PUT paths
* `PhoneEntryList` hides the "Add Another Phone" button once a contact reaches 20 phone entries
* Backend now fires an App Insights `'Phone Number Added'` custom event (via the existing `context.observability` trace pattern, matching `'Trustee Name Edited'`) whenever an update actually persists a net-new phone number — once per phone added, tagged with `contactType: 'internal' | 'staff'`
* No running total is computed or sent — existing App Insights workbooks already tally custom events via `customEvents | summarize count()`, so the workbook does the aggregation, consistent with every other custom event in this codebase

# Testing/Validation

* Added unit tests for the 20-phone validation limit in `common` (validator) and `user-interface` (button visibility)
* Added unit tests for the telemetry in `backend`, covering: trustee internal-contact updates that add phones, staff member creation (all phones counted as new), staff member updates (before/after diff), and confirming no event fires when phones are only removed
* Full `common`/`backend`/`user-interface` suite passes; the only failures present (~79-81 Okta/MSW-related tests, count varies by run) were confirmed pre-existing on the clean baseline before this branch, unrelated to this work

# Notes

Deliberately did not touch the Azure Application Insights workbook (Bicep/JSON under `ops/cloud-deployment/lib/workbooks/`) to add a panel visualizing the new event — that's deployed shared infrastructure, so filed `cams-w5lz` to track it as separate follow-up work rather than bundling an infra change into this PR. The pattern to extend is already established in `trustee-name-management-metrics.json`, which queries a backend-only event (`'Trustee Name Edited'`) the same way.

# Definition of Done:

## Summary by Sourcery

Limit trustee and staff contacts to a maximum of 20 phone numbers and add telemetry for phone number additions.

New Features:
- Enforce a shared 20-phone maximum for trustee internal contacts and staff member contacts via common validation.
- Hide the "Add Another Phone" button in the phone entry UI once a contact reaches 20 phone numbers.
- Emit a "Phone Number Added" App Insights event for each newly added phone number on trustee internal contacts and staff members, tagged by contact type.

Enhancements:
- Introduce a central MAX_PHONE_NUMBERS constant in the shared trustees model for consistent limits across backend and UI.

Tests:
- Add validator tests to ensure the 20-phone limit is accepted at the boundary and rejected when exceeded.
- Add UI tests to verify visibility of the add-phone button below and at the 20-phone threshold.
- Add backend tests to confirm phone addition telemetry fires for staff creation, staff updates, trustee internal-contact updates, and does not fire when phones are only removed.

### DIFF
--- a/backend/lib/use-cases/trustee-staff/trustee-staff.test.ts
+++ b/backend/lib/use-cases/trustee-staff/trustee-staff.test.ts
@@ -179,6 +179,27 @@ describe('TrusteeStaffUseCase', () => {
         }),
       );
     });
+
+    test('should track a Phone Number Added event for each phone on the new staff member', async () => {
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(mockTrustee);
+      vi.spyOn(MockMongoRepository.prototype, 'createStaffMember').mockResolvedValue(
+        createdStaffMember,
+      );
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue(undefined);
+      const completeTraceSpy = vi.spyOn(context.observability, 'completeTrace');
+
+      await trusteeStaffUseCase.createStaffMember(context, trusteeId, validInput);
+
+      const phoneAddedCalls = completeTraceSpy.mock.calls.filter(
+        (call) => call[1] === 'Phone Number Added',
+      );
+      expect(phoneAddedCalls).toHaveLength(validInput.contact!.phones!.length);
+      phoneAddedCalls.forEach((call) => {
+        expect(call[2]).toEqual(
+          expect.objectContaining({ success: true, properties: { contactType: 'staff' } }),
+        );
+      });
+    });
   });
 
   describe('getStaffMember', () => {
@@ -334,6 +355,47 @@ describe('TrusteeStaffUseCase', () => {
           after: updatedStaffMember,
         }),
       );
+    });
+
+    test('should track a Phone Number Added event for each phone added to the staff member', async () => {
+      const mockTrustee = { id: trusteeId, name: 'Test Trustee' };
+      const staffWithOnePhone = MockData.getTrusteeStaff({
+        id: staffId,
+        trusteeId,
+        contact: { phones: [{ type: 'direct', number: '555-000-0001' }] },
+      });
+      const staffWithThreePhones = {
+        ...staffWithOnePhone,
+        contact: {
+          ...staffWithOnePhone.contact,
+          phones: [
+            { type: 'direct' as const, number: '555-000-0001' },
+            { type: 'home' as const, number: '555-000-0002' },
+            { type: 'fax' as const, number: '555-000-0003' },
+          ],
+        },
+      };
+      vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(mockTrustee);
+      vi.spyOn(MockMongoRepository.prototype, 'readStaffMember').mockResolvedValue(
+        staffWithOnePhone,
+      );
+      vi.spyOn(MockMongoRepository.prototype, 'updateStaffMember').mockResolvedValue(
+        staffWithThreePhones,
+      );
+      vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue(undefined);
+      const completeTraceSpy = vi.spyOn(context.observability, 'completeTrace');
+
+      await trusteeStaffUseCase.updateStaffMember(context, trusteeId, staffId, updateInput);
+
+      const phoneAddedCalls = completeTraceSpy.mock.calls.filter(
+        (call) => call[1] === 'Phone Number Added',
+      );
+      expect(phoneAddedCalls).toHaveLength(2);
+      phoneAddedCalls.forEach((call) => {
+        expect(call[2]).toEqual(
+          expect.objectContaining({ success: true, properties: { contactType: 'staff' } }),
+        );
+      });
     });
   });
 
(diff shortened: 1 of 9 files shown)


backend/lib/use-cases/trustee-staff/trustee-staff.ts
```diff
@@ -205,4 +213,22 @@ export class TrusteeStaffUseCase {
       });
     }
   }
+
+  private trackPhoneNumbersAdded(
```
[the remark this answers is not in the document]
cruzjone-flexion: The duplication is small (~8 lines) and these two classes live in separate modules. A shared helper would require either a new utility file or a cross-use-case dependency — both are more coupling than the duplication warrants at two call sites. Leaving as-is.

backend/lib/use-cases/trustees/trustees.ts
```diff
@@ -552,6 +552,13 @@ export class TrusteesUseCase {
       );
     }
 
+    this.trackPhoneNumbersAdded(
```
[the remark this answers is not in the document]
cruzjone-flexion: The duplication is small (~8 lines) and these two classes live in separate modules. A shared helper would require either a new utility file or a cross-use-case dependency — both are more coupling than the duplication warrants at two call sites. Leaving as-is.

common/src/cams/trustees-validators.ts
```diff
@@ -119,7 +119,15 @@ export const typedPhoneNumberSpec: ValidationSpec<TypedPhoneNumber> = {
 
 export const trusteeContactSpec: ValidationSpec<TrusteeContact> = {
   address: [V.optional(V.nullable(V.spec(addressSpec)))],
-  phones: [V.optional(V.arrayOf(V.spec(typedPhoneNumberSpec)))],
+  phones: [
+    V.optional(V.arrayOf(V.spec(typedPhoneNumberSpec))),
+    V.optional(
+      V.maxLength(
+        MAX_PHONE_NUMBERS,
+        `No more than ${MAX_PHONE_NUMBERS} phone numbers are allowed.`,
```
jalba-flexion: Extract message to a constant (can be reused for test also)

jalba-flexion (changes requested):

backend/lib/use-cases/trustee-staff/trustee-staff.test.ts
```diff
@@ -335,6 +356,47 @@ describe('TrusteeStaffUseCase', () => {
...
+        (call) => call[1] === 'Phone Number Added',
+      );
+      expect(phoneAddedCalls).toHaveLength(2);
+      phoneAddedCalls.forEach((call) => {
+        expect(call[2]).toEqual(
+          expect.objectContaining({ success: true, properties: { contactType: 'staff' } }),
+        );
+      });
+    });
   });
```
cruzjone-flexion, line 437: **Missing removal-path test for staff phone tracking**

`trustees.test.ts` has an explicit test verifying no `'Phone Number Added'` event fires when phones are removed, but this file has no equivalent. Since `trackPhoneNumbersAdded` is a separate private method in each class, the removal path here is only covered implicitly.

```typescript
test('should not track a Phone Number Added event when phones are removed', async () => {
  const staffWithTwoPhones = MockData.getTrusteeStaff({
    id: staffId,
    trusteeId,
    contact: {
      phones: [
        { type: 'direct' as const, number: '555-000-0001' },
        { type: 'home' as const, number: '555-000-0002' },
      ],
    },
  });
  const staffWithOnePhone = {
    ...staffWithTwoPhones,
    contact: { ...staffWithTwoPhones.contact, phones: [{ type: 'direct' as const, number: '555-000-0001' }] },
  };
  vi.spyOn(MockMongoRepository.prototype, 'read').mockResolvedValue(mockTrustee);
  vi.spyOn(MockMongoRepository.prototype, 'readStaffMember').mockResolvedValue(staffWithTwoPhones);
  vi.spyOn(MockMongoRepository.prototype, 'updateStaffMember').mockResolvedValue(staffWithOnePhone);
  vi.spyOn(MockMongoRepository.prototype, 'createTrusteeHistory').mockResolvedValue(undefined);
  const completeTraceSpy = vi.spyOn(context.observability, 'completeTrace');

  await trusteeStaffUseCase.updateStaffMember(context, trusteeId, staffId, updateInput);

  const phoneAddedCalls = completeTraceSpy.mock.calls.filter((call) => call[1] === 'Phone Number Added');
  expect(phoneAddedCalls).toHaveLength(0);
});
```

common/src/cams/trustees-validators.ts
cruzjone-flexion: ✅ Done — extracted to `MAX_PHONE_NUMBERS_MESSAGE` in `trustees-validators.ts` and updated the test to reference it.

backend/lib/use-cases/trustee-staff/trustee-staff.test.ts
cruzjone-flexion: ✅ Added the removal-path test to match the one in `trustees.test.ts`.

jalba-flexion (approved):

cruzjone-flexion (approved):

PR #2751, opened 2026-07-24
CAMS-814 Remove default district selection for trustee case list

cruzjone-flexion:
# Purpose

The trustee case list filter was silently preselecting the user's own division codes on load, narrowing the visible case list before the user made any filter choice. This removes that default so the list loads unfiltered.

# Major Changes

- Added `disableDefaultDivisionCodes` prop to `DistrictDivisionComboBox` to opt out of automatic default selection on mount
- Wired the flag into the trustee case list filter
- Refactored `computeDefaultOptions` from an inline component function to a module-level helper, consistent with the other `compute*` functions in the file

# Testing/Validation

Implemented using TDD. New unit tests cover the `disableDefaultDivisionCodes` path — no divisions preselected, `onDefaultsApplied` still fires. All 229 test files pass.

# Notes

`onDefaultsApplied` fires in both paths (defaults applied and defaults disabled) because `TrusteesList` uses it to flip an `isDefaultApplied` ref that gates district-filter analytics. Skipping the callback in the disabled path would silently break analytics tracking.

# Definition of Done:

## Summary by Sourcery

Remove automatic default division selection from the trustee case list by making district/division defaults opt-out and wiring this behavior into the trustee filter.

New Features:
- Add an optional disableDefaultDivisionCodes prop to DistrictDivisionComboBox to allow consumers to suppress automatic default division selection.

Enhancements:
- Refactor default option computation in DistrictDivisionComboBox into a shared computeDefaultOptions helper for clearer defaulting logic.

Tests:
- Add unit tests to verify DistrictDivisionComboBox behavior when default division codes are disabled or explicitly enabled.

### DIFF
--- a/user-interface/src/lib/components/DistrictDivisionComboBox.test.tsx
+++ b/user-interface/src/lib/components/DistrictDivisionComboBox.test.tsx
@@ -192,6 +192,75 @@ describe('DistrictDivisionComboBox', () => {
     });
   });
 
+  describe('disableDefaultDivisionCodes', () => {
+    function sessionWithManhattanOffice() {
+      return {
+        ...MockData.getCamsSession(),
+        user: {
+          ...MockData.getCamsSession().user,
+          offices: [
+            {
+              officeCode: '081',
+              officeName: 'Manhattan',
+              idpGroupName: 'Manhattan',
+              regionId: '02',
+              regionName: 'New York Region',
+              groups: [
+                {
+                  groupDesignator: 'NY',
+                  divisions: [
+                    {
+                      divisionCode: '081',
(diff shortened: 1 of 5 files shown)


cruzjone-flexion:
**Suggestion #1 (disableDefaultDivisionCodes precedence comment):** Skipping. The prop name `disableDefaultDivisionCodes` makes the precedence self-evident — when it's true, defaulting is disabled entirely. Adding a comment would explain the obvious, which conflicts with our convention of only commenting when the WHY is non-obvious. The `divisionCodeAllowList` prop already has its own inline doc in the type definition.

jalba-flexion (approved):

PR #2760, opened 2026-07-27
Normalize ACMS chapter 7A/7N/09/AC on case appointments

btposey:
# Bug

The trustee case list showed "7A"/"7N" verbatim for chapter 7 cases (legacy ACMS asset/no-asset sub-codes), and filtering the list by chapter "7" silently omitted those cases. Root cause: the one-time ACMS-to-CAMS case-appointment migration copied `CURR_CASE_CHAPT` verbatim (only trimming whitespace) instead of normalizing it, and the Mongo filter query does an exact match against the stored chapter value.

# Purpose

Ensure CASE_APPOINTMENT documents always carry a chapter value CAMS recognizes, both for existing data and for any future re-run of the ACMS migration or ongoing DXTR sync, so the trustee case list displays and filters chapter 7 cases consistently regardless of data origin (ACMS vs. DXTR).

# Major Changes

* Normalize ACMS's legacy chapter codes at write time in the migration mapping: `7A`/`7N` -> `7`, `09` -> `9`. Unrecognized codes (e.g. `AC`, which was never meant to be imported into CAMS) now cause the record to be skipped via the migration's existing failure-tracking path instead of being written.
* Introduced a `CaseChapter` literal type (`common/src/cams/trustee-appointments.ts`) and narrowed `CaseAppointment`, `CaseAppointmentInput`, `TrusteeCaseListItem`, and `CaseDenormalizedFields` to it, so invalid chapter values are caught at compile time going forward.
* Added a repository-level runtime guard (`assertValidChapter`) in the Mongo repository, called from `upsert`, `updateCaseAppointment`, and `updateCaseFields` — rejects any write with an invalid chapter before it reaches Mongo. This also protects the recurring DXTR sync path (`export-and-load-case.ts`), which was confirmed to only ever source chapter from DXTR's plain `CS_CHAPTER` column (never ACMS raw codes), but now fails loudly rather than silently if that assumption is ever violated.
* Added `ops/migrations/normalize-case-appointment-chapter.js`, a dry-run-by-default mongosh script to repair existing CASE_APPOINTMENT documents in both partition collections (`case-trustee-appointments` and `trustee-case-appointments`): renames `7A`/`7N` -> `7` and `09` -> `9`, and deletes any `AC`-chapter documents outright.

# Testing/Validation

Implemented with new unit tests alongside the existing suites (TDD-style additions, not full rewrites):
* `migrate-case-appointments.test.ts` — chapter normalization (`7A`/`7N`/`09`) and rejection of unrecognized chapters (`AC`) via `writePage`.
* `trustee-case-appointments.mongo.repository.test.ts` — guard rejects an invalid chapter in both `upsert` and `updateCaseFields` without writing to Mongo.
* `export-and-load-case.test.ts` — guard rejects an invalid DXTR chapter and records it as an event error rather than corrupting a document.

All affected suites pass (350 tests across the touched files; full workspace run: 3428 backend/common tests + 3320 user-interface tests, all green). Backend and common typecheck clean, lint clean, `npm audit`/`knip` show no new issues introduced by this change.

A `/cams-spec-review` pass was run on all three modified test files — no critical issues found (no mocking of code under test); a few pre-existing implementation-detail-coupling and missing-edge-case findings were surfaced but are out of scope for this fix.

# Notes

No Jira ticket for this fix. The ACMS gateway (`acms.gateway.ts`) already had a comment documenting the full valid ACMS chapter set (`09, 11, 12, 13, 15, 7A, 7N, AC`) and treated `7A`/`7N` inclusively as chapter 7 for an unrelated purpose (consolidation lead-case lookup) — this PR is the first place that normalization is applied to the actual case-appointment write path.

A follow-up PR narrowing `CaseBasics.chapter` (DXTR case summary chapter) to the same `CaseChapter` type is being opened separately for isolated review, since it touches a much more widely-used type across the app.

# Definition of Done:

## Summary by Sourcery

Normalize and validate case appointment chapter values across ACMS migrations, DXTR sync, and Mongo writes, and add a migration script to repair existing CASE_APPOINTMENT documents.

Bug Fixes:
- Ensure DXTR case sync rejects and surfaces invalid chapter values instead of updating case fields with bad data.
- Prevent trustee case appointment upsert and case field updates from writing invalid chapter codes to MongoDB.
- Normalize legacy ACMS chapter codes when migrating case appointments so only valid chapters are stored.

Enhancements:
- Introduce a shared CaseChapter type and VALID_CASE_CHAPTERS constant to constrain chapter values across domain types and validators.
- Refactor chapter validation in search predicates to reuse the shared chapter constant.

Deployment:
- Add a mongosh migration script to normalize existing CASE_APPOINTMENT chapter codes and delete records with unsupported 'AC' chapters.

Tests:
- Extend migration, repository, and DXTR sync tests to cover chapter normalization, invalid chapter rejection, and error recording behavior.

### DIFF
--- a/backend/lib/use-cases/gateways.types.ts
+++ b/backend/lib/use-cases/gateways.types.ts
@@ -52,12 +52,10 @@ import { TrusteeNote } from '@common/cams/trustee-notes';
 import {
   CaseAppointment,
   CaseAppointmentInput,
-  CaseChapter,
   CaseDenormalizedFields,
   TrusteeAppointment,
   TrusteeAppointmentInput,
   TrusteeCaseListItem,
-  VALID_CASE_CHAPTERS,
 } from '@common/cams/trustee-appointments';
 import {
   TrusteeMatchVerification,
@@ -284,30 +282,6 @@ export function formatCaseId(div: number, year: number, num: number): string {
 
 export function formatAcmsProfessionalId(group: string, code: number): string {
   return `${group.trim()}-${String(code).padStart(5, '0')}`;
-}
-
-// ACMS's CURR_CASE_CHAPT carries legacy codes that don't match CAMS's chapter
-// vocabulary directly: '7A' (asset) / '7N' (no-asset) are CAMS chapter 7,
-// '9' is unpadded (unlike the '09' used when querying ACMS), and 'AC' is the
-// predecessor to chapter 15 and is not imported into CAMS. See the chapter
-// comment on AcmsGatewayImpl.getLeadCaseIds for the full documented code set.
-// Throws for 'AC' or any other value outside CAMS's valid chapter set so the
-// migration skips/fails the record rather than writing an invalid chapter.
-export function normalizeAcmsCaseChapter(chapter: string): CaseChapter {
-  const trimmed = chapter.trim();
-
-  let normalized = trimmed;
-  if (trimmed === '7A' || trimmed === '7N') {
-    normalized = '7';
-  } else if (trimmed === '09') {
-    normalized = '9';
-  }
-
-  if (!VALID_CASE_CHAPTERS.includes(normalized as CaseChapter)) {
-    throw new Error(`Invalid ACMS chapter value: ${chapter}`);
-  }
-
-  return normalized as CaseChapter;
 }
 
 export interface AcmsGateway {
--- a/backend/lib/use-cases/gateways.types.ts
+++ b/backend/lib/use-cases/gateways.types.ts
@@ -52,10 +52,12 @@ import { TrusteeNote } from '@common/cams/trustee-notes';
 import {
   CaseAppointment,
   CaseAppointmentInput,
+  CaseChapter,
   CaseDenormalizedFields,
   TrusteeAppointment,
   TrusteeAppointmentInput,
   TrusteeCaseListItem,
+  VALID_CASE_CHAPTERS,
 } from '@common/cams/trustee-appointments';
 import {
   TrusteeMatchVerification,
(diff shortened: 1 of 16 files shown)


common/src/cams/trustee-appointments.ts
```diff
@@ -183,14 +183,20 @@ export const TRUSTEE_APPOINTMENTS_INTERNAL_SPEC: Readonly<ValidationSpec<Trustee
     ],
   };
 
+// Valid CAMS case chapters. ACMS's legacy '7A' (asset) / '7N' (no-asset) sub-codes
+// are normalized to '7' before reaching this type — see normalizeAcmsCaseChapter.
+export const VALID_CASE_CHAPTERS = ['7', '9', '11', '12', '13', '15'] as const;
+
+export type CaseChapter = (typeof VALID_CASE_CHAPTERS)[number];
```
btposey: This feels like domain modeling that should have been added to the cases domain, not the trustee-appointment domain. I want this moved.

backend/lib/use-cases/gateways.types.ts
```diff
@@ -284,6 +286,30 @@ export function formatAcmsProfessionalId(group: string, code: number): string {
...
+  } else if (trimmed === '09') {
+    normalized = '9';
+  }
+
+  if (!VALID_CASE_CHAPTERS.includes(normalized as CaseChapter)) {
+    throw new Error(`Invalid ACMS chapter value: ${chapter}`);
+  }
+
+  return normalized as CaseChapter;
+}
```
btposey: This function too is better kept as a helper function in common than added to a "types" file.

common/src/cams/trustee-appointments.ts
btposey: Moved — `CaseChapter`/`VALID_CASE_CHAPTERS` now live in `common/src/cams/cases.ts` alongside the other case domain types, not in the trustee-appointment module.

backend/lib/use-cases/gateways.types.ts
btposey: Moved — `normalizeAcmsCaseChapter` is now a helper function exported from `common/src/cams/cases.ts`, not `gateways.types.ts`.

jalba-flexion (approved):

PR #2762, opened 2026-07-27
CAMS-825 Add typed phones array and flag-gated multi-phone UI for vendor contacts

cruzjone-flexion:
# Purpose

Migrate software vendor contact phone numbers from a single untyped \`phone\` object to a typed \`phones\` array, and add a flag-gated multi-phone UI so staff can add, edit, and remove multiple typed phone numbers on a vendor's contact record. Display is also gated on the flag so the overview card shows all phones with type labels when enabled, and only the first direct phone (no label) when disabled.

# Major Changes

## Data model & migration
- Moved shared phone types (\`PhoneType\`, \`TypedPhoneNumber\`, \`PHONE_TYPES\`, \`PHONE_TYPE_LABELS\`, \`MAX_PHONE_NUMBERS\`, \`sortTypedPhoneNumbers\`) from \`trustees.ts\` to \`contact.ts\`; \`trustees.ts\` re-exports them for backwards compatibility
- Moved \`validateTypedPhones\` from \`trustees/forms/trusteeForms.utils\` to \`lib/components/cams/PhoneEntryList/phoneEntryList.utils\`, colocated with \`PhoneEntryList\`; \`trusteeForms.utils\` re-exports for backwards compatibility
- Added \`phones?: TypedPhoneNumber[]\` to \`SoftwareContactInfo\` in \`common/src/cams/bankruptcy-software.ts\`
- Registered \`SOFTWARE_VENDOR_TYPED_PHONES\` feature flag in \`UseFeatureFlags.ts\` and \`testFeatureFlags\`
- Updated admin seed data to use \`phones\` array
- Created idempotent Cosmos migration script \`ops/migrations/CAMS-825-software-vendor-phone-types.js\`

## Edit form (\`SoftwareVendorContactInfoForm\`)
- Flag on: replaces single phone + extension fields with \`PhoneEntryList\`. Phone entries pre-sorted on load so edit order matches display order.
- Flag off: uses \`DirectPhoneFields\` against the same \`phones\` state (matching the \`TrusteeContactForm\` pattern), eliminating the index-vs-type mismatch and removing the now-redundant separate \`phone\`/\`extension\` state.

## Overview card (\`BankruptcySoftwareDetailOverview\`)
- Flag on: shows all typed phones sorted by type/number/extension with type labels
- Flag off: shows only the first direct phone without a label

## Form types (\`softwareVendorContactInfoForm.types.ts\`)
- Adds \`DEFAULT_PHONE_ENTRY\` constant for initializing empty phone list

# Testing/Validation

Implemented using TDD. 41 form tests and 14 overview tests cover:
- Flag-on: \`PhoneEntryList\` renders, multi-phone save payload, validation blocking save, sort order on load (verified with out-of-order fixture asserting DOM render order)
- Flag-off: \`DirectPhoneFields\` renders, pre-fill from existing direct phone, single-entry array payload on save, empty-phones-array save, phone validation errors
- All 234 UI test files pass; all 230 backend test files pass

# Notes

Deployment order matters: deploy the code first, then run the migration script against each environment's Cosmos connection string:

\`\`\`
mongosh "<connection-string>" ops/migrations/CAMS-825-software-vendor-phone-types.js
\`\`\`

The script is idempotent — it only processes \`BANKRUPTCY_SOFTWARE\` documents where \`contact.phone\` exists and \`contact.phones\` does not, so re-running is safe. Vendors with unmigrated \`phone\` records will show no phone number in the UI until the migration runs.

The legacy \`phone\` field is intentionally kept on \`SoftwareContactInfo\` until a future slice removes it — the form no longer writes it, but the type retains it to avoid breaking any in-flight records.

# Definition of Done:


jalba-flexion (approved):

cruzjone-flexion (commented):
Thanks for the review, @sourcery-ai!

**Suggestion 1 — `phones: undefined` instead of `phones: []`:** Skipping intentionally. Sending `phones: []` is explicit — it means the user cleared the phone field. Sending `undefined` would be ambiguous between "no phone" and "field not included in the payload." An empty array is a valid, intentional state.

**Suggestion 2 — `mongo-upsert.ts` log message:** Not applicable to this PR. The `insertOnly` changes in `mongo-upsert.ts` were unrelated to CAMS-825 and were reverted in a separate commit. That code is not part of this branch's diff.

jalba-flexion (approved):

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
```diff
@@ -36,12 +36,13 @@ export function BankruptcySoftwareDetailOverview({
       }
     : undefined;
 
+  const displayPhones = contact?.phones?.length ? contact.phones : undefined;
```
diabagatekelly: **Should block merge — silently drops phone numbers for any vendor record not yet migrated.**

```ts
const displayPhones = contact?.phones?.length ? contact.phones : undefined;
```
There's no fallback to the legacy `contact.phone` field here. The migration (`ops/migrations/CAMS-825-software-vendor-phone-types.js`) is a separate, manually-run mongosh script, decoupled from deploy. From the moment this code ships until someone runs that script against each environment, every existing vendor's phone number disappears from the Overview page — it'll render `(none)`/no contact info instead.

This is inconsistent within the same PR: the sibling `BankruptcySoftwareDetailAuditHistory.tsx:97-101` *does* fall back to `contact.phone` when `phones` is absent; this component doesn't. `BankruptcySoftwareDetailOverview.test.tsx` has no fixture for a legacy-only record, so the gap is invisible to CI.

Suggest mirroring the audit-history component's fallback here, or if the intent is "migration runs before/during this deploy so this state never exists in prod," document that explicitly and add a test that pins the *documented* behavior.

user-interface/src/lib/hooks/UseFeatureFlags.ts
```diff
@@ -22,6 +22,7 @@ export const TRUSTEE_APPOINTMENT_HISTORY_ENABLED = 'trustee-appointment-history-
 export const TRUSTEE_CASE_LIST = 'trustee-case-list';
 export const TRUSTEE_CHANGE_NOTIFICATIONS = 'trustee-change-notification-enabled';
 export const TRUSTEE_TYPED_PHONES = 'trustee-typed-phones';
+export const SOFTWARE_VENDOR_TYPED_PHONES = 'software-vendor-typed-phones';
```
diabagatekelly: **Should block merge — this flag is declared but wired into nothing.**

Confirmed via repo-wide grep: `SOFTWARE_VENDOR_TYPED_PHONES` appears exactly once in the entire codebase — this declaration. `SoftwareVendorContactInfoForm.tsx`, `BankruptcySoftwareDetailOverview.tsx`, and `BankruptcySoftwareDetailAuditHistory.tsx` all unconditionally run the new `phones[]` behavior regardless of flag state. Compare to `TRUSTEE_TYPED_PHONES`, which is actually read in 4 places to switch UI behavior.

If a gradual rollout was the intent, this flag currently does nothing toward that. If it's meant to be removed before merge, worth cleaning up so it doesn't read as functional when it isn't.

ops/migrations/CAMS-825-software-vendor-phone-types.js
```diff
@@ -0,0 +1,87 @@
...
+    if (phone && phone.number) {
+      const typedPhone = { number: phone.number, type: 'direct' };
+      if (phone.extension) {
+        typedPhone.extension = phone.extension;
+      }
+      typedPhones.push(typedPhone);
+    }
+
+    if (DRY_RUN) {
+      print(`  Would update ${doc._id}: contact.phones = ${JSON.stringify(typedPhones)}`);
```
diabagatekelly: **Should block merge — logs phone numbers (PII) to stdout in the default dry-run path.**

```js
print(`  Would update ${doc._id}: contact.phones = ${JSON.stringify(typedPhones)}`);
```
This is the path an operator is expected to run first per the script's own usage doc (dry-run by default, `CONFIRM=true` required to apply). Every matched document's full phone number and extension gets printed to whatever captures this script's stdout — terminal history, CI logs, ops runbooks. The real-run branch (`Updated N document(s).`) is clean. Suggest printing counts or masked numbers instead in the dry-run branch.

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
```diff
@@ -221,6 +223,17 @@ export class BankruptcySoftwareController implements CamsController {
       });
     }
 
+    if (contact.phones) {
+      for (const phone of contact.phones) {
+        const result = validateObject(typedPhoneNumberSpec, phone);
+        if (!result.valid) {
+          throw new BadRequestError(MODULE_NAME, {
```
diabagatekelly, line 236: Two gaps in this new phone-validation block:

1. **0% test coverage** (confirmed via `vitest --coverage`, lines 227-230 uncovered). Email and website each have a dedicated `'should throw BadRequestError for invalid X in contact'` test (`bankruptcy-software.controller.test.ts:217-231`); phone has no equivalent, despite being the actual subject of this migration.
2. **No `MAX_PHONE_NUMBERS` length cap** on `contact.phones`. The trustee path enforces this via `V.maxLength(MAX_PHONE_NUMBERS, ...)` in `trusteeContactSpec` (and it's well tested there — boundary and over-boundary cases both covered), but this vendor path validates each phone's shape without ever checking array length. This controller is the real trust boundary (a direct API call isn't limited by what the UI sends), so worth adding the same cap while this code is already open.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailAuditHistory.tsx
```diff
@@ -87,13 +88,17 @@ function formatBankAssociations(associations?: SoftwareBankAssociation[]): strin
...
-function formatContact(contact?: BankruptcySoftwareProfile['contact']): string {
+function formatContact(contact?: SoftwareAuditContactInfo): string {
   if (!contact) return '(none)';
   const parts: string[] = [];
   if (contact.contactNames?.length) parts.push(contact.contactNames.join(', '));
   if (contact.emails?.length) parts.push(contact.emails.join(', '));
   if (contact.website) parts.push(contact.website);
-  if (contact.phone?.number) {
+  if (contact.phones?.length) {
+    contact.phones.forEach((p) => {
```
diabagatekelly, line 98: **The *new* `phones[]` rendering branch introduced by this PR has zero test coverage** — confirmed via coverage report (lines 98-99 uncovered) and independently by two review passes. Every phone-related test in `BankruptcySoftwareDetailAuditHistory.test.tsx` constructs `contact.phone` (the legacy singular field) and exercises the `else if` fallback a few lines down; none constructs `contact.phones`. Deleting this `if` branch entirely would not fail any existing test.

Suggest adding a case with `after: { contact: { phones: [{ number: '555-2222', type: 'direct', extension: '9' }] } }` asserting the rendered text includes `555-2222 x9`, plus a multi-phone case.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -249,6 +251,7 @@ export function SoftwareVendorContactInfoForm({
                 ariaDescription="Example: 123-456-7890"
                 value={phone}
                 onChange={handlePhoneChange}
+                errorMessage={fieldErrors['phone']}
```
diabagatekelly: Two things worth addressing while this component is open:

1. **Untested error wiring.** `fieldErrors['phone']` is computed both on mount (line ~73) and on change (line ~128), but neither path has a test — confirmed via coverage. Email and website each get a full trio (error on mount, error on change, clears on correction, Save disabled/re-enabled); phone gets none of that despite identical new wiring.
2. **DRY**: this hand-rolls the single-phone-plus-extension pattern (`PhoneNumberInput` + `Input` for extension, `handlePhoneChange`/`handleExtensionChange`) that `DirectPhoneFields` (`user-interface/src/lib/components/cams/DirectPhoneFields/DirectPhoneFields.tsx`) already encapsulates and that trustee forms reuse. The PR description's "matches the trustee pattern" holds at the type layer but not here — consider rendering `&lt;DirectPhoneFields /&gt;` instead, which would also remove the need for this component's own state/handlers.

common/src/cams/contact-validators.ts
```diff
@@ -13,3 +15,22 @@ export const website = V.skip(
...
+
+export const phoneSpec: ValidationSpec<PhoneNumber> = {
+  number: [phoneNumber],
+  extension: [phoneExtension],
+};
+
+export const typedPhoneNumberSpec: ValidationSpec<TypedPhoneNumber> = {
+  number: [phoneNumber],
+  extension: [phoneExtension],
+  type: [V.checkFirst(V.minLength(1, 'Phone type is required'))],
```
diabagatekelly: MEDIUM — `type` here only checks non-empty string, not membership in the `PHONE_TYPES` enum:
```ts
type: [V.checkFirst(V.minLength(1, 'Phone type is required'))],
```
Pre-existing gap (relocated from `trustees-validators.ts` unchanged), but this PR extends its blast radius to a second document type (software vendors) via the controller. A direct API call could persist `type: "bogus"`, which `PHONE_TYPE_LABELS[type]` (`FormattedContact.tsx`) would render as `undefined`. Worth tightening to `V.oneOf(PHONE_TYPES, ...)` while this file is already touched by this PR.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
```diff
@@ -57,16 +58,10 @@ export function BankruptcySoftwareDetailOverview({
       value: <FormattedContact contact={addressForDisplay} showLinks={false} />,
     });
   }
-  if (commsForDisplay) {
+  if (commsForDisplay || displayPhones?.length) {
```
diabagatekelly: MEDIUM — `commsForDisplay || displayPhones?.length` — the second half of this OR is unreachable, since `commsForDisplay`'s own derivation two lines up already includes `displayPhones?.length` in its condition (line 42). As written, 100% branch coverage of this line is structurally impossible. Simplify to `if (commsForDisplay) {`.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.test.tsx
```diff
@@ -95,4 +95,28 @@ describe('BankruptcySoftwareDetailOverview', () => {
...
+      contact: {
+        phones: [{ number: '212-555-0100', type: 'direct' }],
+      },
+    };
+    renderOverview(softwareWithPhone);
+    expect(screen.queryByTestId('no-contact-info')).not.toBeInTheDocument();
+    expect(screen.getByText('212-555-0100')).toBeInTheDocument();
+  });
+
+  test('should render phone when contact has phones but no email or website', () => {
```
diabagatekelly: MEDIUM — this test and the one directly above it (`'should render phone number from contact.phones array'`, L99-109) use effectively identical fixtures and assert the same outcome (a phone renders, no-contact-info testid absent). Consider consolidating, and using the freed-up test slot for the actually-missing case: a fixture with `website` set and `emails`/`phones` absent, so the `website`-only branch of `commsForDisplay` is exercised by something other than incidental coverage.

diabagatekelly (commented):
Deep review (code-review + tests-audit + cams-spec-review angles). The type-relocation refactor (shared phone types → `contact.ts`) is clean — 100% coverage, all existing trustee-side tests pass unmodified. But the actual feature migration has three issues I'd call blocking, plus the new `phones[]` code path is under-tested across three of the four changed test files (the opposite of what you'd want from a migration PR). Inline comments below cover everything anchorable to a changed line; two lower-value items couldn't be inlined since they point at test code this PR doesn't touch, so noting them here:

- **Wrong test boundary (not inline):** `SoftwareVendorContactInfoForm.test.tsx:380` ("should clear state when state combobox selection is cleared", pre-existing, untouched by this PR) drives the internal DOM of the child `ComboBox` component rather than the parent form's contract — duplicates coverage `ComboBox.test.tsx` should already own. Pre-existing, just flagging since the file was open.
- **Private-method reflection (not inline):** `backend/lib/use-cases/trustees/trustees.test.ts` has ~13 pre-existing tests (lines 1620-1989, untouched by this PR's one-line import change) that cast `trusteesUseCase as unknown as {...}` to call private methods directly and assert on internal return shapes, instead of asserting through the public API. Also pre-existing, flagging for awareness.

Overall: solid refactor, but the Overview-screen regression (below) is the kind of thing that'll get a bug report from a real user the day this ships, unless the migration script is run first in every environment.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
cruzjone-flexion: ✅ Intentionally not adding a fallback. The migration script runs as part of deployment, so legacy-only records will not exist in prod when this code ships. Keeping the code clean and avoiding legacy-support paths that would need to be removed later.

user-interface/src/lib/hooks/UseFeatureFlags.ts
cruzjone-flexion: ✅ Fixed. Removed `SOFTWARE_VENDOR_TYPED_PHONES` from `UseFeatureFlags.ts` and from `testFeatureFlags` in `feature-flags.ts`. This flag belongs to slice 2 and was added here prematurely.

ops/migrations/CAMS-825-software-vendor-phone-types.js
cruzjone-flexion: ✅ Fixed. Removed the dry-run mode, `CONFIRM` parameter, and all logging entirely. The script now applies updates directly on every run. Idempotency is preserved by the existing filter (only matches docs with `contact.phone` and no `contact.phones`).

backend/lib/controllers/bankruptcy-software/bankruptcy-software.controller.ts
cruzjone-flexion: ✅ Fixed both gaps. Added `MAX_PHONE_NUMBERS` length cap before the per-phone loop, and added test coverage for both invalid phone data and phones-exceeding-max-count cases.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailAuditHistory.tsx
cruzjone-flexion: ✅ Fixed. Added two tests covering the `contact.phones` branch: one with a single phone + extension asserting `555-2222 x9`, and one with multiple phones.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
cruzjone-flexion: ✅ Fixed both. Refactored to use `<DirectPhoneFields />` instead of hand-rolling `PhoneNumberInput` + extension `Input` + handlers. Added the missing phone error test coverage (on-mount, on-change, clears-on-correction, Save-disabled) matching the email and website test trios.

common/src/cams/contact-validators.ts
cruzjone-flexion: ✅ Fixed. Replaced `V.minLength(1)` with `V.isInSet([...PHONE_TYPES])` so unrecognized type values are rejected. Also added a direct unit test for `typedPhoneNumberSpec` covering all valid types, unrecognized type, empty type, and invalid number.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
cruzjone-flexion: ✅ Fixed. Simplified to `if (commsForDisplay)` — the `displayPhones?.length` operand was unreachable since `commsForDisplay` already incorporates it.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.test.tsx
cruzjone-flexion: ✅ Fixed. Consolidated the two near-duplicate phones-only tests into one and replaced the freed slot with a website-only fixture, covering the third branch of `commsForDisplay`.

PR #2767, opened 2026-07-27
Fix 7A/7N chapters to 7 dataflow

btposey:
# Purpose

Follow-up to #2760: correct the ops/migrations usage docs (which assumed a mongosh CLI/`load()` workflow that doesn't match how operators actually run these scripts), and replace the manual, single-run chapter-fix script with a proper Azure Functions dataflow that can safely repair the full ~3M+ document CASE_APPOINTMENT collection with paging and 429/RU-throttling backoff.

# Major Changes

**Ops migration script fixes:**
* Corrected the usage instructions in all 6 `ops/migrations/*.js` scripts. `load()` is not available in MongoDB Compass's embedded shell (it returns a `[COMMON-90002]` error), so the actual operator workflow is: open the file, copy its contents, and paste directly into an interactive mongosh-compatible shell (e.g. Compass) connected to the target database — not `mongosh "<connection-string>" <file>` or `load(...)`.
* Removed the `process.env.CONFIRM`/dry-run pattern from the two `CAMS-824` phone-migration scripts and the chapter-fix script, matching the plain, run-directly convention already used by the other migration scripts in this directory — environment-variable toggles require hand-editing the file before running in upper environments, which is undesirable.

**New dataflow: FIX-CHAPTER-7-APPOINTMENTS** (supersedes and removes `ops/migrations/normalize-case-appointment-chapter.js`):
* A start/reader/writer queue topology. `handleStart` dispatches 8 independent, self-draining repair streams — one per (collection × chapter-fix operation): rename `'7A'`/`'7N'`→`'7'`, rename `'09'`→`'9'`, and delete `'AC'` documents, across both the `case-trustee-appointments` and `trustee-case-appointments` collections.
* Each `handleReader` invocation re-queries its own fixed filter for matching document ids, pages them to the writer queue via `pageByByteBudget`, and re-enqueues itself with a 30s delay to let writers drain before the next scan. No cursor or state document is needed — a stream is simply done once its query returns zero matches, since fixed documents naturally drop out of the filter.
* Each `handleWriter` invocation applies the fix to one page of documents by `_id` (`updateMany`/`deleteMany` scoped to that page, not a blanket collection scan), using the established `handleRateLimitRetry` helper for 429/RU-throttling backoff.
* Added `findIdsByChapter` and `applyChapterFix` to `TrusteeCaseAppointmentsRepository`, each operating on exactly one collection per call (not the repository's usual dual-partition write), since each collection is repaired independently for safety.
* Registered but **not** added to `DEFAULT_DATAFLOWS` — must be explicitly enabled via `CAMS_ENABLED_DATAFLOWS` when someone wants to run the repair.

# Testing/Validation

Implemented with unit tests for all new/changed code (repository methods, use-case wrapper, and all three dataflow handlers), following TDD. A `/cams-spec-review` pass surfaced one real gap — `handleReader`'s multi-page writer-enqueue path (via `pageByByteBudget`) wasn't exercised at a large enough scale to actually split across pages — which was closed with a dedicated 3000-id test, along with strengthening two under-specified "routes to DLQ" assertions to check actual payload content instead of just presence.

All affected suites pass: 103 tests directly on the touched/new files; full backend suite (3472 tests) and full workspace typecheck/lint clean. Confirmed via `npm test --workspace=backend` (with the required `DATABASE_MOCK`/`MONGO_CONNECTION_STRING` env vars) rather than ad hoc `vitest run` invocations, after an earlier ad hoc run without those vars produced spurious real-MongoClient-connection failures unrelated to this change.

# Notes

No Jira ticket for this work (matches #2760's original PR). The queue/message design (flat `matchChapter: string` per stream rather than array-based, per-page `_id`-scoped writes rather than a blanket `updateMany(filter)` rescan, and independent per-collection repair rather than assuming shared `_id`s across partitions) was worked out interactively to be safe under concurrent reader/writer execution — duplicate reads across overlapping reader batches are harmless since writer operations are idempotent per-`_id`.

# Definition of Done:


backend/lib/use-cases/dataflows/fix-chapter-7-appointments.ts
```diff
@@ -0,0 +1,40 @@
+import { ApplicationContext } from '../../adapters/types/basic';
+import factory from '../../factory';
+
+export type AppointmentIdPair = { trusteeApptId: string; caseApptId: string };
```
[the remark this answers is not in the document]
btposey: Fixed. `AppointmentIdPair` is now `{ trusteeApptId: string; caseApptId: string | null }`, matching the repository's actual return type and the drift semantics (`null` = missing case-partition counterpart).

backend/lib/testing/mock-gateways/mock-mongo.repository.ts
```diff
@@ -396,6 +396,16 @@ export class MockMongoRepository
     return Promise.resolve([]);
   }
 
+  findAppointmentIdPairsByChapter(
+    ..._ignore: any[]
+  ): Promise<Array<{ trusteeApptId: string; caseApptId: string }>> {
+    throw new Error('Method not implemented.');
+  }
```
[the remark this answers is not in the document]
btposey: Fixed. Updated the mock signature to `Promise<Array<{ trusteeApptId: string; caseApptId: string | null }>>` so the test double matches the real repository return type.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -575,4 +603,286 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+  // matching invalid legacy ACMS chapter codes (e.g. '7A', 'AC') that are not
+  // valid CaseChapter values.
+  //
+  // caseApptId is null when the case partition has no matching document for
+  // an active trustee-partition document (a partition-parity drift — see
+  // heal()'s similar repair for the mechanism this mirrors). The caller
+  // (applyChapterFix) is responsible for repairing that drift by creating the
+  // missing case-partition document — this method only detects the gap, it
+  // never writes.
+  async findAppointmentIdPairsByChapter(
```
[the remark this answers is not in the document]
btposey: Ignoring for now since this is a one-time run.

backend/function-apps/dataflows/migrations/fix-chapter-7-appointments.ts
```diff
@@ -0,0 +1,398 @@
...
+ * re-enqueues itself with a delay so writers can drain before the next query.
+ * When a query returns no id pairs, the stream is complete and nothing is
+ * re-enqueued.
+ *
+ * On 429/RU-throttling (including Cosmos's code-50 "ExceededTimeLimit ...
+ * rate limiting" variant — see isRateLimitTimeoutError), retries with backoff
+ * via handleRateLimitRetry instead of failing straight to DLQ; exhausted
+ * retries route to DLQ. Non-rate-limit errors DLQ immediately, same as before.
+ */
+async function handleReader(message: ReaderMessage, invocationContext: InvocationContext) {
```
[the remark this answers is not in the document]
btposey: Ignoring for now since this is a one-time run.

btposey:
No more changes to this PR @sourcery-ai. It's proven to run in an upper environment as is. This is a one-time use migration and I want to use some of the code to distill patterns for future dataflows.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -575,4 +603,298 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
```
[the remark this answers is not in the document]
btposey: Thanks for the suggestion — holding off on this refactor for now. This is a one-time repair dataflow (not an ongoing/reusable pattern), and we plan to extract shared dataflow boilerplate (retry/backoff, escape-hatch, jittered enqueue, queue wiring) into common library functions in a follow-up rather than restructuring this one-off file in place.

backend/function-apps/dataflows/migrations/fix-chapter-7-appointments.ts
```diff
@@ -0,0 +1,426 @@
...
+  queueName: buildQueueName(MODULE_NAME, 'dlq'),
+  connection: STORAGE_QUEUE_CONNECTION,
+});
+
+// Registered function names
+const HANDLE_START = buildFunctionName(MODULE_NAME, 'handleStart');
+const HANDLE_READER = buildFunctionName(MODULE_NAME, 'handleReader');
+const HANDLE_WRITER = buildFunctionName(MODULE_NAME, 'handleWriter');
+
+/**
```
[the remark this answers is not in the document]
btposey: Thanks for the suggestion — holding off on this refactor for now. This is a one-time repair dataflow (not an ongoing/reusable pattern), and we plan to extract shared dataflow boilerplate (retry/backoff, escape-hatch, jittered enqueue, queue wiring) into common library functions in a follow-up rather than restructuring this one-off file in place.

backend/lib/use-cases/dataflows/fix-chapter-7-appointments.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    totalModified += writeResult.modifiedCount;
+    logger.info(
+      MODULE_NAME,
+      `${logPrefix} iteration=${iteration}: wrote ${writeResult.modifiedCount} of ${idPairs.idPairs.length} id pair(s) — totalModified=${totalModified}, elapsedMs=${Date.now() - startedAt}`,
+    );
+  }
+}
+
+async function readWithRetry(
```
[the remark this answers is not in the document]
btposey: Thanks for the suggestion — holding off on this refactor for now. This is a one-time repair dataflow (not an ongoing/reusable pattern), and we plan to extract shared dataflow boilerplate (retry/backoff, escape-hatch, jittered enqueue, queue wiring) into common library functions in a follow-up rather than restructuring this one-off file in place.

backend/lib/adapters/gateways/mongo/utils/mongo-adapter.ts
```diff
@@ -491,6 +491,20 @@ function isTimeoutError(error: unknown): boolean {
...
+// surfaces as a timeout, but the message confirms it's RU-throttling, not a
+// genuinely slow/unindexed query. Distinct from the generic isTimeoutError
+// check above so a non-rate-limit timeout still becomes a GatewayTimeoutError
+// (not retried indefinitely) while this specific signature is treated as
+// retriable, same as a plain 429.
+export function isRateLimitTimeoutError(error: unknown): boolean {
+  if (!(error instanceof Object)) return false;
+  const err = error as Record<string, unknown>;
+  const message = error instanceof Error ? error.message : '';
+  return err['code'] === 50 && /rate limiting/i.test(message);
```
diabagatekelly: **HIGH — verified independently, this is a real bug, not just a coverage gap.** This regex matches the substring "rate limiting" anywhere in the message, regardless of whether the message says the retry-due-to-rate-limiting flag is `True` or `False`. This function is wired into the **shared** `handleError` (line 442) used by every repository, not just this new dataflow.

A pre-existing, still-passing test in this same file (`mongo-adapter.test.ts` around line 494) already models a **genuine, non-retryable** Cosmos timeout with exactly this boilerplate shape: `'Query failed. Request timed out. Retries due to rate limiting: False.'`. That test only still passes today because its fixture has no `code` property set — but if a real Cosmos `ExceededTimeLimit` (code 50) error for a genuinely slow/unindexed, non-RU-caused query carries this same message with `False`, `isRateLimitTimeoutError` will still return `true`, misclassifying a **fatal** timeout (`GatewayTimeoutError`, HTTP 504) as a **retryable** one (`TooManyRequestsError`, HTTP 429). That's not just an internal label — CAMS-706's search-timeout UX keys off the 504 status specifically, and a caller that treats 429 as retryable could retry a query that's failing for a non-RU reason indefinitely.

Suggested fix: match the boolean explicitly (`/rate limiting:\s*true/i`) rather than the bare substring.

**Also worth closing**: `mongo-adapter.test.ts` has no direct test for this function or for `handleError`'s new branch at all — the only coverage is a duplicate of this same classification logic in `trustee-case-appointments.mongo.repository.ts`'s own catch block (see that file's comment below), which never exercises this shared adapter path. Every other dataflow that writes through `MongoCollectionAdapter` currently has zero regression protection for this fix.

backend/function-apps/dataflows/dataflows-paging.ts
```diff
@@ -32,17 +40,17 @@ export function pageByByteBudget<T>(items: T[], maxPageSize?: number): PagingRes
       continue;
     }
 
-    const wouldExceedByteBudget = currentPageBytes + itemBytes > PAGE_BYTE_BUDGET;
+    const candidatePage = [...currentPage, item];
+    const candidateBytes = Buffer.byteLength(JSON.stringify(candidatePage));
```
diabagatekelly, line 44: **HIGH — the actual production-incident fix has zero regression coverage.** This correctly fixes a real `RequestBodyTooLarge` (413) bug (summing per-item bytes ignored array/separator overhead), but `dataflows-paging.test.ts` was not touched by this PR. Verified by reverting this to the old per-item-sum logic and rerunning the existing test suite plus this PR's new 3000-id test (`fix-chapter-7-appointments.test.ts`) — both still pass unchanged, because neither asserts actual per-page byte totals, only page *count* and lossless reassembly. A revert of the real fix would currently pass every automated check.

Suggested fix: add a test with many small items (not artificially padded ones) that would produce a different, over-budget page under the old logic, and assert the exact per-page byte total stays within `PAGE_BYTE_BUDGET`.

**Also LOW**: re-serializing the whole candidate page (`JSON.stringify(candidatePage)`) on every single item addition makes page-building O(n²) in page size. Bounded today by `WRITER_MAX_PAGE_SIZE=2000` in this dataflow, but this is a shared utility other dataflows could call with larger inputs later — could be made O(n) by tracking cumulative bytes incrementally instead.

backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.ts
```diff
@@ -23,9 +23,17 @@ function isObjectIdString(value: unknown): value is string {
...
  * MongoDB stores _id as ObjectId, so string comparisons won't work correctly.
+ * Also coerces each element of an array value (e.g. the rightOperand of a
+ * CONTAINS/$in condition built via .contains(ids)) — without this, an $in
+ * query against _id with an array of hex strings silently matches nothing,
+ * since none of the raw strings ever equal the stored ObjectId values.
  */
 function coerceObjectId(fieldName: string | number | symbol, value: unknown): unknown {
-  if (fieldName === '_id' && isObjectIdString(value)) {
+  if (fieldName !== '_id') return value;
+  if (isArray(value)) {
```
diabagatekelly, line 33: **HIGH — the mechanism this entire feature depends on has no unit-test coverage.** This array branch of `coerceObjectId` is what makes `applyChapterFix`'s `.contains(trusteeIds)`/`.contains(existingCaseIds)` `$in` queries against `_id` actually match anything — without it, those queries would silently match zero documents and this whole dataflow would find pairs but never fix them. `mongo-query-renderer.test.ts` was not touched by this PR; existing tests only cover the scalar `.equals()`/`.greaterThan()` branch. The only place this is exercised end-to-end is the new integration harness (`test/integration/fix-chapter-7-appointments/scripts/run-tests.ts`), which requires manual setup and is not referenced anywhere in `.github/workflows/` — confirmed via grep, zero hits. A revert of this branch would pass every CI check.

The fix itself is correct and low-risk (verified separately: existing `ObjectId` instances pass through unchanged since `isObjectIdString` requires a string; invalid hex strings are left as-is with no crash; no existing caller in the codebase built a `.contains()`/`$in` condition on `_id` before this PR, so nothing to regress) — it just needs its own direct unit test given how shared and load-bearing it is: an array of valid ObjectId hex strings, an array containing an already-constructed `ObjectId`, a mixed valid/invalid array, and an empty array.

backend/function-apps/dataflows/migrations/fix-chapter-7-appointments.test.ts
```diff
@@ -0,0 +1,544 @@
...
+    vi.spyOn(DataflowTelemetry, 'completeDataflowTrace').mockReturnValue(undefined);
+  });
+
+  describe('handleStart', () => {
+    test('enqueues exactly 4 reader messages, one per chapter-fix operation', async () => {
+      const { handleStart } = await import('./fix-chapter-7-appointments');
+      const invocationContext = makeInvocationContext();
+
+      const mockSendMessage = vi.fn().mockResolvedValue(undefined);
+      vi.spyOn(StorageQueueHumbleObject, 'fromConnectionString').mockReturnValue({
```
diabagatekelly: **HIGH.** `mockReturnValue` (not an implementation keyed on the `queueName` argument) makes every call to `fromConnectionString` — whether constructed with `WRITER.queueName` or `READER.queueName` — return the same `sendMessage` spy (repeated at every `handleReader` test in this file, e.g. lines 72, 107, 204, 225). This means no test can actually assert that a writer message went to the writer queue and a reader continuation went to the reader queue — only that *some* message with the right JSON shape was sent to *some* queue. If `handleReader`'s implementation swapped which queue name it passed to `fromConnectionString` (writer-shaped message to the reader queue, or vice versa), every test in this file would still pass — exactly the kind of bug (wrong queue → orphaned message → silent data-coverage loss) this dataflow's correctness most depends on catching.

Suggested fix: mock `fromConnectionString` with an implementation keyed on its `queueName` argument, returning distinct `sendMessage` spies per queue, then assert against the queue-specific spy in each test.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -575,4 +603,298 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+      // This aggregate runs directly against the collection (bypassing
+      // BaseMongoRepository's adapter), so it must classify RU-throttling
+      // itself: Cosmos signals rate limiting either as a plain 429/16500 or,
+      // for a query aborted mid-execution, as code 50 (ExceededTimeLimit)
+      // with "rate limiting" in the message — both must become a
+      // TooManyRequestsError so the reader's caller knows to back off and
+      // retry rather than treating this as a fatal, non-retriable failure.
+      if (
+        !isCamsError(originalError) &&
+        (isRateLimitError(originalError) || isRateLimitTimeoutError(originalError))
```
diabagatekelly: **MEDIUM.** This is a copy of the exact same condition and message from `mongo-adapter.ts`'s `handleError` (needed here since `findAppointmentIdPairsByChapter` bypasses `MongoCollectionAdapter` to run a raw `.aggregate()` directly) — reasonable given that constraint, but it means the regex bug flagged on `mongo-adapter.ts` now has to be fixed in two places that must stay in lockstep. Consider extracting a shared `classifyMongoError`/`toRetriableCamsError` helper used by both call sites.

test/integration/fix-chapter-7-appointments/scripts/run-tests.ts
```diff
@@ -0,0 +1,500 @@
...
+
+const CASE_COLLECTION = 'case-trustee-appointments';
+const TRUSTEE_COLLECTION = 'trustee-case-appointments';
+
+// ---------------------------------------------------------------------------
+// Env
+// ---------------------------------------------------------------------------
+
+function loadEnv() {
+  const localEnvPath = path.join(HARNESS_DIR, '.env.local');
```
diabagatekelly, line 66: **MEDIUM.** This references `.env.local.template` (also at line 66 and 481, and in the doc header), but no such file — under this name or the repo's established convention name `.env.template` (used by every sibling harness under `test/integration/*/scripts/`) — was added by this PR. A developer following the documented workflow ("copy .env.local.template to .env.local") will find nothing to copy. Either add the template file matching the repo's naming convention, or fix these instructions to point at whatever file is actually meant to exist.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
```diff
@@ -575,4 +603,298 @@ export class TrusteeCaseAppointmentsMongoRepository implements TrusteeCaseAppoin
...
+  // case partition has no document for this trustee-partition appointment — a
+  // partition-parity drift. For 'rename', this is repaired by creating the
+  // missing case-partition document with the chapter already corrected (one
+  // combined write, not create-then-update), copying the trustee-partition
+  // document's own fields as the source of truth, mirroring heal()'s existing
+  // partition-repair mechanism (replaceOneInTrusteePartition) in the opposite
+  // direction. For 'delete', a null caseApptId means there is nothing to delete
+  // on the case side — it's skipped, not repaired, since deletion's goal (no
+  // case-partition document with this chapter) is already satisfied.
+  async applyChapterFix(
```
diabagatekelly, line 756: **MEDIUM.** The PR's core safety argument — "overlapping reader batches double-sending the same ids to the writer is harmless because writes are idempotent per-`_id`" — isn't directly tested anywhere. Verified manually that it *is* true (this method always re-applies the `chapter.equals(matchChapter)` filter, so an already-fixed document simply no longer matches and is silently skipped on retry), but since this is the load-bearing claim for the whole reader/writer concurrency design, it deserves its own explicit test: call `applyChapterFix` twice with the same `idPairs` and assert the second call is a no-op (`modifiedCount: 0`, no throw) rather than relying on that behavior falling out of other tests incidentally.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.test.ts
```diff
@@ -1360,6 +1361,450 @@ describe('TrusteeCaseAppointmentsMongoRepository', () => {
...
+      const context = await createMockApplicationContext();
+      const repo = TrusteeCaseAppointmentsMongoRepository.getInstance(context);
+
+      await expect(repo.applyChapterFix([], 'delete', 'AC')).resolves.toEqual({
+        modifiedCount: 0,
+      });
+      repo.release();
+    });
+
+    test('should fix both partitions per call (not a single-collection write)', async () => {
```
diabagatekelly: **MEDIUM.** This test asserts only `updateManySpy).toHaveBeenCalledTimes(2)`, which is a strict subset of the assertions already made by the `'rename — should call updateMany with $set chapter on both partitions'` test above (same call count, plus it inspects both partitions' actual query/update content). This test adds no coverage beyond that one — candidate for deletion, or fold its intent (if any additional signal was meant) into the existing test.

backend/lib/adapters/gateways/mongo/utils/mongo-adapter.ts
btposey: ✅ Fixed in 6cb4a0f1e. `isRateLimitTimeoutError` now matches `/rate limiting:\s*true/i` explicitly instead of the bare `"rate limiting"` substring, so a genuine fatal timeout with `Retries due to rate limiting: False.` is correctly classified as `GatewayTimeoutError` (504), not `TooManyRequestsError` (429). Also extracted a shared `classifyRateLimitError` helper (used by both this file's `handleError` and the repository's raw-aggregate catch block) so the two copies can't drift again. Added direct unit tests for both the true/false distinction and the code!=50 case, plus a test that reproduces your exact "False" fixture with `code: 50` set (which now correctly stays a GatewayTimeoutError).

backend/function-apps/dataflows/dataflows-paging.ts
btposey: ✅ Added a regression test using many small, unpadded items (not the artificially padded heavy items the existing tests use) sized so the array/comma/bracket overhead alone pushes a page over `PAGE_BYTE_BUDGET` even though the sum of individual item sizes stays under it. Verified by reverting to the old per-item-sum logic — the new test fails as expected against that revert, and passes against the current fix.

backend/lib/adapters/gateways/mongo/utils/mongo-query-renderer.ts
btposey: ✅ Added direct unit tests for the array/`$in` branch of `coerceObjectId`: an array of valid ObjectId hex strings, an already-constructed `ObjectId` passed through unchanged, a mixed valid/invalid array, and an empty array. Verified by removing the array branch — the new tests fail as expected, and pass again with the fix in place.

backend/function-apps/dataflows/migrations/fix-chapter-7-appointments.test.ts
btposey: ✅ Replaced `mockReturnValue` with a `mockImplementation` keyed on the `queueName` argument, returning distinct `sendMessage` spies for the reader vs. writer queue, and updated every affected test to assert against the queue-specific spy. Verified by swapping `WRITER.queueName` for `READER.queueName` in the escape-hatch handler — 2 tests failed as expected, confirming the mock now actually catches a wrong-queue bug.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Extracted a shared `classifyRateLimitError(error, moduleName)` helper in `mongo-adapter.ts`, used by both `handleError` and this raw-aggregate catch block, so the classification logic (including the regex fix from the `mongo-adapter.ts` comment above) only lives in one place now.

test/integration/fix-chapter-7-appointments/scripts/run-tests.ts
btposey: ✅ Investigated further: `.env.local.template` files across every `test/integration/*/` harness are actually gitignored repo-wide (`test/integration/.gitignore` only excepts `.env.template`, not `.env.local.template`) and none of them are tracked in git — this is a pre-existing repo-wide pattern, not unique to this harness. Rather than change that repo-wide convention in this PR, I fixed the docs: `run-tests.ts` now tells developers to create `.env.local` directly with `MONGO_CONNECTION_STRING`/`COSMOS_DATABASE_NAME` listed inline, instead of pointing at a template file that was never actually committed.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.ts
btposey: ✅ Added a unit test modeling this exact claim: `applyChapterFix` called twice with the same `idPairs`, with `updateMany` mocked to return `modifiedCount: 1` on the first call and `modifiedCount: 0` on the second (mirroring real Mongo semantics once the `matchChapter` filter excludes the already-fixed documents). Asserts the second call resolves cleanly with `modifiedCount: 0` rather than throwing.

backend/lib/adapters/gateways/mongo/trustee-case-appointments.mongo.repository.test.ts
btposey: ✅ Deleted this test — its only assertion (`updateManySpy` called twice) was already a strict subset of the "rename — should call updateMany with \$set chapter on both partitions" test, which also inspects the actual query/update content.

PR #2768, opened 2026-07-27
CAMS-825 Add flag-gated multi-phone UI for vendor contacts

cruzjone-flexion:
# Purpose

Enable staff to add, edit, and remove multiple typed phone numbers on a software vendor's contact record. Display is also gated on the flag so the overview card shows all phones with type labels when enabled, and only the first direct phone (no label) when disabled.

# Major Changes

- **Edit form** (`SoftwareVendorContactInfoForm`): gates `PhoneEntryList` on `SOFTWARE_VENDOR_TYPED_PHONES` flag. Flag on replaces the single phone + extension fields with the shared typed phone list component. Flag off preserves existing single-phone behavior. Phone entries are pre-sorted on load so edit order matches display order.
- **Overview card** (`BankruptcySoftwareDetailOverview`): flag on shows all typed phones sorted by type/number/extension with type labels. Flag off shows only the first direct phone without a label.
- **Form types** (`softwareVendorContactInfoForm.types.ts`): adds `DEFAULT_PHONE_ENTRY` constant for initializing empty phone list.

# Testing/Validation

Implemented using TDD. 37 form tests and 13 overview tests cover:
- Flag-on: `PhoneEntryList` renders, add/remove entries, multi-phone save payload, validation blocking save, sort order on load
- Flag-off: single phone + extension fields render, digit stripping, single-entry array payload on save
- All 234 UI test files pass; all 230 backend test files pass

# Notes

Uses the same `validateTypedPhones` / `PhoneEntryList` / `sortTypedPhoneNumbers` pattern as the trustee contact forms — no new patterns introduced.

The `SOFTWARE_VENDOR_TYPED_PHONES` flag was already registered in `UseFeatureFlags.ts` and enabled in `testFeatureFlags` from a prior slice, so tests exercise the flag-on path by default without additional mock setup beyond the explicit per-describe mocks added here.

# Definition of Done:

## Summary by Sourcery

Gate software vendor contact phone handling and display behind the SOFTWARE_VENDOR_TYPED_PHONES feature flag.

New Features:
- Support multiple typed phone numbers for software vendor contacts when the feature flag is enabled, using a shared phone list component.
- Enable the overview card to display all vendor phone numbers with type labels when typed phones are enabled.

Enhancements:
- Preserve legacy single phone + extension behavior and map it into the phones array shape when typed phones are disabled.
- Sort vendor phone numbers for consistent ordering in both the edit form and overview display.
- Extend form validation and Save button disabling logic to account for typed-phone validation errors.

Tests:
- Restructure and expand form and overview tests to cover both flag-on multi-phone behavior and flag-off single-phone behavior, including validation, sorting, and save payloads.


user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -168,12 +180,22 @@ export function SoftwareVendorContactInfoForm({
...
+      ? phones.filter((p) => p.number.trim())
+      : trimmedPhone
+        ? [
+            {
+              number: trimmedPhone,
+              type: 'direct' as const,
+              extension: extension.trim() || undefined,
+            },
+          ]
+        : [];
```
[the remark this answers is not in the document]
cruzjone-flexion: ✅ Fixed! Added `.map()` after `.filter()` to trim `number` and `extension` before sending to the API, matching the normalization already applied on the legacy single-phone path.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.test.tsx
```diff
@@ -62,6 +67,7 @@ describe('SoftwareVendorContactInfoForm', () => {
 
   beforeEach(() => {
     alertHook = TestingUtilities.spyOnGlobalAlert();
+    mockUseFeatureFlags.mockReturnValue({ [SOFTWARE_VENDOR_TYPED_PHONES]: true });
```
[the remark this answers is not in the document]
cruzjone-flexion: ✅ Done! Introduced `mockTypedPhonesOn()` and `mockTypedPhonesOff()` helpers. Moved flag-independent tests (email/website validation, address fields, cancel, contact names) to the top-level describe. Restricted `with typed phones` and `with legacy single phone` blocks to behaviors that genuinely depend on which phone UI is active.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
```diff
@@ -36,7 +40,14 @@ export function BankruptcySoftwareDetailOverview({
...
 
-  const displayPhones = contact?.phones?.length ? contact.phones : undefined;
+  const directPhone = contact?.phones?.find((p) => p.type === 'direct');
+  const displayPhones = typedPhonesEnabled
+    ? contact?.phones?.length
+      ? sortTypedPhoneNumbers(contact.phones)
+      : undefined
+    : directPhone
+      ? [directPhone]
+      : undefined;
```
jalba-flexion: This section is hard to read

btposey (commented):
---

btposey (commented):
## Architecture follow-up (do in this PR)

- [ ] Move `TypedPhoneNumber`, `PhoneType`, `PHONE_TYPE_LABELS`, `sortTypedPhoneNumbers` from `@common/cams/trustees` to `common/src/cams/contact.ts`
- [ ] Move `validateTypedPhones` from `@/trustees/forms/trusteeForms.utils` to a shared forms-utils module near `PhoneEntryList` (`lib/components/cams/`)
- [ ] Update all import sites (trustees + bankruptcy-software) accordingly
- [ ] No behavior change — pure relocation

<details>
<summary>Appendix: prompt for an AI coding agent</summary>

Relocate general-purpose phone-handling utilities out of the `trustees` feature module so `bankruptcy-software` (and any other future consumer) doesn't have a cross-feature dependency on `trustees`.

Context: `SoftwareVendorContactInfoForm.tsx` and `BankruptcySoftwareDetailOverview.tsx` under `admin/bankruptcy-software/` currently import `TypedPhoneNumber` and `sortTypedPhoneNumbers` from `@common/cams/trustees`, and `validateTypedPhones` from `@/trustees/forms/trusteeForms.utils`. These types/utilities aren't trustee-specific — they became homed under `trustees` only because trustees was the first consumer of typed phone numbers. The shared `PhoneEntryList` component already lives in `lib/components/cams/`, so these should live alongside it instead of inside a specific feature's folder.

Do the following as a pure move (no behavior change, no API change):
1. Move `TypedPhoneNumber`, `PhoneType`, and `PHONE_TYPE_LABELS` from wherever they're currently defined in `common/src/cams/trustees.ts` (or similar) to `common/src/cams/contact.ts`.
2. Move `sortTypedPhoneNumbers` alongside them in `common/src/cams/contact.ts`.
3. Move `validateTypedPhones` from `user-interface/src/trustees/forms/trusteeForms.utils.ts` to a new shared forms-utils module colocated with `PhoneEntryList` under `user-interface/src/lib/components/cams/PhoneEntryList/` (or an existing shared forms-utils file if one already exists near there).
4. Update every import site across the codebase (trustees forms/components, bankruptcy-software forms/components, and any tests) to reference the new locations.
5. Run the full frontend and common package test suites and confirm no behavior changes — this should be a mechanical move with updated import paths only.
6. Do not rename any exported symbols; only their module location changes.

</details>

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
```diff
@@ -36,7 +40,14 @@ export function BankruptcySoftwareDetailOverview({
       }
     : undefined;
 
-  const displayPhones = contact?.phones?.length ? contact.phones : undefined;
+  const directPhone = contact?.phones?.find((p) => p.type === 'direct');
```
diabagatekelly: **Should block merge — the legacy-phone-drop bug from #2762 is not fixed, just relocated.**

```ts
const directPhone = contact?.phones?.find((p) => p.type === 'direct');
const displayPhones = typedPhonesEnabled
  ? contact?.phones?.length ? sortTypedPhoneNumbers(contact.phones) : undefined
  : directPhone ? [directPhone] : undefined;
```
`SoftwareContactInfo` no longer declares a `phone` field at all (removed in #2762) — neither branch here, flag-on or flag-off, ever consults it. The migration that converts legacy `contact.phone` → `contact.phones` in Cosmos is a manually-run, dry-run-by-default script, decoupled from deploy. Until it's run with `CONFIRM=true` against every environment, any unmigrated vendor record shows no phone number, in both flag states. This was flagged as a blocker on #2762 (then: `contact.phones?.length ? contact.phones : undefined`); it's the identical defect under a different expression, still with no test covering an unmigrated record in either flag-on or flag-off describe blocks.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -54,6 +61,11 @@ export function SoftwareVendorContactInfoForm({
   const [phone, setPhone] = useState(existingPhone?.number ?? '');
```
diabagatekelly: **Should block merge — new mismatch between this form and the Overview display.**

```ts
const existingPhone = existingContact?.phones?.[0];
```
picks index 0 with no filter on `type`, while `BankruptcySoftwareDetailOverview.tsx` picks `phones.find(p => p.type === 'direct')` for the same flag-off scenario. If a vendor has multiple typed phones (e.g. created while the flag was on) and the flag is later toggled off — plausible, since flags can differ per environment — this form pre-fills whatever's at index 0 (could be a `fax` number) while the overview correctly shows the `direct` one or nothing. Untested: no fixture here uses a multi-entry `phones` array with a non-direct first element.

This goes away if the flag-off path reuses `DirectPhoneFields` against the same `phones` array instead of maintaining its own `phone`/`extension` state — see the comment on the JSX below.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -168,12 +180,28 @@ export function SoftwareVendorContactInfoForm({
...
 
+    const resolvedPhones = typedPhonesEnabled
+      ? phones
+          .filter((p) => p.number.trim())
+          .map((p) => ({
+            ...p,
+            number: p.number.trim(),
+            extension: p.extension?.trim() || undefined,
+          }))
+      : trimmedPhone
```
diabagatekelly: **Confirmed via mutation testing: this branch (`: []` when no phone entered) is not actually verified by any test.** Swapping `[]` for a bogus non-empty array here still passed all 37 tests. The base branch (`CAMS-825-story`) had `'should send empty phones array when no phone is entered'` covering this exact case before the flag/off paths were split; it was preserved only under the `with typed phones` describe (line ~410), not duplicated for `with legacy single phone` (describe starts at line 482) — a real coverage regression on the currently-default (flag-off) path.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
```diff
@@ -243,28 +271,36 @@ export function SoftwareVendorContactInfoForm({
...
-                onChange={handleExtensionChange}
-              />
+          {typedPhonesEnabled ? (
+            <PhoneEntryList
+              phones={phones}
+              onChange={setPhones}
+              errors={validateTypedPhones(phones)}
+            />
+          ) : (
+            <div className="phone-extension-row">
```
diabagatekelly: **DRY not actually fixed for the flag-off path.** This hand-rolls the same `PhoneNumberInput` + extension `Input` pair that `DirectPhoneFields` (`user-interface/src/lib/components/cams/DirectPhoneFields/DirectPhoneFields.tsx`) already provides and that `TrusteeContactForm.tsx` reuses via `typedPhonesEnabled ? <PhoneEntryList/> : <DirectPhoneFields/>`. This PR's description frames the hand-rolled single-phone UI as something being eliminated — it was eliminated for the flag-on case (good, `PhoneEntryList` reuse below) but reintroduced here for flag-off, which is also the root cause of the index-vs-type mismatch flagged at line 61. Swapping this block for `<DirectPhoneFields phones={phones} onChange={setPhones} />` (reading/writing the same `phones` state used by the flag-on branch) would fix both issues at once and let this component drop its separate `phone`/`extension`/`extensionRef` state entirely.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.test.tsx
```diff
@@ -409,4 +336,206 @@ describe('SoftwareVendorContactInfoForm', () => {
...
+      renderForm();
+      const phoneInput = screen.getByLabelText('Phone Number');
+      fireEvent.change(phoneInput, { target: { value: '303' } });
+      await waitFor(() => {
+        expect(screen.getByTestId('button-save-contact-info')).toBeDisabled();
+      });
+    });
+  });
+
+  describe('with legacy single phone', () => {
```
diabagatekelly, line 478: **Two tests missing from this describe block, confirmed by mutation testing to be real gaps, not just style nits:**

1. No test uses `softwareWithContact` (the fixture with an existing phone) under flag-off — every test here uses the contact-less `software` fixture. Swapping `existingContact?.phones?.[0]` for `[1]` at line 61 still passes all 37 tests. Suggest adding:
```ts
test('should pre-fill phone and extension from existing contact when flag is off', () => {
  renderForm(softwareWithContact);
  expect(screen.getByDisplayValue('303-555-1234')).toBeInTheDocument();
  expect(screen.getByDisplayValue('101')).toBeInTheDocument();
});
```
2. No test verifies the empty-phones-array save case under flag-off (see comment on line 191) — this exact case was covered on the base branch before the flag split and needs a flag-off equivalent alongside the one for `with typed phones`.

diabagatekelly (commented):
Deep review (code-review + tests-audit + cams-spec-review angles), specifically checking whether this follow-up genuinely fixes the two structural issues flagged on #2762 (dead `SOFTWARE_VENDOR_TYPED_PHONES` flag, hand-rolled phone UI instead of reusing shared components).

**Verdict:** the flag-wiring fix is genuine and well-tested. Everything else only half-landed — see inline comments. Mutation testing (5 targeted mutations against the flag-off path) found 4 of 5 survive the full 37-test suite unchanged, meaning real bugs in that path would currently ship undetected.

Two lower-priority items couldn't be inlined since they sit outside this PR's diff (pre-existing code/tests, unchanged here), noting for completeness:
- **Mount-time phone validation is untested** (`SoftwareVendorContactInfoForm.tsx:84-85`, `existingContact?.phones?.[0]?.number` fed into `fieldErrors` on mount) — pre-existing from #2762, still no test for an existing contact with an invalid phone number, unlike the equivalent email/website tests.
- **A few Overview tests assert on `FormattedContact`'s rendered text** (e.g. `'(Direct)'`/`'(Fax)'` labels, address text) rather than the props `Overview` computes — pre-existing pattern (lines 91-121, untouched by this PR), duplicates coverage `FormattedContact.test.tsx` should own.

Overall: recommend fixing the flag-off `DirectPhoneFields` reuse (comment below) before merge — it resolves the index-vs-type mismatch and the DRY half-fix in one move — and restoring the three flag-off tests mutation testing showed are currently non-functional.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.test.tsx
```diff
@@ -108,15 +132,55 @@ describe('BankruptcySoftwareDetailOverview', () => {
...
+          phones: [
+            { number: '212-555-0100', type: 'direct' },
+            { number: '212-555-0200', type: 'fax' },
+          ],
+        },
+      };
+      renderOverview(softwareMultiPhone);
+      expect(screen.getByText('212-555-0100')).toBeInTheDocument();
+      expect(screen.getByText('212-555-0200')).toBeInTheDocument();
+      expect(screen.getByText('(Direct)')).toBeInTheDocument();
```
diabagatekelly, line 161: MEDIUM — this test asserts presence only (`getByText`), not order, so it doesn't actually verify `sortTypedPhoneNumbers` is being applied. Confirmed via mutation: deleting the `sortTypedPhoneNumbers()` call in `BankruptcySoftwareDetailOverview.tsx:46` still passes this test, because the fixture already lists `direct` before `fax` — which happens to match the sorted output regardless of whether sorting ran. `FormattedContact.tsx` renders phones by array index (`data-testid="phone-${idx}"`), so order is directly testable — suggest feeding an out-of-order fixture (fax before direct) and asserting render order via `getAllByTestId`.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.test.tsx
```diff
@@ -409,4 +336,206 @@ describe('SoftwareVendorContactInfoForm', () => {
...
+      expect(phoneInputs[1].value).toBe('303-555-0200');
+    });
+
+    test('should pre-fill phone and extension from existing contact', () => {
+      renderForm(softwareWithContact);
+      expect(screen.getByDisplayValue('303-555-1234')).toBeInTheDocument();
+      expect(screen.getByDisplayValue('101')).toBeInTheDocument();
+    });
+
+    test('should add a second phone entry when "Add Another Phone" is clicked', () => {
```
diabagatekelly: LOW/MEDIUM — wrong component boundary: this test drives `PhoneEntryList`'s own "Add Another Phone" button and asserts on its internal DOM (a new blank row appearing), duplicating coverage `PhoneEntryList.test.tsx` should already own for its own add-row behavior. Since `PhoneEntryList` is a pre-existing, already-tested shared component, tests here should assert only this form's integration contract with it (props passed in, `onChange` received, save payload shape) — which the adjacent save-payload tests already do well. Consider removing this test or reworking it to assert something at the parent boundary that isn't already covered.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
cruzjone-flexion: Agreed — we extracted and simplified this logic in the latest commit. We also plan to continue refactoring `FormattedContact` to own phone normalization internally in a follow-up PR, which will make this even cleaner.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.tsx
cruzjone-flexion: Intentionally not adding a `contact.phone` fallback in production code. The migration script (`ops/migrations/CAMS-825-software-vendor-phone-types.js`) is the correct fix for unmigrated records — it runs directly on load via mongosh, applies the `phone` → `phones` transform, and is idempotent. Defending against a post-migration state that should never exist would leave permanent dead code. The migration must be run against each environment before deploy; that is a procedural dependency, not a code fix.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
cruzjone-flexion: ✅ Fixed in c7f420bb8. Replaced the flag-off path with `<DirectPhoneFields phones={phones} onChange={setPhones} errors={validateDirectPhoneFields(phones)} />`, sharing the same `phones` state as the flag-on branch. This eliminates the index-vs-type mismatch (form was picking `phones[0]`, overview was picking `phones.find(type=direct)`) and removes the now-redundant `phone`/`extension`/`extensionRef` state entirely — matching the established pattern from `TrusteeContactForm`.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
cruzjone-flexion: ✅ Fixed in c7f420bb8. The flag-on/flag-off split in `resolvedPhones` is gone — both paths now share a single `phones` state and a unified filter/map. The empty-phones-array save case for flag-off is also covered by the new `"should send empty phones array when no phone is entered and flag is off"` test in the `with legacy single phone` describe block.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.tsx
cruzjone-flexion: ✅ Fixed in c7f420bb8 alongside the line-61 mismatch. The hand-rolled `PhoneNumberInput` + extension block is replaced by `<DirectPhoneFields>`, so both flag states now read/write the same `phones` array and the form no longer maintains separate `phone`/`extension`/`extensionRef` state.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.test.tsx
cruzjone-flexion: ✅ Fixed in c7f420bb8. Added to the `with legacy single phone` describe block: (1) `"should pre-fill phone and extension from existing direct contact phone"` using `softwareWithContact`, and (2) `"should send empty phones array when no phone is entered and flag is off"` covering the empty-save path. Phone validation error tests (`show phone error on mount`, `show phone error on change`, `clear phone error`) are also now correctly scoped under the flag-off block.

user-interface/src/admin/bankruptcy-software/BankruptcySoftwareDetailOverview.test.tsx
cruzjone-flexion: ✅ Fixed in 6f7aa9b30. Added a new test with a reversed fixture (`fax` before `direct`) that asserts DOM render order via `container.querySelectorAll(".phone")`, confirming `phone[0]` is the direct number and `phone[1]` is fax. This test fails if `sortTypedPhoneNumbers` is removed, closing the mutation-testing gap.

user-interface/src/admin/bankruptcy-software/SoftwareVendorContactInfoForm.test.tsx
cruzjone-flexion: ✅ Removed in 5a6cfc495. Agreed — this test was asserting `PhoneEntryList`'s own add-row behavior, which is already covered in `PhoneEntryList.test.tsx`. The form's integration contract with `PhoneEntryList` (props in, onChange received, save payload shape) is fully covered by the adjacent save-payload tests.

jalba-flexion (approved):

cruzjone-flexion:
✅ Done in a87e00ded. Relocated as requested — pure move, no behavior change:

- `sortTypedPhoneNumbers` (+ `compareTypedPhoneNumbers`) moved from `common/src/cams/trustees.ts` → `common/src/cams/contact.ts`, alongside `TypedPhoneNumber`, `PhoneType`, and `PHONE_TYPE_LABELS`
- `validateTypedPhones` moved from `src/trustees/forms/trusteeForms.utils.ts` → new `src/lib/components/cams/PhoneEntryList/phoneEntryList.utils.ts`, colocated with `PhoneEntryList`
- `trustees.ts` and `trusteeForms.utils.ts` now re-export from the new canonical homes so existing callers remain unaffected
- All direct cross-feature imports in `bankruptcy-software` and `lib/components/cams` updated to the new locations

PR #2757, opened 2026-07-27
CAMS-760 Branch deploys use Azure Deployment Stacks for network teardown

bityogi:
# Bug

Branch teardown was deleting the shared dev Key Vault (`kv-ustp-cams-dev`) and its role assignments, breaking secret access for all branches (GH #2749). Root cause: an interim version of this work wrapped the whole app deployment (`main.bicep`) in an Azure Deployment Stack. `main.bicep` deploys resources cross-scope into shared resource groups (the app-config Key Vault + role assignments and SQL vnet rules in `AZURE_RG`; the action group in the analytics RG), and a deployment stack manages every resource its template creates in any RG — so `az stack group delete` removed the shared vault on teardown. This PR's final design does NOT stack the app deploy, eliminating that risk.

# Purpose

Move branch deployment teardown off the blunt `az group delete`-everything model toward a resource-tracked model (Azure Deployment Stacks for the self-contained network tier), as the first slice of CAMS-760 (co-locating branch deploys in stable resource groups for least-privilege RBAC), without ever putting shared resources in a stack's managed set.

# Major Changes

* Extract `network.bicep` as a resource-group-scoped entry template so the per-branch network resources (vnet, subnets, private DNS zone) deploy as their own Azure Deployment Stack. `main.bicep` now consumes the subnets via `existing` references instead of deploying them, and drops the now-unused network params.
* Add `azure-deploy-network.sh` to deploy the network tier (as a stack for branches, plain deployment for main) and own the vnet-existence / `deployVnet` logic previously in `azure-deploy.sh`.
* Keep the app deploy as a plain `az deployment group create` (NOT a stack), because `main.bicep` reaches cross-scope into shared RGs. App resources live in the per-branch app RG and are torn down by deleting that RG.
* Rework `az-delete-branch-resources.sh`: delete the per-branch app and network RGs directly (removing the KV private endpoint + subnets in one operation, avoiding `InUseSubnetCannotBeDeleted`); keep the by-name teardown of e2e Cosmos/SQL and the log analytics workspace; add a safety guard that aborts if any delete target is a shared RG; parameterize the stack unmanage action for the future shared-RG slice.
* RG teardown is scoped strictly to hash-suffixed per-branch RGs; main deploy path behavior is unchanged.

# Testing/Validation

Validated live against the Flexion DOJ USTP dev subscription with a full deploy + teardown cycle:

[...]

Issue #2749, opened 2026-07-24
Branch deployment-stack teardown deletes shared dev Key Vault (kv-ustp-cams-dev)


bityogi:
# Branch deployment-stack teardown deletes the SHARED dev Key Vault (kv-ustp-cams-dev)

## Description

The CAMS-760 (Option E, Slice 1) change that wraps the branch app deployment in an **Azure Deployment Stack** causes branch **teardown** to delete the shared development Key Vault `kv-ustp-cams-dev` and its ~15 role assignments. This takes down secret access for **all** branches (Setup jobs fail resolving `kv-ustp-cams-dev.vault.usgovcloudapi.net`) and required manual recovery of the resource group and re-granting of permissions. It has occurred at least twice.

## Root cause

- `ops/cloud-deployment/main.bicep` includes the `kvSetup` module (`ustp-cams-kv-app-config-setup.bicep`), which **creates** the app-config Key Vault via `module appConfigKeyvault './lib/keyvault/keyvault.bicep'` with `keyVaultName: kvName`, scoped cross-RG to `kvAppConfigResourceGroupName`.
- For **branch** deploys, `kvAppConfigResourceGroupName` = `AZURE_RG` = `bankruptcy-oversight-support-systems` — a **shared** resource group that holds the real `kv-ustp-cams-dev`.
- Slice 1 deploys `main.bicep` as an **Azure Deployment Stack** (`az stack group create`). A deployment stack **manages every resource its template creates in ANY resource group**, so the shared `kv-ustp-cams-dev` (and its role assignments, plus SQL vnet rules in the same RG, and the analytics action group in `rg-analytics`) became part of the stack's managed set.
- Teardown runs `az stack group delete --action-on-unmanage deleteAll`, which deletes **all managed resources across all RGs** — including the shared dev Key Vault.

## Steps to Reproduce

1. Deploy a branch to the dev subscription with `deployBranch=true` (creates `${stackName}-app` stack; its `kvSetup` module manages `kv-ustp-cams-dev` in the shared `bankruptcy-oversight-support-systems` RG).
2. Tear the branch down via `azure-remove-branch.yml` (`az stack group delete --action-on-unmanage deleteAll`).
3. Observe `kv-ustp-cams-dev` and its role assignments deleted from the shared RG.

## Expected Behavior

[...]

PR #2773, opened 2026-07-28
CAMS-760 Co-locate branch app/network into shared RGs

bityogi:
# Purpose

Slice 2 of CAMS-760 (Option E): stop branch deploys from creating their own per-branch `<base>-<hash>` app/network resource groups, and instead deploy into the same stable RGs `main` already uses — distinguished by per-branch-unique resource names. This removes the dynamic per-branch RG that today forces subscription-scope Contributor on the branch deploy identity (Azure RBAC has no wildcard scoping over dynamically-named RGs), setting up Slice 3's RBAC cutover to RG-scoped least privilege.

# Major Changes

* Split `main.bicep` so the app tier can deploy as its own RG-scoped Deployment Stack, symmetric with the network tier from Slice 1. New `app-shared-setup.bicep` holds the genuinely shared cross-scope resources (the app-config Key Vault + its role assignments, and a deduplicated SQL managed identity) as an always-plain (non-stack) deployment.
* Added `az_stack_deploy_func` to `azure-deploy.sh`; branches now deploy the app tier as a stack, main unchanged.
* Added stack-aware teardown for the app tier in `az-delete-branch-resources.sh`, mirroring the network tier's existing `deleteAll`/`deleteResources` conditional from Slice 1.
* The cutover: dropped the `-<hash>` RG-name suffix for branches; fixed the VNet naming override (branches already own their own VNet — they were just forced onto an identical name; now each gets `vnet-<stackName>` again); forced `deployDns=false` for branches on both Private DNS Zones (a newly discovered risk with the same shape as the GH #2749 shared-KV incident — every branch's stack would otherwise manage the same shared DNS zone); switched branch teardown to `deleteResources`; added by-name cleanup for stack-unmanaged Smart Detection alert rules; replaced the nightly-cleanup discovery mechanism (RG tags → network stack tags).

# Testing/Validation

Static gates pass: `az bicep build` clean on all new/changed bicep, `shellcheck` clean on all new/changed scripts, workflow YAML validates, and the repo's `check-kv-secret-parity` pre-commit hook passes (no new/undocumented KV secrets introduced). Full `npm test`/`typecheck`/`coverage:backend`/`knip` also pass (no JS/TS touched by this PR; run for completeness).

**Live Azure validation has NOT been run** — no Azure access during implementation. Before merge, the 8 scenarios listed in the Slice 2 implementation notes (`.ustp-cams-fdp/ai/specs/CAMS-760_branch-deploy-shared-rgs/branch-deploy-shared-rgs.slice-2-implementation-notes.md`) need a real `workflow_dispatch` run: single-branch deploy into the shared RGs, two-branch concurrency (no name/network collisions), shared-resource-untouched checks (KV, SQL identity, both DNS zones), teardown of one branch leaving the other + main intact, idempotent re-teardown, nightly-cleanup discovery, main-unaffected regression, and the slot-swap regression check.

# Notes

* Targets `CAMS-760-azure-resource-groups-for-dev` (Slice 1, PR #2757) as base rather than `main`, since Slice 1 hasn't merged yet. Should be retargeted to `main` if Slice 1 merges first.
* Design + implementation-prompts + implementation-notes docs live in the `.ustp-cams-fdp` submodule under `ai/specs/CAMS-760_branch-deploy-shared-rgs/`.
* Two bugs were found only by tracing the actual code (not anticipated by the design doc) and fixed as part of this PR: `az-delete-branch-resources.sh` was unconditionally suffixing the RG name with the branch hash and gating its idempotent-teardown fast path on raw RG existence — both silently wrong once the RG is shared. `reusable-build-info.yml` independently recomputed the same RG names and did its own VNet-existence check feeding `deployVnet`/`initialDeployment` (which gates the blue-green slot-swap flow) — left unfixed, every branch deploy would have permanently looked like a first-ever deploy.
* Out of scope, deferred separately: the `acsEmail` shared-KV-secret-clobbering issue discovered along the way (pre-existing, unrelated to RG co-location).

# Definition of Done:

## Summary by Sourcery

Co-locate branch deployments into the shared app and network resource groups used by main while isolating per-branch resources via deployment stacks and unique names, and adjust teardown and workflows to respect the new shared-RG model.

New Features:
- Introduce a dedicated app-shared-setup deployment that provisions shared Key Vault, DNS, and SQL managed identity resources outside of per-branch stacks.
- Deploy branch app-tier resources as Azure Deployment Stacks, parallel to the existing network stack approach, enabling stack-scoped lifecycle management.

Bug Fixes:
- Ensure branch teardown correctly handles shared resource groups by using stack existence instead of resource-group existence and by avoiding hash-based RG suffixing in shared-RG mode.
- Fix VNet naming and existence checks so each branch uses and validates its own uniquely named virtual network rather than reusing main’s name.
- Prevent branch deployments from managing shared Private DNS Zones, avoiding accidental deletion of shared DNS configuration on teardown.
- Remove duplicate and inconsistent computation of app and network resource group names between workflows, aligning them on shared, non-suffixed names.
- Ensure stack-unmanaged Smart Detection alert rules created by Application Insights are explicitly cleaned up in shared app resource groups.

Enhancements:
- Refine branch teardown logic to distinguish between legacy per-branch resource groups and shared-RG stack-based cleanup, including post-teardown verification of stacks vs. groups.
- Tag and parameterize deployment stacks and scripts with branch metadata to support discovery, idempotent cleanup, and future least-privilege RBAC scoping.
- Update nightly branch-cleanup workflow to discover orphaned branches via tagged network stacks instead of per-branch tagged resource groups.
- Externalize creation of the read-only SQL managed identity into a single shared deployment and consume it as an existing resource from app templates.
- Simplify app and network deployment workflows by centralizing shared-RG names and stack naming conventions across build and deploy steps.

CI:
- Adjust reusable deploy and build-info GitHub workflows to deploy into stable shared app/network resource groups, pass stack metadata, and control DNS and VNet behavior based on whether the run is for main or a branch.
- Update the branch-removal GitHub workflow to detect and clean up branch environments via deployment stacks in the shared network resource group and to use stack-based unmanage actions during teardown.

Deployment:
- Extend deployment scripts to support both plain and stack-based group deployments, including new helpers for app shared-setup and app-tier stack deployment, and to propagate DNS-related parameters to the appropriate templates.

Documentation:
- Document in code comments the shared-resource vs. per-branch stack design, including how DNS zones, Key Vault, and SQL identities are safely managed across main and branch environments.

### DIFF
--- a/.github/workflows/azure-remove-branch.yml
+++ b/.github/workflows/azure-remove-branch.yml
@@ -31,12 +31,19 @@ jobs:
           environment: ${{ vars.AZURE_ENVIRONMENT }}
       - name: List existing tagged Azure branch resources
         run: |
-          echo "List Azure resource groups tagged as branch deployment resources"
-          az group list --query "[?tags.isBranchDeployment == 'true'].{ Name:name Branch:tags.branchName HashId:tags.branchHashId }" -o table
+          # CAMS-760, Option E / Slice 2: branches no longer create their own
+          # tagged resource group — discover them via their network Deployment
+          # Stack in the shared network RG instead (still carries the same
+          # isBranchDeployment/branchName/branchHashId tags from Slice 1).
+          networkRgBase=$(az keyvault secret show --vault-name kv-ustp-cams-dev --name AZ-NETWORK-RG --query value -o tsv)
+          echo "List Azure network deployment stacks tagged as branch deployment resources"
+          az stack group list -g "${networkRgBase}" --query "[?tags.isBranchDeployment == 'true'].{ Name:name Branch:tags.branchName HashId:tags.branchHashId }" -o table
 
   check:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.hashId != '') || github.event.ref_type == 'branch' }}
+    if:
+      ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' &&
+      inputs.hashId != '') || github.event.ref_type == 'branch' }}
     environment: 'remove-branch'
     permissions:
       contents: read
@@ -76,6 +83,16 @@ jobs:
         run: |
           set -euo pipefail
 
+          # CAMS-760, Option E / Slice 2: branches no longer create their own
+          # tagged resource group — discover them via their network Deployment
+          # Stack in the shared network RG instead (still carries the same
+          # isBranchDeployment/branchName/branchHashId tags from Slice 1). Stack
+          # names are "<stackName>-network" where stackName already carries the
+          # hash (e.g. "ustp-cams-dev-<hash>-network"), so the defensive
+          # name/tag consistency check matches on that suffix instead of the
+          # old "<rg-base>-<hash>" resource-group-name suffix.
+          networkRgBase=$(az keyvault secret show --vault-name kv-ustp-cams-dev --name AZ-NETWORK-RG --query value -o tsv)
+
           if [[ "${GITHUB_EVENT_NAME}" == "schedule" ]]; then
             localHour=$(TZ=America/Los_Angeles date +%H)
             if [[ "${localHour}" != "21" ]]; then
@@ -85,14 +102,14 @@ jobs:
               exit 0
             fi
 
-            branchResources=$(az group list --query "[?tags.isBranchDeployment == 'true' && tags.branchHashId != null].{name:name, branchName:tags.branchName, branchHashId:tags.branchHashId}" -o json)
-            targets=$(echo "${branchResources}" | jq -c '
+            branchStacks=$(az stack group list -g "${networkRgBase}" --query "[?tags.isBranchDeployment == 'true' && tags.branchHashId != null].{name:name, branchName:tags.branchName, branchHashId:tags.branchHashId}" -o json)
+            targets=$(echo "${branchStacks}" | jq -c '
               [
                 .[]
                 | select(.branchHashId | test("^[0-9a-f]{6}$"))
                 | select((.branchName // "") as $branch | $branch != "main" and $branch != "refs/heads/main")
                 | .branchHashId as $branchHashId
-                | select(.name | endswith("-" + $branchHashId))
+                | select(.name | test("-" + $branchHashId + "-network$"))
                 | .branchHashId
               ]
               | unique
@@ -117,14 +134,14 @@ jobs:
             fi
 
             echo "Do existing Azure branch resources with hash $target exist?"
-            branchResources=$(az group list --query "[?tags.isBranchDeployment == 'true' && tags.branchHashId == '$target'].{name:name, branchName:tags.branchName, branchHashId:tags.branchHashId}" -o json)
-            count=$(echo "${branchResources}" | jq '
+            branchStacks=$(az stack group list -g "${networkRgBase}" --query "[?tags.isBranchDeployment == 'true' && tags.branchHashId == '$target'].{name:name, branchName:tags.branchName, branchHashId:tags.branchHashId}" -o json)
+            count=$(echo "${branchStacks}" | jq '
               [
                 .[]
                 | select(.branchHashId | test("^[0-9a-f]{6}$"))
                 | select((.branchName // "") as $branch | $branch != "main" and $branch != "refs/heads/main")
                 | .branchHashId as $branchHashId
-                | select(.name | endswith("-" + $branchHashId))
+                | select(.name | test("-" + $branchHashId + "-network$"))
               ]
               | length
             ')
@@ -217,4 +234,5 @@ jobs:
             --network-resource-group="${AZ_NETWORK_RG_BASE}" \
             --analytics-resource-group="${AZ_ANALYTICS_RG}" \
             --stack-name="${STACK_NAME}" \
-            --short-hash="${SHORT_HASH}"
+            --short-hash="${SHORT_HASH}" \
+            --unmanage-action=deleteResources
--- a/ops/cloud-deployment/app-shared-setup.bicep
+++ b/ops/cloud-deployment/app-shared-setup.bicep
@@ -1,0 +1,115 @@
+// Shared cross-scope resources for the app tier (CAMS-760, Option E / Slice 2).
+//
+// Deploys resources that are genuinely SHARED across main and every branch: the
+// app-config Key Vault (+ its managed identity, its own private DNS zone, and
+// per-secret role assignments) and the read-only SQL managed identity used by
+// the API and dataflows function apps. Always a plain (non-stack) deployment
+// for both main and branches — these resources must never be managed by a
+// branch's Deployment Stack. An earlier version wrapped the whole app deploy
+// (including these cross-scope resources) in a stack, and a branch teardown
+// deleted the shared Key Vault + ~15 role assignments (GH #2749). This template
+// must be deployed before main.bicep (which references the KV identity and SQL
+// identity by name/id, not by bicep dependency, since they now live in a
+// separate deployment).
+targetScope = 'resourceGroup'
+
+param stackName string
+
+param location string = resourceGroup().location
+
+param isUstpDeployment bool = false
+
+@description('Flag: determines the setup of DNS Zone, Link virtual networks to zone.')
+param deployDns bool = true
+
+param networkResourceGroupName string
+
+param virtualNetworkName string = 'vnet-${stackName}'
+
+param privateEndpointSubnetName string = 'snet-${stackName}-private-endpoints'
+
+param privateDnsZoneResourceGroup string = resourceGroup().name
+
+@description('DNS Zone Subscription ID. USTP uses a different subscription for prod deployment.')
+param privateDnsZoneSubscriptionId string = subscription().subscriptionId
+
+@description('Resource group containing the app-config Key Vault')
+param kvAppConfigResourceGroupName string
+
+@description('Name of the app-config Key Vault')
+param kvAppConfigName string = 'kv-${stackName}'
+
+@description('Name of the managed identity with read access to the keyvault storing application configurations.')
+@secure()
+param idKeyvaultAppConfiguration string
+
+param sqlServerName string = ''
+
+param sqlServerResourceGroupName string = ''
+
+@description('Name for managed identity of database server.')
+param sqlServerIdentityName string = ''
+
+param sqlServerIdentityResourceGroupName string = ''
+
+param deployedAt string = utcNow()
+
+var tags = {
+  app: 'cams'
+  component: 'shared-setup'
+  'deployed-at': deployedAt
+}
+
+// The virtual network and its subnets are deployed separately by network.bicep
+// (its own Deployment Stack for branches). This template only needs the
+// private-endpoint subnet, for the Key Vault's private endpoint.
+resource ustpVirtualNetwork 'Microsoft.Network/virtualNetworks@2023-11-01' existing = {
+  name: virtualNetworkName
+  scope: resourceGroup(networkResourceGroupName)
+}
+
+resource privateEndpointSubnetExisting 'Microsoft.Network/virtualNetworks/subnets@2023-11-01' existing = {
+  name: privateEndpointSubnetName
+  parent: ustpVirtualNetwork
+}
+
+module kvSetup './ustp-cams-kv-app-config-setup.bicep' = {
+  name: '${stackName}-kv-setup-module'
+  params: {
+    stackName: stackName
+    location: location
+    deployDns: deployDns
+    kvResourceGroup: kvAppConfigResourceGroupName
+    kvName: kvAppConfigName
+    networkResourceGroup: networkResourceGroupName
+    virtualNetworkName: virtualNetworkName
+    privateEndpointSubnetId: privateEndpointSubnetExisting.id
+    privateDnsZoneResourceGroup: privateDnsZoneResourceGroup
+    privateDnsZoneSubscriptionId: privateDnsZoneSubscriptionId
+    managedIdentityName: idKeyvaultAppConfiguration
+    makeRoleAssignment: !isUstpDeployment
+  }
+}
+
+// The read-only SQL managed identity used by both the API and dataflows
+// function apps to authenticate to the SQL server. Its name comes from a
+// fixed value (the AZ_SQL_IDENTITY_NAME secret) shared by main and every
+// branch, so it is created exactly once here rather than per-function-app.
+// Creating it inside a branch's app stack would let that branch's teardown
+// delete the identity every other branch and main depend on — the same bug
+// shape as the shared Key Vault incident above.
+var sqlIdentityName = !empty(sqlServerIdentityName) ? sqlServerIdentityName : 'id-sql-${stackName}-readonly'
+var sqlIdentityRG = !empty(sqlServerIdentityResourceGroupName)
+  ? sqlServerIdentityResourceGroupName
+  : sqlServerResourceGroupName
+var createSqlIdentity = !empty(sqlServerResourceGroupName) && !empty(sqlServerName) && !isUstpDeployment
+
+module sqlManagedIdentity './lib/identity/managed-identity.bicep' = if (createSqlIdentity) {
+  scope: resourceGroup(sqlIdentityRG)
+  name: '${stackName}-sql-identity-module'
+  params: {
+    managedIdentityName: sqlIdentityName
+    location: location
+    tags: tags
+  }
+}
--- a/ops/cloud-deployment/backend-api-deploy.bicep
+++ b/ops/cloud-deployment/backend-api-deploy.bicep
@@ -471,20 +471,14 @@ module setApiFunctionSqlServerVnetRule './lib/network/sql-vnet-rule.bicep' = if 
   }
 }
 
+// The identity itself is created once, in app-shared-setup.bicep (CAMS-760,
+// Option E / Slice 2) — its name is a fixed value shared by main and every
+// branch, so it must never be created/managed inside a branch's app stack.
+// Referenced here as `existing` only.
 var sqlIdentityName = !empty(sqlServerIdentityName) ? sqlServerIdentityName : 'id-sql-${apiFunctionName}-readonly'
 var sqlIdentityRG = !empty(sqlServerIdentityResourceGroupName)
   ? sqlServerIdentityResourceGroupName
   : sqlServerResourceGroupName
-
-module sqlManagedIdentity './lib/identity/managed-identity.bicep' = if (createSqlServerVnetRule) {
-  scope: resourceGroup(sqlIdentityRG)
-  name: '${apiFunctionName}-sql-identity-module'
-  params: {
-    managedIdentityName: sqlIdentityName
-    location: location
-    tags: tags
-  }
-}
 
 resource sqlIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' existing = {
   name: sqlIdentityName
--- a/ops/cloud-deployment/dataflows-resource-deploy.bicep
+++ b/ops/cloud-deployment/dataflows-resource-deploy.bicep
@@ -736,21 +736,14 @@ module setDataflowFunctionSqlServerVnetRule './lib/network/sql-vnet-rule.bicep' 
   }
 }
 
-// Creates a managed identity that would be used to grant access to function instance
+// The identity itself is created once, in app-shared-setup.bicep (CAMS-760,
+// Option E / Slice 2) — its name is a fixed value shared by main and every
+// branch, so it must never be created/managed inside a branch's app stack.
+// Referenced here as `existing` only.
 var sqlIdentityName = !empty(sqlServerIdentityName) ? sqlServerIdentityName : 'id-sql-${apiFunctionName}-readonly'
 var sqlIdentityRG = !empty(sqlServerIdentityResourceGroupName)
   ? sqlServerIdentityResourceGroupName
   : sqlServerResourceGroupName
-
-module sqlManagedIdentity './lib/identity/managed-identity.bicep' = if (createSqlServerVnetRule) {
-  scope: resourceGroup(sqlIdentityRG)
-  name: '${dataflowsFunctionName}-sql-identity-module'
-  params: {
-    managedIdentityName: sqlIdentityName
-    location: location
-    tags: tags
-  }
-}
 
 resource sqlIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' existing = {
   name: sqlIdentityName
--- a/ops/cloud-deployment/main.bicep
+++ b/ops/cloud-deployment/main.bicep
@@ -10,9 +10,6 @@ param virtualNetworkName string = 'vnet-${stackName}'
 
 param networkResourceGroupName string
 
-@description('Flag: determines the setup of DNS Zone, Link virtual networks to zone.')
-param deployDns bool = true
-
 param privateDnsZoneName string = 'privatelink.azurewebsites.us'
 
 param privateDnsZoneResourceGroup string = networkResourceGroupName
@@ -176,23 +173,12 @@ resource dataflowsFunctionSubnetExisting 'Microsoft.Network/virtualNetworks/subn
   parent: ustpVirtualNetwork
 }
 
-module kvSetup './ustp-cams-kv-app-config-setup.bicep' = {
-  name: '${stackName}-kv-setup-module'
-  params: {
-    stackName: stackName
-    location: location
-    deployDns: deployDns
-    kvResourceGroup: kvAppConfigResourceGroupName
-    kvName: kvAppConfigName
-    networkResourceGroup: networkResourceGroupName
-    virtualNetworkName: virtualNetworkName
-    privateEndpointSubnetId: privateEndpointSubnetExisting.id
-    privateDnsZoneResourceGroup: privateDnsZoneResourceGroup
-    privateDnsZoneSubscriptionId: privateDnsZoneSubscriptionId
-    managedIdentityName: idKeyvaultAppConfiguration
-    makeRoleAssignment: !isUstpDeployment
-  }
-}
+// The app-config Key Vault (+ its managed identity, DNS zone, and secret role
+// assignments) and the SQL managed identity are deployed separately by
+// app-shared-setup.bicep — always a plain (non-stack) deployment, before this
+// template runs, because they are genuinely shared across main and every
+// branch (CAMS-760, Option E / Slice 2; see app-shared-setup.bicep for why).
+// This template references them by the name/id strings passed in as params.
 
 module ustpWebapp 'frontend-webapp-deploy.bicep' = {
     name: '${stackName}-webapp-module'
@@ -227,7 +213,6 @@ module ustpWebapp 'frontend-webapp-deploy.bicep' = {
 
 module acsEmail './lib/email/acs-email.bicep' = {
   name: '${stackName}-acs-email-module'
-  dependsOn: [kvSetup]
   params: {
     stackName: stackName
     kvAppConfigName: kvAppConfigName
@@ -244,7 +229,7 @@ module acsEmail './lib/email/acs-email.bicep' = {
 module ustpApiFunction 'backend-api-deploy.bicep' = {
     name: '${stackName}-function-module'
     scope: resourceGroup(appResourceGroup)
-    dependsOn: [kvSetup, acsEmail]
+    dependsOn: [acsEmail]
     params: {
       deployAppInsights: deployAppInsights
       analyticsWorkspaceId: deployAppInsights ? analyticsWorkspaceId : ''
@@ -290,7 +275,6 @@ module ustpApiFunction 'backend-api-deploy.bicep' = {
 module ustpDataflowsFunction 'dataflows-resource-deploy.bicep' = {
   name: '${stackName}-dataflows-module'
   scope: resourceGroup(appResourceGroup)
-  dependsOn: [kvSetup]
   params: {
     deployAppInsights: deployAppInsights
     analyticsWorkspaceId: deployAppInsights ? analyticsWorkspaceId : ''
--- a/ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
+++ b/ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+
+# Title:        azure-deploy-app-shared-setup.sh
+# Description:  Deploy the USTP CAMS app-tier shared setup resources (the
+#               app-config Key Vault + its managed identity/role assignments,
+#               and the read-only SQL managed identity) into the shared
+#               AZURE_RG. Always a plain resource-group deployment for both
+#               main and branches (CAMS-760, Option E) — these resources are
+#               genuinely shared and must never be managed by a branch's
+#               Deployment Stack.
+#
+# Exitcodes
+# ==========
+# 0   No error
+# 1   Script interrupted
+# 2   Unknown flag or switch passed as parameter to script
+# 10+ Validation check errors
+
+set -euo pipefail # ensure job step fails in CI pipeline when error occurs
+
+deployment_file=''
+resource_group=''
+stack_name=''
+network_rg=''
+vnet_name=''
+kv_app_config_rg=''
+kv_app_config_name=''
+id_keyvault_app_config=''
+sql_server_name=''
+sql_server_rg=''
+sql_server_identity_name=''
+sql_server_identity_rg=''
+is_ustp_deployment=false
+deploy_dns=true
+location=''
+extra_parameters=''
+
+while [[ $# -gt 0 ]]; do
+    case $1 in
+    -f | --file)
+        deployment_file="${2}"
+        shift 2
+        ;;
+    --resource-group)
+        resource_group="${2}"
+        shift 2
+        ;;
+    --stackName)
+        stack_name="${2}"
+        shift 2
+        ;;
+    --networkResourceGroupName)
+        network_rg="${2}"
+        shift 2
+        ;;
+    --virtualNetworkName)
+        vnet_name="${2}"
+        shift 2
+        ;;
+    --kvAppConfigResourceGroupName)
+        kv_app_config_rg="${2}"
+        shift 2
+        ;;
+    --kvAppConfigName)
+        kv_app_config_name="${2}"
+        shift 2
+        ;;
+    --idKeyvaultAppConfiguration)
+        id_keyvault_app_config="${2}"
+        shift 2
+        ;;
+    --sqlServerName)
+        sql_server_name="${2}"
+        shift 2
+        ;;
+    --sqlServerResourceGroupName)
+        sql_server_rg="${2}"
+        shift 2
+        ;;
+    --sqlServerIdentityName)
+        sql_server_identity_name="${2}"
+        shift 2
+        ;;
+    --sqlServerIdentityResourceGroupName)
+        sql_server_identity_rg="${2}"
+        shift 2
+        ;;
+    --isUstpDeployment)
+        is_ustp_deployment="${2}"
+        shift 2
+        ;;
+    --deployDns)
+        deploy_dns="${2}"
+        shift 2
+        ;;
+    -l | --location)
+        location="${2}"
+        shift 2
+        ;;
+    # Space-delimited "key=value" bicep parameters passed straight through
+    -p | --parameters)
+        extra_parameters="${2}"
+        shift 2
+        ;;
+    *)
+        echo "Exit on param: ${1}"
+        exit 2
+        ;;
+    esac
+done
+
+if [[ -z "${deployment_file}" || -z "${resource_group}" || -z "${stack_name}" || -z "${network_rg}" || -z "${vnet_name}" || -z "${kv_app_config_rg}" || -z "${location}" ]]; then
+    echo "Error: --file, --resource-group, --stackName, --networkResourceGroupName, --virtualNetworkName, --kvAppConfigResourceGroupName and --location are required"
+    exit 10
+fi
+
+deployment_parameters="stackName=${stack_name} location=${location} networkResourceGroupName=${network_rg} virtualNetworkName=${vnet_name} kvAppConfigResourceGroupName=${kv_app_config_rg} isUstpDeployment=${is_ustp_deployment} deployDns=${deploy_dns}"
+[[ -n "${kv_app_config_name}" ]] && deployment_parameters="${deployment_parameters} kvAppConfigName=${kv_app_config_name}"
+[[ -n "${id_keyvault_app_config}" ]] && deployment_parameters="${deployment_parameters} idKeyvaultAppConfiguration=${id_keyvault_app_config}"
+[[ -n "${sql_server_name}" ]] && deployment_parameters="${deployment_parameters} sqlServerName=${sql_server_name}"
+[[ -n "${sql_server_rg}" ]] && deployment_parameters="${deployment_parameters} sqlServerResourceGroupName=${sql_server_rg}"
+[[ -n "${sql_server_identity_name}" ]] && deployment_parameters="${deployment_parameters} sqlServerIdentityName=${sql_server_identity_name}"
+[[ -n "${sql_server_identity_rg}" ]] && deployment_parameters="${deployment_parameters} sqlServerIdentityResourceGroupName=${sql_server_identity_rg}"
+if [[ -n "${extra_parameters}" ]]; then
+    deployment_parameters="${deployment_parameters} ${extra_parameters}"
+fi
+
+echo "Deploying app shared-setup resources to ${resource_group} (plain resource-group deployment; always non-stack — see app-shared-setup.bicep)"
+# shellcheck disable=SC2086 # REASON: intentional word-splitting of --parameter
+az deployment group create -w -g "${resource_group}" --template-file "${deployment_file}" --parameter ${deployment_parameters}
+# shellcheck disable=SC2086 # REASON: intentional word-splitting of --parameter
+az deployment group create -g "${resource_group}" --template-file "${deployment_file}" --parameter ${deployment_parameters}
--- a/ops/scripts/utility/az-delete-branch-resources.sh
+++ b/ops/scripts/utility/az-delete-branch-resources.sh
@@ -201,24 +201,30 @@ fi
 
 # Tear down the branch's app and network tiers (CAMS-760, Option E).
 #
-# APP tier: NOT a deployment stack. main.bicep deploys resources cross-scope into
-# SHARED resource groups (the app-config Key Vault + its role assignments and SQL
-# vnet rules in AZURE_RG; the action group in the analytics RG). A deployment stack
-# manages every resource its template creates in ANY resource group, so deleting an
-# app stack would delete those shared resources — this is what deleted the shared
-# kv-ustp-cams-dev (GH #2749). The app resources live in the per-branch app RG, so
-# we tear them down by deleting that resource group directly. Deleting the per-branch
-# app RG cannot touch shared resources, which live in other (shared) RGs.
+# APP tier: a self-contained per-branch deployment stack. The genuinely shared
+# cross-scope resources (the app-config Key Vault + its role assignments, and the
+# SQL managed identity) are deployed separately by app-shared-setup.bicep, always
+# as a plain (non-stack) deployment, so the app stack itself only ever manages
+# app-RG-scoped resources (webapp, functions, app insights, plans, comms/email).
+# An earlier version wrapped those cross-scope resources into the branch's own
+# app stack, and a teardown deleted the shared kv-ustp-cams-dev (GH #2749) — this
+# split is what makes stacking the app tier safe. For per-branch teardown we
+# delete the whole app RG directly rather than deleting the stack first, mirroring
+# the network tier's per-branch behavior below. Only when a shared app RG must be
+# preserved (Slice 2, unmanage_action=deleteResources) do we fall back to a scoped
+# stack delete.
 #
 # NETWORK tier: a self-contained per-branch deployment stack (network.bicep only
 # touches the per-branch network RG). For per-branch teardown we delete the whole
 # network RG directly rather than deleting the stack first: the branch's Key Vault
-# private endpoint (pep-kv-ustp-cams-dev) is created in the network RG by the app-side
-# kvSetup module and is NOT stack-managed, so a stack delete fails with
-# InUseSubnetCannotBeDeleted (the PE still occupies the private-endpoint subnet).
-# `az group delete` removes the PE, subnets, vnet, and stack in one shot regardless of
-# ordering. Only when a shared network RG must be preserved (Slice 2,
-# unmanage_action=deleteResources) do we fall back to a scoped stack delete.
+# private endpoint (pep-kv-ustp-cams-dev) is created in the network RG by
+# app-shared-setup.bicep's kvSetup module and is NOT stack-managed, so a stack
+# delete fails with InUseSubnetCannotBeDeleted (the PE still occupies the
+# private-endpoint subnet). `az group delete` removes the PE, subnets, vnet, and
+# stack in one shot regardless of ordering. Only when a shared network RG must be
+# preserved (Slice 2, unmanage_action=deleteResources) do we fall back to a scoped
+# stack delete.
+appStack="${stack_name}-app"
 networkStack="${stack_name}-network"
 
 function stack_exists() {
@@ -228,8 +234,17 @@ function stack_exists() {
 }
 
 if [[ "${rgAppExists}" == "true" ]]; then
-    echo "Deleting app resource group ${app_rg} (per-branch; contains only branch-owned app resources)"
-    az group delete -n "${app_rg}" --yes
+    if [[ "${unmanage_action}" != "deleteResources" ]]; then
+        # Per-branch app RG: delete the whole RG.
+        echo "Deleting app resource group ${app_rg} (per-branch; contains only branch-owned app resources)"
+        az group delete -n "${app_rg}" --yes
+    elif [[ -n "$(stack_exists "${appStack}" "${app_rg}")" ]]; then
+        # Shared app RG (Slice 2): preserve the RG, remove only this branch's stack.
+        echo "Start deleting app deployment stack ${appStack} (action-on-unmanage=${unmanage_action})"
+        az stack group delete --name "${appStack}" --resource-group "${app_rg}" --action-on-unmanage "${unmanage_action}" --yes
+    else
+        echo "No app deployment stack ${appStack} found; nothing to delete in shared app RG"
+    fi
 fi
 
 if [[ "${rgNetExists}" == "true" ]]; then
@@ -330,13 +345,20 @@ fi
 
 echo "Completed resource clean up operations."
 
-# Verify nothing was left behind. The per-branch app RG is always deleted outright.
-# The network tier's RG is deleted for per-branch RGs (unmanage_action != deleteResources);
-# for a preserved shared network RG (deleteResources, Slice 2) verify the stack is gone.
+# Verify nothing was left behind. Per-branch RGs (unmanage_action != deleteResources)
+# are deleted outright and verified gone; for a preserved shared RG (deleteResources,
+# Slice 2) verify each tier's stack is gone instead.
 failed=false
-if [[ $(az group exists -n "${app_rg}") == "true" ]]; then
-    echo "ERROR: App resource group ${app_rg} still exists after deletion attempt." >&2
-    failed=true
+if [[ "${unmanage_action}" != "deleteResources" ]]; then
+    if [[ $(az group exists -n "${app_rg}") == "true" ]]; then
+        echo "ERROR: App resource group ${app_rg} still exists after deletion attempt." >&2
+        failed=true
+    fi
+else
+    if [[ -n "$(stack_exists "${appStack}" "${app_rg}")" ]]; then
+        echo "ERROR: App deployment stack ${appStack} still exists after deletion attempt." >&2
+        failed=true
+    fi
 fi
 if [[ "${unmanage_action}" != "deleteResources" ]]; then
     if [[ $(az group exists -n "${network_rg}") == "true" ]]; then
--- a/ops/scripts/utility/az-delete-branch-resources.sh
+++ b/ops/scripts/utility/az-delete-branch-resources.sh
@@ -146,48 +146,78 @@ done
 fi
 
 # Check which resources exist (partial cleanup is normal if a previous run partially succeeded)
-app_rg="${app_rg}-${hash_id}"
-network_rg="${net_rg}-${hash_id}"
+#
+# Slice 2 (unmanage_action=deleteResources): app_rg/network_rg are ALREADY the
+# correct shared base names passed in by the caller — do NOT suffix them with
+# the branch hash, that would point at a per-branch RG that no longer exists.
+# Legacy per-branch mode (deleteAll): suffix with the branch hash, as before.
+if [[ "${unmanage_action}" == "deleteResources" ]]; then
+    network_rg="${net_rg}"
+else
+    app_rg="${app_rg}-${hash_id}"
+    network_rg="${net_rg}-${hash_id}"
+fi
 e2e_db="cams-e2e-${hash_id}"
 stack_name="${stack_name}-${hash_id}"
+appStack="${stack_name}-app"
+networkStack="${stack_name}-network"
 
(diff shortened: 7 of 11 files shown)


bityogi:
Responding to Sourcery's review:

1. **`azure-deploy-app-shared-setup.sh` running `az deployment group create` twice** — this is intentional, not a duplicate: the first call (`-w`) is a preview/what-if, the second is the real apply. This is the established preview-then-apply pattern used elsewhere in this repo (`azure-deploy.sh`'s `az_deploy_func`, `azure-deploy-rg.sh`'s `az_deploy_func`) — not something to collapse.

2. **Hardcoded `kv-ustp-cams-dev` in `azure-remove-branch.yml`'s discovery steps** — valid finding, fixed in e20d34afc. Both steps now derive the vault name via the same `GHA_ENVIRONMENT`-based conditional the rest of the file already uses, rather than an inline literal. Functionally identical today (this workflow only ever tears down branch/dev deployments), but keeps the vault-name derivation consistent everywhere.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -147,48 +181,59 @@ done
...
-  fi
 
   if [[ -z "${app_rg:-}" || -z "${db_account:-}" || -z "${db_rg:-}" || -z "${kv_rg:-}" || -z "${net_rg:-}" || -z "${stack_name:-}" || -z "${hash_id:-}" ]]; then
   error "Not all required parameters provided. Run this script with the --help flag for details, or set the appropriate environment variables." 2
 fi
 
 # Check which resources exist (partial cleanup is normal if a previous run partially succeeded)
-app_rg="${app_rg}-${hash_id}"
-network_rg="${net_rg}-${hash_id}"
+#
```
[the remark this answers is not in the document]
bityogi: Already handled — the final verification block checks both the app stack AND the network stack in `deleteResources` mode (see lines ~486-506 as of this commit):

```bash
if [[ "${unmanage_action}" != "deleteResources" ]]; then
    if [[ $(az group exists -n "${network_rg}") == "true" ]]; then
        echo "ERROR: Network resource group ${network_rg} still exists after deletion attempt." >&2
        failed=true
    fi
else
    if [[ -n "$(stack_exists "${networkStack}" "${network_rg}")" ]]; then
        echo "ERROR: Network deployment stack ${networkStack} still exists after deletion attempt." >&2
        failed=true
    fi
fi
```

This mirrors the app-tier check immediately above it, exactly as suggested. No code change needed.

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
```diff
@@ -0,0 +1,157 @@
...
+        attempt=$((attempt + 1))
+        delaySeconds=$((delaySeconds * 2))
+    done
+}
+
+echo "Deploying app shared-setup resources to ${resource_group} (plain resource-group deployment; always non-stack — see app-shared-setup.bicep)"
+# shellcheck disable=SC2086 # REASON: intentional word-splitting of --parameter
+az_deploy_with_retry_func az deployment group create -w -g "${resource_group}" --template-file "${deployment_file}" --parameter ${deployment_parameters}
+# shellcheck disable=SC2086 # REASON: intentional word-splitting of --parameter
+az_deploy_with_retry_func az deployment group create -g "${resource_group}" --template-file "${deployment_file}" --parameter ${deployment_parameters}
```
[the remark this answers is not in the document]
bityogi: Same as the earlier general-comment reply on this PR: intentional, not accidental duplication. The first call (`-w`) is a preview/what-if, the second is the real apply — the established preview-then-apply pattern used elsewhere in this repo (`azure-deploy.sh`'s `az_deploy_func`, `azure-deploy-network.sh`, `azure-deploy-rg.sh`'s `az_deploy_func`). The retry wrapper added around both calls doesn't change that — it just makes each individual call (preview or apply) resilient to a concurrent branch racing on the same shared RG, not a replacement for running the real apply. No code change needed.

bityogi:
## Independent review before requesting human review

Ran a 5-agent adversarial review (pragmatic, DevSecOps, architecture, SRE, product perspectives) against this PR before asking for review, given this touches deploy/teardown logic with a real incident history (GH #2749). Also rebased onto Slice 1's latest fixes. Findings and fixes, all as separate commits:

**2 verified bugs found and fixed:**
- KV private endpoint name collision across branches — was named by the shared `kvName` instead of the per-branch `stackName`, so concurrent/sequential branch deploys collided on one PE resource.
- `deployDns=false` was incorrectly also skipping linking the caller's own VNet to the DNS zone, breaking private DNS resolution for every branch reusing an existing (main-owned) zone.

**Hardening items addressed:**
- Corrected an overclaiming comment in `main.bicep` about what's cross-scope vs. RG-local.
- Guarded the latent `actionGroup` stack-scope risk (safe today only because `createAlerts` is Main-Gov-only).
- Made the shared-RG safety guard in `az-delete-branch-resources.sh` run unconditionally (previously skipped entirely in `deleteResources` mode).
- Stopped the Smart Detection alert-rule cleanup from swallowing genuine `az` CLI failures as "no rules found."
- Added Slack alerting for branch teardown failures (previously silent).
- Added a recovery runbook in the teardown script's header comment.
- Documented the `deleteAll` legacy path as an intentional manual-only escape hatch, not dead code.
- Hardened branch stacks with `--deny-settings-mode denyDelete` (and fixed a now-stale pre-commit guard that would've blocked legitimate app-tier stacking).
- Added retry/backoff to `azure-deploy-app-shared-setup.sh` for concurrent branch deploy races on the shared RG.
- Fixed a secret-masking inconsistency for `AZ-NETWORK-RG` in the nightly cleanup workflow.
- Cross-referenced the duplicated vnet-naming formula across the bicep files and both workflow YAML files.

All changes validated with `shellcheck`, `bash -n`, `az bicep build`, and YAML parsing. Live Azure validation (a real two-branch deploy + teardown cycle) is still outstanding before merge.

ops/cloud-deployment/main.bicep
```diff
@@ -235,7 +249,6 @@ module ustpWebapp 'frontend-webapp-deploy.bicep' = {
 
 module acsEmail './lib/email/acs-email.bicep' = {
```
diabagatekelly, line 258: **BLOCKER — reopens the GH #2749 bug shape this PR exists to prevent.** This module call is unconditional (no `if`, unlike `actionGroup` above which is gated on `createAlerts`). `lib/email/acs-email.bicep`'s two KV-secret submodules write **fixed-name** secrets (`ACS-EMAIL-CONNECTION-STRING`, `ACS-EMAIL-SENDER-ADDRESS`) cross-scope into `kvAppConfigResourceGroupName` — the same shared RG for every branch and main. `main.bicep` is now wrapped in an Azure Deployment Stack for branches (`azure-deploy.sh`'s new `az_stack_deploy_func`). Deployment Stacks track managed resources by resource ID regardless of RG, so these two fixed-name secrets enter *every branch's own* app-stack managed set. Any branch deploy overwrites them with its own values; tearing that branch down (`az-delete-branch-resources.sh`'s `az stack group delete --action-on-unmanage deleteResources`) deletes them from the shared Key Vault — breaking email notifications for main and every other running branch. Unlike this same PR's KV-setup/SQL-identity extraction into `app-shared-setup.bicep` (which correctly avoided exactly this), `acsEmail` was missed.

Verified independently 4 separate times (code-quality audit, DevSecOps audit, manual read, and an adversarial `security-review` verification pass scoring 8/10 confidence) — all four traced the identical mechanism from scratch.

Suggested fix: move `acsEmail`'s two KV-secret-writing submodules (or the whole module) into `app-shared-setup.bicep` (always-plain deployment), or branch-qualify the secret names if per-branch email config is actually intended.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -274,8 +328,40 @@ if [[ "${rgAppExists}" == "true" ]]; then
...
-        az group delete -n "${app_rg}" --yes
+
+        if [[ "${unmanage_action}" != "deleteResources" ]]; then
+            # Per-branch app RG: delete the whole RG.
+            echo "Deleting app resource group ${app_rg} (per-branch; contains only branch-owned app resources)"
+            az group delete -n "${app_rg}" --yes
+        else
+            # Shared app RG (Slice 2): preserve the RG, remove only this branch's stack.
+            echo "Start deleting app deployment stack ${appStack} (action-on-unmanage=${unmanage_action})"
+            az stack group delete --name "${appStack}" --resource-group "${app_rg}" --action-on-unmanage "${unmanage_action}" --yes
```
diabagatekelly: **BLOCKER — every nightly branch teardown will fail here.** The branch's Key Vault private endpoint (`pep-${stackName}`) is created by `app-shared-setup.bicep` (always a plain, non-stacked deployment) inside this same shared network RG, in the branch's own private-endpoint subnet. Nothing in this script ever deletes it: `kv_rg`/`--kv-resource-group` is parsed but used *only* for the shared-RG safety-guard comparison earlier in this file, never to locate/delete the PE. When the network-tier teardown (below) runs `az stack group delete` on the branch's network stack, Azure will refuse to delete the subnet with `InUseSubnetCannotBeDeleted` because the PE still occupies it — the same failure mode the pre-existing comment two blocks up (webapp/functionapp vnet-integration removal) was specifically added to avoid for the *app* tier's own subnet, just not for this one.

Suggested fix: add an explicit, existence-checked delete of `pep-${stack_name}` (and its DNS vnet link, in the KV's own RG) before the network-stack teardown runs, mirroring the vnet-integration-removal pattern already used above in this same function.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -274,8 +328,40 @@ if [[ "${rgAppExists}" == "true" ]]; then
...
+
+            # Application Insights auto-creates Smart Detection alert rules that are
+            # never declared in bicep, so the stack never manages/deletes them. The
+            # per-branch path sweeps these up for free via the whole-RG delete above;
+            # the shared-RG path cannot do that, so clean them up by name instead.
+            echo "Checking for stack-unmanaged Smart Detection alert rules for ${stack_name}"
```
diabagatekelly, line 453: **HIGH — ordering risk.** This Smart Detection alert-rule cleanup (added specifically because these rules are stack-unmanaged and can block deletion) runs *after* the app-stack delete two lines above, in the same `set -e` subshell. If a stack-unmanaged Smart Detection rule can actually block `az stack group delete` (as the runbook comment near the top of this file implies for a similar case), this cleanup code will never execute, because the failure it's meant to route around happens first and aborts the subshell before reaching it.

Worth confirming against real Azure behavior: does `az stack group delete` actually fail when an unmanaged Smart Detection rule references the App Insights resource being deleted, or does the rule just become orphaned non-blocking? If it can block, this cleanup needs to run *before* the stack delete, not after.

ops/cloud-deployment/main.bicep
```diff
@@ -14,13 +14,14 @@ param appResourceGroup string = resourceGroup().name
 // lookups below (ustpVirtualNetwork, *SubnetExisting) will fail to find what
 // network.bicep actually created. Fails loudly (deploy error), not silently,
 // but keep both files' naming formulas in lockstep when changing either one.
+// app-shared-setup.bicep also duplicates this same virtualNetworkName default
+// (its own `existing` vnet/subnet lookups), and reusable-deploy.yml /
+// reusable-build-info.yml duplicate it again for their vnet-existence checks
+// (PR #2773 review) — same lockstep requirement applies to all of them.
 param virtualNetworkName string = 'vnet-${stackName}'
```
diabagatekelly: **HIGH.** This `vnet-${stackName}` naming formula is now duplicated across five files: `main.bicep` (here), `network.bicep`, the new `app-shared-setup.bicep`, `reusable-deploy.yml`'s `AZ_NETWORK_VNET_NAME` derivation, and `reusable-build-info.yml`'s vnet-existence check. All five are comment-cross-referenced (good defensive documentation), but the cross-referencing is manual and mechanically unenforced — no shared source of truth, no automated check. This PR's own "Notes" section admits two real bugs were found only by tracing code, not the design doc, and both were exactly this class of drift (RG-suffix duplication, VNet-existence-check duplication). Worth considering a shared Bicep naming-function file or a single reusable workflow step, now that the duplication has grown from 2 files (Slice 1) to 5.

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
```diff
@@ -0,0 +1,157 @@
...
+    deployment_parameters="${deployment_parameters} ${extra_parameters}"
+fi
+
+# Every branch (plus main) deploys this template to the SAME shared resource
+# group, and `az deployment group create` defaults its deployment name to the
+# template's base filename when --name is not given, so concurrent branch
+# pushes routinely race on the same deployment name and can hit a transient
+# 409 AnotherOperationInProgress. Retry with backoff rather than failing the
+# whole pipeline run on what is usually just a timing collision.
+function az_deploy_with_retry_func() {
```
diabagatekelly, line 142: **MEDIUM.** This retry wrapper retries *any* non-zero exit from the wrapped command up to 3 times with backoff (15s/30s), not just the `AnotherOperationInProgress`/409 conflicts it's meant to paper over. A genuine template validation error (bad parameter, bad Bicep) will now silently eat ~45s of pointless retries before finally surfacing, on every single deploy (main and every branch). Consider inspecting the error text/exit condition for the specific conflict signature before retrying — `az deployment group create`'s exit code alone doesn't distinguish a real error from a timing collision.

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
diabagatekelly: **MEDIUM.** No `--name` is passed to this deployment, so `az deployment group create` defaults the deployment *record* name to the template's base filename (`app-shared-setup`) — every branch and main race on that same name when deploying concurrently, relying on the retry wrapper above to paper over the resulting 409s rather than avoiding the collision. Passing `--name "${stack_name}-shared-setup"` would remove this whole class of contention rather than retrying through it, reducing latency and log noise on every deploy.

.pre-commit-config.yaml
```diff
@@ -57,8 +57,8 @@ repos:
         language: python
         files: '^ops/cloud-deployment/lib/cosmos/mongo/cosmos-collections\.bicep$'
       - id: guard-app-deploy-not-stacked
-        name: Guard app deploy against re-stacking (GH #2749)
+        name: Guard against shared KV re-entering the app stack (GH #2749)
         entry: .pre-commit-hooks/guard-app-deploy-not-stacked.sh
         language: script
-        files: '^ops/scripts/pipeline/azure-deploy\.sh$'
+        files: '^ops/cloud-deployment/main\.bicep$'
```
diabagatekelly: **MEDIUM.** This guard now only fires when `main.bicep` itself changes, and (per the hook script) only greps for KV-setup references. It would not catch: (a) the SQL-managed-identity bug class reappearing in `backend-api-deploy.bicep`/`dataflows-resource-deploy.bicep` (both are part of `main.bicep`'s stacked module tree but wouldn't trigger this `files:` filter), or (b) the `acsEmail` finding above — a fixed-name cross-scope write that isn't KV-setup-shaped at all. Consider widening `files:` to also match `backend-api-deploy\.bicep$|dataflows-resource-deploy\.bicep$|app-shared-setup\.bicep$|lib/email/acs-email\.bicep$`, and widening the grep to a more general "module .*managed-identity\.bicep.*= if\(" / fixed-secret-name pattern rather than just the two known KV-setup strings.

.github/workflows/azure-remove-branch.yml
```diff
@@ -30,13 +30,34 @@ jobs:
...
+            KV="kv-ustp-cams-dev"
+          fi
+          # CAMS-760, Option E / Slice 2: branches no longer create their own
+          # tagged resource group — discover them via their network Deployment
+          # Stack in the shared network RG instead (still carries the same
+          # isBranchDeployment/branchName/branchHashId tags from Slice 1).
+          networkRgBase=$(az keyvault secret show --vault-name "$KV" --name AZ-NETWORK-RG --query value -o tsv)
+          echo "::add-mask::${networkRgBase}"
+          echo "List Azure network deployment stacks tagged as branch deployment resources"
+          az stack group list -g "${networkRgBase}" --query "[?tags.isBranchDeployment == 'true'].{ Name:name Branch:tags.branchName HashId:tags.branchHashId }" -o table
```
diabagatekelly, line 57: **MEDIUM.** Nightly-cleanup discovery is now keyed solely on network-stack tags in the shared network RG. `az-delete-branch-resources.sh` still independently checks both the app stack and network stack's existence and warns-but-continues if one tier is missing — so a branch whose app stack exists but whose network stack was never created (or was torn down out-of-band) would never be discovered by this nightly sweep at all, since discovery never looks at the app RG. Given network is deployed before app in the pipeline, this is a narrow window, but it would leak an app stack silently forever with no alert. Worth also discovering via app-stack tags (union of both) rather than network-stack tags alone.

ops/cloud-deployment/main.bicep
bityogi: ✅ Fixed in b1de6bfab. Branch-qualified both secret names (`ACS-EMAIL-CONNECTION-STRING-${stackName}` / `ACS-EMAIL-SENDER-ADDRESS-${stackName}`) rather than moving the module into app-shared-setup.bicep — communicationService/emailService above are already stackName-qualified per-branch resources, so moving just the secrets would make the connection string shared while the ACS resource it points at stays per-branch. Updated `backend-api-deploy.bicep`'s two `@Microsoft.KeyVault(...)` app settings to read the same qualified name (main.bicep now also passes `stackName` through to that module call, which it previously didnt). Also widened the `guard-app-deploy-not-stacked` pre-commit hook to catch this whole bug class mechanically going forward (see reply on the MEDIUM finding below).

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: ✅ Fixed in b1de6bfab. Added existence-checked deletes of both `pep-${stack_name}` (via `az resource list`/`az resource delete`, in `network_rg`) and its DNS zone vnet link `privatelink.vaultcore.usgovcloudapi.net-vnet-link-${stack_name}` (via `az network private-dns link vnet list`/`az resource delete`, in `kv_rg`), placed right before the network stack delete in the `deleteResources` branch — mirrors the existing vnet-integration-removal pattern as suggested.

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: ✅ Fixed in b1de6bfab — reordered so the Smart Detection cleanup now runs before the app stack delete, not after, so it always executes regardless of whether an unmanaged rule can block the stack delete.

ops/cloud-deployment/main.bicep
bityogi: ✅ Fixed in b1de6bfab. `app-shared-setup.bicep` now imports `virtualNetworkName`/`privateEndpointSubnetName` from `lib/naming.bicep` instead of duplicating the literal formulas, same as main.bicep/network.bicep already do (2757, commit 9d48bd479). That leaves reusable-deploy.yml/reusable-build-info.yml as the only remaining duplication — those derive the name in bash for their vnet-existence checks, so cant import the bicep function; left a comment in app-shared-setup.bicep pointing at them so future changes stay in lockstep manually.

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
bityogi: ✅ Fixed in b1de6bfab. The retry wrapper now captures output, checks for the `AnotherOperationInProgress`/`409` signature specifically, and fails immediately (no retry) on anything else — verified against 3 mocked scenarios (immediate success, genuine validation error, real 409 conflict) before landing. Output is still echoed in full either way so nothing is lost from CI logs, just no longer streamed live (traded off for the ability to inspect it before deciding whether to retry).

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
bityogi: ✅ Fixed in b1de6bfab — both `az deployment group create` calls now pass `--name "${stack_name}-shared-setup"`, so concurrent branches/main deploying to the shared RG no longer race on the CLI's default (the template basename).

.pre-commit-config.yaml
bityogi: ✅ Fixed in b1de6bfab. Widened `files:` to also match backend-api-deploy.bicep/dataflows-resource-deploy.bicep/frontend-webapp-deploy.bicep/lib/email/acs-email.bicep, and rewrote the hook itself to scan all of them (a loop over the target list, not just main.bicep). Also widened the check beyond the two known kvSetup strings to a general rule: any `secretName:` with a quoted value that contains NO `${...}` interpolation is flagged as a fixed/shared name, which is exactly the acsEmail bug shape above — verified it catches a reintroduction of that exact bug (reverted the fix locally, confirmed the hook fails, restored it) before landing.

.github/workflows/azure-remove-branch.yml
bityogi: ✅ Fixed in b1de6bfab. Discovery (both the scheduled sweep and the single-hash check) now unions app-stack and network-stack tags via jq (`($net + $app) | ... | unique`) rather than network-stack tags alone, so a branch missing either tier's stack still gets discovered. Also updated the purely-informational `list` job to display both tiers' tagged stacks.

ops/scripts/utility/az-delete-branch-resources.sh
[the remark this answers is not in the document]
bityogi: ✅ Already fixed — this landed via the merge from 2757 (commit 0f7673b67): the final verification block now checks the network stack too (in the `else`/deleteResources branch), using the same `set +e; ...; rc=$?; set -e` pattern as the app-stack check so a probe failure is treated as "not verified clean" rather than skipped. See the current file around the final verification section.

bityogi:
Addressed the BLOCKER that couldn't be anchored inline (the affected files aren't part of this PR's diff): 7 other workflow files still independently derived `AZ_APP_RG`/`AZ_NETWORK_RG` with the legacy `-<hash>` suffix this PR eliminates.

Fixed all 7 in b1de6bfab, matching `reusable-deploy.yml`/`reusable-build-info.yml`'s already-correct derivation (always the raw KV-sourced base value, no hash suffix):
- `reusable-infrastructure-deploy.yml` (`AZ_NETWORK_RG`)
- `sub-deploy-code.yml` (4 sites)
- `sub-deploy-code-slot.yml` (7 sites)
- `reusable-build-frontend.yml`
- `reusable-e2e.yml`
- `reusable-endpoint-test.yml`
- `reusable-dast.yml`

Also removed the now-dead `ENVIRONMENT_HASH`/`BRANCH_HASH_ID` env-var declarations in the steps where they'd become unused. Verified all touched workflow YAML with `pre-commit run check-yaml` and confirmed no other references to the dropped `AZ_APP_RG_BASE`/`AZ_NETWORK_RG_BASE` intermediate vars remain (`grep -c` returns 0 across all 7 files).

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
```diff
@@ -0,0 +1,171 @@
...
+    while true; do
+        set +e
+        output=$("$@" 2>&1)
+        rc=$?
+        set -e
+        echo "${output}"
+        if [[ ${rc} -eq 0 ]]; then
+            return 0
+        fi
+        if [[ ${attempt} -ge ${maxAttempts} ]] || ! grep -qi "AnotherOperationInProgress\|409" <<< "${output}"; then
```
bityogi: **Self-review finding (SRE agent) — HIGH.** The `--name "${stack_name}-shared-setup"` pinning (from the earlier fix round) shifted the collision class from "any two branches racing on the default deployment name" to "the same branch redeploying while its own prior deployment is still active." That specific ARM error is `(DeploymentActive) Unable to edit or replace deployment 'X': previous deployment from '...' is still active...` — no literal `409` in the printed text, and it doesn't match `AnotherOperationInProgress` either. So the dominant collision this wrapper now exists to survive fails the grep and aborts after attempt 1 instead of retrying — arguably worse than the pre-fix "retry on anything" behavior for this exact scenario. Widening the pattern to also match `DeploymentActive`.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -393,19 +489,45 @@ fi
...
+            kvPrivateDnsZoneName='privatelink.vaultcore.usgovcloudapi.net'
+            pepName="pep-${stack_name}"
+            pepId=$(az resource list -g "${network_rg}" --resource-type Microsoft.Network/privateEndpoints --query "[?name=='${pepName}'].id" -o tsv)
+            if [[ -n "${pepId}" ]]; then
+                echo "Deleting KV private endpoint ${pepName} in ${network_rg}"
+                az resource delete --ids "${pepId}"
+            else
+                echo "No KV private endpoint ${pepName} found in ${network_rg}; nothing to delete"
+            fi
+            vnetLinkId=$(az network private-dns link vnet list --resource-group "${kv_rg}" --zone-name "${kvPrivateDnsZoneName}" --query "[?name=='${kvPrivateDnsZoneName}-vnet-link-${stack_name}'].id" -o tsv)
```
bityogi, line 540: **Self-review finding (DevSecOps agent) — MEDIUM/HIGH, flagging not fixing.** This (and the PE delete above it) is the first place this script *mutates* `kv_rg` (previously read-only, used only in the shared-RG string-comparison guard) into an actual delete target. The teardown identity now needs `Microsoft.Network/privateEndpoints/delete` and `Microsoft.Network/privateDnsZones/virtualNetworkLinks/delete` (plus list) permission scoped to the shared KV resource group. Nothing in this repo confirms that identity has been granted this. If it hasn't, every `deleteResources`-mode teardown now hard-fails here (not silently — `set -euo pipefail` trips `failed=true`, so it's at least loud). If it *has* broad Contributor on this RG to make this work, that re-widens the exact blast radius GH #2749 was about. This needs a live Azure RBAC check I can't perform from a coding session — @diabagatekelly can you (or whoever owns the OIDC federated credential setup) confirm the scoped role assignment exists, ideally as narrowly as a custom role limited to these two resource types?

.pre-commit-hooks/guard-app-deploy-not-stacked.sh
```diff
@@ -1,38 +1,66 @@
...
 
+# Scans every file in main.bicep's own stacked module tree, not just
+# main.bicep itself — the bug this guard exists to catch (a fixed-name,
+# cross-scope shared resource inside a per-branch stack) has already
+# reappeared once in a module main.bicep calls (acs-email.bicep's KV
+# secrets), not in main.bicep's own text.
 REPO_ROOT="$(git rev-parse --show-toplevel)"
-TARGET="$REPO_ROOT/ops/scripts/pipeline/azure-deploy.sh"
+TARGETS=(
+  "$REPO_ROOT/ops/cloud-deployment/main.bicep"
```
bityogi, line 24: **Self-review finding (Architecture agent) — MEDIUM.** This list is already stale relative to `main.bicep`'s actual module graph: `main.bicep` also calls `lib/monitoring-alerts/alert-action-group.bicep` cross-scope (`scope: resourceGroup(analyticsResourceGroupName)`) — structurally the same bug shape this guard exists to catch. It's currently dormant because `createAlerts` defaults false for branches, but `main.bicep`'s own adjacent comment explicitly flags "if you ever want createAlerts true for a branch... do not just flip the flag" as an anticipated future change, with zero mention that this guard's `TARGETS` would also need updating — it would silently provide no coverage until someone remembers. Adding it to `TARGETS` now, before that flag ever flips.

ops/cloud-deployment/lib/email/acs-email.bicep
```diff
@@ -39,12 +39,26 @@ resource communicationService 'Microsoft.Communication/communicationServices@202
...
+// across branches even though the ACS resource it points at is not.
+// backend-api-deploy.bicep's app settings must reference the exact same
+// stackName-qualified name — keep both in lockstep.
 module acsConnectionStringSecret '../keyvault/keyvault-secret.bicep' = {
   name: '${stackName}-acs-connection-string-secret'
   scope: resourceGroup(kvAppConfigResourceGroupName)
   params: {
     keyVaultName: kvAppConfigName
-    secretName: 'ACS-EMAIL-CONNECTION-STRING' // pragma: allowlist secret
+    secretName: 'ACS-EMAIL-CONNECTION-STRING-${stackName}' // pragma: allowlist secret
```
bityogi: **Self-review finding (Architecture agent) — MEDIUM.** This (and the sender-address secret below) hardcodes the `-${stackName}` suffix formula, and `backend-api-deploy.bicep`'s `@Microsoft.KeyVault(...)` app settings independently re-derive the identical string literal to build the reference. The only thing preventing drift is a comment ("keep both in lockstep"). This is the exact anti-pattern this same commit fixes for vnet/subnet naming by centralizing formulas into `lib/naming.bicep` — which `app-shared-setup.bicep` already imports in this same diff. Adding `acsConnectionStringSecretName`/`acsSenderAddressSecretName` to `naming.bicep` and importing in both files.

.github/workflows/reusable-build-frontend.yml
```diff
@@ -77,12 +77,11 @@ jobs:
...
-          if [[ "${GHA_ENVIRONMENT}" == "Main-Gov" ]]; then
-            echo "AZ_APP_RG=${AZ_APP_RG_BASE}" >> "$GITHUB_ENV"
-          else
-            echo "AZ_APP_RG=${AZ_APP_RG_BASE}-${ENVIRONMENT_HASH}" >> "$GITHUB_ENV"
-          fi
+          # CAMS-760, Option E / Slice 2: branches deploy into the SAME stable
+          # app RG main uses, distinguished by per-branch-unique resource
+          # names, rather than a per-branch hash-suffixed RG (matches
+          # reusable-deploy.yml's derivation — keep both in lockstep).
+          echo "AZ_APP_RG=${_val}" >> "$GITHUB_ENV"
```
bityogi, line 80: **Self-review finding (Pragmatic agent) — MEDIUM.** This fix collapsed the only place in this file that read `inputs.environmentHash` (removed the `ENVIRONMENT_HASH` env var and the branch that used it a few lines up). But the `environmentHash` input itself is still declared in this workflow's `on: workflow_call: inputs:` block, and both callers (`sub-build.yml`, `sub-deploy.yml`) still pass it in — its now write-only, accepted and threaded through but never consumed. Removing the input and both callers pass-through.

ops/cloud-deployment/lib/email/acs-email.bicep
```diff
@@ -39,12 +39,26 @@ resource communicationService 'Microsoft.Communication/communicationServices@202
   }
 }
 
+// BLOCKER fix (CAMS-760, GH #2749 bug shape): this module is called
```
bityogi: **Self-review finding (Pragmatic agent) — LOW.** "BLOCKER fix" is a review-severity label, not a fact about the code — six months from now nobody reviewing this file cares that it was once tagged BLOCKER in a review. The rest of this comment is legitimately earning its keep (explains a non-obvious "why here, not app-shared-setup.bicep" decision); just dropping the "BLOCKER fix" framing and leading with the ticket/incident reference, matching the convention used everywhere else in this PR chain.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -535,16 +657,33 @@ fi
...
-    echo "ERROR: App resource group ${app_rg} still exists after deletion attempt." >&2
-    failed=true
+# Verify nothing was left behind. Per-branch RGs (unmanage_action != deleteResources)
+# are deleted outright and verified gone; for a preserved shared RG (deleteResources,
+# Slice 2) verify each tier's stack is gone instead. Reuses the same `failed` flag
+# the teardown loop above set, so a teardown failure isn't masked even if the
+# resource happens to look gone here.
+if [[ "${unmanage_action}" != "deleteResources" ]]; then
+    if [[ $(az group exists -n "${app_rg}") == "true" ]]; then
+        echo "ERROR: App resource group ${app_rg} still exists after deletion attempt." >&2
```
bityogi: **Self-review finding (Pragmatic agent) — LOW.** This block (app-stack verify) and the network-stack verify block below it are near-identical, including a verbatim 6-line comment repeated twice. This is exactly the kind of duplication `stack_exists()` above already demonstrates the value of factoring out — extracting a `verify_stack_gone()` wrapper would cut ~20 lines and remove one copy of the comment.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -393,19 +489,45 @@ fi
...
+            # Shared network RG (Slice 2): the network stack never manages the
+            # KV private endpoint or its DNS zone vnet link — app-shared-setup.bicep
+            # creates both as a plain (non-stack) deployment, per-branch-named
+            # (pep-${stack_name} / <zone>-vnet-link-${stack_name}), in the shared
+            # network RG and the KV's own RG respectively. Unlike the per-branch
+            # path above (whole-RG delete sweeps these for free), the shared RG
+            # survives, so they must be deleted explicitly here, BEFORE the stack
+            # delete: the PE still occupying the private-endpoint subnet is exactly
+            # what makes `az stack group delete` fail with InUseSubnetCannotBeDeleted.
+            kvPrivateDnsZoneName='privatelink.vaultcore.usgovcloudapi.net'
```
bityogi, line 531: **Self-review finding (DevSecOps agent) — LOW.** `privatelink.vaultcore.usgovcloudapi.net` is hardcoded both here and in `ustp-cams-kv-app-config-setup.bicep` (`keyvaultPrivateDnsZoneName`). If one changes, the vnet-link delete silently stops matching and falls to the "nothing to delete" branch, leaking the link. Cant easily share a literal across bash/bicep without a codegen step — adding a cross-reference comment so a future editor of either side knows to check the other.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -15,6 +15,35 @@
 # 1   Unrecognized parameter provided
 # 2   Required parameter not provided
 # 10+ Validation check errors
+#
+# Recovery (CAMS-760, Slice 2, shared RGs)
```
bityogi, line 19: **Self-review finding (SRE agent) — MEDIUM.** This top-of-file recovery guidance doesnt mention the new PE/DNS-vnet-link partial-failure state this commit introduces (PE deleted but link remains, or neither touched, depending on which of the two `az resource delete` calls failed). Neither reachable state is worse than pre-fix (previously neither was ever deleted), and a re-run correctly resumes since both are existence-checked — but an on-call engineer diagnosing a failed run has to infer that today. Adding a note.

ops/scripts/pipeline/azure-deploy-app-shared-setup.sh
bityogi: ✅ Fixed in 45f91e2ac. Widened the pattern to also match `DeploymentActive`, and verified against a mocked message matching the real ARM error text.

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: Still open — this needs a live Azure RBAC check I cant perform from here. @diabagatekelly (or whoever owns the OIDC federated credential setup for the teardown identity) can you confirm the scoped role assignment exists before this merges?

.pre-commit-hooks/guard-app-deploy-not-stacked.sh
bityogi: ✅ Fixed in c88c29fe6. Added `alert-action-group.bicep` to `TARGETS`, plus a comment documenting that the list is manually maintained with no automatic derivation from main.bicep's module graph.

ops/cloud-deployment/lib/email/acs-email.bicep
bityogi: ✅ Fixed in 99c9f9c0a. Added `acsConnectionStringSecretName`/`acsSenderAddressSecretName` to `naming.bicep` and imported them here and in `backend-api-deploy.bicep`, matching the same centralization pattern already used for vnet naming.

.github/workflows/reusable-build-frontend.yml
bityogi: ✅ Fixed in 9612bbc4d. Traced the pass-through up the call chain and removed it at all three levels (this file, sub-build.yml's own now-dead input, and continuous-deployment.yml's `build` job) rather than just here — leaving the parent levels would've just moved the dead code up one hop. `deploy`/sub-deploy.yml's own environmentHash usage (concurrency group, infra/network deploys) is unrelated and untouched.

ops/cloud-deployment/lib/email/acs-email.bicep
bityogi: ✅ Fixed in 99c9f9c0a — dropped "BLOCKER fix" from the comment, kept the rest of the rationale (it was earning its keep).

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: ✅ Fixed in 3520f7e62. Extracted `verify_stack_gone()`, parameterized on label/RG/stack name and whether a "skipped this run" message applies — cut ~20 lines and removed the duplicated comment. Verified against both the clean and skipped-with-still-there scenarios.

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: ✅ Fixed in 05c8e3aaa. Added a cross-reference comment on both sides (here, and `keyvaultPrivateDnsZoneName` in `ustp-cams-kv-app-config-setup.bicep`) — cant share the literal across bash/bicep without a codegen step, so this is the cheapest way to keep an editor of either one from missing the other.

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: ✅ Fixed in 58593e2df. Added a bullet documenting the new PE/DNS-vnet-link partial-failure state and that a re-run resumes cleanly either way.

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: Checked this directly via `az role assignment list` / `az role definition list` against the real Flexion Gov subscription (I have read access as pgovinda) rather than waiting on a manual confirmation.

**Answer: it won't hard-fail — but not because a scoped grant exists.**

The teardown identity (`cams-remove-branch-oidc`, app id `30e5afbf-8ac9-41d1-b7e9-dccb56dd6bf5`, backing the `remove-branch` GH environment / "Clean up Flexion Azure Resources" workflow) has held **subscription-wide `Contributor`** since 2026-06-03 (granted by James Brooks, well before this PR). Contributor's `notActions` only exclude `Microsoft.Authorization/*` writes/deletes and a few unrelated actions (confirmed via `az role definition list`) — `Microsoft.Network/privateEndpoints/delete` and `Microsoft.Network/privateDnsZones/virtualNetworkLinks/delete` are both covered by the `*` actions grant.

So this lands on the second branch you called out: **it does have broad Contributor**, and it's actually wider than "on this RG" — it's subscription-wide. This PR doesn't introduce that grant (it predates CAMS-760 entirely, likely because pre-Slice-2 per-branch RGs had unpredictable hash-suffixed names that made a narrower scope hard to express), but it is the first place a script exercises delete permissions that happen to be covered only by that legacy broad grant rather than anything scoped.

For contrast, this identity's other role assignments are already properly scoped — Key Vault Secrets User on 5 individual secrets in `kv-ustp-cams-dev` (granted 2026-07-24, post-dates the Contributor grant). It's specifically this network-delete path that's riding on the old broad grant.

Filed cams-q0dq4 to replace the subscription-wide Contributor with a role scoped to just the RGs/resource types teardown actually needs, now that CAMS-760 gives it fixed shared-RG names to scope against (which wasn't true under the old per-branch-RG naming).

Given the script is confirmed to work today and the broad grant is pre-existing rather than newly introduced by this PR, I don't think this needs to block merge on its own — but leaving this thread open for your call, since you're the one who flagged it and tagged yourself for the confirmation.

.github/workflows/slack-notification.yml
```diff
@@ -2,14 +2,25 @@ on:
...
+    # actual infra deploy, and runs on every branch push — alerting on every
+    # branch's failure there would mostly be normal PR-review noise (a failing
+    # test, not an infra problem), so it stays scoped to main. Clean up
+    # Flexion Azure Resources (CAMS-760) only ever does Azure teardown work —
+    # any failure in it is an infra problem worth alerting on regardless of
+    # which branch's resources it was cleaning up.
+    if: >
+      ${{ github.event.workflow_run.conclusion == 'failure' &&
+      (github.event.workflow_run.name == 'Clean up Flexion Azure Resources' ||
+      github.event.workflow_run.head_branch == 'main') }}
```
diabagatekelly, line 23: **HIGH — this `if:` change makes a pre-existing shell-injection sink reachable with attacker-controlled input.**

Further down this same job (the `printf` building `full_message`), `${{ github.event.workflow_run.head_branch }}` is interpolated directly into the `run:` script body rather than via `env:` — a classic GitHub Actions template-injection sink (git ref names permit backtick/`$()`/`;`/`|`). That line is unchanged by this PR, and it was previously safe *only* because this `if:` required `head_branch == 'main'` — a fixed, trusted literal.

This diff adds `workflow_run.name == 'Clean up Flexion Azure Resources'` as an alternative, with no branch restriction on that path. Since the Cleanup workflow runs for any branch, a branch name containing shell metacharacters that triggers/fails cleanup now reaches that `head_branch` interpolation with a fully attacker-chosen string — in a job holding `secrets.SLACK_WEBHOOK_URL` and `secrets.SLACK_USER_MAPPING` (contributor emails/Slack IDs).

Suggested fix: move `head_branch`/`head_sha` into the step's `env:` block (they already do this for `workflow_name` in this same diff) and reference as `"$head_branch"` inside the script instead of `${{ }}` interpolation in the run body.

.pre-commit-config.yaml
```diff
@@ -57,10 +57,10 @@ repos:
         language: python
         files: '^ops/cloud-deployment/lib/cosmos/mongo/cosmos-collections\.bicep$'
       - id: guard-app-deploy-not-stacked
-        name: Guard app deploy against re-stacking (GH #2749)
+        name: Guard against shared KV re-entering the app stack (GH #2749)
         entry: .pre-commit-hooks/guard-app-deploy-not-stacked.sh
         language: script
-        files: '^ops/scripts/pipeline/azure-deploy\.sh$'
+        files: '^ops/cloud-deployment/(main|backend-api-deploy|dataflows-resource-deploy|frontend-webapp-deploy|lib/email/acs-email)\.bicep$'
```
diabagatekelly: **HIGH — this file filter doesn't cover a file the guard script itself already flags as a live risk.**

`guard-app-deploy-not-stacked.sh`'s own `TARGETS` array also scans `lib/monitoring-alerts/alert-action-group.bicep`, and the script's own comment calls this out directly: *"the actionGroup module above is already one example, currently dormant only because createAlerts defaults false for branches."* In other words, the guard's author already identified this file as equivalent to the GH #2749 bug shape this hook exists to catch — but this `files:` regex is what actually decides whether the hook runs at all (independent of the script's internal `TARGETS` list), and it doesn't include this path. A commit touching only `alert-action-group.bicep` never triggers this guard.

Suggested fix: add `|lib/monitoring-alerts/alert-action-group` to the regex.

ops/cloud-deployment/app-shared-setup.bicep
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// The read-only SQL managed identity used by both the API and dataflows
+// function apps to authenticate to the SQL server. Its name comes from a
+// fixed value (the AZ_SQL_IDENTITY_NAME secret) shared by main and every
+// branch, so it is created exactly once here rather than per-function-app.
+// Creating it inside a branch's app stack would let that branch's teardown
+// delete the identity every other branch and main depend on — the same bug
+// shape as the shared Key Vault incident above.
+var sqlIdentityName = !empty(sqlServerIdentityName) ? sqlServerIdentityName : 'id-sql-${stackName}-readonly'
```
diabagatekelly: **MEDIUM — fallback name formula here disagrees with the two `existing`-reference consumers.**

This is now the identity's actual creator, falling back to `id-sql-${stackName}-readonly` when `sqlServerIdentityName` is empty. `backend-api-deploy.bicep` and `dataflows-resource-deploy.bicep` (both `existing`-only lookups) still fall back to `id-sql-${apiFunctionName}-readonly` — a different string (`apiFunctionName = ${stackName}-node-api`). Before this PR each consumer created its own identity with its own name, so the mismatch was harmless (self-consistent per file). Now creation is centralized here but the fallback wasn't updated to match in the two consumers.

Currently masked because `reusable-deploy.yml` always passes `--sqlServerIdentityName` explicitly to all three deploy paths — but any path that omits it (local repro, a future workflow, a missing secret) makes the `existing` lookup in the consumers target a resource that was never created here, failing with `ResourceNotFound`.

This PR already solved this exact problem correctly elsewhere (`acsConnectionStringSecretNameFor` in `naming.bicep`) — suggest the same treatment: add a `sqlIdentityNameFor(stackName)` function to `naming.bicep` and import it in all three files instead of hand-rolling divergent literals.

ops/cloud-deployment/backend-api-deploy.bicep
```diff
@@ -471,21 +482,15 @@ module setApiFunctionSqlServerVnetRule './lib/network/sql-vnet-rule.bicep' = if
   }
 }
 
+// The identity itself is created once, in app-shared-setup.bicep (CAMS-760,
+// Option E / Slice 2) — its name is a fixed value shared by main and every
+// branch, so it must never be created/managed inside a branch's app stack.
+// Referenced here as `existing` only.
 var sqlIdentityName = !empty(sqlServerIdentityName) ? sqlServerIdentityName : 'id-sql-${apiFunctionName}-readonly'
```
diabagatekelly: **MEDIUM — fallback name here (`id-sql-${apiFunctionName}-readonly`) doesn't match the identity's actual creator.**

`app-shared-setup.bicep` (where this identity is now actually created) falls back to `id-sql-${stackName}-readonly` when `sqlServerIdentityName` is empty — a different string. See the comment on `app-shared-setup.bicep`'s `sqlIdentityName` var for the full explanation and suggested fix (a shared `sqlIdentityNameFor(stackName)` helper in `naming.bicep`). Same issue is duplicated in `dataflows-resource-deploy.bicep`.

ops/cloud-deployment/dataflows-resource-deploy.bicep
```diff
@@ -736,22 +736,15 @@ module setDataflowFunctionSqlServerVnetRule './lib/network/sql-vnet-rule.bicep'
   }
 }
 
-// Creates a managed identity that would be used to grant access to function instance
+// The identity itself is created once, in app-shared-setup.bicep (CAMS-760,
+// Option E / Slice 2) — its name is a fixed value shared by main and every
+// branch, so it must never be created/managed inside a branch's app stack.
+// Referenced here as `existing` only.
 var sqlIdentityName = !empty(sqlServerIdentityName) ? sqlServerIdentityName : 'id-sql-${apiFunctionName}-readonly'
```
diabagatekelly: **MEDIUM — same fallback-name mismatch as `backend-api-deploy.bicep`.**

This falls back to `id-sql-${apiFunctionName}-readonly`, while `app-shared-setup.bicep` (the actual creator) falls back to `id-sql-${stackName}-readonly`. See the comment on `app-shared-setup.bicep`'s `sqlIdentityName` var for the full explanation and suggested fix.

ops/scripts/utility/az-delete-branch-resources.sh
```diff
@@ -15,6 +15,43 @@
...
+#   still fails: check whether the DNS zone vnet link (pep-<stack>'s sibling,
+#   in kv_rg) also got deleted — the two are existence-checked independently,
+#   so a transient failure on either one leaves the other in place (PE gone
+#   but link remains, or neither touched, never the reverse order). Neither
+#   partial state is worse than before this delete existed, and a re-run
+#   resumes cleanly either way, but it's worth knowing which one is still
+#   there before assuming the whole PE/DNS-link step is broken.
+# - To manually confirm a shared RG (app_rg/network_rg) is not left with
+#   orphaned per-branch resources after a hash's teardown, filter by tag:
+#   `az resource list -g <rg> --query "[?tags.branchHashId=='<hash>']"`.
```
diabagatekelly: **MEDIUM — this verification command will always return empty, giving a false "all clean" signal.**

Individual resources (webapp, api/dataflows function apps, VNet/subnets, ACS comms/email resources) never carry a `branchHashId` tag — only the deployment *stack* object gets tagged with `branchHashId` (via `az stack group create --tag ...` in `azure-deploy.sh`/`azure-deploy-network.sh`). So `az resource list -g <rg> --query "[?tags.branchHashId=='<hash>']"` will return empty regardless of whether teardown actually succeeded — exactly the wrong failure mode for a runbook step whose purpose is confirming shared-RG blast-radius containment during an incident.

Suggest either tagging individual resources with `branchHashId`, or rewriting this runbook command to query stacks (`az stack group list`) instead.

ops/cloud-deployment/ustp-cams-kv-app-config-setup.bicep
```diff
@@ -28,10 +28,17 @@ param deployedAt string = utcNow()
...
+//   USTP/ADO            no                      no
+// Both actions are excluded by Contributor's NotActions, but not identically
+// across identities — don't assume one implies the other for a future
+// Microsoft.Authorization/*-gated resource.
 @description('When false, no role assignments are created (used for USTP deployments where the ADO service principal lacks role assignment permissions).')
 param makeRoleAssignment bool = true
 
-@description('When true, deploys CanNotDelete resource locks on the shared Key Vault and its managed identity (GH #2749 defense-in-depth). Defaults to false because no deploy identity currently in use — Flexion or USTP — has been granted Microsoft.Authorization/locks/write; enable only once that permission has actually been granted to the deploying identity.')
+@description('When true, deploys CanNotDelete resource locks on the shared Key Vault and its managed identity. Defaults to false — see the permissions matrix above.')
 param enableResourceLocks bool = false
```
diabagatekelly, line 52: **MEDIUM — this safeguard is currently dormant for every environment, which the surrounding comments don't quite convey.**

Several comments in this PR (and the permissions matrix directly above this param) frame `CanNotDelete` resource locks as a defense-in-depth layer against a repeat of the GH #2749 incident. But `enableResourceLocks` defaults to `false` here, and per the matrix above, neither the Flexion nor USTP deploy identity currently has `Microsoft.Authorization/locks/write` — so this layer isn't actually active anywhere today. Not a regression (the default predates this PR), but worth flagging so reviewers don't read the lock code as an independently-enforced control on top of the structural fix (plain non-stack deployment) and the script/pre-commit guards — right now those three are the *only* active protection.

ops/cloud-deployment/main.bicep
```diff
@@ -22,13 +22,14 @@ param appResourceGroup string = resourceGroup().name
 // network.bicep imports too — so the `existing` lookups below
 // (ustpVirtualNetwork, *SubnetExisting), for resources network.bicep creates,
 // can no longer silently drift out of sync.
+// app-shared-setup.bicep still duplicates this formula as a literal string
```
diabagatekelly: **LOW — this line is no longer accurate.**

`app-shared-setup.bicep` already imports and uses `virtualNetworkNameFor(stackName)` from `naming.bicep` (see its own header comment: "This default is computed via the shared functions in lib/naming.bicep"). Only the bash duplication in `reusable-deploy.yml`/`reusable-build-info.yml` remains as the actual lockstep risk. Suggest updating this comment so a future maintainer isn't misdirected toward `app-shared-setup.bicep` as the remaining drift risk when it's already been migrated.

.github/workflows/slack-notification.yml
bityogi: ✅ Fixed in a3cbc5bb8. Moved `head_branch`/`head_sha`/`name` into the step's `env:` block and reference them as shell variables (`$HEAD_BRANCH` etc.) instead of `${{ }}` interpolation in the run body. `head_branch` was the actual attacker-controlled sink; moved the other two too for consistency.

.pre-commit-config.yaml
bityogi: ✅ Fixed in 8c1685b14. Added `|lib/monitoring-alerts/alert-action-group` to the `files:` regex, matching the guard script's own `TARGETS` array.

ops/cloud-deployment/app-shared-setup.bicep
bityogi: ✅ Fixed in 111c5cc19. Added a `sqlIdentityName(stackName)` function to `naming.bicep` (same treatment as `acsConnectionStringSecretName`) and imported it here and in `backend-api-deploy.bicep`/`dataflows-resource-deploy.bicep`, so all three now agree on the fallback formula.

ops/cloud-deployment/backend-api-deploy.bicep
bityogi: ✅ Fixed in 111c5cc19 — same commit as the `app-shared-setup.bicep` comment. Now imports and uses `sqlIdentityName` from `naming.bicep` instead of the divergent `apiFunctionName`-based literal.

ops/cloud-deployment/dataflows-resource-deploy.bicep
bityogi: ✅ Fixed in 111c5cc19 — same fix as `backend-api-deploy.bicep`, now imports `sqlIdentityName` from `naming.bicep`.

ops/scripts/utility/az-delete-branch-resources.sh
bityogi: ✅ Fixed in de45e8184. Rewrote the runbook command to query `az stack group list` instead of `az resource list`, since `branchHashId` is only ever set on the deployment stack object, never on individual managed resources.

ops/cloud-deployment/ustp-cams-kv-app-config-setup.bicep
bityogi: ✅ Fixed in 0be06c0da. Added a comment above `enableResourceLocks` making explicit that with the current default and permissions matrix, this layer is dormant in every environment today — not an independently-enforced control alongside the structural fix and guards.

ops/cloud-deployment/main.bicep
bityogi: ✅ Fixed in e520db80a. Updated the comment to reflect that app-shared-setup.bicep already imports virtualNetworkNameFor from naming.bicep — only reusable-deploy.yml/reusable-build-info.yml's bash duplication remains as the actual drift risk.

bityogi:
Addressed all of Kelly's Aug 6 review feedback:

**General notes (not anchored to a diff line):**
- PR description's Notes section wrongly said the `acsEmail` fix was "out of scope, deferred separately" — corrected it to reflect that the branch-qualified-secret-name fix is actually part of this diff.
- `ustp-cams-kv-app-config-setup.bicep`'s header comment was stale (said `main.bicep` invokes it directly) — fixed in 0d46761ba to reflect that `app-shared-setup.bicep` invokes it before `main.bicep`.

**Inline findings — replied + resolved individually, one commit per fix:**
- a3cbc5bb8 — branch-name shell-injection sink in Slack notify (HIGH)
- 8c1685b14 — pre-commit guard regex gap for `alert-action-group.bicep` (HIGH)
- 111c5cc19 — centralized SQL identity fallback name formula across 3 files (MEDIUM)
- de45e8184 — fixed runbook `branchHashId` tag query to target stacks, not resources (MEDIUM)
- 0be06c0da — clarified KV resource locks are dormant in every environment today (MEDIUM)
- e520db80a — fixed stale vnet-naming drift comment in `main.bicep` (LOW)

All changed bicep files rebuild clean (`az bicep build`), the changed workflow YAML validates, and shellcheck passes on the modified script.

ops/cloud-deployment/dataflows-resource-deploy.bicep
bityogi: Follow-up: the fix in 111c5cc19 above (adding `sqlIdentityNameFor(stackName)`) introduced a regression here that a targeted architecture review caught — `main.bicep`'s `ustpDataflowsFunction` module call never passed `stackName`, so it was silently falling back to `dataflows-resource-deploy.bicep`'s `'ustp-cams'` param default instead of the real branch stack name. Once the SQL identity fallback started depending on `stackName`, that meant branches omitting `sqlServerIdentityName` would resolve to `id-sql-ustp-cams-readonly` instead of their actual identity — the same `ResourceNotFound` failure mode, just relocated. Fixed in 6642cafec by passing `stackName: stackName` explicitly, matching the sibling `ustpApiFunction` module call.

PR #2779, opened 2026-07-29
CAMS-856 Fix silent trustee-notification delivery failures

fmaddenflx:
# Bug

Trustee change-notification emails were silently not delivered to some recipient groups because their NOTIFICATION_ROUTING records stored a wrong-but-syntactically-valid domain (@UST.DOJ.GOV instead of @usdoj.gov). ACS accepted the send since the address was well-formed, so nothing in CAMS ever signaled a delivery problem. Separately, a nested-loop bug meant a single bad recipient's send failure aborted every remaining recipient/address in that same notification run.

# Purpose

Close the observability and error-isolation gaps that let a bad recipient address fail silently: one recipient's send failure should never block delivery to the rest, obviously-wrong domains should be caught before they're saved, and delivery-status data should actually be flowing somewhere it can be inspected.

# Major Changes

* Isolated per-recipient/per-address send failures in `TrusteeChangeNotificationUseCase.notify()` so one bad address no longer aborts the rest of the batch; each failure is now logged individually
* Added DNS domain validation (MX with A/AAAA fallback) to the admin notification-routing PUT endpoint — a definitively nonexistent domain is rejected at save time, and an indeterminate DNS failure is saved but surfaced as a UI warning
* Enabled ACS Communication Services diagnostic logging (`EmailSendMailOperational`, `EmailStatusUpdateOperational`) forwarded to Log Analytics, giving per-recipient send/delivery visibility instead of only an aggregate metric count
* Renamed `notify()`'s internal recipient/mailing-list terminology (`recipient` → `mailingList`) to match what the resolved objects actually represent, and extracted the inner per-address send loop into its own `sendToMailingList` method for clarity
* Removed a dangling `MODULE_NAME` export left over from an earlier refactor

# Testing/Validation

Unit tests added for: per-recipient send-failure isolation in `trustee-change-notification.test.ts`; DNS domain validation branches (not-found, indeterminate, valid) in `notification-routing.controller.test.ts`. Full backend/common/user-interface test suites pass (2 pre-existing, unrelated flaky timeouts in `migrate-trustees*.test.ts` under full-suite load, confirmed to pass standalone and unaffected by this branch). Typecheck, lint, knip, and prettier all clean on changed files.

# Notes

This PR covers the send-time isolation and prevention pieces of CAMS-856. Two follow-up items remain, tracked separately: auditing/correcting the actual wrong-domain routing records across all environments (chapter 12/13 were fixed live in one environment during triage; chapter 7, chapter 11, subchapter-v, and zoom-341 are still unverified), and building admin/infra alerting on top of the new diagnostic logs for real bounces (an async ACS delivery-status signal that this PR's synchronous fixes cannot catch on their own).

# Definition of Done:

## Summary by Sourcery

Improve trustee notification reliability and observability by validating notification-routing email domains, isolating per-address send failures, and surfacing diagnostic signals end-to-end.

Bug Fixes:
- Prevent a single trustee notification send failure from aborting delivery to other addresses and mailing lists.
- Reject notification-routing updates that use domains that do not appear to accept email, avoiding silently undeliverable trustee notifications.

Enhancements:
- Add DNS-based domain validation (MX with A-record fallback and indeterminate handling) to the admin notification-routing update endpoint and flow warnings back to the UI.
- Propagate backend response warnings through the admin notification-routing UI so operators see when domain verification could not be completed.
- Refine trustee change notification mailing-list terminology and logging to clearly identify which routed mailing lists and chapters are affected by send failures.
- Remove an unused trustees module name export after refactoring notification error handling.

Build:
- Wire ACS Communication Services email diagnostic settings into the deployment templates, optionally sending send/delivery logs to a Log Analytics workspace.
- Extend the shared API response shape to support optional warnings alongside data and metadata.

Tests:
- Add unit tests for DNS domain validation behavior in the notification-routing controller, including not-found, fallback, indeterminate, and unique-domain lookup cases.
- Add unit tests ensuring trustee notifications continue sending after a failed address and emit appropriate error logs.
- Adjust trustees use-case tests to assert the new trustee change notification error logging behavior.

### DIFF
--- a/backend/lib/controllers/admin/notification-routing.controller.test.ts
+++ b/backend/lib/controllers/admin/notification-routing.controller.test.ts
@@ -1,4 +1,5 @@
 import { vi, Mocked } from 'vitest';
+import * as dns from 'node:dns/promises';
 import { ApplicationContext } from '../../adapters/types/basic';
 import { createMockApplicationContext } from '../../testing/testing-utilities';
 import { NotificationRoutingController } from './notification-routing.controller';
@@ -9,7 +10,20 @@ import { NotificationRoutingRecord } from '@common/cams/notifications';
 import HttpStatusCodes from '@common/api/http-status-codes';
 
 vi.mock('../../factory');
+vi.mock('node:dns/promises');
 const mockFactory = factory as Mocked<typeof factory>;
+
+function notFoundError(): NodeJS.ErrnoException {
+  const error = new Error('queryA ENOTFOUND') as NodeJS.ErrnoException;
+  error.code = 'ENOTFOUND';
+  return error;
+}
+
+function timeoutError(): NodeJS.ErrnoException {
+  const error = new Error('queryMx ETIMEOUT') as NodeJS.ErrnoException;
+  error.code = 'ETIMEOUT';
+  return error;
+}
 
 describe('NotificationRoutingController', () => {
   let context: ApplicationContext;
@@ -34,6 +48,11 @@ describe('NotificationRoutingController', () => {
     } as unknown as Mocked<NotificationRoutingRepository>;
 
     mockFactory.getNotificationRoutingRepository = vi.fn().mockReturnValue(mockRepo);
+
+    vi.mocked(dns.resolveMx)
+      .mockReset()
+      .mockResolvedValue([{ exchange: 'mail.example.com', priority: 10 }]);
+    vi.mocked(dns.resolve).mockReset().mockResolvedValue(['1.2.3.4']);
 
     context = await createMockApplicationContext();
     context.session.user.roles = [CamsRole.SuperUser];
@@ -241,6 +260,93 @@ describe('NotificationRoutingController', () => {
       await expect(controller.handleRequest(context)).rejects.toThrow(
         expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
       );
+    });
+  });
+
+  describe('handlePut domain validation', () => {
+    test('should throw BadRequestError when a domain has neither MX nor A records', async () => {
+      vi.mocked(dns.resolveMx).mockRejectedValue(notFoundError());
+      vi.mocked(dns.resolve).mockRejectedValue(notFoundError());
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = { recipientAddresses: ['someone@UST.DOJ.GOV'] };
+
+      await expect(controller.handleRequest(context)).rejects.toThrow(
+        expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
+      );
+      expect(mockRepo.updateRoutingRecord).not.toHaveBeenCalled();
+    });
+
+    test('should fall back to an A-record lookup when a domain has no MX records', async () => {
+      vi.mocked(dns.resolveMx).mockRejectedValue(notFoundError());
+      vi.mocked(dns.resolve).mockResolvedValue(['1.2.3.4']);
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = { recipientAddresses: ['someone@usdoj.gov'] };
+      mockRepo.updateRoutingRecord.mockResolvedValue({
+        ...mockRecord,
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+
+      const result = await controller.handleRequest(context);
+
+      expect(result.statusCode).toBe(HttpStatusCodes.OK);
+      expect(mockRepo.updateRoutingRecord).toHaveBeenCalledWith('chapter-7-oversight', {
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+    });
+
+    test('should save and return a warning when the DNS lookup is indeterminate rather than confirming the domain is missing', async () => {
+      vi.mocked(dns.resolveMx).mockRejectedValue(timeoutError());
+      vi.mocked(dns.resolve).mockRejectedValue(timeoutError());
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = { recipientAddresses: ['someone@usdoj.gov'] };
+      mockRepo.updateRoutingRecord.mockResolvedValue({
+        ...mockRecord,
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+
+      const result = await controller.handleRequest(context);
+
+      expect(result.statusCode).toBe(HttpStatusCodes.OK);
+      expect(mockRepo.updateRoutingRecord).toHaveBeenCalledWith('chapter-7-oversight', {
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+      expect(result.body.warnings).toEqual([expect.stringContaining("'usdoj.gov'")]);
+    });
+
+    test('should not include a warnings field on the response when domain validation succeeds cleanly', async () => {
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = { recipientAddresses: ['someone@usdoj.gov'] };
+      mockRepo.updateRoutingRecord.mockResolvedValue({
+        ...mockRecord,
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+
+      const result = await controller.handleRequest(context);
+
+      expect(result.statusCode).toBe(HttpStatusCodes.OK);
+      expect(result.body.warnings).toBeUndefined();
+    });
+
+    test('should only check each unique domain once across multiple recipient addresses', async () => {
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = {
+        recipientAddresses: ['first@usdoj.gov', 'second@usdoj.gov'],
+      };
+      mockRepo.updateRoutingRecord.mockResolvedValue({
+        ...mockRecord,
+        recipientAddresses: ['first@usdoj.gov', 'second@usdoj.gov'],
+      });
+
+      const result = await controller.handleRequest(context);
+
+      expect(result.statusCode).toBe(HttpStatusCodes.OK);
+      expect(dns.resolveMx).toHaveBeenCalledTimes(1);
+      expect(dns.resolveMx).toHaveBeenCalledWith('usdoj.gov');
     });
   });
 });
--- a/backend/lib/controllers/admin/notification-routing.controller.test.ts
+++ b/backend/lib/controllers/admin/notification-routing.controller.test.ts
@@ -9,13 +9,17 @@ import factory from '../../factory';
 import { NotificationRoutingRecord } from '@common/cams/notifications';
 import HttpStatusCodes from '@common/api/http-status-codes';
 
-vi.mock('../../factory');
 vi.mock('node:dns/promises');
-const mockFactory = factory as Mocked<typeof factory>;
 
 function notFoundError(): NodeJS.ErrnoException {
   const error = new Error('queryA ENOTFOUND') as NodeJS.ErrnoException;
   error.code = 'ENOTFOUND';
+  return error;
+}
+
+function noDataError(): NodeJS.ErrnoException {
+  const error = new Error('queryMx ENODATA') as NodeJS.ErrnoException;
+  error.code = 'ENODATA';
   return error;
 }
 
@@ -39,6 +43,8 @@ describe('NotificationRoutingController', () => {
   };
 
   beforeEach(async () => {
+    vi.restoreAllMocks();
+
     mockRepo = {
       getAll: vi.fn(),
       updateRoutingRecord: vi.fn(),
@@ -47,7 +53,7 @@ describe('NotificationRoutingController', () => {
       release: vi.fn(),
     } as unknown as Mocked<NotificationRoutingRepository>;
 
-    mockFactory.getNotificationRoutingRepository = vi.fn().mockReturnValue(mockRepo);
+    vi.spyOn(factory, 'getNotificationRoutingRepository').mockReturnValue(mockRepo);
 
     vi.mocked(dns.resolveMx)
       .mockReset()
@@ -57,10 +63,6 @@ describe('NotificationRoutingController', () => {
     context = await createMockApplicationContext();
     context.session.user.roles = [CamsRole.SuperUser];
     controller = new NotificationRoutingController(context);
-  });
-
-  afterEach(() => {
-    vi.restoreAllMocks();
   });
 
   describe('authorization', () => {
@@ -188,14 +190,50 @@ describe('NotificationRoutingController', () => {
       expect(result.statusCode).toBe(HttpStatusCodes.METHOD_NOT_ALLOWED);
     });
 
-    test('should throw BadRequestError when recipientAddresses is missing', async () => {
+    test.each([
+      { description: 'recipientAddresses is missing', body: {} },
+      {
+        description: 'recipientAddresses contains an invalid address',
+        body: { recipientAddresses: ['valid@example.com', 'not-an-email'] },
+      },
+      { description: 'body is null', body: null },
+      {
+        description: 'recipientAddresses is a non-array value',
+        body: { recipientAddresses: 'not-an-array' },
+      },
+      {
+        description: 'recipientAddresses contains non-string values',
+        body: { recipientAddresses: [1, null, {}] },
+      },
+      {
+        description: 'recipientAddresses mixes valid strings and non-strings',
+        body: { recipientAddresses: ['valid@example.com', 42] },
+      },
+    ])('should throw BadRequestError when $description', async ({ body }) => {
       context.request.method = 'PUT';
       context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = {};
+      context.request.body = body;
 
       await expect(controller.handleRequest(context)).rejects.toThrow(
         expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
       );
+    });
+
+    test('should trim leading/trailing whitespace from addresses before validating and saving', async () => {
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = { recipientAddresses: ['  someone@usdoj.gov  '] };
+      mockRepo.updateRoutingRecord.mockResolvedValue({
+        ...mockRecord,
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+
+      const result = await controller.handleRequest(context);
+
+      expect(result.statusCode).toBe(HttpStatusCodes.OK);
+      expect(mockRepo.updateRoutingRecord).toHaveBeenCalledWith('chapter-7-oversight', {
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
     });
 
     test('should allow clearing all recipients by passing an empty array', async () => {
@@ -211,65 +249,53 @@ describe('NotificationRoutingController', () => {
         recipientAddresses: [],
       });
     });
-
-    test('should throw BadRequestError when any address in recipientAddresses is invalid', async () => {
-      context.request.method = 'PUT';
-      context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = { recipientAddresses: ['valid@example.com', 'not-an-email'] };
-
-      await expect(controller.handleRequest(context)).rejects.toThrow(
-        expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
-      );
-    });
-
-    test('should throw BadRequestError when body is null', async () => {
-      context.request.method = 'PUT';
-      context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = null;
-
-      await expect(controller.handleRequest(context)).rejects.toThrow(
-        expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
-      );
-    });
-
-    test('should throw BadRequestError when recipientAddresses is a non-array value', async () => {
-      context.request.method = 'PUT';
-      context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = { recipientAddresses: 'not-an-array' };
-
-      await expect(controller.handleRequest(context)).rejects.toThrow(
-        expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
-      );
-    });
-
-    test('should throw BadRequestError when recipientAddresses contains non-string values', async () => {
-      context.request.method = 'PUT';
-      context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = { recipientAddresses: [1, null, {}] };
-
-      await expect(controller.handleRequest(context)).rejects.toThrow(
-        expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
-      );
-    });
-
-    test('should throw BadRequestError when recipientAddresses mixes valid strings and non-strings', async () => {
-      context.request.method = 'PUT';
-      context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = { recipientAddresses: ['valid@example.com', 42] };
-
-      await expect(controller.handleRequest(context)).rejects.toThrow(
-        expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
-      );
-    });
   });
 
   describe('handlePut domain validation', () => {
-    test('should throw BadRequestError when a domain has neither MX nor A records', async () => {
-      vi.mocked(dns.resolveMx).mockRejectedValue(notFoundError());
-      vi.mocked(dns.resolve).mockRejectedValue(notFoundError());
+    test.each([
+      { description: 'ENOTFOUND', mxError: notFoundError(), aError: notFoundError() },
+      { description: 'ENODATA', mxError: noDataError(), aError: noDataError() },
+    ])(
+      'should throw BadRequestError when a domain has neither MX nor A records ($description)',
+      async ({ mxError, aError }) => {
+        vi.mocked(dns.resolveMx).mockRejectedValue(mxError);
+        vi.mocked(dns.resolve).mockRejectedValue(aError);
+        context.request.method = 'PUT';
+        context.request.params = { routingId: 'chapter-7-oversight' };
+        context.request.body = { recipientAddresses: ['someone@UST.DOJ.GOV'] };
+
+        await expect(controller.handleRequest(context)).rejects.toThrow(
+          expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
+        );
+        expect(mockRepo.updateRoutingRecord).not.toHaveBeenCalled();
+      },
+    );
+
+    test('should fall back to an A-record lookup when the MX lookup resolves an empty list', async () => {
+      vi.mocked(dns.resolveMx).mockResolvedValue([]);
+      vi.mocked(dns.resolve).mockResolvedValue(['1.2.3.4']);
       context.request.method = 'PUT';
       context.request.params = { routingId: 'chapter-7-oversight' };
-      context.request.body = { recipientAddresses: ['someone@UST.DOJ.GOV'] };
+      context.request.body = { recipientAddresses: ['someone@usdoj.gov'] };
+      mockRepo.updateRoutingRecord.mockResolvedValue({
+        ...mockRecord,
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+
+      const result = await controller.handleRequest(context);
+
+      expect(result.statusCode).toBe(HttpStatusCodes.OK);
+      expect(mockRepo.updateRoutingRecord).toHaveBeenCalledWith('chapter-7-oversight', {
+        recipientAddresses: ['someone@usdoj.gov'],
+      });
+    });
+
+    test('should reject a domain when the A-record lookup resolves an empty list', async () => {
+      vi.mocked(dns.resolveMx).mockRejectedValue(notFoundError());
+      vi.mocked(dns.resolve).mockResolvedValue([]);
+      context.request.method = 'PUT';
+      context.request.params = { routingId: 'chapter-7-oversight' };
+      context.request.body = { recipientAddresses: ['someone@usdoj.gov'] };
 
       await expect(controller.handleRequest(context)).rejects.toThrow(
         expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST }),
--- a/backend/lib/controllers/admin/notification-routing.controller.ts
+++ b/backend/lib/controllers/admin/notification-routing.controller.ts
@@ -1,3 +1,4 @@
+import * as dns from 'node:dns/promises';
 import { ApplicationContext } from '../../adapters/types/basic';
 import { CamsHttpResponseInit, httpSuccess } from '../../adapters/utils/http-response';
 import { CamsController } from '../controller';
@@ -59,7 +60,7 @@ export class NotificationRoutingController implements CamsController {
         message: `Unknown routing ID: '${routingId}'. Must be one of: ${NOTIFICATION_ROUTING_DEFINITIONS.map((d) => d.id).join(', ')}`,
       });
     }
-    const input = this.validateUpdateInput(context.request.body);
+    const { input, warnings } = await this.validateUpdateInput(context, context.request.body);
     const definition = NOTIFICATION_ROUTING_DEFINITIONS.find((d) => d.id === routingId)!;
     const existing = await this.repository.findRecipientByRoutingKey(definition.covers[0]);
     const record = await this.repository.updateRoutingRecord(routingId, input);
@@ -73,11 +74,18 @@ export class NotificationRoutingController implements CamsController {
     });
     return httpSuccess({
       statusCode: HttpStatusCodes.OK,
-      body: { meta: { self: context.request.url }, data: record },
+      body: {
+        meta: { self: context.request.url },
+        data: record,
+        ...(warnings.length > 0 ? { warnings } : {}),
+      },
     });
   }
 
-  private validateUpdateInput(body: unknown): NotificationRoutingUpdateInput {
+  private async validateUpdateInput(
+    context: ApplicationContext,
+    body: unknown,
+  ): Promise<{ input: NotificationRoutingUpdateInput; warnings: string[] }> {
     const input = body as { recipientAddresses?: unknown } | null;
     if (!input) {
       throw new BadRequestError(MODULE_NAME, { message: 'Request body is required.' });
@@ -100,7 +108,60 @@ export class NotificationRoutingController implements CamsController {
         message: `Invalid email address(es): ${invalid.join(', ')}`,
       });
     }
-    return { recipientAddresses: trimmed };
+
+    const { invalidDomains, warnings } = await this.checkDomains(context, trimmed);
+    if (invalidDomains.length > 0) {
+      throw new BadRequestError(MODULE_NAME, {
+        message: `Domain(s) do not appear to accept email and were rejected: ${invalidDomains.join(', ')}`,
+      });
+    }
+
+    return { input: { recipientAddresses: trimmed }, warnings };
+  }
+
+  private async checkDomains(
+    context: ApplicationContext,
+    addresses: string[],
+  ): Promise<{ invalidDomains: string[]; warnings: string[] }> {
+    const domains = new Set(
+      addresses.map((address) => address.slice(address.lastIndexOf('@') + 1).toLowerCase()),
+    );
+    const invalidDomains: string[] = [];
+    const warnings: string[] = [];
+
+    for (const domain of domains) {
+      const result = await this.resolveDomain(domain);
+      if (result === 'not-found') {
+        invalidDomains.push(domain);
+      } else if (result === 'indeterminate') {
+        const message = `Could not verify that the domain '${domain}' accepts email; the DNS lookup failed rather than confirming the domain doesn't exist. Saved without domain verification for this address.`;
+        context.logger.warn(MODULE_NAME, message);
+        warnings.push(message);
+      }
+    }
+
+    return { invalidDomains, warnings };
+  }
+
+  private async resolveDomain(domain: string): Promise<'valid' | 'not-found' | 'indeterminate'> {
+    try {
+      const mxRecords = await dns.resolveMx(domain);
+      if (mxRecords.length > 0) return 'valid';
+    } catch (error) {
+      if (!this.isDomainNotFoundError(error)) return 'indeterminate';
+    }
+
+    try {
+      const addresses = await dns.resolve(domain);
+      return addresses.length > 0 ? 'valid' : 'not-found';
+    } catch (error) {
+      return this.isDomainNotFoundError(error) ? 'not-found' : 'indeterminate';
+    }
+  }
+
+  private isDomainNotFoundError(error: unknown): boolean {
+    const code = (error as NodeJS.ErrnoException)?.code;
+    return code === 'ENOTFOUND' || code === 'ENODATA';
   }
 
   private requireSuperUser(context: ApplicationContext): void {
--- a/backend/lib/use-cases/notifications/trustee-change-notification.ts
+++ b/backend/lib/use-cases/notifications/trustee-change-notification.ts
@@ -23,39 +23,48 @@ export class TrusteeChangeNotificationUseCase {
   async notify(context: ApplicationContext, changeSet: TrusteeChangeSet): Promise<void> {
     if (changeSet.fields.length === 0) return;
 
-    const recipients = await this.resolveRecipients(context, changeSet);
-    if (recipients.length === 0) return;
+    const mailingLists = await this.resolveMailingLists(context, changeSet);
+    if (mailingLists.length === 0) return;
 
     const compiled = compileTrusteeChangeTemplate(changeSet);
     const replyTo = changeSet.author?.email
       ? { address: changeSet.author.email, displayName: changeSet.author.name }
       : undefined;
 
-    for (const recipient of recipients) {
-      for (const address of recipient.recipientAddresses) {
-        const notification: Notification = {
-          to: address,
-          toDisplayName: recipient.displayName,
-          subject: compiled.subject,
-          html: compiled.html,
-          text: compiled.text,
-          correlationId: context.invocationId,
-          replyTo,
-        };
-        try {
-          await this.notificationGateway.send(notification);
-        } catch (error) {
-          context.logger.error(
-            MODULE_NAME,
-            `Failed to send trustee change notification to '${address}' (covers: ${recipient.covers.join(', ')}).`,
-            error,
-          );
-        }
+    for (const mailingList of mailingLists) {
+      await this.sendToMailingList(context, mailingList, compiled, replyTo);
+    }
+  }
+
+  private async sendToMailingList(
+    context: ApplicationContext,
+    mailingList: NotificationRecipient,
+    compiled: { subject: string; html: string; text: string },
+    replyTo: Notification['replyTo'],
+  ): Promise<void> {
+    for (const address of mailingList.recipientAddresses) {
+      const notification: Notification = {
+        to: address,
+        toDisplayName: mailingList.displayName,
+        subject: compiled.subject,
+        html: compiled.html,
+        text: compiled.text,
+        correlationId: context.invocationId,
+        replyTo,
+      };
+      try {
+        await this.notificationGateway.send(notification);
+      } catch (error) {
+        context.logger.error(
+          MODULE_NAME,
+          `Failed to send trustee change notification to '${address}' (covers: ${mailingList.covers.join(', ')}).`,
+          error,
+        );
       }
     }
   }
 
-  private async resolveRecipients(
+  private async resolveMailingLists(
     context: ApplicationContext,
     changeSet: TrusteeChangeSet,
   ): Promise<NotificationRecipient[]> {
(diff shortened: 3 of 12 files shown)


btposey:
## Orchestrated Review (pragmatic + software-architect + devsecops + sre)

**The function & this branch's changes** (recap): `TrusteeChangeNotificationUseCase.notify()` sends trustee-change emails to routed mailing lists. This branch (CAMS-856) fixes two real bugs — a single bad recipient no longer aborts the whole send batch (new `sendToMailingList()` with per-address try/catch), and the admin notification-routing endpoint now DNS-validates domains at save time (MX→A/AAAA fallback, reject-if-nonexistent / warn-if-indeterminate) plus enables ACS diagnostic logging for real send/delivery visibility.

**Orchestrator review results** — 4 agents (pragmatic, software-architect, devsecops, sre), all clean on the core bug-fix logic itself, converging on one real theme: **the fix stops the batch-abort but doesn't fully close the "silent" failure loop the PR title promises.**

### High severity

1. **Observability gap (SRE)** — the new per-address catch only calls `context.logger.error` (console), never `observability.completeTrace`/`trackEvent`. Worse: if *every* send in a batch fails, `notify()` still returns `void` with no exception, so the pre-existing outer catch in `dispatchChangeNotification` — the only code path wired to App Insights — never fires. A total-ACS-outage scenario is now *less* visible than before this PR.
2. **No alert wired to new diagnostic logs (SRE)** — `EmailStatusUpdateOperational` now flows to Log Analytics, and reusable alert-rule modules already exist in this repo, but none references the new tables. Real bounces produce data nobody queries.
3. **Recipient PII into shared Log Analytics workspace (DevSecOps)** — `EmailStatusUpdateOperational` populates `RecipientId` with the literal recipient address, undocumented in the Bicep, in a workspace shared with broader ops/monitoring RBAC than the primary datastore.

### Medium severity

- DNS resolution isn't isolated behind an adapter (Invasive Species violation) — it's the only direct network-I/O call in `backend/lib`, which is why the test resorts to `vi.mock('node:dns/promises')` instead of the repo's preferred `vi.spyOn()`.
- `notify()`'s `void` return loses the total-failure signal (companion to finding #1).
- Unbounded, sequential DNS lookups in the PUT request path — no timeout, so a slow resolver can hang an admin save.
- No cap on `recipientAddresses` length before triggering DNS lookups (resource-exhaustion defense-in-depth gap behind the SuperUser gate).

### Low severity

- `main.bicep`'s `acsEmail` module bypasses the `deployAppInsights` gating convention every sibling call site uses.
- Test coverage handles one simultaneous failure but not multiple concurrent failures.

### No issues found

SuperUser gating, DTO/`warnings` boundary design, `sendToMailingList` extraction, the `ResponseBody.warnings` field design, YAGNI/dead-code cleanup (`MODULE_NAME` removal), and the DNS three-way classification logic itself — all were explicitly reviewed and confirmed sound.

### Bottom line

This PR is a legitimate, well-scoped fix for the batch-abort bug and adds real save-time validation, but as currently written it trades one silent-failure mode for a quieter one. The two High findings worth raising in review are (1) wiring the new per-address failure path (and a total-failure summary from `notify()`) into `observability.completeTrace`, and (2) adding an alert rule against the new ACS diagnostic tables — both are explicitly flagged in the PR description as "tracked separately," so this may be intentional scoping, but it's worth confirming that's understood rather than assumed.

---

## For AI Agents

Detailed, self-contained prompts an AI coding agent can use to address each finding. Each prompt assumes a fresh session with no memory of this review — file paths and context are included so the agent doesn't need to rediscover them. Click a heading to expand.

<details>
<summary>1. Wire per-address send failures (and total-failure summary) into Application Insights — <strong>High</strong></summary>

```
In backend/lib/use-cases/notifications/trustee-change-notification.ts, the
sendToMailingList() method's catch block (around line 55) only calls
context.logger.error(...) when a single address's send fails — this never
reaches Application Insights, only console/log-stream output.

Additionally, notify() (the public method, ~line 23) returns Promise<void>
with no indication of partial or total failure. Its caller,
dispatchChangeNotification() in backend/lib/use-cases/trustees/trustees.ts
(~line 656), has an outer catch that DOES call
context.observability.completeTrace(trace, 'Trustee Change Notification',
{ success: false, ... }) — but that catch can now never fire, because
sendToMailingList's per-address try/catch swallows every failure before it
can propagate.

Task:
1. Change notify() to return a summary object instead of void, e.g.:
   { attempted: number; failed: number; failedAddresses: string[] }
   Track attempted/failed counts across all mailingLists/addresses in
   sendToMailingList and aggregate them in notify().
2. Update dispatchChangeNotification() in trustees.ts to inspect that
   summary and call context.observability.completeTrace(trace,
   'Trustee Change Notification', { success: summary.failed === 0, ... })
   (match the existing completeTrace call signature/conventions already
   used in that file) so both partial and total failures produce a
   queryable App Insights event, not just a console log line.
3. Follow this repo's TDD convention (Vitest, vi.spyOn() for mocking,
   test() not it()) — see backend/lib/use-cases/notifications/
   trustee-change-notification.test.ts and backend/lib/use-cases/trustees/
   trustees.test.ts for existing patterns to extend. Add/update tests
   that assert completeTrace is called with success:false on both a
   single-address failure and an all-addresses-failed scenario.
4. Run `npm run test:backend` and `npm run typecheck` (backend workspace)
   to confirm nothing regresses.

Reference: backend/lib/adapters/services/observability.ts for the
completeTrace/trackEvent implementation this should plug into.
```

</details>

<details>
<summary>2. Add an alert rule for the new ACS diagnostic log categories — <strong>High</strong></summary>

```
ops/cloud-deployment/lib/app-insights/diagnostics-settings-email.bicep
(new in this branch) enables the EmailSendMailOperational and
EmailStatusUpdateOperational diagnostic categories on the ACS
Communication Services resource, forwarding them to the Log Analytics
workspace referenced by analyticsWorkspaceId in
ops/cloud-deployment/main.bicep. No alert rule currently queries these
new tables.

This repo already has reusable alert-rule Bicep modules under
ops/cloud-deployment/lib/monitoring-alerts/ (look for
scheduled-query-alert-rule.bicep, log-alert-rule.bicep,
metrics-alert-rule.bicep) and an existing alert-action-group module
instantiated in main.bicep (~line 157).

Task:
1. Read ops/cloud-deployment/lib/monitoring-alerts/
   scheduled-query-alert-rule.bicep to understand its parameters and
   find an existing call site in main.bicep as a template for the
   KQL query shape and alert-action-group wiring.
2. Add a new scheduled-query-alert-rule instance in main.bicep that
   queries EmailStatusUpdateOperational for bounce/delivery-failure
   statuses (Delivered != true or equivalent per the ACS log schema —
   see https://learn.microsoft.com/en-us/azure/communication-services/
   concepts/analytics/logs/email-logs for the field schema) over a
   rolling window (e.g. last 1 hour), firing through the existing
   alert-action-group.
3. Follow this repo's Bicep conventions: conditional deployment via
   `if (!empty(analyticsWorkspaceId))` or similar guards already used
   by sibling diagnostic-settings files, and match the existing
   deployAppInsights gating pattern used at other main.bicep call sites
   (see prompt 7 below for the one place that currently doesn't).
4. Validate with `az bicep build` or however this repo's CI validates
   Bicep (check ops/cloud-deployment for a lint/validate script) before
   considering this done.

If the team decides this is legitimately out of scope for this PR
(the PR description does mention "building admin/infra alerting... "
as tracked follow-up work), do not implement — instead confirm this
finding is a known, accepted gap rather than an oversight.
```

</details>

<details>
<summary>3. Address recipient PII flowing into the shared Log Analytics workspace — <strong>High</strong></summary>

```
ops/cloud-deployment/lib/app-insights/diagnostics-settings-email.bicep
enables the EmailStatusUpdateOperational diagnostic log category on the
ACS Communication Services resource. Per Microsoft's documented schema
(https://learn.microsoft.com/en-us/azure/communication-services/concepts/
analytics/logs/email-logs), this category's RecipientId field contains
the literal recipient email address for delivery-status events. This
data now flows into the shared Log Analytics workspace referenced by
analyticsWorkspaceId, which also receives Cosmos/SQL/web-app diagnostics
and may have broader RBAC read access (ops/monitoring personnel) than
the primary trustee data store.

This is a data-governance question as much as a code question — do NOT
just silently change the Bicep. Task:
1. First, ask a human reviewer (or the PR author) to confirm with the
   USTP data-governance/FedRAMP compliance owner whether recipient email
   addresses in Log Analytics are acceptable given this workspace's
   existing access model and retention policy
   (ops/cloud-deployment/lib/analytics/log-analytics-workspace.bicep
   sets retentionInDays, currently defaulted around 30 days — check the
   current value).
2. If accepted: add a comment block in
   diagnostics-settings-email.bicep directly above the
   EmailStatusUpdateOperational category entry documenting that this
   category contains recipient PII (RecipientId) and noting the
   retention/RBAC model it now falls under, so future reviewers aren't
   surprised.
3. If NOT accepted: either (a) look into Log Analytics workspace
   transformation / Data Collection Rule (DCR) support for redacting the
   RecipientId field before ingestion, or (b) disable the
   EmailStatusUpdateOperational category and rely solely on
   EmailSendMailOperational (aggregate counts, no PII) plus the
   application-level per-address error logging already added in
   trustee-change-notification.ts for delivery-failure diagnosis.
4. Whatever is decided, do NOT proceed with code changes to Bicep
   without that data-governance answer — this finding is flagged
   specifically because it needs a human policy decision, not just an
   engineering fix.
```

</details>

<details>
<summary>4. Isolate DNS domain resolution behind a gateway interface — <strong>Medium</strong></summary>

```
backend/lib/controllers/admin/notification-routing.controller.ts imports
node:dns/promises directly and calls dns.resolveMx/dns.resolve inside
private methods resolveDomain(), checkDomains(), and
isDomainNotFoundError() (~lines 122-165). This is the only place in
backend/lib that performs live network I/O without going through an
adapter/gateway interface — every other integration (ACS notifications,
Okta, DXTR, ACMS, storage) is wrapped behind an interface defined in
backend/lib/use-cases/gateways.types.ts and obtained via
factory.get*(context), per this repo's Option-Enabling Architecture
conventions (see .ustp-cams-fdp/skills/cams-architecture/SKILL.md,
specifically the "Invasive Species" principle and PROCEDURE 6).

A direct symptom of this gap: backend/lib/controllers/admin/
notification-routing.controller.test.ts has to use
vi.mock('node:dns/promises') to test this logic — a pattern this repo's
own testing convention (.ustp-cams-fdp/skills/cams-testing-unit/SKILL.md)
calls out as a last resort, preferring vi.spyOn() against an injected
dependency.

Task:
1. Define a new interface in backend/lib/use-cases/gateways.types.ts,
   e.g.:
   export interface DomainVerificationGateway {
     verifyMailDomain(domain: string): Promise<'valid' | 'not-found' | 'indeterminate'>;
   }
2. Create backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts
   implementing this interface, moving resolveDomain()'s and
   isDomainNotFoundError()'s logic (the node:dns/promises calls) into it.
   Follow the existing gateway file conventions in
   backend/lib/adapters/gateways/ (e.g. the notifications/ or okta/
   subdirectories) for structure and naming.
3. Add a factory function (getDomainVerificationGateway) to
   backend/lib/factory.ts following the existing hybrid DI pattern used
   for NotificationGateway/NotificationRoutingRepository in this same
   controller — check context.config.get('dbMock') or an equivalent
   flag for whether to return a mock in test mode, and register the
   real implementation with deferRelease() if applicable (check whether
   DNS gateways need cleanup — likely not, so this may not apply).
4. Update NotificationRoutingController to obtain this gateway via
   factory.getDomainVerificationGateway(context) in its constructor
   (matching the existing pattern for this.repository), and replace
   direct dns.* calls in checkDomains() with calls to
   this.domainVerificationGateway.verifyMailDomain(domain).
5. Update notification-routing.controller.test.ts to use vi.spyOn()
   against a mock DomainVerificationGateway implementation instead of
   vi.mock('node:dns/promises'), consistent with this repo's testing
   conventions.
6. Run `npm run test:backend`, `npm run typecheck`, and `npm run lint`
   (backend workspace) to confirm everything passes.

This also sets up prompt 5 (moving domain-check business logic out of
the controller into a use-case) as a natural follow-on, since the
controller would then depend only on the interface.
```

</details>

<details>
<summary>5. Add a timeout and parallelize DNS lookups in the admin PUT path — <strong>Medium</strong></summary>

```
backend/lib/controllers/admin/notification-routing.controller.ts's
resolveDomain() method (~lines 146-160) calls dns.resolveMx() and, on
fallback, dns.resolve() from node:dns/promises with no explicit timeout
— it relies entirely on Node's default resolver retry/timeout behavior,
which is not tightly bounded. Additionally, checkDomains() (~lines
122-144) awaits resolveDomain() sequentially in a for-loop over each
unique domain in the request, so total worst-case latency for N domains
is N x (MX timeout + A timeout), fully synchronous inside a single HTTP
PUT request handler.

Task:
1. Wrap each dns.resolveMx/dns.resolve call in resolveDomain() with an
   explicit timeout — use Promise.race() against a short timer (e.g.
   2-3 seconds) — and treat a timeout the same as any other
   'indeterminate' result (the existing fail-open behavior for
   non-ENOTFOUND/ENODATA errors is correct and should be preserved;
   just make sure a slow/hanging resolver hits that path quickly
   instead of waiting on Node's default retry behavior).
2. In checkDomains(), change the sequential for-loop to resolve all
   unique domains concurrently, e.g.:
   const results = await Promise.all(
     Array.from(domains).map((domain) => this.resolveDomain(domain))
   );
   then process the results array to build invalidDomains/warnings,
   preserving the existing per-domain classification logic.
3. This depends on (but doesn't strictly require) prompt 4's gateway
   extraction — if that's done first, put the timeout logic inside the
   new DomainVerificationGateway implementation instead of the
   controller.
4. Add/update tests in notification-routing.controller.test.ts to cover:
   a timeout scenario classified as 'indeterminate', and confirm domains
   are resolved concurrently (e.g. via timing assertions or call-order
   assertions with vi.spyOn(), consistent with this repo's mocking
   conventions — avoid vi.mock() per cams-testing-unit skill guidance).
5. Run `npm run test:backend` and `npm run typecheck` (backend
   workspace) to confirm no regressions.
```

</details>

<details>
<summary>6. Cap the number of recipient addresses accepted before triggering DNS lookups — <strong>Medium</strong></summary>

```
backend/lib/controllers/admin/notification-routing.controller.ts's
validateUpdateInput() method (~line 93) accepts any array length for
recipientAddresses — only type and email-format (EMAIL_REGEX) are
validated before checkDomains() performs live DNS lookups against every
unique domain in the array. This endpoint is gated by requireSuperUser()
(confirmed intact, called unconditionally first in handleRequest), so
exploitation requires an already-privileged account, but there's
currently no defense-in-depth cap.

Task:
1. In validateUpdateInput(), after confirming input.recipientAddresses
   is an array of strings, add a reasonable upper-bound check (e.g.
   reject if trimmed.length exceeds some threshold like 50 — check
   NOTIFICATION_ROUTING_DEFINITIONS in @common/cams/notifications for
   any existing guidance on realistic mailing-list sizes, or ask the
   PR author/product owner what a sane real-world limit looks like for
   a single routing record).
2. Throw a BadRequestError (matching the existing error-handling
   pattern already used in this method, e.g.
   `throw new BadRequestError(MODULE_NAME, { message: '...' })`) with a
   clear message if the limit is exceeded.
3. Add a test in notification-routing.controller.test.ts asserting the
   PUT endpoint rejects a request with more than the configured limit
   of addresses, following existing test patterns in that file (Vitest,
   vi.spyOn() for any mocking needed).
4. Run `npm run test:backend` and `npm run typecheck` (backend
   workspace) to confirm no regressions.

This is complementary to prompt 5 (concurrent lookups with timeouts)
— both reduce the same class of resource-exhaustion/hang risk from
different angles.
```

</details>

<details>
<summary>7. Fix the <code>deployAppInsights</code> gating inconsistency in main.bicep — <strong>Low</strong></summary>

```
In ops/cloud-deployment/main.bicep, the acsEmail module call (~line 252)
passes `analyticsWorkspaceId: analyticsWorkspaceId` unconditionally,
while every other module call site in this file (~lines 218, 267, 313)
gates the same parameter behind
`deployAppInsights ? analyticsWorkspaceId : ''`. This is currently
harmless (acs-email.bicep's diagnostic-settings module is itself guarded
with `if (!empty(analyticsWorkspaceId))`, and the parameter defaults to
''), but it's an inconsistency that will confuse a future reader
reconciling why this one call site behaves differently.

Task:
1. Open ops/cloud-deployment/main.bicep and locate the acsEmail module
   call (~line 252) alongside the other call sites that use the
   `deployAppInsights ? analyticsWorkspaceId : ''` pattern (~lines 218,
   267, 313) to confirm the exact existing syntax.
2. Update the acsEmail call to match: change
   `analyticsWorkspaceId: analyticsWorkspaceId` to
   `analyticsWorkspaceId: deployAppInsights ? analyticsWorkspaceId : ''`
   — UNLESS there's a deliberate reason email diagnostics should be
   independent of the deployAppInsights flag, in which case instead add
   a one-line comment directly above the parameter explaining why (e.g.
   "// Email diagnostics are intentionally independent of
   deployAppInsights because ...").
3. Validate the Bicep compiles (check ops/cloud-deployment for its
   lint/validate script, or use `az bicep build --file main.bicep`).
4. This is a low-risk, mechanical fix — no test changes should be
   needed since Bicep templates aren't covered by the Vitest suites.
```

</details>

<details>
<summary>8. Add a multi-simultaneous-failure test for send isolation — <strong>Low</strong></summary>

```
backend/lib/use-cases/notifications/trustee-change-notification.test.ts
has a test (~line 319, "continues sending to addresses after a failed
send, both within and across recipients") that fails exactly one
address (the first) and asserts the remaining addresses/mailing lists
still get sent. This is good coverage for the core fix, but there's no
test where multiple addresses fail simultaneously (e.g. the first
address of the first mailing list AND the last address of the second
mailing list both throw), which would harden confidence that no
shared/reused state leaks between loop iterations in sendToMailingList()
and resolveMailingLists() — arguably the more realistic real-world
trigger for this bug class (a stale mailing list with several dead
addresses, not just one).

Task:
1. Read the existing test at ~line 319 in
   trustee-change-notification.test.ts to match its setup/assertion
   style exactly (vi.spyOn() usage, how NotificationGateway.send is
   mocked to fail for specific addresses, how context.logger.error
   calls are asserted).
2. Add a new test where at least 2 sends fail across different mailing
   lists/positions in the loop (e.g. first address of the first list,
   and the last address of the last list), and assert:
   - All successful sends still complete (assert notificationGateway.send
     was called for every non-failing address)
   - Both failures are logged individually via context.logger.error
     with the correct address/covers in the message
   - No cross-contamination between the two failure cases (e.g. the
     second failure's error object isn't accidentally logged against
     the first failure's address)
3. Use test() not it() (enforced by this repo's Sourcery config), and
   follow the file's existing describe/test structure.
4. Run `npm run test:backend` to confirm the new test passes and
   existing tests aren't broken. Check coverage with
   `npm run coverage:backend` if you want to confirm this improves
   branch coverage for sendToMailingList().
```

</details>

backend/lib/controllers/admin/notification-routing.controller.ts
```diff
@@ -100,7 +108,60 @@ export class NotificationRoutingController implements CamsController {
...
+    return { invalidDomains, warnings };
+  }
+
+  private async resolveDomain(domain: string): Promise<'valid' | 'not-found' | 'indeterminate'> {
+    try {
+      const mxRecords = await dns.resolveMx(domain);
+      if (mxRecords.length > 0) return 'valid';
+    } catch (error) {
+      if (!this.isDomainNotFoundError(error)) return 'indeterminate';
+    }
```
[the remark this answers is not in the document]
fmaddenflx: ✅ Fixed. `resolveDomain` (now `DnsDomainVerificationGateway.verifyMailDomain`) no longer short-circuits to `indeterminate` on a transient MX error — it always attempts the A-record lookup afterward, and only returns `not-found` when both MX and A lookups affirmatively confirm absence. See `backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts`.

ops/cloud-deployment/lib/app-insights/diagnostics-settings-email.bicep
```diff
@@ -0,0 +1,57 @@
...
+param communicationServiceName string
+
+@description('Resource id of the Log Analytics workspace to send diagnostic data to.')
+param analyticsWorkspaceId string
+
+resource communicationService 'Microsoft.Communication/communicationServices@2023-04-01' existing = {
+  name: communicationServiceName
+}
+
+resource setting 'Microsoft.Insights/diagnosticSettings@2021-05-01-preview' = {
```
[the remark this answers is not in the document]
fmaddenflx: Not applied — every other diagnostic-settings bicep file in this repo (`diagnostics-settings-appi.bicep`, `-func.bicep`, `-webapp.bicep`, `-cosmos.bicep`, `-sql.bicep`, all pre-existing) uses the preview API version `2021-05-01-preview`, not the stable one. Switching just this file to stable would make it the outlier rather than align it with the rest of the deployment as suggested. Left as `2021-05-01-preview` for consistency with the existing convention.

backend/lib/controllers/admin/notification-routing.controller.ts
```diff
@@ -73,11 +74,18 @@ export class NotificationRoutingController implements CamsController {
...
-      body: { meta: { self: context.request.url }, data: record },
+      body: {
+        meta: { self: context.request.url },
+        data: record,
+        ...(warnings.length > 0 ? { warnings } : {}),
+      },
     });
   }
 
-  private validateUpdateInput(body: unknown): NotificationRoutingUpdateInput {
```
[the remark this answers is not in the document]
fmaddenflx: ✅ Fixed. Split `validateUpdateInput` into a pure/synchronous shape-and-format validator plus a separate `validateDomains` orchestrator called from `handlePut`, and decomposed domain checking into `extractDomains`/`buildDomainWarning` helpers. Also extracted the MX/A resolution logic itself out of the controller entirely into a new `DnsDomainVerificationGateway` (see reply on the resolveDomain comment below and the Medium-severity gateway-isolation finding from the human review) — so the controller no longer contains any DNS-specific branching at all.

fmaddenflx:
Addressed the review findings, per priority discussed with the team:

**High #1 (observability gap)** — Fixed. `TrusteeChangeNotificationUseCase.notify()` now returns an `{ attempted, failed, failedAddresses }` summary instead of `void`. `dispatchChangeNotification` in `trustees.ts` always calls `observability.completeTrace` with that result (success = `failed === 0`), not just on exception — a total-batch-failure scenario is now visible in App Insights instead of only console/log-stream output.

**High #2 (no alert on new diagnostic logs)** — Deferred, not dropped. This is now tracked as its own vertical slice (cams-wh5m / branch `CAMS-856-slice-a-actionable-bounce-alert`), split off from the original single-slice design so it can ship independently once ready.

**High #3 (recipient PII into shared Log Analytics workspace)** — Deferred pending a data-governance decision, per this comment's own recommendation. Not something to resolve unilaterally in code.

**Medium — DNS gateway isolation** — Fixed. Extracted `DnsDomainVerificationGateway` (`backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts`) behind a new `DomainVerificationGateway` interface in `gateways.types.ts`, wired through `factory.ts`. The DNS resolver itself is now injected (defaulting to `node:dns/promises`) rather than imported directly, since that module's exports can't be `vi.spyOn()`'d — injection is what actually made `vi.mock()` unnecessary; a plain `import * as dns` inside a gateway class wouldn't have solved the underlying ESM spy limitation. Controller tests now inject a mock gateway instead of mocking the dns module.

**Medium — `notify()`'s void return** — Fixed as part of the High #1 fix above.

**Medium — unbounded/sequential DNS lookups** — Fixed. Added a 3s timeout per lookup inside the gateway, and `checkDomains` now resolves all unique domains concurrently via `Promise.all` instead of a sequential loop.

**Medium — no cap on recipientAddresses** — Fixed. Added a 50-address limit, checked before any DNS lookups fire.

**Low — bicep gating inconsistency** — Fixed. `acsEmail`'s `analyticsWorkspaceId` param now uses the same `deployAppInsights ? analyticsWorkspaceId : ''` pattern as its sibling call sites in `main.bicep`.

**Low — missing multi-failure test** — Fixed. Added a test with two simultaneous failures across two different mailing lists (first address of one list, last address of another), asserting all other sends still complete and both failures are logged independently without cross-contamination.

All new/changed code covered by tests (155 passing across the affected files), typechecks clean, lints clean, bicep compiles.

backend/lib/controllers/admin/notification-routing.controller.ts
```diff
@@ -100,9 +118,62 @@ export class NotificationRoutingController implements CamsController {
         message: `Invalid email address(es): ${invalid.join(', ')}`,
       });
     }
+
     return { recipientAddresses: trimmed };
   }
 
+  private async validateDomains(
```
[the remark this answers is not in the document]
fmaddenflx: ✅ Fixed. Collapsed `validateDomains`/`checkDomains`/`extractDomains`/`buildDomainWarning` into a single `validateAndCollectDomainWarnings` method as suggested.

backend/lib/use-cases/gateways.types.ts
```diff
@@ -418,6 +418,13 @@ export interface NotificationGateway {
   send(notification: Notification): Promise<void>;
 }
 
+export type DomainVerificationResult = 'valid' | 'not-found' | 'indeterminate';
+
+export interface DomainVerificationGateway {
+  /** Checks whether a domain appears able to accept email (MX, falling back to A/AAAA). */
```
diabagatekelly, line 432: **MEDIUM — confirmed by two independent reviewers.** This docstring (and the PR description) say "MX, falling back to A/AAAA," but `checkA()` in `dns-domain-verification.gateway.ts` only calls `resolver.resolve(domain)`, which defaults to A-record lookups only (Node's `dns.promises.resolve` rrtype defaults to `'A'`) — AAAA is never queried. In practice this is a narrow edge case (an IPv6-only mail domain with no MX and no A record would be misclassified `not-found` and falsely rejected at save time), but it's worth either implementing the AAAA fallback (`resolver.resolveAaaa` / `resolve(domain, 'AAAA')` after A comes back not-found) or fixing this docstring + the PR description to say "A only" so a future reader doesn't rely on behavior that isn't there.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts
```diff
@@ -0,0 +1,79 @@
...
+    const mxResult = await this.checkMx(domain);
+    if (mxResult === 'valid') return 'valid';
+
+    const aResult = await this.checkA(domain);
+    if (aResult === 'valid') return 'valid';
+
+    // Only report 'not-found' when both lookups affirmatively confirm the domain
+    // doesn't exist. If either was merely indeterminate (timeout, SERVFAIL, etc.),
+    // we haven't actually confirmed absence, so fail open rather than reject.
+    return mxResult === 'not-found' && aResult === 'not-found' ? 'not-found' : 'indeterminate';
```
diabagatekelly: **MEDIUM.** One combination of this truth table is untested: MX indeterminate (e.g. timeout) + A confirms-absent (`ENOTFOUND`/`ENODATA`). The existing test file covers the other 3 combinations. This specific combination is behaviorally distinct — worth asserting explicitly that it still correctly falls through to `'indeterminate'` (fail-open), since a coverage tool would likely show 100% branch coverage on this line while still missing this case.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts
```diff
@@ -0,0 +1,79 @@
...
+  resolveMx: typeof nodeDns.resolveMx;
+  resolve: typeof nodeDns.resolve;
+}
+
+function isDomainNotFoundError(error: unknown): boolean {
+  const code = (error as NodeJS.ErrnoException)?.code;
+  return code === 'ENOTFOUND' || code === 'ENODATA';
+}
+
+function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
```
diabagatekelly: **LOW.** `withTimeout()` races a timer against the real DNS promise but never cancels/aborts the underlying `resolveMx`/`resolve` call when the timer wins — on a genuinely slow resolver, the real lookup keeps running detached after this function has already returned `'indeterminate'`. Not exploitable (bounded by the OS-level DNS client's own timeout) and likely fine to leave as-is given this is a low-frequency, admin-only endpoint — but worth a one-line comment noting it's an accepted tradeoff, or switching to `dns.Resolver` with `.cancel()` if this ever becomes a concern under load.

backend/lib/controllers/admin/notification-routing.controller.ts
```diff
@@ -100,9 +118,48 @@ export class NotificationRoutingController implements CamsController {
...
+    );
+
+    const invalidDomains: string[] = [];
+    const warnings: string[] = [];
+    for (const [domain, result] of results) {
+      if (result === 'not-found') {
+        invalidDomains.push(domain);
+      } else if (result === 'indeterminate') {
+        const message = `Could not verify that the domain '${domain}' accepts email; the DNS lookup failed rather than confirming the domain doesn't exist. Saved without domain verification for this address.`;
+        context.logger.warn(MODULE_NAME, message);
```
diabagatekelly, line 149: **MEDIUM.** No test spies on this `logger.warn` call — if this line were deleted entirely, no test would fail. Compare with `trustee-change-notification.test.ts`, which does spy on the equivalent `logger.error` call for its per-address failure logging. Suggest adding `vi.spyOn(context.logger, 'warn')` to the existing "should save and return a warning when domain verification is indeterminate" test and asserting it was called with a message containing the domain.

backend/lib/controllers/admin/notification-routing.controller.test.ts
```diff
@@ -192,55 +235,122 @@ describe('NotificationRoutingController', () => {
         recipientAddresses: [],
       });
     });
+  });
 
-    test('should throw BadRequestError when any address in recipientAddresses is invalid', async () => {
+  describe('handlePut domain validation', () => {
+    test('should throw BadRequestError when a domain is not found', async () => {
```
diabagatekelly, line 241: **MEDIUM.** This test only checks `status === BAD_REQUEST` and that `updateRoutingRecord` wasn't called — the `BadRequestError` message content (which domain, from which array) is never asserted. A mutation that kept the same control flow (still throws) but corrupted the message text (wrong domain interpolated, wrong list joined) would survive undetected. Suggest strengthening to `expect.objectContaining({ status: HttpStatusCodes.BAD_REQUEST, message: expect.stringContaining('ust.doj.gov') })` or similar. A companion test for a *mixed* request (one domain not-found, one indeterminate, in the same PUT) would also be worth adding, to confirm rejection takes priority and the indeterminate domain's warning is correctly never surfaced — exactly the kind of interaction bug class this PR was written to fix.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  return error;
+}
+
+describe('DnsDomainVerificationGateway', () => {
+  let gateway: DnsDomainVerificationGateway;
+  let resolver: Mocked<DnsResolver>;
+
+  beforeEach(() => {
+    resolver = {
+      resolveMx: vi.fn(),
```
diabagatekelly, line 28: **LOW.** No test asserts `resolver.resolveMx` was called with the actual `domain` argument (only the `resolve` call's argument is checked, at line 68, and only in one test). Every test in this file uses the same literal domain (`'example.com'`), so a mutation that hardcoded a wrong string into the `resolveMx` call wouldn't be caught. Suggest adding `expect(resolver.resolveMx).toHaveBeenCalledWith('example.com')` to at least one test, and varying the domain value in one other test to make argument-passing verifiable rather than incidental.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
```diff
@@ -118,14 +121,18 @@ export function NotificationRouting() {
...
           });
+          if (response.warnings?.length) {
+            warnings.push(...response.warnings);
+          }
         } catch {
           failedDefinitions.push(def.displayName);
         }
       }
     }
+    setSaveWarnings(warnings);
```
diabagatekelly, line 135: **LOW.** When multiple routing categories are saved in the same click and only some have domain warnings, all warnings across all categories get flattened into one array and rendered in a single generic Alert (`saveWarnings.join(' ')` below). The warning text does include the offending domain, so it's still actionable, but the user can't tell *which routing category* triggered it. Minor UX gap — consider attributing each warning to its `def.displayName` before surfacing, if that's easy to wire through.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts
fmaddenflx: dns.Resolver with .cancel() sounds like a way to go, if a timeout is reached.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
fmaddenflx: Might be a good idea.

backend/lib/factory.ts
```diff
@@ -555,6 +557,10 @@ const getNotificationRoutingRepository = (
   return repo;
 };
 
+const getDomainVerificationGateway = (): DomainVerificationGateway => {
+  return new DnsDomainVerificationGateway();
+};
+
```
cruzjone-flexion, line 574: **`getDomainVerificationGateway` allocates a new instance on every call**

Every other gateway factory in this file uses a module-level singleton variable (e.g. `let notificationGateway: NotificationGateway | null = null`) so the instance is created once and reused. `DnsDomainVerificationGateway` holds no request-scoped state — its resolver is a fixed injected object — so there's no reason to allocate a fresh one per call.

```typescript
let domainVerificationGateway: DomainVerificationGateway | null = null;

const getDomainVerificationGateway = (): DomainVerificationGateway => {
  if (!domainVerificationGateway) {
    domainVerificationGateway = new DnsDomainVerificationGateway();
  }
  return domainVerificationGateway;
};
```

Not a bug (DNS lookups are stateless), but inconsistent with every other factory method in this file.

backend/lib/use-cases/notifications/trustee-change-notification.ts
```diff
@@ -20,47 +26,86 @@ export class TrusteeChangeNotificationUseCase {
     this.notificationGateway = factory.getNotificationGateway(context);
   }
 
-  async notify(context: ApplicationContext, changeSet: TrusteeChangeSet): Promise<void> {
-    if (changeSet.fields.length === 0) return;
+  async notify(
```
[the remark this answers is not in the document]
fmaddenflx: Restructured `notify()`/`sendToMailingList` to build one `TrusteeChangeNotificationSummary` from a single flat list of per-address results, instead of two `TrusteeChangeNotificationSummary` objects that get merged field-by-field. That removes the "summary of summaries" boilerplate this comment flagged, while keeping the per-address send/catch logic sequential (no behavior change).

We intentionally kept `sendToMailingList` (and `resolveMailingListsForCategory`) as separate methods rather than fully inlining per the second suggestion — each names a distinct unit of iteration and keeps `sendToMailingList` independently unit-testable, which we felt outweighed the extra line count from the split. Happy to revisit if the team feels otherwise.

Addressed in b85a6a361.

backend/lib/use-cases/gateways.types.ts
fmaddenflx: Implemented the AAAA fallback (`resolver.resolve6`) rather than narrowing the docstring — after MX and A both come back non-valid, we now also check AAAA before deciding `not-found` vs `indeterminate`. The docstring's original "MX, falling back to A/AAAA" claim now matches actual behavior.

Addressed in b85a6a361.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts
fmaddenflx: Added an explicit test for this combination (extended to the new 3-way MX/A/AAAA truth table): MX times out (indeterminate) while A and AAAA both affirmatively confirm absence. Confirmed it still falls through to `indeterminate`, since MX itself was never confirmed absent.

Addressed in b85a6a361.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.ts
fmaddenflx: Implemented — `withTimeout()` now calls `resolver.cancel()` when the timer wins. We went further than a single shared `dns.promises.Resolver`: each of `checkMx`/`checkA`/`checkAaaa` now builds its own resolver instance per call (via an injected factory) rather than sharing one on the gateway. That's because `Resolver.cancel()` aborts *every* outstanding query on that instance — a shared resolver would let a timeout on one domain's lookup cancel a different, unrelated domain's concurrently in-flight lookup (the controller checks multiple distinct domains concurrently via `Promise.all`). Per-call construction closes that hole and, as a side effect, made it safe to memoize the gateway as a singleton (see the `factory.ts` thread below).

Addressed in b85a6a361.

backend/lib/controllers/admin/notification-routing.controller.ts
fmaddenflx: Added a `vi.spyOn(context.logger, 'warn')` to the existing "should save and return a warning when domain verification is indeterminate" test, asserting it's called with a message containing the domain.

Addressed in b85a6a361.

backend/lib/controllers/admin/notification-routing.controller.test.ts
fmaddenflx: Strengthened the not-found test to assert the `BadRequestError` message contains the domain, and added a new test for a mixed request (one not-found domain + one indeterminate domain in the same PUT) confirming rejection wins and the indeterminate domain's warning is never surfaced.

Addressed in b85a6a361.

backend/lib/adapters/gateways/dns/dns-domain-verification.gateway.test.ts
fmaddenflx: Added a test asserting `resolver.resolveMx` is called with the actual domain argument, using a distinct domain value (`ust.doj.gov`) rather than the shared `example.com` literal, so a hardcoded-argument mutation would be caught.

Addressed in b85a6a361.

user-interface/src/admin/notification-routing/NotificationRouting.tsx
fmaddenflx: Each save warning is now prefixed with its routing category's `displayName` before being surfaced, so a mixed-category save with multiple warnings no longer flattens them into one unattributed alert. Added tests for both the single-category and multi-category warning cases.

Addressed in b85a6a361.

backend/lib/factory.ts
fmaddenflx: Good catch, and consistent with the rest of the file — memoized `getDomainVerificationGateway` as a module-level singleton to match the other gateway factories.

This is safe specifically because of the resolver-per-call change made in the same commit (see the `withTimeout()` thread above): each lookup now builds its own `dns.promises.Resolver` rather than the gateway sharing one instance. Singleton-izing the gateway *before* that fix would have let a timeout in one concurrent request's DNS lookup cancel an unrelated request's in-flight lookup, since `Resolver.cancel()` aborts every outstanding query on the shared instance.

Addressed in b85a6a361.

mattstankey:
Reviewed the full diff. Solid, disciplined work overall — per-recipient isolation is correct and well-tested, and the commit history shows real iteration on prior findings. Flagging four **significant** items (mostly architectural/operational, not correctness):

**1. CorrelationId ↔ messageId join is unverified against live Azure.** `acs-bounce-query.gateway.ts` projects `CorrelationId` and treats it as the ACS `messageId` (the `beginSend()` poller `result.id`) used as the archive key. `deployment.md` itself notes the table/column names were unverified against live data. If `ACSEmailStatusUpdateOperational.CorrelationId` isn't the same value returned by `beginSend()`, every `readArchivedEmail` lookup misses and the poller silently reports rows as `expired` — reproducing the exact silent-failure class this PR exists to kill, one layer deeper. This should be a hard gate: a real test bounce confirming the join before trusting this on Main-Gov.

**2. Admins likely get two notifications per bounce.** Both wired to the same `adminNotificationEmail`: `main.bicep`'s `acsBounceAlert` scheduled query fires a bare Azure Monitor alert email, and the `poll-notification-bounces` dataflow independently forwards a reconstructed email for the same rows ~15 min later. May be intentional (prompt heads-up vs. full content), but it isn't stated anywhere — worth confirming intent or suppressing one before it becomes a "why am I getting double emails" ticket.

**3. Head-of-line blocking in the poll cursor.** `bounce-poll.ts` advances across a contiguous prefix and `break`s on the first non-expired failure, so a single poison row blocks all newer bounces from reaching the admin until it succeeds or its 7-day archive expires. The docstring acknowledges this as a deliberate "never silently drop" tradeoff, but given bounces are the signal admins need promptly, a stuck row delaying newer alerts for up to a week is a real operational risk. Consider a follow-up for a retry cap / dead-letter.

**4. Scope.** This bundles ~5 features (send isolation, DNS validation + UI, diagnostic logging, Slice A alert infra, Slice B poll/archive/reconstruct pipeline) into 2,785 lines against `main`. Slices A/B were their own merged PRs (#2783, #2787). Not fixable now — just flagging that landing slices independently first keeps each review tractable.

The send-isolation and DNS-validation pieces I'd approve without hesitation. I'd want #1 and #2 resolved before trusting the bounce pipeline end-to-end.

mattstankey:
## Automated multi-perspective review (High/Medium findings)

Ran a parallel review across pragmatic/simplicity, security, SRE/reliability, and architecture lenses. Posting the High and Medium severity findings below; Low-severity notes omitted here.

### Headline / scope issue
**This PR appears to ship a feature its own description defers.** The PR body states that admin/infra alerting on top of the new diagnostic logs "is tracked separately," but the diff includes the full bounce-poll dataflow, a Mongo archive repository with TTL, a KQL bounce-query gateway, a reconstruction use-case, and ~6 new Bicep modules (alert rule, action group, Log Analytics reader role, workspace/ACS resource locks), plus significant new parameterization in `app-shared-setup.bicep`. Recommend splitting: land the bug fix (per-recipient send isolation, DNS validation on save, the `mailingList` rename) on its own, and let the bounce-alerting subsystem go through review separately so its retention policy, RBAC scope, and failure modes can be judged on their own merits.

### High
1. **`updateTrustee()` / `updateTrusteeAppointment()` now block the API response on the full notification send.** Dispatch changed from fire-and-forget (`.catch()` on an unawaited promise) to `await`ed. Sends are sequential per mailing list / per address, and each ACS send has its own ~30s poll timeout, so a slow-but-not-failing ACS can now directly multiply request latency on the trustee-update endpoint — a behavior/latency change not called out in "Major Changes." Recommend parallelizing sends (`Promise.allSettled`) with an overall time budget, or explicitly documenting the tradeoff if blocking is intentional.
   - `backend/lib/use-cases/trustees/trustees.ts` (`updateTrustee`), `backend/lib/use-cases/trustees/trustee-appointments.ts`, `backend/lib/use-cases/notifications/trustee-change-notification.ts` (`notify`/`sendToMailingList`)

2. **Bounce-poll can silently wedge on a single bad row for up to 7 days.** The cursor only advances past rows that succeeded or TTL'd out as "expired"; any other failure breaks the loop, leaving all later (including unrelated, newer) bounces unprocessed until the blocking row expires. This failure mode is not covered by the new alert rule (`acsSendFailureAlert`'s query only matches `[ERROR] [ACS-NOTIFICATION-GATEWAY]` / `[ERROR] [TRUSTEE-CHANGE-NOTIFICATION]`, not `[ERROR] [BOUNCE-POLL]`), so a stuck cursor is only visible via manual trace inspection.
   - `backend/lib/use-cases/notifications/bounce-poll.ts` (`pollAndReconstruct`), `ops/cloud-deployment/main.bicep` (alert query)

### Medium
3. **PII (`RecipientId`) now flows into Log Analytics with no documented retention or access-scoping decision.** This PR is the first to forward `EmailStatusUpdateOperational` diagnostics, which carry recipient email addresses, into the shared workspace. No `retentionInDays` is set/verified for the target workspace(s) alongside this new PII category, and no table-level RBAC is introduced.
   - `ops/cloud-deployment/lib/app-insights/diagnostics-settings-email.bicep`, `ops/cloud-deployment/main.bicep` / `app-shared-setup.bicep`

4. **New Log Analytics Reader role is granted to the shared `appConfigIdentity`** (also used broadly for Key Vault secret access) rather than a purpose-scoped identity, widening the blast radius of that identity to read access over the whole workspace, not just the new ACS email tables.
   - `ops/cloud-deployment/app-shared-setup.bicep`, `ops/cloud-deployment/lib/analytics/log-analytics-reader-role-assignment.bicep`

5. **Domain-validation business logic lives directly in the controller** (deriving domains, deciding hard-reject vs. soft-warning, building rejection messages) rather than a use-case. Pre-existing pattern for this controller, but it's growing — worth extracting a `NotificationRoutingUseCase` (or similar) as a fast-follow so the controller stays a thin translation layer.
   - `backend/lib/controllers/admin/notification-routing.controller.ts` (`validateAndCollectDomainWarnings`, `validateUpdateInput`)

6. **The `recipient` → `mailingList` rename is accurate but incomplete.** The use-case layer now says `mailingList` throughout, but the shared type (`NotificationRecipient` in `common/src/cams/notifications.ts`) and the repository method (`findRecipientByRoutingKey`) still use "recipient" language — leaving two names for the same domain concept depending on which file you're in.
   - `common/src/cams/notifications.ts`, `backend/lib/use-cases/notifications/trustee-change-notification.ts`

7. **No timeout on the Log Analytics bounce query**, and **no retry for transient (`connection`-classified) ACS send failures.** Note: these two findings partially conflict with a simplicity finding that the `isConnectionFailure` classification is currently unused for behavior — recommend either wiring up a short retry for `connection` failures (giving the classification a purpose) or removing the classification if retry isn't desired, rather than leaving it as unused dead-weight classification.
   - `backend/lib/adapters/gateways/monitor/acs-bounce-query.gateway.ts` (`queryBounces`), `backend/lib/adapters/gateways/notifications/acs-notification.gateway.ts` (`isConnectionFailure`)

---
*Generated via a 4-agent parallel review (pragmatic/simplicity, DevSecOps/security, SRE/reliability, software-architect/OeSA compliance).*

